PR #5, opened 2017-02-06
Clean root dir

kolja-esders:
Issue #4.


michitux:
Isn't `mbe.py` used for building without scons? And why did you remove the dependency information?

kolja-esders:
Yes, `mbe.py` is supposedly used for this purpose. In the file itself it is noted that it is "A very minimalistic Build Environment that _aims_ to replace SCons" and that "It remains unclear if this build environment will be extended to fully replace SCons."

Based on this it seems to me that this might not be fully functional yet and is not being used. Do you know more about that?

michitux:
It is used in `setup.py` when scons is not present, see [here](https://github.com/kit-parco/networkit/blob/ee76da936f05bb107f6ceabff36a610cd9a8fffd/setup.py#L181). It doesn't do any kind of dependency tracking and therefore causes a full rebuild every time you build NetworKit, but apart from that it is working.

PR #13, opened 2017-04-11
Update setup.py to cythonize in a Git environment

paddya:
The setup script currently only checks if the Mercurial directory (.hg) is present to determine whether to use the pre-cythonized _NetworKit.cpp. Therefore, the Cython wrapper is never re-built if you checked out NetworKit in Git.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -272,7 +272,7 @@ do_cythonize = False
 # depending on the role in which the setup script is called, it will be determined if _NetworKit.pyx will be cythonized.
 build_ext_cmd = None
 # the `build_ext` command depends on the role of the setup script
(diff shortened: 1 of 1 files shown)


michitux:
Why is this now in master and not in Dev? Shouldn't master only be updated for releases?

kolja-esders:
Yes you are correct. PR was for master and I didn't find an option to change the dest branch to Dev. For this minor change I thought it was okay to merge the PR and merge back to Dev afterwards. If you think we should be completely strict on this we can obviously remove the changes from master again. Future PRs should always target Dev though.

michitux:
No,  I don't think we should remove this. What I think would be useful is to set the Dev branch as default branch in GitHub. Then pull requests will be against Dev by default.

PR #18, opened 2017-04-24
Faster GCE, also for weighted graphs

michitux:
This contains some improvements for GCE which implement support for weighted graphs and make the algorithm a lot faster (not just constants, also asymptotically). In particular, the weights between a node and the community (and the rest of the graph) are stored for every node in the shell which makes the selection of the next candidate for quality measure "M" a lot faster as only one unordered map needs to be scanned instead of asking for every neighbor of all nodes in the shell if it is part of the community.


michitux:
This introduces an error with the quality measure L, please do not merge yet. I'll provide a fix soon.

michitux:
I've just pushed a fix. However, the error was not introduced by my changes but the calculation has actually always been wrong. I've checked the original paper, the measure "L" wants definitely the boundary of the community in the sense of the nodes of the community that have external neighbors. All previous versions a) use nodes of the shell in the boundary and b) do not update the shell before using it which leads to wrong deltas being added to the quality measure, i.e., calculating the quality measure from scratch gives a different result.

The good news is though that a) this should already be faster in practice and b) it can be made even faster, I think we can achieve almost the same speed as GCE M (which was much faster previously). I'll provide another update to this pull request for making it faster.

kolja-esders:
Alright, will wait for the update before merging.

kolja-esders:
@michitux Just to double-check before finally merging in the changes. Would you like to add the update you referred to?

Already merged it but reverted for now to wait for your response.

michitux:
All updates are in the commits from May 4th. I have further changes though that depend on this pull request and #19 which is why I haven't pushed them yet.

PR #26, opened 2017-05-01
Rename LFRGenerator to ReCoNGenerator

kolja-esders:
Will help readers of 'Generating realistic scaled complex networks' from Staudt et al. find the corresponding generator.


michitux:
The LFR generator should not be renamed as this really is an implementation of the LFR generator. Instead, a separate, new class that also directly takes an existing graph to replicate should be added IMHO.

hmeyerhenke:
I forgot to ask Kolja to coordinate with Michael... Michael is right in that the original LFR name should not go away completely. After all, LFR is way more etablished and an asset to NetworKit, in particular after Michael has tuned the code. Apparently both LFR and ReCon were realized in the same class, which I did not remember.

Michael, as one of the code authors, how would you approach this concretely?

Also, for each publication that contributes code to NetworKit we should have in the (near) future a Jupyter notebook with some explanation on how to find and use the corresponding code. The ReCon paper is a good starting point since -- as of now -- it is really difficult to find and use it. Kolja, could you please prepare one with Michael's help?

PR #28, opened 2017-05-04
ApproxBetweeness2 deprecation

### LINKED ISSUE #27: Profiling module broken due to renaming?
I pulled the latest Dev branch this morning, compiled, and started a new local Jupyter NB with the user guide. The user guide does not perform profiling successfully due to the missing ApproxBetweenness2.

Aren't there wrapper functions to not run into this issue? We should have that at least for one release together with a "deprecated" statement.

kolja-esders:
Resolves #27.


ebergamini:
@kolja-esders Since ApproxBetweenness2 used to be what now is called EstimateBetweenness, I think you should write "DEPRECATED: Please use EstimateBetweeness instead", right?

kolja-esders:
Which part are you referring to exactly? I believe the deprecation notice is always referring to EstimateBetweenness.

PR #42, opened 2017-06-12
Signed 0 default in Octree

timostrunk:
GCC 7 gives a warning about a signed bit-shift in Octree.h, which (in my opinion) poses a possible bug:

0 is signed by default, therefore ~0 is not (signed 0) and implementation specific, first comment here:
https://stackoverflow.com/a/2999588/2419265

The intended value here, if I understand correctly, is UINT_MAX or 0xffffffff, therefore it should be 0U, however in the bitness of index: therefore static_cast<index>(0). Thoroughly check this commit before merging.


kolja-esders:
LGTM. @michaelwegner?

PR #51, opened 2017-07-25
Deprecate SSSP::getStack and replace with getDistanceVector

### LINKED ISSUE #50: Rename SSSP::getStack to reflect current functionality
`SSSP::getStack` is currently not returning a stack but a **vector** of nodes in increasing distance from the source.

`getStack` should be deprecated and instead there should be a `getDistanceVector` method that has the same functionality.

kolja-esders:
Resolves #50.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -1198,6 +1198,7 @@ cdef extern from "cpp/distance/SSSP.h":
 		vector[node] getPath(node t, bool forward) except +
 		set[vector[node]] getPaths(node t, bool forward) except +
 		vector[node] getStack(bool moveOut) except +
+		vector[node] getDistanceVector(bool moveOut) except +
 		double _numberOfPaths(node t) except +
 
 cdef class SSSP(Algorithm):
@@ -1253,7 +1254,38 @@ cdef class SSSP(Algorithm):
 		return result
 
 	def getStack(self, moveOut=True):
+		""" DEPRECATED: Use getDistanceVector instead.
+		
+		Returns a vector of nodes ordered in increasing distance from the source
+
+		Parameters
+		----------
+		moveOut : bool
+			If set to true, the container will be moved out of the class instead of copying it; default=true.
+
+		Returns
+		-------
+		vector
+			Nodes ordered in increasing distance from the source
+		"""
+		from warnings import warn
+		warn("getStack is deprecated; use getDistanceVector instead.", DeprecationWarning)
 		return (<_SSSP*>(self._this)).getStack(moveOut)
+
+	def getDistanceVector(self, moveOut=True):
+		""" Returns a vector of nodes ordered in increasing distance from the source
+
+		Parameters
+		----------
+		moveOut : bool
+			If set to true, the container will be moved out of the class instead of copying it; default=true.
+
+		Returns
+		-------
+		vector
+			Nodes ordered in increasing distance from the source
+		"""
+		return (<_SSSP*>(self._this)).getDistanceVector(moveOut)
 
 	def numberOfPaths(self, t):
 		return (<_SSSP*>(self._this))._numberOfPaths(t)
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -1258,6 +1258,9 @@ cdef class SSSP(Algorithm):
 		
 		Returns a vector of nodes ordered in increasing distance from the source.
 
+		For this functionality to be available, storeNodesSortedByDistance has to be set to true in the constructor.
+		There are no guarantees regarding the order of two nodes that have the same distance to the source.
+
 		Parameters
 		----------
 		moveOut : bool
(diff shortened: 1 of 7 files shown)


kolja-esders:
I think it should still be discussed whether having a `getDistanceVector` and a `getDistances` method might lead to confusion for the user.

Just looking at the names it is hard to grasp what the difference between the methods is supposed to be.

If we keep the naming, the corresponding parameter of the constructor (currently `storeStack=false`) should probably also change.

kolja-esders:
Upon further adjusting the naming, I got a chance to look at the actual algorithm and realized that the functionality of storing the stack/vector is actually not implemented.

As a consequence, we should discuss whether we want to get rid of the `storeStack` functionality or whether we would like to implement it.

rionda:
@kolja-esders : I got an email saying you add a comment about the implementation of the functionality of storing the vector being missing. The implementation is in the classes that inherit from this one (e.g., Dijkstra and BFS).

networkit/_NetworKit.pyx
```diff
@@ -1253,8 +1254,39 @@ cdef class SSSP(Algorithm):
 		return result
 
 	def getStack(self, moveOut=True):
+		""" DEPRECATED: Use getDistanceVector instead.
+		
+		Returns a vector of nodes ordered in increasing distance from the source
```
rionda: Please add a comment about how ties are broken (likely arbitrarily?)

networkit/cpp/distance/SSSP.cpp
```diff
@@ -76,12 +76,15 @@ std::set<std::vector<node>> SSSP::getPaths(node t, bool forward) const {
...
 }
 
-
-std::vector<node> SSSP::getStack(bool moveOut) {
+std::vector<node> SSSP::getDistanceVector(bool moveOut) {
 	if (!storeStack) {
 		throw std::runtime_error("stack has not been stored");
 	}
-	return (moveOut)?std::move(stack):stack;
+	return moveOut ? std::move(stack) : stack;
```
rionda: What happens if I call sssp.getDistanceVector(true) and then sssp.getDistanceVector()? 

Should an exception been thrown if the vector has been moved out and the function is called again (with or without "true" as the parameter)

networkit/cpp/distance/SSSP.h
```diff
@@ -98,13 +97,22 @@ class SSSP: public Algorithm {
 	bigfloat getNumberOfPaths(node t) const;
 
 	/**
-	* Returns a stack of nodes ordered in decreasing distance from the source
+	* Returns a vector of nodes ordered in increasing distance from the source
 	*
 	* @param moveOut If set to true, the container will be moved out of the class instead of copying it; default=true.
-	* @return stack of nodes
+	* @return vector of nodes ordered in increasing distance from the source
```
rionda: Please add a comment about how ties are broken (likely arbitrarily?)

networkit/cpp/distance/SSSP.h
```diff
@@ -98,13 +97,22 @@ class SSSP: public Algorithm {
 	bigfloat getNumberOfPaths(node t) const;
 
 	/**
-	* Returns a stack of nodes ordered in decreasing distance from the source
+	* Returns a vector of nodes ordered in increasing distance from the source
```
rionda: Please add a comment about how ties are broken (likely arbitrarily?)

networkit/cpp/distance/SSSP.cpp
kolja-esders: Good point. Thoughts on newest changes there?

networkit/_NetworKit.pyx
```diff
@@ -1258,6 +1258,9 @@ cdef class SSSP(Algorithm):
 		
 		Returns a vector of nodes ordered in increasing distance from the source.
 
+		For this functionality to be available, storeNodesSortedByDistance has to be set to true in the constructor.
+		There are no guarantees regarding the order of two nodes that have the same distance to the source.
```
rionda: Small edits: replace "order" with "relative ordering" and "that have" with "with". Here and elsewhere.

networkit/cpp/distance/SSSP.cpp
```diff
@@ -76,12 +74,23 @@ std::set<std::vector<node>> SSSP::getPaths(node t, bool forward) const {
...
+std::vector<node> SSSP::getNodesSortedByDistance(bool moveOut) {
+	if (!storeNodesSortedByDistance) {
+		throw std::runtime_error("Nodes sorted by distance have not been stored. Set storeNodesSortedByDistance in the constructor to true to enable this behaviour.");
+	} else if (nodesSortedByDistance.empty()) {
+		throw std::runtime_error("The container has already been moved or run() has not been called yet. Please call run() first.");
+	}
+	if (moveOut) {
+		auto movedNodes = std::move(nodesSortedByDistance);
+		nodesSortedByDistance.clear();
+		return movedNodes;
```
rionda: Are you sure this works, in the sense that it is actually returning a rval reference to a non-empty vector?

kolja-esders: I double-checked and this should correctly return a rval ref.

PR #52, opened 2017-07-26
Shared library build in SCons

### LINKED ISSUE #48: NetworKit as a shared library
Right now it is only possible to build a **static library** by executing `scons --target=Lib --optimize=Opt`.

Add another another target (e.g. `Shared` or `SharedLib`) which ensures NetworKit will be build as a shared library.

kolja-esders:
Resolves #48.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 .DS_Store
 *.pyc
 *.a
+*.dylib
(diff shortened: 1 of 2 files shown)


kolja-esders:
Sure thing, documentation will be updated on the website.

PR #54, opened 2017-07-28
(Re-)Add the BTER wrapper class

michitux:
This provides a wrapper class for the BTER implementation FEASTPACK (http://www.sandia.gov/~tgkolda/feastpack/). A previous version of this class was already included in NetworKit but got lost in some merge commit.


bokveizen:
I believe the new version is here https://www.mathsci.ai/feastpack/

Issue #69, opened 2017-09-06
Use of functions deprecated in C++17


rionda:
With C++17 almost here, any use of functions removed (and possibly even deprecated) in this version of the standard should be removed from the tree.

Luckily, the only such function in the NetworKit tree is "std::random_shuffle", which is used in two places, plus one commented out. 

Of the two non-commented-out uses, one is in RandomSpanningTree and on in PseudoRandomSpanningTree. Both these classes are marked as deprecated and should probably be removed. The commented-out use is in NeighborhoodFunctionHeuristic.

Removing the two above classes  *SpanningTree result in successful compilation with --std=c++17 (after 19005a18180d227f3306b3d71c4ca8901b420a5b).

PR #71, opened 2017-09-07
Remove deprecated classes under spanning/.

rionda:
A by-product of this change is that NetworKit can now be compiled with
--std=c++17, so #69 is fixed.


rionda:
@kolja-esders , can you look at this please?

PR #77, opened 2017-10-10
SConstruct: fixes for SCons 3

michitux:
This fixes the indentation for Python 3 and migrates from (Python 2) `ConfigParser` to (Python 3) `configparser`. This allows building with scons 3.0 with Python 3 (yes, this means no more dependency on Python 2). For Python 2 support, installing the `configparser` backport from pip is necessary now. I fixed this for Travis, but probably this should also be documented or automated somewhere?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
           packages:
             - g++-4.8
             - scons
+            - python-configparser
             - libgtest-dev
       install:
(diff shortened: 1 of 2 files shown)


rionda:
I would argue that this pull request should be merge before #67, and then have #67 reworked for Scons3 (should be minor). My 2 cents.

PR #89, opened 2017-12-13
Configparser backwards compatibility

kolja-esders:
Since we're now using the built-in `configparser` module in Python3, there should be a backport for python2.7 users. For now we can simply install the `configparser` package. This way, on python3 `import configparser` will use the built-in parser while on python3 this import will use the installed package.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,7 +1,14 @@
 import os
 import subprocess
 import fnmatch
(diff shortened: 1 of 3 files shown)


kolja-esders:
Doesn't work reliably on all systems. Need to find a better solution.

PR #105, opened 2018-01-29
Feature/fix dijkstra

eug93:


eug93:
Travis failed due to an error in the installation of gcc49 on a MacOS system. All the other tests were successful. Shall we merge anyway?

PR #107, opened 2018-01-30
Travis: Fix custom gcc installation on macOS images

kolja-esders:
Travis updated their macOS images to include `oclint` which will provide a default gcc binary. This leads to failure during installation of a custom gcc binary through Homebrew.

This PR addresses the issue by uninstalling oclint beforehand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,8 @@ matrix:
       before_install:
         - brew update
         - brew tap homebrew/versions
+        - brew cask uninstall --force oclint # See https://github.com/travis-ci/travis-ci/issues/8826 for details.
         - brew install gcc49
-        - brew link --overwrite gcc@4.9
         - brew install scons
         - pip2 install configparser
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,14 +22,14 @@ matrix:
        - make
        - cd -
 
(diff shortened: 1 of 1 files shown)


kolja-esders:
Note:
There is another issue regarding compatibility between gcc < 6.3 and XCode 8.0 which is installed on the new images. Will lead to the following *warning* being repeated **very often**:

```
/var/tmp//ccyf0y8L.s:47334:11: warning: section "__const_coal" is deprecated
        .section __DATA,__const_coal,coalesced
                 ^      ~~~~~~~~~~~~
```

This leads to the log file exceeding 4MB and thus Travis terminating the build assuming that it it erroneous:

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

The issue was addressed in gcc >= 6.3. So the easiest fix for now is to update the gcc version.

If we wanted to explicitly test older versions (GCC 4.9 is '14) this would need to be revisited.

See also:

https://stackoverflow.com/questions/38193043/disable-assembler-warning-section-text-textcoal-nt-coalesced-pure-instruct

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71767

https://github.com/Homebrew/homebrew-core/issues/4812

PR #108, opened 2018-01-31
Fix travis failures due to testTimer

avdgrinten:
testTimer causes Travis CI to fail sporadically.
We cannot rely on the OS' scheduler to wake up the test within a certain time frame. This is especially unreliable on busy systems (like the Travis servers).

Turning this test into a benchmark allows us to still run it manually without causing CI failures.

@kolja-esders Please check if this makes sense.

### DIFF
--- a/networkit/cpp/auxiliary/test/AuxGTest.cpp
+++ b/networkit/cpp/auxiliary/test/AuxGTest.cpp
@@ -117,23 +117,6 @@ TEST_F(AuxGTest, testRandomProbability) {
 }
 
 
-TEST_F(AuxGTest, testTimer) {
-	int64_t sleepTime = 1000; // sleep time in ms
-	int64_t tolerance = 20;
-
-	Aux::Timer timer;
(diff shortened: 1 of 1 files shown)


kolja-esders:
What are your thoughts on completely removing the test/benchmark instead?

Changing it to be a benchmark seems suboptimal to me since it leaves the impression we're *performance* benchmarking something here. Since the implementation is really just a decorator for the chrono clocks there is also not really any advanced logic that needs to be tested.

Might make sense to test for coherence (any inconsistencies between startTime, stopTime, elapsedMilliseconds, elapsedMicroseconds, ...) though.

Also when looking at the usage of the timers in the code, it might make sense to provide a Timer interface and a ManualTimer implementation where we can set the elapsedTime during testing to check if the code using the timer can properly handle certain elapsed times.

avdgrinten:
Thanks for your feedback. I will delete the test completely. Consistency tests can be added in future PRs.

I will merge the PR as soon as the Travis build completes.

PR #129, opened 2018-03-15
Fix/centrality ranking

eug93:


avdgrinten:
I do not like reformatting whole files just for formatting sake. This breaks "git blame" and other useful tools. For now I'm merging this; however, it should be avoided in the future (or discussed in a broader context, i.e. the mailing list).

PR #139, opened 2018-03-29
io/KONECTGraphReader: Fix default constructor

avdgrinten:
The default constructor in the KONECTGraphReader class did not
properly initialize all member variables. Remove the default
constructor and assign default values to the non-default
constructor so that it can be called without arguments.

This commit also cleans up unused variables from the class.


rbange:
Totally agree with this commit.
Calling the existing default constructor (which is and should be a usual procedure for graph readers) results in a segmentation fault or malloc error, which is a super odd behaviour.

PR #142, opened 2018-04-01
Added MahmoodyGroupBetweenness,GTest and updated BucketPQ

Marvin812:
This commit contains the MahmoodyGroupBetweenness-feature and updates BucketPQ with a getKey-Method which is used by MahmoodyGroupBetweenness. GTests for MahmoodyGroupBetweenness are also included.


avdgrinten:
I already did this but Travis does not pick up existing PRs. It should work for new ones though.

eug93:
What about creating a new pull request? It is safer in case something breaks.

PR #146, opened 2018-04-10
MahmoodyGroupBetweenness + Updated BucketPQ

MarvinPogoda:
This commit contains the MahmoodyGroupBetweenness-feature and updates BucketPQ with a getKey-Method which is used by MahmoodyGroupBetweenness. GTests for MahmoodyGroupBetweenness are also included.


avdgrinten:
The networkit/cpp/centrality/test/MahmoodyGroupBetweennessGTest.cpp test fails. Can you debug that or do you know why that happens?

Marvin812:
The test works fine locally. Is it possible, that the fixed seed gives a different result depending on the compiler used in the Travis build?

avdgrinten:
Yes. The order of items that have the same value in the priority queue depends on the environment. Can we make a test that does not depend on that order?

avdgrinten:
Look good.

PR #149, opened 2018-04-12
Prevent index out of range exception while counting isolated nodes

### LINKED ISSUE #148: List index out of range exception in overview(G)
How to reproduce:

```python
G = Graph(10)
overview(G)
```

This will throw an index out of range exception since the node sequence is accessed before checking whether the index is valid while determining the number of isolated nodes.

kolja-esders:
Resolves #148.


avdgrinten:
Good catch. Thank you for the fix.

PR #150, opened 2018-04-16
MaximalCliques: fix callback handling

michitux:
Previously, the callback has always been deleted even if the instantiation succeeded (no idea why it ever worked).


avdgrinten:
Thank you for the fix. Yes, it seems strange that this worked at all.

One comment about the code: Can't we catch something more specific than BaseException?

michitux:
My goal was specifically to catch everything to ensure that no memory is leaked. Note that the exception is raised again so no exception is lost. The constructor currently doesn't throw anything though I suspect it should throw a runtime error if the graph type doesn't fit. Further, the memory allocation might fail.

avdgrinten:
Makes sense. I'm merging this.

PR #154, opened 2018-04-24
CMake

manpen:
*Do not merge, do not read the code*
On the NetworKit day (and on other occasions) there were discussions to migrate from `scons` to the `cmake`. There are obvious benefits from that:
- CMake seems to be omni-present on developer machines (maybe that's a sampling bias); it not its available as a pip module
- CMake supports a great variety of build systems (including make for gcc/clang and MSVC) and IDEs (e.g. CLion, XCode, MSVS)
- No hassle with multiple Python environments ...
- The cool kids use it ;)

The initial commit allows to build NetworKit with tests (and the example); but is in no way complete or tested. Before I continue working on this project, I want to discuss a few points:

- Is this still needed/desired and still an open problem?

- Is `cmake` intended as a replacement for the `scons` build-path or merely an addition. If the latter, what system should be the default. IMHO using a single path (in contrast to the cmake, scons, and mbe paths) would be preferable.

- Currently `scons` seems to be responsible for compiling the Core static lib, which is then linked to the Cyphon code in `setup.py`. Should we keep it this way? It might be a good starting point to not handle the Cyphon compilation in the beginning. This could be added in a second step.

- Which options do we need? Currently the following are provisioned (but not fully supported)

```cmake
option(NETWORKIT_OPENMP "Build with OpenMP support (requires OpenMP)" ON)
option(NETWORKIT_BUILD_STATIC "Build as static library" OFF)
option(NETWORKIT_BUILD_SHARED "Build as shared library" ON)
option(NETWORKIT_BUILD_TESTS "Build with tests (requires GTest)" OFF)
option(NETWORKIT_CXX14 "Build with standard C++14" ON)
```
`NETWORKIT_CXX14` is included as it is part of the `scons` configuration, but I would suggest to query it from the compiler directly.

Additionally, we can use `CMAKE_BUILD_TYPE` to distinguish between Opt, Dbg and Pro.


avdgrinten:
Hi @manpen, thanks for your work. There is interest in a new build system as scons is very slow (running scons when there is nothing to do still takes multiple seconds on a fast machine).

Personally, I would prefer meson over cmake, as meson seems to gain a lot of traction right now (basically the whole Linux desktop is moving from autotools to meson, including systemd, x.org and so on), its dependency management is really good and its syntax is much simpler than the syntax of cmake. Still, cmake would be a great improvement over scons.

However, if we make a change, I think that we should atomically move to the new build system instead of maintaining both scons and cmake at the same time. That means, that a new build system should be able to build the Python package, too.

About the patch itself, I dislike the automated build file generation. This could be good to generate the initial build files, but after that, I would prefer editing them by hand (as that gives us more flexibility regarding the code organization).

manpen:
Hi @avdgrinten,

thank you for your feedback. I did an (non-representative) experiment and looked at the configuration of four different computer systems maintained by different groups. All of them had CMake installed, non meson. Similarly three guys I talked to had used CMake before, none of them has experience in meson. So, if it's fine with you I would prefer staying on the CMake track, as it currently seems to be better supported.

Regarding the automatic generation of files: I remove the warning at the headers and will change the generation script to not override files by default.

avdgrinten:
Thank you for your work. I acknowledge that cmake is (currently) more widespread than meson. As I said, I would also merge a cmake build system (once it is ready). Going from 

What do you think about integrating the Python package build and cmake? To make this work, scons would need to be replaced by cmake in setup.py (but that should not be a lot of work). As I stated earlier, I would prefer to get rid of scons atomically (and not maintain scons for pip + cmake for the libs).

manpen:
I agree, but never did it before. So I will give it a shot and we'll see whether it's practical. Just a warning: It may take a couple of days.

manpen:
As I'm not an CMake expert, it took some time to reply, but I think, I now have a quite clear picture of how to approach the issue. Please let me know if you consider it worth-while:

There seems to be no cmake-ish way of performing source-to-source compilations (yet) which is an issue for the pyc -> cpp translation. There exists packages (e.g., [scikit-build](https://github.com/scikit-build/scikit-build)) to ease this step but this adds dependencies.

To get us started, I thus suggest the following approach:
- Compile all CPP parts using CMake. This also includes a build target for the Python shared-library, but requires a previous run of Cython (either invoked by setup.py or shipped with the source archive). Installation of NetworKit as a CPP library and CPP unit tests are also managed by CMake.

- Preparation and installation of the Python module remains part of the `setup.py`. So we basically replace the invocation of `scons` with `cmake`. It seems awkward to replicate (and maintain) setup_utils infrastructure. Also I suggest to drop the `mbe` path, as it's not too much to ask to have `cmake` installed (especially if we can automatically install it with `pip`).

In contrast to my original proposal, I would like to try to make the internal dependency structure of the NetworKIT's CPP parts more explicit in CMake. It seems pragmatic to have a module per directory, but in future we could make it more fine-grain. This should not affect the resulting binary library, but allows us to compile the unit test individually. Also it anticipate some of the work required to modularize of Python-modules as proposed by @michitux.

Hence we would end up with a couple of binaries for the unit tests at the advantage of not having to link the whole NetworKit library each time, we make a small change to an isolated algorithm. This should speed-up the development/optimization cycle of new algorithms.

Using `CTest` we can still run all tests with a single command invocation.

manpen:
Just to give you some update:

While the CMake build infrastructure is not integrated into setup.py yet, it can build NetworKit as a static or shared library and generate a single GTest file as before.

If `NETWORKIT_MONOLITH=OFF` is set, each directory (an arbitrary granularity that I selected, see below) is linked into its own static library (e.g. `networkit_io`) and each Unit-Tests CPP is compiled into its own executable. This requires explicit marking of internal dependencies. I consider this a feature which is also easy to maintain. As an example the declaration of the module `networkit_graph` (containing all files in the directory `graph`) looks as follows:

```cmake
networkit_add_module(graph
    Graph.cpp
    GraphBuilder.cpp
    GraphTools.cpp
    KruskalMSF.cpp
    RandomMaximumSpanningForest.cpp
    RandomSpanningForest.cpp
    Sampling.cpp
    SpanningForest.cpp
    UnionMaximumSpanningForest.cpp
    )

networkit_module_link_libraries(graph
    PRIVATE OpenMP::OpenMP_CXX)

networkit_module_link_modules(graph
    auxiliary base components structures)
```

Unit-Tests may have dependencies exceed those of the module they are testing. This is also explicitly marked, e.g.

```cmake
networkit_add_test(graph SpanningGTest io)
```

In case the project is build monolithically (which is the default), the macros above extract the required information (such as source files) and discard the dependency annotations. Hence, it's not necessary to explicitly maintain two build path.

Generating modules at the granularity of directories does not seem to be a good choice, since it cyclic dependencies which disallows generating shared libraries for the modules.The dependency graph for the Unit-Test mention above can be found here: https://raw.githubusercontent.com/manpen/networkit/cmake_infos/Spanning.png

This is easily fixable by further splitting modules. It however seems to me that this is not a blocking issue since it can be resolved later.

I'm currently testing the scheme while porting our Curveball Implementation.

Until now it works just fine: For instance, CLion interacts well with NetworKit and one can select the Unit-Test one wishes to build and execute directly. Rebuilding  and running my (small) Unit Test takes well below 5s.

avdgrinten:
I pulled the PR into the `cmake` branch. @rbange will rewrite the `setup.py` to call cmake instead of scons. He will also integrate Cython properly into this new build process.

rbange:
Hi @manpen,

I just tried working on the cmake branch in order to adapt the setup.py file and encountered multiple problems. 

1. The CXX compiler identification does not work as expected on MacOS. MacOS doesn't not ship with a GNU compiler, but a ~~shi*y~~, eh not so well `Apple LLVM version 9.1.0 (clang-902.0.39.2)`. This compiler does not seem to support OpenMP and therefore a GNU compiler has to be installed before compiling NetworKit - so far so good. The problem is that CMake prefers this compiler instead of e.g. `g++-8` and then fails when checking for OpenMP support.
When running: `cmake ..` (which should be the preferred command for beginners who compile networkit for the first time)
```
-- The CXX compiler identification is AppleClang 9.1.0.9020039
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
Use Release Build Type as default
CMake Error at /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindOpenMP.cmake:442 (find_package_handle_standard_args)
  CMakeLists.txt:20 (find_package)
```
Of course this problem can be mitigated by specifying the used compiler like `CXX=g++-8 cmake ..`, but that might not be a suitable solution in the setup.py, because we either have to add a new parameter for this (then setup.py does not work out of the box for macOS) or stick with the build.conf (we really want to get rid of this). Is there a way to optimise the compiler identification in order to solve the described problem?

2. How does cmake try to find gtest libraries (previously specified in build.conf)? Does it look into expected paths e.g. /usr/lib/? Is there a way to specify a custom path?

3. At least for the latest branch status, `CXX=g++-8 cmake -NETWORKIT_BUILD_TESTS=ON ..` does not generate a make file which produces an executable like before, instead I get the normal library without any errors. Maybe its because my gtest-lib was not found? At least I don't get any errors which indicate that cmake is not able to to build the test executable.

I don't think I only speaking for myself if I say that we are not experts into the cmake topic. Would your mind give us a brief description on how to use your implementation? (as long as we don't have an usage documentation or special --help command)

Thank you!

manpen:
Hi, 

first of all, thank you @rbange for taking over the integration part. I'm really happy to not go through it ;) It might however be interesting to have a fork to which we both have write access, which could make it a little bit easier to integrate missing features (or you take over completely, which is also fine with me ;) ). For instance, I do not consider the modularization done, there are missing parameters, and also the is no packaging support yet. 

1.) As far as I know, specifying `cmake -DCMAKE_CXX_COMPILER=g++` ... or `CXX=g++ cmake ..` is *the* way to deal with this kind of situations. There's a system-wide default for a reason, and it might be a good thing to be explicit about deviating from that. I suggest to add a warning including a pointer to a solution in case a user uses a non-supported compiler.

2.) @avdgrinten, @eug93 and me discussed on Thursday that `gtest` (and TLX) should be added as a git submodules, e.g. have a path `extlibs/gtest`. Then NetworKit is independent of the system's version (which is typically missing on vanilla machines). In this case importing `gtest` boils down to an `add_subdirectory` statement. 

3.) As also discussed on our meeting, the project is not yet done; there are parameters missing and many cases are untested. The only build path I was interested in, is building `-DNETWORKIT_MONOLITH=OFF -DNETWORKIT_BUILD_TESTS=ON` to get small unit tests for a single algorithm only. 

BTW: I suggest using `ninja` rather than `make` (if available) as this far more efficient such large projects and we do not have to worry figuring out the right number of CPUs to compile with.

Since I have two deadlines next week, I won't be able to do anything next week. Maybe we should have a Skype meeting in the first week of July to discuss what's left to do and how we do it??

PR #156, opened 2018-04-26
Implement equality and hashing for graph events

michitux:
This makes it possible to put GraphEvent instances into a set in order to e.g. quickly calculate the union or intersection of two sets of graph events.

Note that modifying a GraphEvent object that is stored in a set or used as key in a dict is undefined behavior.

Note that I have modified the Cython import of the GraphEvent type enum as otherwise it did not compile for me.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -7366,8 +7366,8 @@ cdef class LaplacianCentrality(Centrality):
 
 # Module: dynamic
 
-cdef extern from "cpp/dynamics/GraphEvent.h":
-	enum _GraphEventType "NetworKit::GraphEvent::Type":
+cdef extern from "cpp/dynamics/GraphEvent.h" namespace "NetworKit::GraphEvent::Type":
+	cdef enum _GraphEventType "NetworKit::GraphEvent::Type":
 		NODE_ADDITION,
 		NODE_REMOVAL,
 		NODE_RESTORATION,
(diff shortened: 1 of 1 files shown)


avdgrinten:
Thanks for the patch. Regarding the UB: Would it be a good idea to make GraphEvents immutable in the Python interface? Of course, they are mutable in C++ (as that fits the C++ programming model better) but I think it might be beneficial to make them immutable and avoid the UB here. On the other hand, that is a breaking change and we should avoid that in the short term.

michitux:
I think making graph events immutable would make sense. In the meantime, would you be more comfortable with only implementing equality? According to the [Python documentation](https://docs.python.org/3.6/reference/datamodel.html#object.__hash__), implementing equality but not hashing is the recommended approach for mutable objects. Note that this explicitly prevents adding items to a set or using them as keys of a dict. Without this change, there are default equality and hash implementations that basically use the pointer which is not very intuitive.

If you would be more happy with just equality, I can update this PR to implement just equality (which would still be much better than the current state imho).

avdgrinten:
I guess that depends on your requirements. If you use/need hashing right now, I'd be happy to merge the PR as-is (with the warning to users that objects that are used as keys in collections must not be modified). Otherwise, only implementing equality makes sense.

michitux:
I am using that code (as I have submitted it) to evaluate a set of edge editing solutions. However, we are talking about small solutions with at maximum 70 edits in each solution so using a "brute force" intersection/union method would be totally fine I think so I could easily work around this.

I assume that for graph events implementing hashing could be useful for more applications than just my evaluation but I'm not sure how useful it would really be. If you intend to make graph events immutable anyway I would suggest to deprecate writing to these attributes now and add hashing support.

avdgrinten:
Closed in favor of #268.

PR #157, opened 2018-04-26
Implement equality and hashing of partitions

michitux:
This allows to quickly check partitions on equality. Further, it allows to store partitions in a Python set. This allows to quickly compute the number of different partitions in a large set of partitions.

Note that modifying partitions that are stored in a set or even used as keys in a dict leads to undefined behavior. A cleaner way to implement this would thus be to have some kind of "frozen" partition but I'm not sure how to implement that or if it is woth the effort.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -3767,6 +3767,65 @@ cdef class Partition:
 		"""
 		return self._this.getSubsetIds()
 
+	def __eq__(self, Partition other not None):
+		""" Compare self to other partition.
+
+		Equality is independent of the used partition
+		ids. This tries to construct a mapping between the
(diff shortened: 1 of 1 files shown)


michitux:
Similar to #156, I guess a safer approach would be to just implement equality and not hashing. This would still simplify comparing partitions a lot without causing any undefined or unexpected behavior. The only question would then be if you agree with my definition of equality, which is the same as equality in terms of one of the difference measures that are implemented in NetworKit but which is not equality in terms of the actual partition ids.

avdgrinten:
The only problem that I can see here is that the equality function is quite slow. OTOH caching its result is not really feasible as the C++ Partition object can change.

avdgrinten:
Closing in favor of #267.

PR #158, opened 2018-05-03
Add algorithm to compute the difference of two graphs

michitux:
The GraphDifference algorithm computes a series of graph events that transform one given graph into another given graph. This makes it possible to construct a series of events from a given series of graphs.

Further, this marks `GraphEvent::toString()` as `const` and marks event stream passed to `GraphUpdater` as `const`.


avdgrinten:
Looks good. Very nice addition to NetworKit. I'm merging this.

PR #159, opened 2018-05-08
Faster Partition::compact() implementation

michitux:
This uses parallel sorting instead of a `std::map`. From a theoretical perspective, the total work is unchanged but while previously the sequential part was O(n log(n)), it is now only O(n) - a single sequential scan to determine unique ids.

`Partition::compact()` is the main bottleneck when comparing partitions using e.g. normalized mutual information or the adjusted rand index.

Apart from this optimization I also tried to simply replace the  `std::map` by a `std::unordered_map`. While this causes less total work than the implementation proposed in this PR, this implementation is faster  than the `std::unordered_map` variant when using multiple cores even on my laptop.

### DIFF
--- a/networkit/cpp/structures/Partition.cpp
+++ b/networkit/cpp/structures/Partition.cpp
@@ -6,6 +6,7 @@
  */
 
 #include "Partition.h"
+#include "../auxiliary/Parallel.h"
 #include <atomic>
 
 namespace NetworKit {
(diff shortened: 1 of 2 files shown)


michitux:
Example benchmark on a system 16 cores + hyperthreading on a graph with 128 million nodes comparing a partition with 58886 clusters to a partition with 58887 clusters (that are largely identical) using the adjusted rand measure needs 26.48 seconds with the old code while it only needs 3.83 seconds with the proposed change.

PR #160, opened 2018-05-08
RFC: Faster graph readers for NetworKit

avdgrinten:
Many of NetworKit's graph readers are quite slow. Reading very large graphs (billions of edges) can take hours of time. The reason for that is not so much the inherent I/O cost (which is also quite high) but the inefficient implementation of some graph readers in NetworKit.

For example, take a look at the SNAP and KONECT readers. The SNAP reader processes the entire input twice: Once to map SNAP node IDs to NetworKit IDs and once to retrieve the actual edges. During both passes, every line is loaded into a std::string and that string is split into a std::vector<std::string>. This approach requires at least 4 allocations *per line*. The KONECT reader is structured similarly. This PR (which is not intended to be merged as-is) tries to improve upon this situation.

Instead of using string classes to read lines, we mmap() the entire file at once and then only operate on characters, without requiring any allocations. Instead of performing two passes, we allocate node IDs on-the-fly.

### DIFF
--- a/networkit/cpp/io/FastSNAPGraphReader.cpp
+++ b/networkit/cpp/io/FastSNAPGraphReader.cpp
@@ -1,0 +1,109 @@
+/*
+ * FastSNAPGraphReader.cpp
+ *
+ *  Created on: 04.05.2018
+ *      Author: Alexander van der Grinten
+ */
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "FastSNAPGraphReader.h"
+#include "../auxiliary/StringTools.h"
+#include "../auxiliary/Log.h"
+
+namespace NetworKit {
+
+Graph FastSNAPGraphReader::read(const std::string &path) {
+	Graph G;
+
+	// Maps SNAP node IDs to consecutive NetworKit node IDs.
+	auto mapNode = [&] (node in) -> node {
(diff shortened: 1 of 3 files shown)


networkit/cpp/io/FastSNAPGraphReader.cpp
```diff
@@ -0,0 +1,109 @@
...
+		if(it != nodeIdMap.end())
+			return it->second;
+		auto result = nodeIdMap.insert({in, G.addNode()});
+		assert(result.second);
+		return result.first->second;
+	};
+
+	// This function modifies the graph on input.
+	auto handleEdge = [&] (node source, node target) {
+		if(!G.hasEdge(source, target))
```
michitux: This check can also be quite slow as it is a linear scan over all neighbors of that node. If memory does not matter, it is probably faster to store the edges in one large vector of node pairs and then sort this vector twice using bucket sort (or just parallel sorting). The graph could then be created using that vector.

Just a small but still significant improvement for undirected graphs is to execute hasEdge on the node with the smaller degree.

rbange: Unfortunately I think memory does matter regarding I/O. Especially if the library is used on consumer machines with low RAM. 

Nevertheless the second idea can be very useful. Is there a reason this improvement is not already implemented in the ```Graph.cpp```?

avdgrinten:
Replaced by #168 and #169.

PR #161, opened 2018-05-08
Add various binary partition and graph readers and writers

michitux:
This adds some readers and writers for binary graph and partition files. Basically, this is the code we used for the evaluation of our paper [Distributed Graph Clustering using Modularity and Map Equation](https://arxiv.org/abs/1710.09605) (code available at [kit-algo/distributed_clustering_thrill](https://github.com/kit-algo/distributed_clustering_thrill)). The [external memory LFR implementation](https://github.com/massive-graphs/extmem-lfr) supports the same formats, too.

The binary partition formats are straightforward generalizations of the standard partition formats, just using binary integers. In [Thrill](http://project-thrill.org/), this corresponds to a DIA of integers or pairs of integers. The graph format consists of one entry per node. Each node begins with its degree as variant and then the neighbors. The graph is undirected but neighbors are only stored in one direction. In Thrill this corresponds to a DIA of a `std::vector` of node ids. Due to this correspondence, unfortunately, no further metadata is supported. Currently, for graphs, only 32-bit node ids are implemented. As Thrill writes these files in parallel on multiple hosts, more than one input file is supported.

I can understand if you do not want to include these quite custom formats in NetworKit. Then, I would just continue maintaining a fork with them (and transform that into a separate module that can be built against NetworKit when separate Cython modules will be supported in NetworKit).

### DIFF
--- a/networkit/cpp/io/ThrillGraphBinaryReader.cpp
+++ b/networkit/cpp/io/ThrillGraphBinaryReader.cpp
@@ -48,12 +48,12 @@ NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::vector<std:
 
 	if (!paths.empty()) {
 		std::ifstream is;
-		count _file_index = 0;
+		count file_index = 0;
 
 		auto next_input = [&]() {
 			is.close();
-			if (_file_index < paths.size()) {
-				is.open(paths[_file_index++]);
+			if (file_index < paths.size()) {
+				is.open(paths[file_index++]);
 			}
 		};
 
--- a/networkit/cpp/io/ThrillGraphBinaryReader.cpp
+++ b/networkit/cpp/io/ThrillGraphBinaryReader.cpp
@@ -1,5 +1,6 @@
 #include "ThrillGraphBinaryReader.h"
 #include "../graph/GraphBuilder.h"
+#include <algorithm>
 
 NetworKit::ThrillGraphBinaryReader::ThrillGraphBinaryReader(count n) : n(n) {};
 
@@ -59,6 +60,8 @@ NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::vector<std:
 
 		next_input();
 
+		node max_id = 0;
+
 		for (node u = 0; is.good() && is.is_open(); ++u) {
 			// Add node if it does not exist yet, only one may be missing
 			if (u >= gb.upperNodeIdBound()) {
@@ -71,6 +74,8 @@ NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::vector<std:
 					throw std::runtime_error("I/O error while reading next neighbor");
 				}
 
+				max_id = std::max<node>(max_id, v);
+
 				gb.addHalfEdge(u, v);
 			}
 
@@ -78,7 +83,11 @@ NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::vector<std:
 				next_input();
 			}
 		}
+
+		if (max_id >= gb.upperNodeIdBound()) {
+			throw std::runtime_error("Maximum read node id larger than number of nodes read.");
+		}
 	}
-	
+
 	return gb.toGraph(true, true);
 };
--- a/networkit/cpp/io/ThrillGraphBinaryReader.cpp
+++ b/networkit/cpp/io/ThrillGraphBinaryReader.cpp
@@ -1,0 +1,84 @@
+#include "ThrillGraphBinaryReader.h"
+#include "../graph/GraphBuilder.h"
+
+NetworKit::ThrillGraphBinaryReader::ThrillGraphBinaryReader(count n) : n(n) {};
+
+namespace {
+	template <typename stream_t>
+	uint64_t GetVarint(stream_t &is) {
+		auto get_byte = [&is]() -> uint8_t {
+			uint8_t result;
+			is.read(reinterpret_cast<char*>(&result), 1);
+			return result;
+		};
+		uint64_t u, v = get_byte();
+		if (!(v & 0x80)) return v;
+		v &= 0x7F;
+		u = get_byte(), v |= (u & 0x7F) << 7;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 14;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 21;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 28;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 35;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 42;
+		if (!(u & 0x80)) return v;
+		u = get_byte(), v |= (u & 0x7F) << 49;
+		if (!(u & 0x80)) return v;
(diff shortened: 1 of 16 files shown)


avdgrinten:
In principle, I see no problem with supporting custom formats - as long as they are maintained. As you're a regular contributor, I do not mind merging them. Do you think this is ready to be merged into the new NetworKit release?

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
```diff
@@ -0,0 +1,115 @@
+#include "ThrillGraphBinaryWriter.h"
+
+namespace {
+    //! Append a varint to the writer.
+template <typename stream>
+stream& PutVarint(stream& s, uint64_t v) {
```
avdgrinten: Uh. Manually unrolling this loop certainly is ugly. Are compilers not smart enough to optimize the following?
```
if(!v)
   s << uint8_t(0);
while(v) {
    auto u = v & 0x7F;
    v >>= 7;
    s << uint8_t(u | (v ? 0x80 : 0));
}
```

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
```diff
@@ -0,0 +1,115 @@
...
+					if (u <= v) {
+						neighbors.push_back(v);
+					}
+			});
+
+			PutVarint(out_stream, neighbors.size());
+			
+			for (uint32_t v : neighbors) {
+				static_assert(std::is_same<uint32_t, decltype(v)>::value, "Node type is not uint32 anymore, adjust code!");
+				out_stream.write(reinterpret_cast<const char*>(&v), 4);
```
avdgrinten: Okay, we write here in host byte order which means that the readers (that I commented on) work correctly on any machine. However, transferring the files from little endian to big endian or vice-versa is not possible. Is that intended?

avdgrinten:
By the way, did you examine the performance difference between the text and binary edge list readers? I would expect the binary reader to be considerably faster.

networkit/cpp/io/ThrillGraphBinaryReader.cpp
```diff
@@ -0,0 +1,84 @@
...
+NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::string &path) {
+	return read(std::vector<std::string>(1, path));
+};
+
+NetworKit::Graph NetworKit::ThrillGraphBinaryReader::read(const std::vector<std::string> &paths) {
+	GraphBuilder gb(n);
+
+	if (!paths.empty()) {
+		std::ifstream is;
+		count _file_index = 0;
```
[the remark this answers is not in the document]
michitux: Fixed in c4598e076c520335cf2e28a9ab9f32f0e3b62494.

networkit/cpp/io/ThrillGraphBinaryReader.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+		for (node u = 0; is.good() && is.is_open(); ++u) {
+			// Add node if it does not exist yet, only one may be missing
+			if (u >= gb.upperNodeIdBound()) {
+				gb.addNode();
+			}
+
+			for (count deg = GetVarint(is); deg > 0 && is.good(); --deg) {
+				uint32_t v;
+				if (!is.read(reinterpret_cast<char*>(&v), 4)) {
```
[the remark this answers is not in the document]
michitux: Fixed in 8c7963507ced841458708fc9dd8912946a919de0.

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
michitux: Fixed in a7ff6f55733400afb172d0a24669715cb7fea8d2.

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
```diff
@@ -103,11 +16,25 @@ void NetworKit::ThrillGraphBinaryWriter::write( const NetworKit::Graph &G, const
...
+				out_stream << uint8_t(u | (deg ? 0x80 : 0));
+			}
+
 			for (uint32_t v : neighbors) {
-				static_assert(std::is_same<uint32_t, decltype(v)>::value, "Node type is not uint32 anymore, adjust code!");
-				out_stream.write(reinterpret_cast<const char*>(&v), 4);
+				// write neighbor as little endian
+				for (size_t i = 0; i < sizeof(uint32_t); ++i) {
+					out_stream << uint8_t(v);
+					v >>= 7;
```
avdgrinten: This seems to be a typo. While we want to shift only 7 bits for varints, we're throwing away one bit of each byte here.

avdgrinten:
Looks good. Do you think this is ready for merge, assuming CI tests pass?

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
michitux: I was aware of that, but as far as I understand the Thrill code, it is doing the same (i.e., uses the host's endianness). Nevertheless I think it would be a good idea to fix the endianness, so a7ff6f55733400afb172d0a24669715cb7fea8d2 now reads individual bytes. I chose little endian because it is what will be used on like most current compute clusters where you execute Thrill or NetworKit and is thus also the format my current graph files use.

networkit/cpp/io/ThrillGraphBinaryWriter.cpp
michitux: Yes, this was a typo which is already fixed in 8ba00589be305a5a514ef68bfb10a3db047767ba.

michitux:
I haven't fixed anything for the partition writers/readers yet, only the graph readers/writers. My suggestion would be to fix them first, too. Further, I have not done any tests on the performance impact of these changes. I would hope of course that compilers are able to optimize a lot but I have not checked anything. Therefore, I would suggest to still wait a bit until this is done. I will try to do this as soon as possible (i.e., ideally still today) but I cannot guarantee anything at the moment.

avdgrinten:
That is fine - we can wait until you are ready. We're planning to do a NetworKit release on next Monday so if you want this work to be in the release, it would make sense to have a completed version by then.

michitux:
As far as I can see I have made all readers and writers independent of the endianness now. Further, I have added a `BinaryPartitionWriter` to complement the `BinaryPartitionReader`. I have also added at least some basic test cases that managed to catch the bug that the `BinaryEdgeListPartitionWriter` did not obey the `firstNode` parameter (which is fixed now). I have also removed all templates. In various places I have also added additional sanity checks. A logical next step would be to use memory mapping to improve performance but from my side this is good to be merged as it is at the moment.

Some tiny first benchmark results on a partition with 256 million nodes:
* `PartitionReader`: 19.1 seconds
* `BinaryPartitionReader`: 8.32 seconds
* `BinaryEdgeListPartitionReader` with randomized node order: 30.5 seconds, with sequential node order: 19.1 seconds.

I couldn't test `EdgeListPartitionReader` as there is no corresponding writer class from what I could see.

These numbers shouldn't be totally trusted as they come from my laptop with local SSD with transparent zstd encryption which leads to less than 100% CPU usage for NetworKit in some cases. Still I think this suggests that there is room for improvement as e.g. the (node, partition)-pair version (`BinaryEdgeListPartitionReader`, maybe we should rename that) shouldn't take twice as long I think.

avdgrinten:
Thank you. As all comments have been addressed and the performance seems to be very good, I'm merging this now.

PR #164, opened 2018-05-23
Cleanup of input folder

rbange:
As discussed, this PR removes unused graph files and thereby decreases the size of the input folder from 83MB to 16.5 MB.

In addition the remaining graphs are now fairly documented, which allows new users of the NetworKit to get started more quickly.

**Nevertheless** we might consider to further decrease the git size. I used this [post](http://stevelorek.com/how-to-shrink-a-git-repository.html) to determine the top sized files in the git and found out, that the top 3 files are already removed graphs. They make up 138 MB of space which is nearly the half of the whole git. Im aware that this is an unusual procedure and one should think some times about it, but I still wanted to bring up this point


rionda:
Also, it may be worth to tell users to do a shallow clone, unless they really need the history.

avdgrinten:
Yes, use shallow clones if you need to save space. We will NOT do a filter-branch as that will rewrite the entire history, change SHA identifiers and break everyones local clones.

It is nice that this commit reduced the size of NetworKit's working tree. Shrinking the disk usage of the repository is not necessary as everybody has enough disk space anyway. Sometime in the far future it might be possible to get rid of "early" commits but for now we're not going to do that.

PR #167, opened 2018-06-01
faster hasEdge() for undirected graphs

rbange:
As proposed by @michitux here #160, i implemented a more efficient hasEdge function. There is no reason this feature should be limited to Graph Readers alone so i applied the changes to the Graph.cpp itself.

Nevertheless, the changes do not really speed up the performance of graph readers, but are efficient in other use cases - especially under power law conditions. For example the LFRGenerator uses the hasEdge method very extensively. I benchmarked the generation of a graph with power law degrees and recognized performance increases of around 10%.

### DIFF
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -745,8 +745,11 @@ void Graph::swapEdge(node s1, node t1, node s2, node t2) {
 bool Graph::hasEdge(node u, node v) const {
 	if (!directed && outDeg[u] > outDeg[v]){
 		return indexInOutEdgeArray(v, u) != none;
+	}else if (directed && outDeg[u] > inDeg[v]){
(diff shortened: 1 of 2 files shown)


rbange:
BTW the build works. Travis fails in its last attempt by at downloading ( `brew install llvm37`) for whatever reason? Is there a way to restart Travis?

kolja-esders:
I manually restarted the job. Seems to be a temporary network issue.

michitux:
This could also be implemented for directed graphs by using the in-degree and `indexInInEdgeArray`.

rbange:
I conducted the changes and made use of `indexInInArray`, but then the test suite failed.
After conducting several tests on my own, I looked up the failing tests and i'm still not sure why this test passed in the first place for directed graphs, because nodes were swapped.

There might be a problem at the GraphBuilder in matching half edges!? (error was not recognised because only `indexInOutArray` was used)

PR #168, opened 2018-06-04
Faster SNAPGraphReader implementation

rbange:
This PR is the extension of https://github.com/kit-parco/networkit/pull/160.

The new SNAP Graph Reader performs **twice** as good as the old one by using mmap() and iterating just once over the file. 
In addition, it can now be used for directed graphs too and there are parameters to further speed up the performance if necessary.

I also modified the Graph class in order to reserve space, but not create nodes [(here)](https://github.com/rbange/networkit/tree/feature/reserveSpaceInGraph). However this change has shown no difference in benchmarking with larger graphs


networkit/cpp/io/FastSNAPGraphReader.h
```diff
@@ -0,0 +1,52 @@
...
+	 * Passing a nodeCount might result in a higher memory use for the output graph
+	 * but reduces runtime. Leaving this parameter on its default value, the reader
+	 * is forced to remap the nodes and thereby decreases the memory use on cost of
+	 * slighty higher runtime.
+	 *
+	 * @param[in]	directed	reads in the graph as directed, if set to true
+	 * @param[in]	nodeCount	nodeCount is used to preallocated memory for the number of nodes
+	 * @param[in]	remapNodes	indicates whether nodes should be remapped to other node ids in order to create consecutive node ids
+	 */
+	FastSNAPGraphReader(const bool& directed = false, const count& nodeCount = 0, const bool& remapNodes = true);
```
michitux: Passing a `bool` as const reference is not really useful as here you replace a 1 byte value by an 8 byte pointer. I would suggest to just pass them by value.

networkit/cpp/io/FastSNAPGraphReader.cpp
```diff
@@ -0,0 +1,134 @@
...
+FastSNAPGraphReader::FastSNAPGraphReader(const bool& directed, const count& nodeCount, const bool& remapNodes) :
+directed(directed), nodeCount(nodeCount), remapNodes(remapNodes){}
+
+Graph FastSNAPGraphReader::read(const std::string &path) {
+	Graph graph(0,false,directed);
+
+	//In the actual state this parameter has very little influence on the reader performance.
+	//There can be a significant boost if it is possible to reserve space in the graph initialization
+	if(nodeCount != 0)
+		nodeIdMap.reserve(nodeCount);
```
michitux: Shouldn't memory only be reserved when actually needed, i.e., when `remapNodes` is true, too?

networkit/cpp/io/FastSNAPGraphReader.h
```diff
@@ -0,0 +1,52 @@
...
+	bool remapNodes;
+
+public:
+	/**
+	 * Default constructor for the FastSNAPGraphReader.
+	 * NOTE: Keep in mind that many SNAP graphs do not have consecutive node ids.
+	 * Passing a nodeCount might result in a higher memory use for the output graph
+	 * but reduces runtime. Leaving this parameter on its default value, the reader
+	 * is forced to remap the nodes and thereby decreases the memory use on cost of
+	 * slighty higher runtime.
```
michitux: This comment doesn't seem to correspond to the implementation as `nodeCount` is not used in the graph initialization in any way but only used for reserving memory for `nodeIdMap` which is not part of the final graph.

If you say this doesn't impact the performance I wonder if it really makes sense to bother the user with an additional parameter that makes usage much more complicated in many cases.

networkit/cpp/io/FastSNAPGraphReader.h
rbange: True, I fixed it in https://github.com/kit-parco/networkit/pull/168/commits/f944ddb670fdac2161c583d90acf6a2e0414604d

networkit/cpp/io/FastSNAPGraphReader.cpp
rbange: True, I fixed it in https://github.com/kit-parco/networkit/pull/168/commits/f944ddb670fdac2161c583d90acf6a2e0414604d

networkit/cpp/io/FastSNAPGraphReader.h
rbange: Indeed, the comment was still for an old version. I wrote down the correct behaviour in https://github.com/kit-parco/networkit/pull/168/commits/f944ddb670fdac2161c583d90acf6a2e0414604d.

I did not remove the NodeCount parameter, because I am still hoping we can get performance increases through reserving graph space. Do you have an idea why the changes in https://github.com/rbange/networkit/commit/fb3f089f36965734fb5170286ac903163f4bd470 does not show any performance increase? I tested the reader with bigger graphs (>1 mio nodes). Is reserving space in C++ not as relevant as thought?

michitux: The problem is here you are only reserving space for the vectors over all nodes, which is a small number of vectors. The majority of storage space is in the neighborhood and weight information - for a million nodes, you have at least a million of these vectors (assuming an unweighted, undirected graph). All these 1 million vectors need memory allocations for every resize - and there are many, as each of them will be resized several times as more and more edges are added. I assume this costs much more time than resizing the vectors over all nodes a few times (note that resizes always double the size, so this won't happen too often after a certain number of nodes have been seen). Further, sequential memory accesses are fast. Note that storing something like the 24 bytes that are needed per vector needs just 24MB for a million nodes, so resizing this vector something like 20 times shouldn't have a big performance impact (further, most of these resizes happen when the vector is still much smaller).

avdgrinten:
@rbange This PR needs a rebase. Can you rebase it, resolve the conflicts in IOGTest and update the PR by force pushing your branch?

PR #169, opened 2018-06-04
Faster KONECTGraphReader implementation

rbange:
The new KONECTGraphReader implementation also uses mmap() as proposed by @avdgrinten here: https://github.com/kit-parco/networkit/pull/160.

In benchmarks with bigger graphs, this implementation performs up to 4 times faster then its old counterpart. This performance increase basically results from iterating just once over the file, using mmap() and using the second property line of KONECT Graph files more safe and efficiently. 

In addition, this implementation offers the possibility to handle graphs with multiple edges properly, by offering a parameter to specify a method.


networkit/cpp/io/FastKONECTGraphReader.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include "../auxiliary/Log.h"
+
+#include "FastKONECTGraphReader.h"
+
+namespace NetworKit{
+	FastKONECTGraphReader::FastKONECTGraphReader(const bool& remapNodes, MultipleEdgesHandling handlingmethod):
```
michitux: Same as in #168 - why is the `bool` passed by reference?

networkit/cpp/io/FastKONECTGraphReader.cpp
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+
+		Graph graph((secondPropertyLine ? numberOfNodes : 0), weighted, directed);
+
+		//Map nodes and increase graph size if no second property is defined
+		auto mapNode = [&] (node in) -> node {
+			if(secondPropertyLine){
+				if (in > numberOfNodes){
+					return graph.hasNode(in) ? in : graph.addNode();
```
michitux: Why are new nodes added dynamically here? And if you see the same node id above the range, why are always new nodes added? Is it somewhere defined (I have never dealt with KONECT graphs) that these nodes outside the range always occur only once?

networkit/cpp/io/GraphReader.h
```diff
@@ -14,12 +14,26 @@
...
  * Abstract base class for graph readers.
  */
 class GraphReader {
 public:
+
+	enum MultipleEdgesHandling
+	{
+		 DISCARD_EDGES, //Reads and selects the first edge which occurs and discards all following
+		 SUM_WEIGHTS_UP, //If an edge occurs again, the weight of it is added to the existing edge
+		 KEEP_SHORTEST_PATH //The edge with the lowest weight is kept
```
michitux: Wouldn't be something like `KEEP_MINIMUM_WEIGHT` more intuitive?

avdgrinten: Agreed.

networkit/cpp/io/FastKONECTGraphReader.cpp
rbange: Fixed in https://github.com/kit-parco/networkit/pull/169/commits/2af7dbc977feb8f954a6706ba93c009295ecf4e0

networkit/cpp/io/FastKONECTGraphReader.cpp
rbange: Fixed in https://github.com/kit-parco/networkit/pull/169/commits/2af7dbc977feb8f954a6706ba93c009295ecf4e0

I agree, that the behaviour there was misleading and could lead to misbehaviour in these special cases where the file header are corrupted. Now, I implemented a "fallback" as if no header was given in first place

networkit/cpp/io/GraphReader.h
rbange: Fixed in https://github.com/kit-parco/networkit/pull/169/commits/109b56668449c7a52d332d2c7b50168cbd298800

rbange:
The latest commit replaces the KONECTGraphReader with the FastKONECTGraphReader and exposes the reader, as well as the new enum in `GraphReader` to python.

Regarding the old constructor parameters: `separator` was quite useless in the first place, because KONECT graphs are separated by Tab and normally well formatted. 
`ignoreLoops` might be a useful parameter because no `hasEdge` or `addEdge` has to be called, but 
1. there is no special need for this parameter in the KONECT graph format
2. I think graphs should be read in as intended by the creator of the file and modification to the graph should be made by the user after reading
3. The performance increase does not even scale up to the share of loops in the total amount of edges, because the lines still have to be parsed

btw: travis failed due to internal errors

networkit/_NetworKit.pyx
```diff
@@ -3021,6 +3021,17 @@ cdef extern from "cpp/io/GraphReader.h":
 		_GraphReader() nogil except +
 		_Graph read(string path) nogil except +
 
+cdef extern from "cpp/io/GraphReader.h" namespace "NetworKit::GraphReader":
+	cdef enum MultipleEdgesHandling "NetworKit::GraphReader::MultipleEdgesHandling":
+		DISCARD_EDGES,
+		SUM_WEIGHTS_UP,
+		KEEP_MINIUM_WEIGHT
+
+class _MultipleEdgesHandling(object):
```
michitux: Shouldn't this be the class without "_" (and above with "_")? And why does this inherit from `object`, is this really needed? Then the import into the `graphio` module would also be cleaner.

networkit/cpp/io/KONECTGraphReader.cpp
```diff
@@ -1,226 +1,279 @@
...
+
+
+		Graph graph((secondPropertyLine ? numberOfNodes : 0), weighted, directed);
+
+		//Map nodes and increase graph size if no second property is defined
+		std::function<node(node)> mapNode = [&] (node in) -> node {
+			if(secondPropertyLine){
+				if (in > numberOfNodes){ //if file is corrupted
+					//secondPropertyLine is made useless
+					ERROR("Given number of nodes by file does not match actual graph");
```
michitux: Wouldn't it be better to just throw an exception here? Or is it so common that this property line is wrong? Users might not notice this message and then get wrong results as their graph was broken.

networkit/cpp/io/KONECTGraphReader.cpp
```diff
@@ -1,226 +1,279 @@
...
+					secondPropertyLine = false;
+					if(remapNodes){ // if remapNodes is selected true, the map has to be initalized with the existing nodes
+						nodeIdMap.reserve(numberOfNodes);
+						graph.forNodes([&](node u) {
+							nodeIdMap.insert({u,u});
+						});
+					}
+					return mapNode(in);
+				}else{
```
michitux: Shouldn't there `remapNodes` also be considered when the number of nodes is given in the graph file? I.e., shouldn't this be the other way around, if `remapNodes` is set always use `nodeIdMap` and only otherwise possibly add nodes to the graph? At least this is what I would expect from the documentation of the constructor.

networkit/cpp/io/KONECTGraphReader.cpp
avdgrinten: It is indeed common. IMHO ERROR is reasonably easy to notice; I don't see what we could do better here.

networkit/cpp/io/KONECTGraphReader.cpp
rbange: Well, the whole use of the remapNodes parameter is to allow the user to read in KONECT graphs with non-consecutive ID's. If the number of nodes is given in the second property line, there is basically no need to remap Nodes (because we can assume that the id's are consecutive with a VERY high probability), which would cost more runtime and memory to additionally map nodes in a map.

The reason I added the remap behaviour in this error handling is, that if the second property line is "lying" about the actual number of nodes, the user can still remap the nodes and thereby decrease the amount if isolated nodes and size of the graph, without greatly modifying the graph file itself.

networkit/cpp/io/KONECTGraphReader.cpp
rbange: In addition, this behaviour is consistent with the one of the old KONECTGraphReader so we might stick with it.

networkit/_NetworKit.pyx
rbange: Agreed, that style would fit the .pyx way better: https://github.com/kit-parco/networkit/pull/169/commits/94564f867bcd3d25dd8a14cc9b11f860495f7680

michitux:
I assume this doesn't work on Windows, does it? So with this merged, Windows support is dropped also for the C++ core?

avdgrinten:
This does not work on Windows but I'm under the impression that NetworKit does not really work on Windows anyway. The current recommendation is to run it using WSL. This patch will not break WSL as WSL is able to handle mmap() just fine. OTOH if first-class Windows support is desired, this reader can easily be ported to use Windows' file mapping facilities like [MapViewOfFile](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366761(v=vs.85).aspx).

@michitux Do you have a use case for Windows?

@kolja-esders can you comment on the current state of Windows support?

michitux:
@avdgrinten No, I don't have any use case. I was just under the impression that Windows compatibility is a (low-priority) goal of NetworKit.

hmeyerhenke:
Two quick comments:
1) Windows support should not go below the current standard (which is not native but close enough from what I was told).
2) Make sure the interface is not changed if a functionality is replaced. If the new default KONECT reader is the fast one, it should get the name of the old one (unless there are reasons not to do so).

michitux:
https://github.com/mandreyel/mio seems to be a nice cross-platform library for memory-mapping under MIT license. I'm asking about this as I'm considering using memory mapping for the binary file reading (and possibly also writing), too.

avdgrinten:
My suggestion is to just use the raw mmap() interface for now. Writing a wrapper for Windows compatibility should be trivial and can be added in the future; we do not need a 3rd party library for that. As the rest of NetworKit does not build on Windows right now, writing such a wrapper right now is not desirable.

After we merge cmake, we could try to focus on native Windows compatibility in the next release.

rbange:
Im not sure which error one could get when compiling the NetworKit C++ core on Windows, but as a compromise I could add some preprocessor directives like in `io/DibapGraphReader.cpp`. This will probably result in a more intuitive error handling

avdgrinten:
For now, I do not want to introduce `#ifdef`s. We can still do that after the Monday release. In particular, I want to avoid adding untestable code now and we can indeed not test Windows code as the rest of NetworKit does not compile on Windows.

In addition to that, this PR also needs a rebase.

PR #170, opened 2018-06-14
Test suite rework

rbange:
This PR intends to improve and clean up the test suite by ensuring that tests 

1. use assertions (if possible assertions were added)
2. unnecessary runtime is further reduced
3. output into INFO log level is reduced to results only

Test which do not use GTest assertions are moved to a new "run" prefix class. For more info see https://github.com/kit-parco/networkit/commit/2c1ba39f1330995f35e9f5988cc4868878acceec.

Overall, most test cases use assertions properly, but 65 cpp classes are not tested yet.


networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -914,25 +934,24 @@ TEST_F(GeneratorsGTest, tryGiganticCollectionOfHyperbolicUnitDiskGraphs) {
 }
 
 TEST_F(GeneratorsGTest, testLFRGenerator) {
-	Aux::Random::setSeed(42, true);
```
michitux: Not using a static seed for the LFR generator test case is a very bad idea as the parameters are relatively tight and depending on the actual randomness used this test case might fail randomly because the drawn sequence is not realizable. I understand that this might not be desired that an algorithm might fail randomly but imho it is better to fail instead of trying to repair the sequence as it is not easy to see if a failure is (almost) unavoidable with the given parameters or if failure is unlikely. The original LFR generator does some strange repairing operations that might give you a completely different community size sequence than what you expected.

networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -172,15 +173,16 @@ TEST_F(GeneratorsGTest, testStaticPubWebGenerator) {
 	DEBUG("Modularity of clustering: " , modVal);
 	DEBUG("Total edge weight: " , G.totalEdgeWeight());
 	EXPECT_TRUE(G.checkConsistency());
+
+	//reset seed to previous
+	Aux::Random::setSeed(seed, false);
```
michitux: This does not restore the random number generator to the previous state but instead restores it to the previously set static seed. I don't understand why this is done here. Could you please explain why the seed should be restored? If there had been no seed set previously, this will set a seed but the generator will be in a completely different state. If there had been a seed set, it restores the random number generator to this state. However, I don't see why anyone would want to have their seed restored.

Imho every test that uses randomness should set a seed in order to ensure that test results are reproducible. Otherwise test results are random/depend on how many and which other tests were executed before the current test.

rbange: Well, as far as I understand the concept of this test suite, there is an initially non static seed at the beginning, but because some tests require (or at least is preferred to) set a see, this non static seed is overwritten with some other value e.g. 42. 

The problem here is that a following test, which is written in mind that there is a non static seed, is now run with a seed of 42 because the previous test set it so. 

Sooner or later, the whole test suite runs with a static seed because some test initially set it to a specific value. Ideally I would like to have every test to pass with a non-static seed (I see that this is not possible for every test), but this behaviour is impossible at the moment.

I admit, that this function is not part of the branch at the moment, but I think a good design would be to run the whole test suite with a random, but static seed which is printed in front of the output. This way, test which are well written run with a truly random seed while others can set their own seed. In addition, if a test fails for a random seed one can reproduce this bug because the seed is provided in the output.

Ideally this behaviour can also be implemented without this construct of preserving the seed (e.g by implemented an automatic reset after every test).

networkit/cpp/generators/test/GeneratorsGTest.cpp
rbange: I see your point, in this specific case we might need a solution like explained in https://github.com/kit-parco/networkit/pull/170#discussion_r195463533. 

The question is, is the algorithm really implemented to fail randomly? If so, then the seed should definitely set here back and this behaviour should be written down in the comments. This would not be a desirable behaviour if used in a productive environment and has to be caught there.

networkit/cpp/generators/test/GeneratorsGTest.cpp
michitux: While I agree that tests should be pass with random seeds, imho tests should also be reproducible. It is not nice if you implement a feature, submit a pull request and then some random, completely unrelated unit test fails because of the random seed that was used.

Having the possibility to specify a static seed and otherwise printing the used random seed is indeed a good alternative as it means random test failures are reproducible. Note that this would need to happen before every single test function as otherwise the order and number of tests that are executed plays a role (and one shouldn't need to re-run the whole test suite to reproduce a failure). Further, there might be algorithms that uses randomness inside multiple threads that use a different number of random numbers/thread depending on the scheduling which might leave the random generator in a state that depends on the scheduling afterwards. For continuous integration, one could also just use a fixed seed in order to avoid the unrelated test failures described above.

Some time ago, fixed seeds were added in NetworKit at least to some tests in order to make the test suite more reproducible and to avoid random test failures. You are right that they have effects on other tests, too. However, your change does not avoid these effects. It just sets the random number generator to a different seed. Either to the previously set seed, or, if there was no previous seed, to a random seed. Note, however, that you only get a random seed once. After setting a seed once, `getSeed` will always return the previously set seed.

My personal suggestion would be to leave static seeds as they are currently (i.e., remove the changes to random seeds from this PR) and implement random seeds that are reset for every test and printed later.

michitux:
Have you read the discussion in #7 concerning test cases without google test assertions where the executed code still contains assertions and why it is very difficult to write assertions for some of these algorithms? I'm just asking to ensure that the old discussion is not lost.

rbange:
I actually did not see this discussion, but it seems that the new handling (renaming them to "run") fits the desired test behaviour quite well (keep in mind we still run these tests in travis). 

An actual problem is that, the "try" tests are not able to be run automatically because many of them require a manual input of graph files or there is a reference to graphs in the contributors file system. Until these "try" tests are not reworked too (they should either be transformed to test*, run* or removed completely), a 4th group of tests (test*, try*, benchmark*, run*) might be acceptable.

networkit/cpp/generators/test/GeneratorsGTest.cpp
michitux: I have opened #171 to discuss this in more detail but would suggest to not to remove this seed in this PR.

networkit/cpp/generators/test/GeneratorsGTest.cpp
avdgrinten: I agree that tests should be reproducible. On the other hand, they should not hard code results that depend on a random seed. A test that checks if some algorithm output equals X if random seed S was used is useless (as X was clearly obtained by just running the algorithm with the given seed) and will break if the random number generator changes.

We should remove all cases of the latter behavior from the test suite. Still, we should use random seeds for every test to avoid the spurious failures that @michitux mentioned. If it is feasible, we can loop in a test and check different fixed random seeds.

avdgrinten:
Yes, that is the current resultion. Tests are divided into test*, run*, try* (I think we agreed to call that debug* but names do not matter so much) and benchmark*. test* denotes real tests that check if algorithm behavior is correct. run* just runs algorithms and does not check the results (i.e. they just check if the algorithm does not throw/segfault). This is appropriate for heuristic algorithms, where we cannot insert assertions about the results. test* and run* are execute on CI, while the others are intended to be invoked manually.

networkit/cpp/generators/test/GeneratorsGTest.cpp
rbange: https://github.com/kit-parco/networkit/pull/170/commits/75c0e0fd4cc6761505f8e0027ee8be0ed1469315 reverts the changes

networkit/cpp/generators/test/GeneratorsGTest.cpp
rbange: https://github.com/kit-parco/networkit/pull/170/commits/75c0e0fd4cc6761505f8e0027ee8be0ed1469315 reverts the changes

rbange:
The latests commits add back the seed for the LFRGeneratorTest, removes the seed preserving behaviour, and refactors try* tests to debug* to better describe their behaviour. 

I also started in implementing the behaviour of resetting the seed after each test, unfortunately this is far more work then expected. We have to implement a new base class for all tests and override the SetUp() method in `::testing::Test`, which is already in use of some of the NetworKits test cases. In addition, all .h files in the test folders have to be refactored. So IF this behaviour is implemented, I think it will be part of a new PR.

To reduce the risk of wrong tests by "cascading" setting of a seed, I activated the internal shuffle function of gtest, which runs the tests in a random order (random order of test cases and random order of tests in each case).

avdgrinten:
The problem of resetting the seed can be avoided if any test uses one (or more) fixed seed(s). I suggest to add fixed seeds to all tests that depend on randomness (and remove all "resets" if we have any additional ones). But yes, that should be done in a different PR.

avdgrinten:
Also: Can we print/set the gtest seed for shuffling so that tests can be reproduced exactly?

rbange:
Yes, one can append `--gtest_random_seed=[NUMBER]` to the end to run the tests in the desired order.

Btw the gtest suite can still handle all the native commands, someone just decided to overwrite the parameters with the NetworKit specific parameter selection:


```
Test Selection:
  --gtest_list_tests
      List the names of all tests instead of running them. The name of
      TEST(Foo, Bar) is "Foo.Bar".
  --gtest_filter=POSTIVE_PATTERNS[-NEGATIVE_PATTERNS]
      Run only the tests whose name matches one of the positive patterns but
      none of the negative patterns. '?' matches any single character; '*'
      matches any substring; ':' separates two patterns.
  --gtest_also_run_disabled_tests
      Run all disabled tests too.

Test Execution:
  --gtest_repeat=[COUNT]
      Run the tests repeatedly; use a negative count to repeat forever.
  --gtest_shuffle
      Randomize tests' orders on every iteration.
  --gtest_random_seed=[NUMBER]
      Random number seed to use for shuffling test orders (between 1 and
      99999, or 0 to use a seed based on the current time).

Test Output:
  --gtest_color=(yes|no|auto)
      Enable/disable colored output. The default is auto.
  --gtest_print_time=0
      Don't print the elapsed time of each test.
  --gtest_output=(json|xml)[:DIRECTORY_PATH/|:FILE_PATH]
      Generate a JSON or XML report in the given directory or with the given
      file name. FILE_PATH defaults to test_details.xml.

Assertion Behavior:
  --gtest_death_test_style=(fast|threadsafe)
      Set the default death test style.
  --gtest_break_on_failure
      Turn assertion failures into debugger break-points.
  --gtest_throw_on_failure
      Turn assertion failures into C++ exceptions for use by an external
      test framework.
  --gtest_catch_exceptions=0
      Do not report exceptions as test failures. Instead, allow them
      to crash the program or throw a pop-up (on Windows).
```
visible by calling e.g. `./NetworKit-Tests-Dbg -t --gtest_repeat="thisIsNotAnInteger"`

Maybe we should rework the Unittest-X.cpp to point out that these parameters are still available.

avdgrinten:
OK. The seed is also printed automatically. Thank you for your work, I'm merging this PR now as all comments have been addressed.

PR #174, opened 2018-06-20
Feature curveball

manpen:
Don't merge yet.
Description following.


hthetran:
Just FYI: I removed an old deprecated todo, added an option for parallel graph building after the Curveball algorithm has run and fixed an issue in the docstrings for the TradeGenerators where the order of the class documentation and the cdef was not correct.

networkit/cpp/graph/Graph.h
```diff
@@ -80,6 +83,7 @@ class Graph final {
 
 	friend class ParallelPartitionCoarsening;
 	friend class GraphBuilder;
+	friend class CurveballDetails::CurveballMaterialization;
```
avdgrinten: While this is OK for now, it would be desirable to extract the materialization class from the Curveball namespace - it is not really Curveball specific, right?

networkit/cpp/randomization/Curveball.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "../auxiliary/Random.h"
+
+#include "Curveball.h"
+#include "CurveballImpl.h"
+
+namespace NetworKit {
+
+Curveball::Curveball(const NetworKit::Graph &G) :
+    impl_(new CurveballDetails::CurveballIM{G})
```
avdgrinten: Existing NetworKit code uses camelCase even for internal variables. While I can see the point of the underscore, consistency should be preferred here. Can you change the internal variables to use camelCase?

networkit/cpp/randomization/Curveball.h
```diff
@@ -0,0 +1,52 @@
...
+
+	virtual bool isParallel() const {
+		return false;
+	}
+
+	count getNumberOfAffectedEdges() const;
+
+
+private:
+	std::unique_ptr<CurveballDetails::CurveballIM> impl_;
```
avdgrinten: Why is a private implementation necessary here? There is only a single CurveballIM implementation, right? Why not just use that here without a heap allocation?

networkit/cpp/randomization/CurveballUniformTradeGenerator.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef RANDOMIZATION_CURVEBALL_UNIFORM_TRADE_GENERATOR_H
+#define RANDOMIZATION_CURVEBALL_UNIFORM_TRADE_GENERATOR_H
+
+#include "../Globals.h"
+
+namespace NetworKit {
+
+class CurveballUniformTradeGenerator {
+public:
+	using value_type = std::vector<std::pair<node, node> >;
```
avdgrinten: value_type is very generic. Can we use some more specific name? Also, I don't think _type has precedent in NetworKit. I would like the suffix to be dropped.

networkit/cpp/randomization/CurveballImpl.h
```diff
@@ -0,0 +1,245 @@
...
+using count = node;
+
+using trade_vector = std::vector<TradeDescriptor>;
+using nodepair_vector = std::vector <std::pair<node, node> >;
+
+constexpr node INVALID_NODE = std::numeric_limits<node>::max();
+constexpr count LISTROW_END = std::numeric_limits<count>::max();
+constexpr tradeid_t TRADELIST_END = std::numeric_limits<tradeid_t>::max();
+
+struct edge_t : public std::pair<node, node> {
```
avdgrinten: I do not like the _t suffixes. node and count do not have suffices. As naming the type just `edge` might not be a good idea, I suggest to rename it. On the other hand, this struct adds little over std::pair<node, node>.

networkit/cpp/randomization/CurveballImpl.h
```diff
@@ -0,0 +1,245 @@
...
+            std::swap(first, second);
+    }
+
+    bool is_invalid() const {
+        return (*this == invalid());
+    }
+};
+
+
+class CurveballAdjacencyList {
```
avdgrinten: I guess this class is needed to preallocate adjacency lists if you know the right degree. Such a class could be useful in a more general context. Could this be added to the Graph class instead?

If there is not enough time until Monday to do this, consider it for the future.

networkit/cpp/randomization/GlobalCurveball.h
```diff
@@ -0,0 +1,64 @@
...
+    Graph getGraph();
+
+    virtual std::string toString() const override final;
+
+    virtual bool isParallel() const {
+        return false;
+    }
+
+private:
+    std::unique_ptr<CurveballDetails::GlobalCurveballImpl> impl_;
```
avdgrinten: Same remark as in the non-global case.

networkit/cpp/randomization/GlobalCurveballImpl.h
```diff
@@ -0,0 +1,369 @@
...
+                        // u is processed before v
+                        current_pq.emplace_in_bucket(hint_u, hashed_u, v);
+                    } else {
+                        current_pq.emplace(hashed_v, hashed_v, u);
+                    }
+                });
+            });
+
+            loadTimer.stop();
+            INFO("Loading graph took ", loadTimer.elapsedMilliseconds(), "ms.");
```
avdgrinten: Consider turning this into DEBUG. Lots of INFO messages can be annoying e.g. when you run the test suite.

networkit/cpp/randomization/GlobalCurveballImpl.h
```diff
@@ -0,0 +1,369 @@
...
+        return input_graph_;
+    }
+
+protected:
+    bool hasRun_ {false};
+    tfp_queue_type pq_;
+
+    const NetworKit::Graph& input_graph_;
+
+    void common_disjoint_neighbour_sortsearch(std::vector<node>& neighbourhood_of_u,
```
avdgrinten: Internal methods are camelCase, too.

networkit/cpp/graph/Graph.h
manpen: Yes, it's very similar to the GraphBuild with the major difference that the GraphBuilder implements an in situ push interface, while this class takes an edge list. Also, the new implementation is much faster for our application. So following your suggestion I won't change it until the next release.

networkit/cpp/randomization/Curveball.h
manpen: The main reason to do so is to follow the pImpl-Idiom to provide a compilation barrier; this is useful since the .pyx file includes all public headers. If we do not use the barrier, a lot of Curveball specific header-files need to be parsed w/o benefit to the .pyx compilation process. The indirection won't affect the performance, since only a very small number of functions calls happen.

networkit/cpp/randomization/GlobalCurveball.h
manpen: Here pImpl is even more important, since the GlobalCurveball Implementation is templated and hence completely resides in the Header files. It also adds header-only dependencies to e.g. the RadixHeap and hence accounts for several thousand lines of code, which do not have to be parsed for the pyx build.

networkit/cpp/randomization/CurveballUniformTradeGenerator.h
manpen: We commonly use this type in our implementations mimicking STL's conventions. The generic name is by design and since it's the same for all generators, transformers and containers you do not think about how the type is called when chaining together templates. In this case it's not really used. So if you do not like it, I suggest completely dropping the type alias.

manpen:
Thank you for your review. I will fix all comments to which I did not answer directly.

manpen:
The PR should now comply to NetworKit's naming conventions, and I removed the _t types, and speed-up tests. Thus, I currently see nothing in the way of merging the PR.

avdgrinten:
Yes. Thank you for your work! I will merge the PR before the release.

Issue #175, opened 2018-06-21
Test failure in testHashingOverlapperForCorrectness


avdgrinten:
After the merge of #170, one of the tests fails on CI. It did not fail on the PR's branch; the failure might have been caused by the merge. @rbange can you look into that?

rbange:
Yea, indeed this is caused by https://github.com/kit-parco/networkit/pull/159 which somehow changes the order of elements in the output partition of the `HashingOverlapper::run` method. Because https://github.com/kit-parco/networkit/pull/159 was not merged into https://github.com/kit-parco/networkit/pull/170, the test failed. 
I rebased my branch and fixed the assertion in https://github.com/kit-parco/networkit/pull/176

avdgrinten:
Thank you.

PR #176, opened 2018-06-21
fix assertion in testHashingOverlapperForCorrectness

rbange:
Fixes the issue raised in https://github.com/kit-parco/networkit/issues/175.

The issue is fixed by comparing if the output vector is a permutation of the desired one.


avdgrinten:
Look good. Fixes #175.

PR #184, opened 2018-07-06
Fix/fix module deps

manpen:
This PR accompanies PR #154. 

It seems like we can realize NetworKit submodules with an acyclic (binary) dependency graph by moving only two classes, namely `graph/RandomSpanningForest` and `dynamics/AffectedNodes` to `components` and `distance` respectively.

All remaining dependencies are Header-only and hence do not prevent shared-libs.


avdgrinten:
Thank you. As discussed earlier, we will move those classes to support non-monolithic builds. As there are no Python bindings, we expect this not to break too many existing applications using NetworKit.

PR #189, opened 2018-07-10
Improved ApproxGroupBetweenness.cpp

MarvinPogoda:
Slightly increased the accuracy, since we only need to add the internal nodes of a shortest path to the hyperedge.


avdgrinten:
Looks good, thank you.

PR #191, opened 2018-07-13
Fix usage of EffectiveDiameterApproximation

### LINKED ISSUE #199: Effective diameter approximation in profiling fails
Right now, the effective diameter approximation during a complete profiling fails (returns `N/A`), because the class is called by a wrong name.

https://github.com/kit-parco/networkit/blob/a91e1ce0e70df7b22da88f56a2fc783cd2aa1d29/networkit/profiling/profiling.py#L757

The name of the class is `EffectiveDiameterApproximation`, not `ApproxEffectiveDiameter`.

PFischbeck:
Until now, measuring the approximated effective diameter in profiling always failed due to an incorrect function call.

This fixes #199.

### DIFF
--- a/benchmark/nk.py
+++ b/benchmark/nk.py
@@ -93,11 +93,11 @@ class bEffectiveDiameter(Algo):
 		return networkit.distance.EffectiveDiameter(G).run().getEffectiveDiameter()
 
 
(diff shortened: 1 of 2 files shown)


avdgrinten:
That is of course a bug. Thank you for the PR.

PR #197, opened 2018-07-23
CMake build environment

rbange:
This PR is carrying on the work of @manpen in https://github.com/kit-parco/networkit/pull/154 and is based upon contribution of @manpen @avdgrinten and myself.

**Major changes:**
- CMake is replacing SCons
- Googletest is now a git submodule and automatically built
- setup.py is rewritten completely
- python requirements are no hard dependencies for the installation of networkit as a python module
- the build system [ninja](https://ninja-build.org/) or make can be used within python module setup

Minor changes:
- `.gitignore` is structured better and all kind of build directory names are added
- old build environment is removed 
    - `mbe.py` is removed because cmake is hard dependency now
    - `ez_setup.py` is removed because python setuptools is hard dependency now
    - build.conf is not necessary anymore

Missing features:
- Fixing the usage of openmp + clang and cmake
- Update documentation
- Maybe we should ship a good script for automatically adding .cpp 's to CMakeList files?


CMakeLists.txt
```diff
@@ -0,0 +1,272 @@
...
+    message(SEND_ERROR "In-source builds are not allowed.")
+endif("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
+
+# BUILD OPTIONS
+option(NETWORKIT_BUILD_TESTS "Build with tests" OFF)
+option(NETWORKIT_LOGGING "Build with logging support" ON)
+option(NETWORKIT_MONOLITH "Build single library (and tests is requested; required for shared lib)" ON)
+option(NETWORKIT_NATIVE "Optimize for native architecture (often better performance)" OFF)
+option(NETWORKIT_WARNINGS "Issue more warnings" OFF)
+option(NETWORKIT_INCLUDESYMLINK "Create symlink to cpp directory" ON)
```
avdgrinten: Why do we need those symlinks?

rbange: There is no direct necessity on build such symlinks. They are quite handy when you use the networkit cpp library inside your own program, because you link your libraries as well as the include files to your build directory. (Otherwise you have link the library, but the .h files to the cpp source directory)

This behaviour was copied from the old scons build environment

manpen: Another quite drastic but much cleaner variant would be to move networkit/cpp to cpp/networkit. This has HUGE advantages:
- No symlinks (by the way, i ported the CPP project to MSVC so this will become relevant)
- If we use Networkit as a subproject in another project by including it as a git submodule, the 3rd-party just includes the header files as #include <networkit/....>
- Includes WITHIN Networkit can be written the same way rather than relative #include "../"

avdgrinten: I agree that the current directory structure is suboptimal with respect to the header files. However, I do not think that this should be changed in this PR. Let us transition to cmake first before we change the directory structure.

Regarding a MSVC port: Do you plan to make that public? IMO this would be a huge addition to NetworKit. If you want to discuss a MSVC port, we could do that via email.

CMakeLists.txt
```diff
@@ -0,0 +1,272 @@
...
+                gtest_main gtest
+                networkit
+        )
+        set_target_properties(networkit_tests PROPERTIES
+            CXX_STANDARD 11
+            COMPILE_FLAGS "${NETWORKIT_CXX_FLAGS}"
+            LINK_FLAGS "${OpenMP_CXX_FLAGS}")
+
+        add_test(networkit_tests networkit_tests)
+		add_custom_target(symlink_input_folder ALL COMMAND ${CMAKE_COMMAND} -E create_symlink "${PROJECT_SOURCE_DIR}/input" "input")
```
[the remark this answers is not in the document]
rbange: Warning for windows users is missing, indeed. 

Indentation is off here too, right. @manpen defaults his editor to spaces while I set mine to Tab.  

Of course they could run the networkit_tests executable from the networkit root directory and can make it without a symlink, but this is an odd behaviour when working with cmake because you expect it to work without moving the executable. The real problem is that ALL graph paths in the tests are given like `input/graph`. The only "clean" way to do this, would be changing the paths of the tests to be depended on a variable which would be the path from the build directory to the source directory.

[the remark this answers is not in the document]
avdgrinten: True, it is odd behavior. But it only affects potential Windows users who cannot even run NetworKit without patches right now anyway.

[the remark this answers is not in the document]
manpen: It not necessary to move the executable? The way I currently run tests is using

```bash
cd build
[...]
make 
cd ..
build/networkit_tests -t
```
I think to more CMake-ish way to do it, is using CTest which allows to specify a working directory:

```cmake
 add_test(
                    NAME networkit_tests
                    COMMAND networkit_test -t
                    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 )

 add_test(
                    NAME networkit_tests_no_assertions
                    COMMAND networkit_test -r
                    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 )
```
and then invoke `ctest -V` after compilation.

avdgrinten:
Look good. I'm merging this to the `cmake-pr` branch. We will fix the remaining issues and then merge into `Dev`.

PR #202, opened 2018-07-24
Cumulative Updates to CMake-PR

manpen:
This PR combines several improvements which I consider okay, since it's not targeting the Dev-branch yet.
- Rather than using OpenFlags as an parameter to `LINK_FLAGS` directly, this PR introduces the `NETWORKIT_LINK_FLAGS` proxy. This is partly to make MSVC porting easier, which treats the Flags subtle differently.
- Introduce `NETWORKIT_CXX_STANDARD` as parameter to make it possible to test NetworKit against higher standards during CI.
- `setup.py` now only invokes Cython if `_Networkit.cpp` is either missing or out-dated (its timestamp is older than `_Networkit.pyx`). This significantly speeds up recompilation.
- In monolithic builds define two tests with correctly set working path, to avoid the necessity of symlinks to the input/output directories in the build directory. Consequently, Travis hence does not call `networkit_tests` directly, but executes `ctest -V`.
- Travis now tests now also include a g++8 with CXX in {11,14,17} to spot future issues early on during CI
- Travis now also test the PIP installation. Here we have a slight issue since Travis worker expose 48 virtual cores, which causes OOM-issues during build. Unfortunatly passing `-j2` to setup.py requires `pip install --install-options="-j2"` which are also passed to all dependencies triggering fatal errors. This seems to be a known issues with PIP and I did not find a solution. For this reason, setup.py now also checks whether the environment variable `NETWORKIT_PARALLEL_JOBS` is set, and defaults the number of parallel jobs to it.

Remark: executing the unit tests in `networkit/tests/*` caues SegFaults; hence the failing Travis tests. I consider fixing this out-of-scope of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,8 +80,8 @@ script:
 
  - python3 -m venv pyenv && . pyenv/bin/activate
  - pip3 install --upgrade pip
+ - pip3 install ipython numpy scipy cython # jobs that conflict with --install-options
  - pip3 install --install-option="-j 2" -e . 
- - pip3 install ipython # required in networkit/profiling/profiling
  - python3 -m unittest discover -v networkit/test/
 
  - mkdir debug_test && cd "$_"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: generic
 
+env:
+  global:
+    - export OMP_NUM_THREADS=2
+    - export CXX_STANDARD=11
+
 matrix:
   include:
     - env: CC=gcc-4.9 CXX=g++-4.9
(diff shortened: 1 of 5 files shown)


avdgrinten:
This PR breaks the previously successful tests. Can you comment on why that happens / what would be necessary to get them working again? I'm hesitant to merge the PR without a fix for the tests.

manpen:
The previous tests were partly successful because they never tested the Python interface. Observe however that currently the branch has failing tests for clang 3.7 and 3.8 (https://travis-ci.org/kit-parco/networkit/builds/407254831), these are still failing after this PR.

The Travis tests now also execute the tests in `network/test`. Here I found two unrelated bugs. The first one was an unnecessary assertions in Curveball (sorry). The second issue accounted for in commit 723bc5f caused SegFaults; since I'm no Cython-expert please check this commit and my explanation:

The Cython docu states:

> When subclassing extension types, be aware that the `__dealloc__()` method of the superclass will always be called, even if it is overridden. This is in contrast to typical Python behavior where superclass methods will not be executed unless they are explicitly called by the subclass. (http://cython.readthedocs.io/en/latest/src/userguide/special_methods.html)

The issue suffaced first in `ConnectedComponent`, which does:
```cython
cdef class ConnectedComponents(Algorithm):
	#[ ... ]
	def __dealloc__(self):	
		del self._this # free memory but do not set _this to NULL
```

In `Algorithm` we have the following destructor:
```cython
	def __dealloc__(self):
		if self._this != NULL:
			del self._this
		self._this = NULL
```

This in combination (sometimes?, consistently on my machines) leads to a double-free segfault.
If this is really the explanation for the bug, I do not understand, why it did not come up before.

There are several destructors (mostly ConnectedComponent related) where this occurs. While I was at it, I also removed correct but redundant destructors, which would delete `self._this` and set it to `NULL` which the superclass would to the same job with less clutter.

avdgrinten:
Yes, the Cython fix looks good (as far as I can tell). The rest is also fine. Merging.

PR #203, opened 2018-07-25
Fix macOS clang build on Travis

avdgrinten:
Apparently, NetworKit hits an internal LLVM error with LLVM 3.7. Upgrade the macOS build to LLVM 4 to fix this issue.


avdgrinten:
There have been no complaints, so I'm merging this.

PR #205, opened 2018-07-26
Fix Linux clang build on Travis

avdgrinten:
While the macOS clang build was broken because of LLVM bugs, the Linux clang build was broken because of NetworKit bugs. In particular, NetworKit relied on the undefined `std::vector<std::atomic<...>>` pattern. This PR fixes this issue.

Based on #203.


avdgrinten:
There have been no complaints, so I'm merging this.

PR #208, opened 2018-07-30
MacOS OpenMP linking

rbange:
As mentioned by @manpen the `OpenMP::OpenMP_CXX` field is not supported in CMake 3.5, which is our minimum required version because of Ubuntu 16.04 support.

Probably no macOS user is stuck on a 2.5 year old version, but nevertheless this workaround should be added regarding consistency and stability.

The commit is tested and compatible with the travis changes proposed here: https://github.com/kit-parco/networkit/pull/203 ([Travis](https://travis-ci.org/rbange/networkit/builds/409777088))


avdgrinten:
Should we really link to libgomp? What if we're compiling with a compiler (e.g. clang) that uses a different OpenMp implementation? I'd suggest to replace that by `OpenMP_CXX_FLAGS` (which should include `-lgomp`, right?). It seems to be possible to test for the existence of OpenMP::OpenMP_CXX using `if(NOT TARGET OpenMP::OpenMP_CXX)`, so we could fall back to `OpenMP_CXX_FLAGS` only in that case.

By the way, doesn't the `set_target_properties()` in the `APPLE` case prevent the `NETWORKIT_LINK_FLAGS` from being used? I guess that would be a bug.

avdgrinten:
The the `if(NOT TARGET OpenMP::OpenMP_CXX)` check can probably be done OS-independently. Instead of calling `set_target_properties()` we should change `NETWORKIT_LINK_FLAGS`.

rbange:
This is clearing the better solution! For whatever reason I thought a second `set_target_properties` call would add to the existing linking flags, because I found the file this way. 
But in the old version, the existing `-fopenmp` was overwritten and thereby OpenMP was never linked.

The new commit should cleanup this issue. Seems that there is not need at all for checking after `OpenMP::OpenMP_CXX`.

In addition I rechecked the OpenMP linking and saw that `link_directories("${OPENMP_LIBRARIES}")` is also not supported by CMake 3.5. However, it can be removed because `-fopenmp` should do the trick here too.

avdgrinten:
Can you rebase this and fix the merge conflicts on the current cmake-pr?

avdgrinten:
Travis fails only due to a timeout, so I'm merging this.

Issue #195, opened 2018-07-21
Windows binaries. How much could be difficult? 


nico1996it:
I really need to build the windows binaries of this library, but before trying, I ask you if it could be trivial or not. 
Because i saw that there's not strange dependencies or anything like that, and so I'm thinking about why there's no official binaries.
Thanks

avdgrinten:
The main reason why Windows binaries do not exist is because none of the maintainers uses Windows. There were some attempts to make NetworKit run on Windows but they did not make it to a full PR.

Regarding the code: Yes, there are no strange libraries, so porting NetworKit to Windows should not be too hard. Some classes use Linux specific functions but I do not think it would be much work to add Windows-specific workarounds. Work is underway to switch to cmake as a build system which is able to generate Visual Studio projects. This should make it much more easy to build NetworKit on Windows.

However, instead of porting NetworKit directly to Windows, you might want to consider using WSL. Could you tell us more about your exact use case and why it is necessary to build the library on Windows?

avdgrinten:
It would be acceptable to have a Windows port with some missing/disabled features first. Once we have that, we could iteratively fix all features. If you have code cleanups to make the code more conforming to C++ standards, you could also post that as a separate PR before posting the full Windows support.

In my opinion a critical feature regarding Windows support would be CI for Windows. This would prevent us from breaking Windows support in the future even if we do not primarily develop on Windows. Unfortunately, Travis does not support this; however, other CI services (like AppVeyor) do.

manpen:
I will try to filter standard-related issues, but there are only a few. 
Regarding failing tests: I will look at them more detailed, and if there something a cannot fix quickly I will disable the feature. I few implementations will be less efficient, since omp task is not supported by MSVC. Hence I disabled parallelism there, which could be fixed using explicit task queues (as included in TLX), but I currently do not see the point.
AppVeyor is already implemented.

fabratu:
Closing this issue, since Windows binaries are supported by now.

PR #209, opened 2018-07-31
MSVC support

manpen:
This PR brings basic support for MSVC. In its current state, the NetworKit's CPP library builds with MSVC 2017, and all tests run successfully. As such, it partly addresses Issue #195.

The PR also introduce an AppVeyor configuration file, so you might want to enable this CI for the main repository.

It's worth noting that commits 0ddfc9c and d59e47b introduce the new type `omp_index` as MSVC requires signed loop variable for `#pragma omp parallel for`. Hence, this PR will add additional signness-related warnings.


networkit/cpp/io/test/MemoryMappedFileGTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+		using value_type = char;
+		using const_iterator = const value_type*;
+
+		TemporaryFile(size_t bytes, std::default_random_engine& prng) : 
+			dataBegin(new value_type[bytes]),
+			dataEnd(dataBegin.get() + bytes)
+		{
+			// obtain temporary file name (C++17 offers std::filesystem, which we currently cannot rely on)
+			while (true) {
+				std::uniform_int_distribution<unsigned> distr;
```
avdgrinten: I'm not a fan of this approach. I must admit that it works cross-platform, though. However, platform-dependent functions like `mkostemp` provide better guarantees (i.e. that a new file is created).

networkit/cpp/io/test/MemoryMappedFileGTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+				path = "tempfile_" + std::to_string(distr(prng));
+
+				std::ifstream f(path);
+				if (!f.good()) break;
```
avdgrinten: There is a TOCTTOU condition here (the file might be created by a different program after the check succeeds).

CMakeLists.txt
```diff
@@ -75,15 +95,22 @@ endif()
 
 # NETWORKIT MODULES
 if (NETWORKIT_MONOLITH)
-	add_library(networkit SHARED networkit/cpp/networkit.cpp)
+	if (NETWORKIT_PYTHON)
```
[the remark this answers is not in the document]
manpen: The idea is to not force CMake to build a shared library if it's not necessary. The user can still select a shared library by passing `-DBUILD_SHARED_LIBS=ON` to CMake.

manpen:
Thanks @avdgrinten for your thorough feedback. I agree with all your remarks and will change the PR accordingly. Questions I did not answer directly on GitHub will get an comment in source; I'll let you know, as soon as this is done.

networkit/cpp/io/MemoryMappedFile.h
```diff
@@ -0,0 +1,54 @@
...
+struct MemoryMappedFileState;
+
+/**
+ * This class is a wrapper to os-dependend file mapping implementations.
+ * It supports to map a read-only file into virtual address space.
+ * After successful opening, pointers to the begin/end of the virtual
+ * address space are accessible via cbegin() / cend(). All file and mapping
+ * handlers are freed automatically.
+ */
+class MemoryMappedFile {
```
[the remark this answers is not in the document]
manpen: The class already supported RAII (if used with the constructor that takes a path). `open` and  `close` are only there to make it more flexible (similarly to `std::fstream`).

Regarding moving: I implemented it using move-constructors/assignments, as it's more verbose and IMHO states the intent better. Swap is then implicitly possible.

networkit/cpp/io/test/MemoryMappedFileGTest.cpp
manpen: The reason I did not use `mkostemp` is that it's not in the C++ standard and (at least) not available on Windows. Hence there is now a compiler switch with OS dependent code.

manpen:
The new commits should address all remarks; there's nothing I currently want to add to the PR.

avdgrinten:
OK. I plan to merge cmake-pr into Dev first; afterwards, this PR can also be merged into Dev.

networkit/cpp/io/MemoryMappedFile.cpp
```diff
@@ -6,6 +6,24 @@ namespace NetworKit {
 	MemoryMappedFile::MemoryMappedFile(const std::string &path) { open(path); }
 
 	MemoryMappedFile::~MemoryMappedFile() { close(); }
+
+    MemoryMappedFile::MemoryMappedFile(MemoryMappedFile&& o) noexcept {
+		*this = std::move(o);
+	}
+
+    MemoryMappedFile& MemoryMappedFile::operator=(MemoryMappedFile&& o) noexcept {
+	    close();
```
avdgrinten: This fails if someone this `x = std::move(x);` for a `MemoryMappedFile` `x`. Move assignment should almost always be implemented using `swap` to avoid such problems. Other than that, the PR is fine.

avdgrinten:

manpen:
I changed the base of this PR to Dev. Please re-open it, as I think we should keep the discussion history and not hide it behind a new PR.

avdgrinten:
Huh, I was not aware it was possible to retarget PRs. Sure, opened again.

avdgrinten:
Seems that there are some merge conflicts now. If those can be resolved, appveyor should pick up the PR and we can merge it.

networkit/cpp/io/MemoryMappedFile.cpp
manpen: I fixed it by adding a check into the move-assignment operator. This is a no-brainer, since we are not talking about hot-code. Nevertheless I would argue that this is not an issue:
 - either a user move a unique temporary, then this cannot happen
 - if one writes `x = std::move(x)` he's clearly saying on the RHS that `x` is allows to be invalidated

AFAIK for this reason STL forbids these constructs in [res.on.arguments].

PR #212, opened 2018-08-09
CMake: AppleClang support

rbange:
This PR aims at supporting the AppleClang compiler which is the Clang derivation of Apple and is shipped as the default compiler on all MacOS systems.

The core problem is that the OpenMP lib is not part of AppleClang. The installation of openmp is no real problem on MacOS (`brew install libomp`), but the linking afterwards is kind of annoying. This blog entry summaries well on how to use OpenMP with AppleClang: https://iscinumpy.gitlab.io/post/omp-on-high-sierra/

With CMake 3.12, finally support for AppleClang was introduced by implementing the correct compiler flags. Nevertheless, linking the appropriate OpenMP lib is still necessary when not using the `OpenMP::OpenMP_CXX ` target. 

So the way I implemented this, was linking omp if available and if CMake version is greater or equal to 3.12 . 
It would be also possible to link against OpenMP like `target_link_libraries(networkit PRIVATE OpenMP::OpenMP_CXX)`, but this has to be added for all targets then and mess up the code.

I additionally added a travis case for testing against their default macOS image. I would like to see this as an replacement of the MacOS clang4 test case, but the travis script might need a review anyway before or directly after the cmake merge.

This PR is not aimed to pulled as is, but more to open the discussion (especially with MacOS users), if AppleClang support is a desired feature or if everyone is fine with just installing g++8 and specifying it everywhere.


avdgrinten:
Thank you for this work. I restarted the Travis build after merging #210. If all tests pass, I will merge this.

avdgrinten:
You said that this PR is not intended to be merged as-is. However, it does not seem to be intrusive at all and supporting more compilers is always a good thing. What would be the problem with merging the PR?

rbange:
No the commit is not intrusive at all because it is designed this way. 

Nevertheless the advantage of using the shipped AppleClang vs a GNU compiler on MacOS can be discussed.

Disadvantage would be that with AppleClang every piece of code based on networkit would need `-XClang -fopenmp -lomp` flags to compile while g++-8 just needs `-fopenmp`. 

On the other side users just have to install omp (`brew install libomp` ~ 10 mb) instead of gcc (`brew install gcc`~ 350 mb) and there would be no special need to specify a compiler in cmake or python.

And in the end, if the cmake required minimum version is increased at some point of time, the changes do not matter at all because all OpenMP linking should be perfectly handled with a structure like this:  `target_link_libraries(networkit PRIVATE OpenMP::OpenMP_CXX)`

avdgrinten:

PR #215, opened 2018-08-09
Race conditions in Aux::Random

manpen:
Aux::Random offers `thread_local` references to random number generators and distributions. If the code does not recognize that the compiler supports the `thread_local` keyword, it silently uses non-`thread_local` instances. The test however only checked for GCC>4.8, Clang never used thread_local!

If a user assumes (for good reason) that its references are thread-local and the detection fails, we get race-conditions and false-sharing, i.e. slow code broken code. I did not check which algorithms are affected, but my new code certainly was.

IMHO we should expect `thread_local` from a compiler nowadays. If its not supported we should be very verbose about it, and have a compile-time error.

Another issue in Aux::Random was the treatment of seed-values, which involved inter-thread communication between via non-atomic global variables. I removed one unnecessary variable and made the remaining once atomic.

The PR also includes a small benchmark to quantify the overhead cause when using convenient functions such as `Aux::Random::probability()`. They all involve static variable, extra instructions are issued by the compiler to check proper initialization. Also, if LTO is not used, additional calls are required. All together, using something like the following is roughly 30% faster than repeated calls to `Aux::Random::*`:
```cpp
auto& gen = getURNG();
std::uniform_read_distribution distr {...};
for (.... large number ....) {
   auto prob = distr(gen);
}
```

I hence add a warning in the documentation.

### DIFF
--- a/networkit/cpp/auxiliary/Random.cpp
+++ b/networkit/cpp/auxiliary/Random.cpp
@@ -8,43 +8,60 @@
 #include <cmath>
 #include <omp.h>
 #include <limits>
+#include <atomic>
 
 #include "Random.h"
 
 namespace Aux {
 namespace Random {
 
-static bool staticSeed = false;
-static uint64_t seedValue = 0;
-static uint64_t globalSeedGeneration = 0; // global seed generation, updated on every setSeed-call
-static bool seedUseThredId = false;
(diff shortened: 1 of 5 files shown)


avdgrinten:

PR #216, opened 2018-08-10
Install to PREFIX/lib by default

avdgrinten:
As suggested by @rbange, installing into a flat directory hierarchy is not intuitive. This PR aims to fix that. We still need the possibility to install into a flat hierarchy to build the Python package.


avdgrinten:
Let's restart the build by closing/reopening the issue.

PR #217, opened 2018-08-10
CMake: AppleClang support

rbange:
Same changes as in https://github.com/kit-parco/networkit/pull/212, but now resubmitted against Dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,14 @@ matrix:
         - brew install libomp
         - if brew outdated | grep -q python; then brew upgrade python; fi
 
+    - name: macOS, AppleClang
+      env: CC=cc CXX=c++
+      os: osx
+      sudo: false
+      before_install:
+        - brew install libomp
+        - brew upgrade cmake # cmake on osx images (xode9.4) is on 3.11.4 - 3.12+ required 
+
 
     - name: C++14 conformance
       env: CC=gcc-8 CXX=g++-8 CXX_STANDARD=14
@@ -91,18 +99,11 @@ matrix:
       env: CC=gcc-8 CXX=g++-8 CXX_STANDARD=17
       os: linux
       addons: *gcc8
-  
+
   allow_failures:
     - name: macOS, modern Clang
     - name: macOS, legacy Clang 4
-
-    - name: MacOS default image + libomp + cmake >=3.12
-      env: CC=cc CXX=c++
-      os: osx
-      sudo: false
-      before_install:
-          - brew install libomp
-          - brew upgrade cmake # cmake on osx images (xode9.4) is on 3.11.4
+    - name: macOS, AppleClang
 
 script:
  - $CXX --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,9 +96,18 @@ matrix:
     - name: macOS, modern Clang
     - name: macOS, legacy Clang 4
 
+    - name: MacOS default image + libomp + cmake >=3.12
+      env: CC=cc CXX=c++
+      os: osx
+      sudo: false
+      before_install:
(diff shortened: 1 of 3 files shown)


avdgrinten:
I do not think the slightly increased complexity in CMakeLists.txt is hard to maintain. Not having to install a new compiler seems to be a real improvement for macOS users. Do you think the PR can be merged as-is or do you still want to change something?

rbange:
From a technical point of view I have nothing to change, but for me there are still 2 open points:

1. Should we remove the compiler detection from the `setup.py`? It was initially introduced to tackle the problem of CMake + AppleClang
2. I think the travis script is expanding too hard. In contrast to Linux, it is quite common to work with the latest versions of compiler and programs on MacOS. I would recommend to remove the MacOS clang4 test and pump up the other MacOS test up to g++-8 instead of 6

rbange:
PR is rebased to https://github.com/kit-parco/networkit/commit/07aceae5a180f2232c9f16eba700c2f0b718e465 (MSVC commit)

avdgrinten:
Thanks for the rebase. Yes, the Travis script should be reduced, but that should be done in a different PR. The same applies for the compiler detection - but I am not entirely sure that we should remove that. I will accept the PR in this state (once the failing tests on Dev are fixed), IMHO it is a great improvement over the current state. Thank you for your work.

rbange:
While looking into this PR, I realised that AppleClang would never be recognised by the determination function in `setup.py`. So if additional GNU or Clang compiler is installed, the installation would fail. The new commit first checks if a GNU/Clang compiler is installed and if not, tries to work with AppleClang.

So when I saw travis failing, I looked into the log and saw that the our own compiler identification never worked in the first place on travis. It just passed an empty "-DCMAKE_CXX_COMPILER=" flag and cmake fell back to the CXX environment variable. I don't know why the compiler identification fails on Travis, because I cannot recreate the bug. Nevertheless, i recreated the old behaviour in 
https://github.com/kit-parco/networkit/pull/217/commits/054b8ec17961e676188513abddca67b7cab5272 by using the CXX variable if it exists (which should be an expected behaviour - e.g. cmake and several other build systems do this by default)

btw Travis fails for MacOS+gcc6 build as trying to be fixed here: https://github.com/kit-parco/networkit/pull/229

avdgrinten:
Maybe we should just use `NETWORKIT_OVERRIDE_CXX` in the Travis script then? Is there a need to support `CXX` and `NETWORKIT_OVERRRIDE_CXX` at the same time?

avdgrinten:
Closing/reopening the commit to let Travis re-parse the configuration.

avdgrinten:
One question is whether the new Travis configuration builds as intended. If it does, this PR can be merged.

rbange:
> Maybe we should just use `NETWORKIT_OVERRIDE_CXX` in the Travis script then? Is there a need to support `CXX` and `NETWORKIT_OVERRRIDE_CXX` at the same time?

Well, I am not sure why we need a `NETWORKIT_OVERRRIDE_CXX` in the first place, because setting `CXX` should be the most common way to specify a compiler (e.g. also accepted by cmake). But yes, theoretically we could specify `NETWORKIT_OVERRRIDE_CXX` instead of `CXX` in Travis

avdgrinten:
I thought we were already using `NETWORKIT_OVERRIDE_CXX` everywhere. Turns out that we do not; and it does not make sense anyway. I guess `NETWORKIT_OVERRIDE_CXX` can be removed.

This PR is ready to be merged. Let's wait for the CI to finish first, though.

avdgrinten:
I am merging this. Thank you for your work, Roman.

PR #221, opened 2018-08-10
Feature Parallel ErdosRenyi Generator

manpen:
Apologies for issuing this PR without a description.

Remark: NetworKit calls G(n,p) graphs Erdos-Renyi-Graphs. It is my understanding that G(n,p) are infact Gilbert graphs, while ER introduced the G(n,m) model. For consistency, I will keep on refering to G(n,p) as ER.

> The G(n, p) model was first introduced by Edgar Gilbert in a 1959 paper studying the connectivity threshold mentioned above.[3] The G(n, M) model was introduced by Erdős and Rényi in their 1959 paper. As with Gilbert, their first investigations were as to the connectivity of G(n, M), with the more detailed analysis following in 1960.  https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93R%C3%A9nyi_model

The primary reason for this PR is the fact that the GraphBuilder benchmarks use ER graphs. Unfortunately it relied on a quadratic time sampling that required (for the parameters used) expected 1000 random floats for each edge generated. Hence, the runtime of the benchmarks was primarily dominated by the generation of test data and did not actually measure the GraphBuilder itself.

After migrating away from the quadratic time sampling to the parallel ER model, the runtime dropped from ~1900ms to ~30ms. This includes sampling the data and maintaining the GraphBuilder. I hence increased the problem size by a factor of 100 to get meaningful data. Also observe that previously the parallel benchmark did not measure a parallel structure as it passed `parallel=false` to the GraphBuilder. I also moved the benchmarks from `generators/test` to `graph/test` as the GraphBuilder is part of the latter.

As the benchmark cannot accept a `Graph` data structure, the ER generator itself is implemented as a StreamGenerator; it has a `forEach` callback and expects (e.g.) a lambda to be called for each edge. It basically uses the linear time algorithm previously implemented in `ErdosRenyiGenerator` with a few tweaks giving a sequential speed-up between 3x and 5x; it scales almost perfectly, as the hot-section runs pleasingly parallel.

I now demoted the original ErdosRenyiGenerator to a wrapper to the Enumerator. Here the speed-ups rather small, since the runtime is dominated by the GraphBuilder.


networkit/cpp/generators/ErdosRenyiEnumerator.h
```diff
@@ -0,0 +1,258 @@
...
+
+#include "../Globals.h"
+#include "../auxiliary/Random.h"
+#include "../auxiliary/SignalHandling.h"
+
+namespace NetworKit {
+
+/**
+ * Generates a stream of edges of a G(n,p) graph. The edges are not
+ * written two memory, but emitted only via usual @a forEdges semantics
```
rionda: "two" -> "to"

manpen: Thanks!

manpen:
I forgot to ask my important question:

The old ER generator generates only edge (u, v) with u > v. While I agree with this for *undirected* graphs, a *directed* G(n,p) graph may also contain edges (v < u) and esp. self-loops. The old generator was unable of emitting them. The new enumerator implements this, as it is the mathematically correct way to do it.

There is however code that relies on directed graphs with (u < v), e.g. `CentralityGTest.testGroupDegreeDirected` which cannot deal with self-loops.

IMHO the other code should be fixed, and the generator should produce graphs according to definition even if this breaks user-code. What's your opinion?

manpen:
What do think about the following solution:

- Deprecate ErdosRenyiGenerator (as its broken for the directed case and has a misleading name)
- Rename ErdosRenyiEnumerator (and its wrapper) to GilbertGNPEnumerator/GilbertGNPGenerator
- (later) Introduce an ErdosRenyiGNMGenerator (e.g. the simple parallel algorithm in https://arxiv.org/abs/1710.07565)

manpen:
Just to let you know: I included all changes discussed last week (optional support to generated directed graphs without self loops and support for the extreme cases of p=0.0 and p=1.0); in case of p>0.5 there's also a further optimisation that the implementation does not draw a random number of each edge but for each edge NOT included (which are expected to be less). This does not yield an asymptomatic improvement, but is faster in practice.

I consider all tests passing, as the one failing test case in Travis is not related to this PR and hence feel it is completed from a technical point of view. I did however not change the interface of the Python code to expose the flag for self-loop-free instances, as the naming issue is still open.

avdgrinten:
I did not have the time to look at this PR yet. It also has a few merge conflicts. Could you resolve the conflicts while I review the PR?

manpen:
Yes, sure.

avdgrinten:
We will probably not rename the generator to GilbertGenerator (or similar). Even if ER is technically the wrong name, it has been in use for too long to change it now. If we want to add a true ER generator, we will have to use a different name for that.

networkit/cpp/generators/test/GeneratorsBenchmark.cpp
```diff
@@ -38,15 +48,12 @@ TEST_F(GeneratorsBenchmark, benchmarkGraphBuilder) {
 
 	// prepare a random generator for each possible thread
 	int maxThreads = omp_get_max_threads();
-	std::vector< std::function<double()> > randomPerThread;
```
avdgrinten: I assume that the reason why we do not use NetworKit's PRNG is that we want this particular distribution?

networkit/cpp/generators/test/GeneratorsBenchmark.cpp
```diff
@@ -26,87 +27,58 @@
...
-    explicit UniformReal(int seed) : gen(seed) {}
-    double operator()() {return dist(gen);}
-
-private:
-    std::mt19937_64 gen;
-    std::uniform_real_distribution<double> dist{0.0, std::nexttoward(1.0, 2.0)};
-};
-
-
 TEST_F(GeneratorsBenchmark, benchmarkGraphBuilder) {
```
avdgrinten: I understand that the old code did not really benchmark the GraphBuilder. Does the new code do that? Is the running time actually dominated by the GraphBuilder?

networkit/cpp/graph/test/GraphBuilderBenchmark.h
```diff
@@ -20,7 +20,7 @@ class GTestBenchmark : public testing::Test {
 public:
 	static void measureInMs(std::function<int()> func, int iterations = 1) {
 		Aux::Timer timer;
-		std::vector<uint64_t> runningTimes(iterations);
+		std::vector<long long unsigned> runningTimes(iterations);
```
avdgrinten: What is the problem here? Nitpick: I would prefer `unsigned long long` to `long long unsigned`.

networkit/cpp/generators/test/GeneratorsBenchmark.cpp
manpen: This point is not important anymore, since the code was removed a couple of commits later. (nevertheless, I think it's okay to not rebase it, since it documents the development process. This is one of those cases, where in the future we should have a clear Guide on how to do it.) The reason I did not use `Random::probability` at that point, was that I was not yet clear on the performance impact (which is now measured and document in  69dcccc).

networkit/cpp/generators/test/GeneratorsBenchmark.cpp
manpen: Yes it does; for detailed benchmarks, we can use the `GraphBuilderBenchmark->benchmarkGraphBuilderBaseline` benchmark, which generates an identical graph w/o the GraphBuilder. Subtracting that time from a GraphBuild benchmarks then yields a good approximation of the GraphBuilder's runtime. If I remember correctly, the Enumerator's runtime is completely insignificant.

networkit/cpp/graph/test/GraphBuilderBenchmark.h
manpen: Thank you for asking. The underlying issue is that the old code uses printf(...) to format the output and `uint64_t` is incompatible to `%llu` -- this causes warnings. I do not recall why I fixed it this way (probably a hack, which made it into the commit; sorry!) and will change it back to `uint64_t` and replace the `printf` with `ostream`.

manpen:
A small question regarding the name: is it okay if I rename the Enumerator part into `GilbertEnumerator`. This is new code, and the (correct) name does not break existing code; the only odd part is then that the `ErdosRenyiGenerator` uses the `GilbertEnumerator` ... I feel this is the right way to do it, but I completely understand if you prefer `ErdosRenyiEnumerator`.

avdgrinten:
I think I'd still prefer `ErdosRenyiEnumerator` for consistency. If ER generator did not exist yet it would be a different matter...

manpen:
This failed CI test is again a sporadic Travis bug during system setup. I dealt with all remarks and consider this PR ready. Please let me know if there's anything else.

avdgrinten:
LGTM

PR #224, opened 2018-08-14
Fix virtual warnings

manpen:
Clang rightfully complains, that the virtual abstract class `SelectiveCommunityDetector` inherited by `PageRankNibble` does not contain a virtual destructor despite it's required in the Unit Test. This PR adds it.

Also, Clang thinks `Curveball::isParallel` should be annotated with `override`. So why not ;)


avdgrinten:
Thank you. I do not really like the virtual destructors warning (it's only a problem if you delete through a pointer to the base class) but fixing it also does not require much effort. Merging.

Issue #214, opened 2018-08-09
Header files for tests?


manpen:
Most of the test cases consist of a header file along the lines of 

```cpp
/*
 * FooBar.h
 *
 *  Created on: look, it's a date
 *      Author: VIP
 */

#ifndef FOOBAR_H_
#define FOOBAR_H_

#include <gtest/gtest.h>
class FooBarGTest: public testing::Test {};

#endif /* FOOBAR_H_ */
```

The only two meaningful lines here are 
```cpp
#include <gtest/gtest.h>
class FooBarGTest: public testing::Test {};
```

I do not see any advantage from having these header files, since the unit tests are meant to be self-contained. On the contrary: they clutter any directory listing, are a nuisance to create, and typically no one reads them.

Am I missing something? Do they have a purpose? Are they required? 
If not, should we get rid of them (or at least do not create them for new code)?

avdgrinten:
I agree. Those headers are useless. We can remove them in the future and they are definitely not needed for new code.

avdgrinten:
This has been resolved.

PR #225, opened 2018-08-15
[Low Prio] Remove Test Headers

manpen:
While this PR targets many files, it's straight-forward: as discussed in #214, it removes the header files of unit tests to reduce the clutter and motivate people to write more tests.

There are three commits targetting the tests, but do something different (if you want, I can separate it into another PR, but I do not really see the point):

- 9e5020c Removes a "four-liner" benchmark for `GraphBuilder` by reading a graph file. Since the file is not included in the repo, the benchmark is not reproducible, nor do I think, that parsing a file (where a significant amount of not-fully deterministic work is done somewhere else and not accounted for) is a good benchmark.

- b1b3853 Fixes a bug in AuxGTest: The test for `Aux::Random::index` used a `using namespace Aux::Random` and then did not fully qualify `index`. Therefore, after rearranging headers `index(10)` was interpret as the number 10 stored as data type index. The unittest therefore now uses fully-qualified names everywhere.

- 9dd7d64 In non-monolithic builds, each individual test was previously linked again `gtest_main`. This commit introduces a new library `networkit_gtest_main` to use the `Unittests-X.cpp` infrastructure. This filtering now allows to run `ctest` also in non-monolithic builds and is thus tested with Travis.


avdgrinten:
The PR looks fine. I'm restarting the build by closing/reopening the PR.

avdgrinten:
All relevant Travis checks passed, I'm merging this. Thanks again for the code!

PR #226, opened 2018-08-16
Group degree fix

eug93:
Signed-off-by: Eugenio Angriman <eugenio.angriman@informatik.hu-berlin.de>


avdgrinten:
The Windows build fails because of the data type of a parallel for loop. Please use the new `omp_index` type as in the rest of the code base.

avdgrinten:
The remaining problems are not related to the group degree algorithm. I would prefer to rebase (i.e. squash) this patch series into a single commit. (With a single commit message to explain all changes.) Can you do that or do you want me to merge it as-is?

eug93:
I would also prefer to squash it, let me do that before the merge.

PR #229, opened 2018-09-07
travis: Fix build by avoiding High Sierra

avdgrinten:
The High Sierra image on Travis is broken. Just fix the xcode (and macOS) version to xcode 8 for now to work around the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,13 @@ matrix:
 
     - env: CC=gcc-6 CXX=g++-6
       os: osx
+      osx_image: xcode8
       sudo: false
       before_install:
         - brew update
         - brew cask uninstall --force oclint # See https://github.com/travis-ci/travis-ci/issues/8826 for details.
         - brew install gcc@6
-        - brew upgrade python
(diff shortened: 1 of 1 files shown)


rbange:
Mhh, at least the clang build works fine with at my PR(https://github.com/kit-parco/networkit/pull/217). 

While this issue with the travis image also exists for gcc8, is there any particular reason why networkit is tested with gcc6 instead of gcc8 on MacOS? If its just for legacy reasons, you might consider increasing on the latest version, since support for legacy compiler is ensured by the linux tests.

rbange:
Because this PR also addresses the long homebrew update times: 

The default images get updated quite regularly, so `brew update`, is not really necessary if the existing version works fine (existing python is 3.6 so should work well too).

So if the existing issue is fixed with the latest image and the test environment is pushed back to the latest version, 
```
      before_install:
         - brew cask uninstall --force oclint # See https://github.com/travis-ci/travis-ci/issues/8826 for details.
         - brew install gcc@6
```
should be sufficient.

avdgrinten:
Finally, this can be merged.

PR #238, opened 2018-10-02
Kadabra algorithm for betweenness approximation

eug93:
Implementation of the Kadabra algorithm by Borassi et al.: KADABRA is an ADaptive Algorithm for Betweenness via Random Approximation (2016).

### DIFF
--- a/networkit/cpp/auxiliary/PQVector.h
+++ b/networkit/cpp/auxiliary/PQVector.h
@@ -1,0 +1,100 @@
+/*
+ * PQVector.h
+ *
+ *  Created on: 21.09.2018
+ *      Author: Eugenio Angriman
+ */
+
+#ifndef PQVECTOR_H_
+#define PQVECTOR_H_
+
+#include <algorithm>
+#include <vector>
+
+#include "../auxiliary/Log.h"
+
+namespace Aux {
+/*
+ * A simple prio queue for Kadabra that allows inspecting the ranking.
+ */
+class PQVector {
+private:
+	std::vector<std::pair<uint64_t, double>> elements;
+	std::vector<uint64_t> position;
+	uint64_t virtualSize;
+	const uint64_t size;
+	const uint64_t maxKey;
+
+public:
+	PQVector(const uint64_t size, const uint64_t maxKey);
+	void insert(const uint64_t newElement, const double newValue);
+	double getValue(const uint64_t i) const { return elements[i].second; }
+	uint64_t getElement(const uint64_t i) const { return elements[i].first; }
+	uint64_t getSize() const { return virtualSize; }
+	void clear();
+};
+
+inline PQVector::PQVector(const uint64_t size, const uint64_t maxKey)
+		: size(size), maxKey(maxKey) {
+	if (maxKey < size) {
+		throw std::runtime_error("maxKey must be bigger than the size.");
+	}
+	clear();
+}
+
+inline void PQVector::clear() {
+	elements.resize(size);
+	position.resize(maxKey);
+	uint64_t i;
+	for (i = 0; i < size; ++i) {
+		elements[i] = std::make_pair(i, 0.);
+		position[i] = i;
+	}
+	for (i = size; i < maxKey; ++i) {
+		position[i] = i;
+	}
+	virtualSize = 0;
+}
+
+inline void PQVector::insert(const uint64_t newElement, const double newValue) {
+	uint64_t ub =
+			std::upper_bound(
+					elements.begin(), elements.begin() + virtualSize, newValue,
+					[&](const double val, const std::pair<uint64_t, double> pair) {
+						return val > pair.second;
+					}) -
+			elements.begin();
+
+	uint64_t oldPos;
+	// We assume that if the same key is inserted again, its value will be
+	// greater than before (i.e., oldPos >= ub).
+	if (ub < size) {
+		oldPos = std::min(position[newElement], size - 1);
+		if (position[newElement] < size) {
+			assert(elements[ub].second <= newValue);
+		}
+		if (virtualSize < size && oldPos >= virtualSize) {
+			++virtualSize;
+		}
+
+		if (ub < oldPos) {
+			std::rotate(elements.begin() + ub, elements.begin() + oldPos,
+									elements.begin() + oldPos + 1);
+
+			if (oldPos == size - 1) {
+				++position[elements[ub].first];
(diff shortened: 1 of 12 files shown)


avdgrinten:
The code seems to segfault on Windows.

eug93:
The issue concerned`PQVector.h`, the indexes for the vector rotation were not handled correctly in the case where the position of an element does not need to be updated in the priority queue. Now it is works correctly.

networkit/cpp/graph/Graph.cpp
```diff
@@ -718,6 +718,32 @@ void Graph::removeAllEdges() {
...
+		outDeg[u] = 0;
+		outEdges[u].clear();
+		if (this->weighted) {
+			outEdgeWeights[u].clear();
+		}
+		if (this->isDirected()) {
+			inDeg[u] = 0;
+			inEdges[u].clear();
+			if (this->weighted) {
+ 				for (node u : nodes()) {
```
eug93: This is a loop with just some assertions. Does the optimizer discard it?

networkit/cpp/auxiliary/PQVector.h
```diff
@@ -0,0 +1,100 @@
...
+#include <algorithm>
+#include <vector>
+
+#include "../auxiliary/Log.h"
+
+namespace Aux {
+/*
+ * A simple prio queue for Kadabra that allows inspecting the ranking.
+ */
+class PQVector {
```
avdgrinten: This class should have some explanation about why it is useful. Something like:

> This PQ stores at most k elements; if more than k elements are inserted, elements with smallest keys are dropped. The PQ is implemented on top of a vector, thus the insert operation takes O(k) time.

I am not sure if we should call this class really a "priority queue" as it has no `extractMax()`. Maybe `SortedList` would be more appropriate?

networkit/cpp/centrality/KadabraBetweenness.h
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	void fillPQ() {
+		for (count i = 0; i < n; ++i) {
+			top->insert(i, approxSum[i]);
+		}
+	}
+};
+
+inline void KadabraBetweenness::fillResult() {
```
avdgrinten: Why is this in a header file?

networkit/cpp/graph/Graph.h
```diff
@@ -645,6 +645,8 @@ class Graph final {
 	 */
 	void removeEdge(node u, node v);
 
+	void removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet);
```
avdgrinten: This method should have a comment to describe what it does. You can just copy the one from the Python interface.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1434,4 +1435,67 @@ TEST_F(CentralityGTest, testGroupCloseness) {
 
 	EXPECT_NEAR(gc.scoreOfGroup(apx), 1.0, 1e-5);
 }
+
+TEST_F(CentralityGTest, testKadabraAbsolute) {
```
avdgrinten: The two tests here can fail e.g. if we change the random number generator, right? Even though it is unlikely, there could be more than `delta * n` errors.

networkit/cpp/centrality/KadabraBetweenness.cpp
```diff
@@ -0,0 +1,526 @@
...
+	// Getting diameter upper bound
+	int32_t diameter = diam.getDiameter().second;
+	omega =
+	    0.5 / err / err * (std::log2(diameter - 1) + 1 + std::log(0.5 / delta));
+
+	const count tau = omega / startFactor;
+
+	if (unionSample == 0) {
+		if (absolute) {
+			unionSample = n;
```
avdgrinten: Let's add a comment that this is different from the original code.

networkit/cpp/centrality/KadabraBetweenness.cpp
```diff
@@ -0,0 +1,526 @@
...
+	for (count i = 0; i < unionSample; ++i) {
+		node v = status.top[i];
+		deltaLGuess[v] = std::exp(-b * errL[i] * errL[i] / bet[i]) +
+		                 delta * balancingFactor / 4. / (double)n;
+		deltaUGuess[v] = std::exp(-b * errU[i] * errU[i] / bet[i]) +
+		                 delta * balancingFactor / 4. / (double)n;
+	}
+}
+
+void KadabraBetweenness::computeApproxParallel(const bool normalize) {
```
avdgrinten: We sometimes need a sequential version of this method (e.g. in line 343), right? Why don't we add a parameter to control if it parallel as in `getStatus()`?

networkit/cpp/centrality/KadabraBetweenness.h
```diff
@@ -0,0 +1,237 @@
...
+
+#include <atomic>
+
+#include "../auxiliary/PQVector.h"
+#include "../base/Algorithm.h"
+#include "../graph/Graph.h"
+
+namespace NetworKit {
+
+class Status {
```
avdgrinten: In contrast to the original code, we only use `Status` on thread zero. Can we get rid of this? Or would that require too much effort?

networkit/cpp/centrality/KadabraBetweenness.cpp
eug93: We use this function in line 315 and 358, both calls are executed in a sequential section of the code. I think we can keep it as it is.

networkit/cpp/centrality/KadabraBetweenness.h
eug93: Yes we can get rid of it but I doubt that it would bring significant performance improvements, so I would prefer to leave it as a future development.

avdgrinten: OK, let's do this after the initial merge.

networkit/cpp/centrality/KadabraBetweenness.cpp
avdgrinten: I see. OK, let's keep it that way, then.

networkit/cpp/graph/Graph.h
eug93: Addressed in f096e901f296a2e293463c6e32f910172fcd3882

Issue #236, opened 2018-09-28
Reducing CI latency


avdgrinten:
Our Travis configuration needs quite a lot of time to finish execution. I would like to reduce the latency of the build.

Current for every OS/compiler we do the following:

- Install the Python package (includes building with CMake).
- Run the Python tests.
- Re-build NetworKit core in debug mode.
- Run the C++ tests in debug mode.
- Re-build NetworKit core in release mode.
- Run the C++ tests in release mode.
- Re-build NetworKit core in non-monolithic release mode.
- Run the C++ tests in non-monolithic release mode.

IMHO this is a bit too much (a single execution of that script takes 45 min on Travis). What I propose to do instead is:

- Build the C++ library in release mode.
- Build the Python package (potentially re-using the C++ library built before).
- Run the C++ tests.
- Run the Python tests.

These changes would bring down the Travis latency to about 15 min (= the time required for a C++ build + running the tests). Such a script could be run for multiple OS/compiler combinations.

Assuming that those things work, more exotic configurations (e.g. non-monolithic builds, debug builds or C++14/17 builds) are likely to work, too. Thus, we do not need to test for them on many OS/compiler configurations - one or two additional configurations that check them should be sufficient. Increasing the number of configurations while reducing the work/configuration makes sense as it increases the amount of parallelism in the CI pipeline.

Any objections or comments? Pinging @manpen as he wrote the majority of the current configuration.

manpen:
I agree in most parts, but I think it's important to build the library in both in Release and Debug mode, as the code is not identical (due to preprocessor directives) and we should make sure that either compiles.
Hence I suggest to build and test the C++ lib in debug mode (as it include more assertions).
Then we can build the Python library as release (as it's the default).
In my opinion we should keep both compilation runs separate (i.e. do not reuse stuff in PIP) to make sure that the PIP are doing okay.

avdgrinten:
Good point about keep the debugging configurations. In PR #237 I kept them as separate configurations.

I still suggest to teach `setup.py` to use an existing `libnetworkit.so` so that we do not have to build the C++ library twice. `setup.py`/`pip` would still build the `_NetworKit.something.so` so that this part of the build system would still be tested. Doing this would almost halve the time required for most configurations.

avdgrinten:
After #240, Travis takes about 20 minutes for a full build. 10 minutes are required to compile NetworKit (in release mode). About one minute could be avoided if TLX was not built twice (once for the core library and once for the Python extension).

[...]

PR #240, opened 2018-10-04
Support Python build with external NetworKit core library

avdgrinten:
This is another PR to reduce Travis' build times (see #236); however, the functionality implemented in this PR might also be interesting on its own:

We extend the build system to allow building the Python extention without building the NetworKit core (i.e. C++) library. This is done by passing -DNETWORKIT_BUILD_CORE=OFF to cmake and requires an external prebuilt copy of the core library. Our Travis script is modified to take advantage of this feature.

I should remark that after this PR, NetworKit is built as a shared library by default. This can be deactivated using -DNETWORKIT_STATIC but is no longer coupled to -DNETWORKIT_PYTHON.

This PR reduces Travis build times from [3 h 50 min](https://travis-ci.org/kit-parco/networkit/builds/436732224) to [3 h 5 min](https://travis-ci.org/kit-parco/networkit/builds/436805470); it also decreases latency and the maximal time for a single build considerably.

Pinging @manpen as he gave some comments in #236.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -145,14 +145,20 @@ script:
  # cython is required because git does not contain _NetworKit.
  # ipython is required for tests.
  - pip3 install cython ipython
- - NETWORKIT_PARALLEL_JOBS=2 travis_wait pip3 install -e .
- - python3 -m unittest discover -v networkit/test/
 
- - mkdir release_test && cd "$_"
+ # First, build the C++ core library and run C++ tests.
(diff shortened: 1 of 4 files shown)


avdgrinten:
Apparently, Windows requires static builds. I will push a commit to add -DNETWORKIT_STATIC=ON to AppVeyor.

avdgrinten:
This has been open long enough. There are no comments, so I am going to merge it.

PR #243, opened 2018-10-09
Removed deprecated class PrioQueueForInts and its tests

eug93:
The `PrioQueueForInts` class is deprecated and there is no NetworKit class using it.


avdgrinten:
This is uncontroversial. I will just merge it.

PR #246, opened 2018-10-11
Removed allowed failures on Travis

eug93:
Travis macos image seems to have been fixed so we shall not allow those builds to fail anymore.


avdgrinten:
This is uncontroversial. Let's merge this.

PR #247, opened 2018-10-11
Removed deprecated header

eug93:
This header file is deprecated and not used by any other class in NetworKit.


avdgrinten:
Thank you.

PR #250, opened 2018-10-24
Feature/closeness disconnected graphs

eug93:
Now Closeness also supports disconnected graphs using the general definition (see [here](https://arxiv.org/pdf/1308.2140.pdf) for further details).

Also, re-defined the C++ booleans in `_NetworKit.pyx` as `bool_t`. Previously the line: https://github.com/kit-parco/networkit/blob/9cb72da6db56a368a0a63b5ccf2f1c04a29e5773/networkit/_NetworKit.pyx#L27
was overriding the python `bool` as a cpp bool, and thus operations like `isinstance(foo, bool)` were not possible.

### DIFF
--- a/networkit/cpp/centrality/Closeness.h
+++ b/networkit/cpp/centrality/Closeness.h
@@ -12,6 +12,8 @@
 
 namespace NetworKit {
 
+enum ClosenessVariant { standard = 0, generalized = 1};
+
 /**
  * @ingroup centrality
  */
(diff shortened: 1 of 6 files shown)


networkit/cpp/centrality/Closeness.h
```diff
@@ -12,6 +12,8 @@
 
 namespace NetworKit {
 
+enum ClosenessVariant { standard, generalized };
```
avdgrinten: Can we make that a `enum class`?

networkit/cpp/centrality/Closeness.h
```diff
@@ -28,11 +30,10 @@ class Closeness : public Centrality {
 	 * @param normalized Set this parameter to <code>false</code> if scores should
 	 * not be normalized into an interval of [0, 1]. Normalization only for
 	 * unweighted graphs.
-	 * @param	checkConnectedness	turn this off if you know the graph is connected
 	 *
 	 */
 	Closeness(const Graph &G, bool normalized = true,
-	          bool checkConnectedness = true);
```
avdgrinten: Please keep the old constructor so that we do not have a breaking API change. The old constructor can delegate to the new one. Add `checkConnectedness` to the new constructor to make that functionality optional.

avdgrinten:
Looks good now.

PR #251, opened 2018-10-29
Fix the documentation build

avdgrinten:
The documentation of NetworKit has been neglected in the last few months. Now that the build system is stable, we should fix this. This PR cleans up a few old files, and, more importantly, adds a `docs` target for CMake so that we can easily build the docs.


avdgrinten:
There are probably not so many contributors that care about the build system of the documentation so I will just merge this to enable follow-up PRs.

PR #263, opened 2018-11-07
Cover: add extend() method similar to Partition

michitux:


eug93:
This looks fine, I am merging it.

PR #264, opened 2018-11-07
Partition.mergeSubsets: return id of merged part in Python

michitux:
Note that the documentation of the method actually already promised the return value, just that no value was returned.


avdgrinten:
Looks good. (AppVeyor problems are unrelated to this PR.)

PR #265, opened 2018-11-08
Classes that extend Algorithm now use the 'assureFinished()' method.

eug93:
NetworKit classes extending `Algorithm` should use the `assureFinished()` method instead of checking every time whether `hasRun` is true (and then eventually throw a runtime error) before returning results.


eug93:
Merged as discussed with @avdgrinten

PR #268, opened 2018-11-08
Implement equality of partitions

eug93:
Based on #157 by @michitux. Same reasons as in #267.


avdgrinten:
PR title and commit message mention hashing but hashing is not implemented. Can you edit that?

PR #270, opened 2018-11-18
Max in/out degree

eug93:
New parallel methods in the Graph class to get the maximum in/out degree.


avdgrinten:
AppVeyor tests fail because MSVC does not support the OpenMP reduction. Please disable parallelism on Windows using `#ifdef` (@manpen what is the `#ifdef` that we should use for that?).

eug93:
Thank you, I just used that and it worked.

PR #271, opened 2018-11-19
Fix/travis allowed failures and new macOS builds

eug93:
Now the travis script for osx builds on osx 10.13/12 (xcode 10.1 and 9.2), with gcc 8, llvm 7, and Apple clang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,90 @@ env:
 
 matrix:
   include:
+    - name: macOS 10.13, llvm 7
+      env: CC=/usr/local/opt/llvm@7/bin/clang
+      env: CXX=/usr/local/opt/llvm@7/bin/clang++
+      os: osx
+      osx_image: xcode10.1
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - llvm@7
+            - libomp
+            - python
+          update: true
+
+    - name: macOS 10.13, GCC 8
+      env: CC=/usr/local/opt/gcc@8/bin/gcc-8
+      env: CXX=/usr/local/opt/gcc@8/bin/g++-8
+      os: osx
+      osx_image: xcode10.1
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - gcc@8
+            - python
+          update: true
+      before_install:
+        - brew link gcc@8
+
+    - name: macOS 10.13, Apple Clang
+      env: CC=cc
+      env: CXX=c++
+      os: osx
+      osx_image: xcode10.1
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - libomp
+            - python
+          update: true
+
+    - name: macOS 10.12, llvm 7
+      env: CC=/usr/local/opt/llvm@7/bin/clang
+      env: CXX=/usr/local/opt/llvm@7/bin/clang++
+      os: osx
+      osx_image: xcode9.2
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - llvm@7
+            - libomp
+            - python
+          update: true
+
+    - name: macOS 10.12, GCC 8
+      env: CC=/usr/local/opt/gcc@8/bin/gcc-8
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,6 +7,74 @@ env:
...
+
+    - name: macOS 10.13, GCC 8
+      env: CC=/usr/local/opt/gcc@8/bin/gcc-8
+      env: CXX=/usr/local/opt/gcc@8/bin/g++-8
+      os: osx
+      osx_image: xcode10.1
+      sudo: false
+      before_install:
+        - brew update
+        - brew install libomp
```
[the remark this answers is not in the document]
eug93: In this image gcc-8 is preinstalled. However, this may lead to errors if one just updates the xcode image without checking the gcc version. In c0f5510f43b03bd98d1d3af9f7b72731937afcaf I am using the homebrew addon as you suggested and added gcc-8 to the packages to be installed.

.travis.yml
```diff
@@ -7,6 +7,74 @@ env:
...
+
+    - name: macOS 10.12, Apple Clang
+      env: CC=cc CXX=c++
+      os: osx
+      osx_image: xcode9.2
+      sudo: false
+      before_install:
+        - brew update
+        - brew install libomp
+        - brew upgrade cmake
```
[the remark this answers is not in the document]
eug93: True, addressed in c0f5510f43b03bd98d1d3af9f7b72731937afcaf via the travis homebrew addon.

.travis.yml
```diff
@@ -7,6 +7,94 @@ env:
...
+      env: CC=/usr/local/opt/gcc@8/bin/gcc-8
+      env: CXX=/usr/local/opt/gcc@8/bin/g++-8
+      os: osx
+      osx_image: xcode10.1
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - gcc@8
+            - libomp
```
rbange: can be removed here

.travis.yml
```diff
@@ -7,6 +7,94 @@ env:
...
+      os: osx
+      osx_image: xcode9.2
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - llvm@7
+            - libomp
+            - python
+            - cmake
```
rbange: ```cmake``` could be removed here

.travis.yml
```diff
@@ -7,6 +7,94 @@ env:
...
+      os: osx
+      osx_image: xcode9.2
+      sudo: false
+      addons:
+        homebrew:
+          packages:
+            - gcc@8
+            - libomp
+            - python
+            - cmake
```
rbange: in this build ```libomp``` as well as ```cmake```could be removed

.travis.yml
eug93: Addressed in c0f5510

.travis.yml
eug93: Addressed in c0f5510

.travis.yml
eug93: Addressed in 24e2c62e9f851be72a5e3bd4a19d8eb7d744e2a5

eug93:
About the first point of https://github.com/kit-parco/networkit/pull/271#issuecomment-440222076: I would agree that it is worth to avoid updating python in every build. Are you suggesting to remove it from the homebrew package lists?

rbange:
Well that kinda depends on which version is pre-installed on the respective images. Sadly this is not listed anywhere so it comes down to experimenting. 
I would assume that that the Xcode 10.x images have python 3.7.x pre-installed and the xcode9.x ones have python 3.6.x. In my opinion it could be removed in every build.

At least the conservative way would be to keep the entries in the xcode10 builds (as an upstream build) and remove them from the xcode9 builds (because they are legacy builds to ensure compatibility)

eug93:
@rbange let's move the discussion https://github.com/kit-parco/networkit/pull/271#issuecomment-440229732 in #275

PR #275, opened 2018-11-20
Avoid travis python update

eug93:
Avoid to update homebrew and python saves a few minutes on every osx build. Removed macOS 10.12 build with Apple Clang as it required cmake to be updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ matrix:
           packages:
             - llvm@7
             - libomp
-            - python
-          update: true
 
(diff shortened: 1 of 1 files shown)


eug93:
Seems like that for macOS 10.12 we cannot avoid to update homebrew and install python (python3 is not even installed by default).

eug93:
@rbange Thank you for pointing that out. I would agree that to maintain consistency with the other boulds we shall keep gcc-8 also for macOS 10.12.

PR #278, opened 2018-11-27
Add community.CoverF1Similarity

michitux:
This adds an algorithm to compare two covers using the F1 score. This is a typical similarity measure used to compare the found overlapping community structure to a ground truth community structure. Each cluster is compared to the best-matching reference cluster (in terms of highest F1 score). A value of 1 indicates perfect agreement while a while of 0 indicates complete disagreement. An example where this measure is used is the following paper:

Alessandro Epasto, Silvio Lattanzi, and Renato Paes Leme. 2017. Ego-Splitting Framework: from Non-Overlapping to Overlapping Clusters. In Proceedings of the 23rd ACM SIGKDD International Conference on Knowledge Discovery and Data Mining (KDD '17). ACM, New York, NY, USA, 145-154. DOI: https://doi.org/10.1145/3097983.3098054

Note that this breaks with the tradition of having dissimilarity measures implementing `DissimilarityMeasure` with a basically empty constructor and a `getDissimilarity()`-method that does all the computation. Instead, this class implements the `LocalCoverEvaluation` interface. This is for three reasons:

* Dissimilarity is not very intuitive as in the literature all of these measures are considered as similarities.
* The interface of the dissimilarity measures does not follow the `Algorithm` interface in NetworKit as they are very old.
* Following the `LocalCoverEvaluation` interface makes it easy to generate more fine-grained evaluations by inspecting the F1 score of individual clusters. Further, this class inherits from `Algorithm`.

If you do not like the interface I'm also willing to adapt the algorithm to the traditional interface e.g. by adding a wrapper class that implements `DissimilarityMeasure`.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -5314,6 +5314,45 @@ cdef class StablePartitionNodes(LocalPartitionEvaluation):
 			raise RuntimeError("Error, object not properly initialized")
 		return (<_StablePartitionNodes*>(self._this)).isStable(u)
 
+
+cdef extern from "cpp/community/CoverF1Similarity.h":
+	cdef cppclass _CoverF1Similarity "NetworKit::CoverF1Similarity"(_LocalCoverEvaluation):
+		_CoverF1Similarity(_Graph G, _Cover C, _Cover reference) except +
+
+cdef class CoverF1Similarity(LocalCoverEvaluation):
+	"""
+	Compare a given cover to a reference cover using the F1 measure.
+	This is a typical similarity measure used to compare the found
+	overlapping community structure to a ground truth community
+	structure. Each cluster is compared to the best-matching reference
+	cluster (in terms of highest F1 score). A value of 1 indicates
+	perfect agreement while a while of 0 indicates complete
+	disagreement. An example where this measure is used is the
+	following paper:
+
+	Alessandro Epasto, Silvio Lattanzi, and Renato Paes
+	Leme. 2017. Ego-Splitting Framework: from Non-Overlapping to
+	Overlapping Clusters. In Proceedings of the 23rd ACM SIGKDD
+	International Conference on Knowledge Discovery and Data Mining
+	(KDD '17). ACM, New York, NY, USA, 145-154. DOI:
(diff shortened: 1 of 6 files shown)


avdgrinten:
The PR looks good. Regarding `LocalCoverEvaluation` vs. `DissimilarityMeasure`: It makes sense to use the `LocalCoverEvaluation` class here .For future refactoring, would it be possible to rewrite all dissimilarity algorithms to use `LocalCoverEvaluation`? My first guess would be "no", as `LocalCoverEvaluation` seems to be more expressive than `DissimilarityMeasure`. Is that correct?

michitux:
Yes, this is correct. The idea of `LocalCommunityEvaluation` is to have one score per community (i.e., a part of a `Partition` or a `Cover`). This makes sense for measures that evaluate a community e.g. concerning its size, edge density, its number of triangles or its cut or conductance to the rest of the graph. The idea is to allow a fine-grained evaluation that also considers correlations between such properties and thus allows a much more fine-grained understanding of the found community structure (or a ground truth community structure). This is kind of similar to node or edge centrality measures. Many dissimilarity (or similarity) measures for partitions and covers (like NMI or ARI), however, do not work on such a local level and their final score is not the average of one score per community, so there this design doesn't work. Note also that even in the case of F1 score there are actually several different variants how this per-community score is turned into one global score. First of all, some papers use weighted averages while others use unweighted averages. Then, some only consider one direction (i.e., consider for every found community the best ground truth community or consider for every ground truth community the best found community). Other papers again use the average of both directions. This class makes all of this possible (for both directions you would need to call it twice), but it is not really a similarity (or dissimilarity) measure.

To be honest I am also not completely happy with the design of `LocalCommunityEvaluation` algorithms, however, for a different reason. Having one score per class means that there will be a lot of duplicated calculations. E.g. when you calculate the F1 score, at the same time you could also calculate precision, recall and Jaccard index without additional overhead. Unfortunately, I have no idea how to make this possible without having one giant class that calculates everything.

avdgrinten:
Alright. I am merging this now. There duplicated calculation can be improved at a later time if we find a good strategy to do that.

PR #303, opened 2019-03-20
Removed neighbors() method from Graph and fixed affected classes

angriman:
This removes the `neighbors(v)` method from the graph class; the method creates a copy of the `outEdges[v]` vector which is not an efficient way to iterate over neighbors.

Declared the getter methods of the neighbor iterators of the `Graph` class as `const` so they can be also be used by const refs of Graph objects.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -445,10 +445,29 @@ class Graph final {
     /**
      * Class to iterate over the in/out neighbors of a node.
      */
-    class NeighborIterator
-        : public std::iterator<std::forward_iterator_tag, const node> {
+    class NeighborIterator {
 
       public:
+        // The value type of the neighbors (i.e. nodes). Returned by
+        // operator*().
+        typedef node value_type;
+
+        // Reference to the value_type, required by STL.
+        typedef value_type &reference;
+
+        // Pointer to the value_type, required by STL.
+        typedef value_type *pointer;
+
+        // STL iterator category.
+        typedef std::forward_iterator_tag iterator_category;
+
+        // Signed integer type of the result of subtracting two pointers,
+        // required by STL.
+        typedef ptrdiff_t difference_type;
+
+        // Own type.
+        typedef NeighborIterator self;
+
         NeighborIterator(const Graph &G, node u,
                          std::vector<node>::const_iterator nodesIter)
             : G(G), u(u), nIter(nodesIter) {}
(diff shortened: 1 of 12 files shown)


avdgrinten:
While I think that it is a good idea to remove this method, we should deprecate it first and only remove it after two releases.

Of course, it does makes sense to get rid of internal users now.

networkit/cpp/centrality/GroupDegree.cpp
```diff
@@ -58,10 +58,7 @@ void GroupDegree::run() {
 		updateGroup();
 	}
 
-	std::vector<node> neighbors = G.neighbors(group.back());
-#pragma omp parallel for
-	for (omp_index i = 0; i < neighbors.size(); ++i) {
-		node u = neighbors[i];
+	for (node u : G.neighborRange(group.back())) {
```
avdgrinten: Here, we get rid of the parallelism -- which is probably not an issue as there are not so many neighbors.

networkit/cpp/clique/test/MaximalCliquesGTest.cpp
```diff
@@ -17,7 +17,10 @@ TEST_F(MaximalCliquesGTest, testMaximalCliques) {
 	Graph G = reader.read("input/hep-th.graph");
 
 	node seed = 2;
-	auto sn = G.neighbors(seed);
+    std::vector<node> sn;
+    sn.reserve(G.degree(seed));
+    for (node u : G.neighborRange(seed))
+        sn.push_back(u);
```
avdgrinten: `std::vector` has a constructor that takes iterators; this can be used to turn those 4 lines into a single line. The indentation seems to be off, too.

angriman: This needed some fixes in the `NeighborIterator` and `NeighborWeightIterator` classes (9ce72ee); addressed in fbcc2b7.

include/networkit/graph/Graph.hpp
```diff
@@ -445,11 +445,10 @@ class Graph final {
     /**
      * Class to iterate over the in/out neighbors of a node.
      */
-    class NeighborIterator {
+    class NeighborIterator
+        : public std::iterator<std::forward_iterator_tag, const node> {
```
avdgrinten: `std::iterator` is deprecated in C++17. It does nothing except for defining the member types listed in <https://en.cppreference.com/w/cpp/iterator/iterator>, so it should be easy to avoid.

avdgrinten:
Looks good to me now.

PR #304, opened 2019-03-21
Add macOS builds with llvm 8

angriman:
This adds two new builds to travis: macOS 10.12/13 with the newly released llvm 8.


avdgrinten:
Do we really need two more builds? My suggestion would be to only add a llvm 8 build for the more recent version of macOS.

PR #310, opened 2019-04-02
travis/mac-os-10.14

angriman:
Support for mac os 10.14 builds on travis; drop support for mac os 10.12

PR #312, opened 2019-04-09
MacOS 10.14

rbange:
This PR incorporates the new macOS travis config introduced here: https://github.com/kit-parco/networkit/pull/310

Also `CMakeLists.txt` is adapted to manually set compile and linking flags for clang-like compiler.

Regarding https://github.com/kit-parco/networkit/pull/310#issuecomment-481229629:
We could get rid of the `find_library` and `find_path` parts by assuming the installed libomp is properly linked (as e.g. homebrew does by default), but that might be less robust.

Also, the OpenMP::OpenMP_CXX target is manually created if it does not exist, although that's never tested in CI. The travis linux images got a package update and cmake has version 3.12.4 now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,49 @@ env:
 
 matrix:
   include:
+    - name: macOS 10.14, Apple Clang
+      env:
+        - CC=cc
+        - CXX=c++
+      os: osx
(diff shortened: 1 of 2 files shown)


avdgrinten:
Great, thank you very much Roman!

PR #315, opened 2019-05-10
New Graph method: toLargestConnectedComponent()

angriman:
New method that deletes all nodes and edges not in the largest connected component of a graph. Only works for undirected graphs.

### DIFF
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -9,7 +9,10 @@
 #include <cmath>
 #include <random>
 #include <sstream>
+#include<unordered_map>
+#include<unordered_set>
 
+#include "../../include/networkit/components/ConnectedComponents.hpp"
 #include "../../include/networkit/graph/Graph.hpp"
 #include "../../include/networkit/graph/GraphBuilder.hpp"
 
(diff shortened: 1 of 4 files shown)


networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+		throw std::runtime_error(
+			"Method available for undirected graphs only.");
+
+	ConnectedComponents cc(*this);
+	cc.run();
+	auto comps = cc.getComponents();
+	std::sort(comps.begin(), comps.end(),
+		[&](const std::vector<node> &x, const std::vector<node> &y) {
+			return x.size() > y.size();
+	});
```
michitux: Why is this code using `std::sort` to find the maximum?

networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+	std::unordered_map<node, node> map;
+	std::unordered_set<node> set(comps[0].begin(), comps[0].end());
```
michitux: Why are you storing the nodes in an unordered set? You can already:
* Iterate over the nodes in the component in `comps[0]`
* Get the component id of any of these nodes and compare it to the connected component of any node to check if that other node is in the largest connected component.

The first is in linear time, the second in constant time, there is no need to use a potentially large and inefficient `unordered_set`.

networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+	node idx = 0;
+	for (node u : comps[0])
+		map[u] = idx++;
```
michitux: Why are you using an unordered map here instead of a simple `std::vector` over all nodes which should give a much better running time? We have O(n) storage used anyway (note that the `subgraphFromNodes` below even copies the whole graph).

networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+	std::vector<std::pair<node, node>> edges;
+	std::vector<edgeweight> weights;
+	this->subgraphFromNodes(set).forEdges([&](node x, node y, edgeweight w) {
+		if (map[x] != map[y]) {
```
michitux: Why are you comparing the values of the map here? What is the purpose of this test anyway? Shall this exclude self-loops? If yes, why?

networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+			return x.size() > y.size();
+	});
+	std::unordered_map<node, node> map;
+	std::unordered_set<node> set(comps[0].begin(), comps[0].end());
+	node idx = 0;
+	for (node u : comps[0])
+		map[u] = idx++;
+	std::vector<std::pair<node, node>> edges;
+	std::vector<edgeweight> weights;
+	this->subgraphFromNodes(set).forEdges([&](node x, node y, edgeweight w) {
```
michitux: This creates a copy of the whole graph just to enumerate all edges where both endpoints are in the given set! Why does this not simply iterate over all nodes in the connected component and all of its neighbors and checks if both endpoints are in the largest connected component?

networkit/cpp/graph/Graph.cpp
```diff
@@ -1065,4 +1068,52 @@ Graph Graph::subgraphFromNodes(const std::unordered_set<node> &nodes) const {
...
+	inEdgeWeights.resize(n);
+	outEdgeWeights.clear();
+	outEdgeWeights.resize(n);
+	exists.clear();
+	exists.resize(n, true);
+
+	for (count e = 0; e < edges.size(); ++e) {
+		this->addEdge(edges[e].first, edges[e].second);
+		if (this->isWeighted())
+			this->setWeight(edges[e].first, edges[e].second, weights[e]);
```
michitux: This innocent-looking `setWeight` is an O(deg(u)) complexity! Why does the code not simply give the weight to `addEdge`?

PR #317, opened 2019-05-12
New method: extractLargestConnectedComponent

angriman:
Even though `extractLargestComponent` already exists in `workflows.py`, this algorithm is not available in C++. This commit brings a more efficient C++ implementation of the same algorithm, which can additionally "compact" the graph.

### DIFF
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -1,7 +1,10 @@
-#include <networkit/graph/GraphTools.hpp>
-#include <unordered_map>
-#include <networkit/graph/Graph.hpp>
+#include <algorithm>
 #include <random>
+#include <unordered_map>
+
+#include <networkit/components/ConnectedComponents.hpp>
+#include <networkit/graph/GraphTools.hpp>
+#include <networkit/graph/Graph.hpp>
 
 namespace NetworKit {
 
@@ -24,7 +27,6 @@ std::unordered_map<node,node> getContinuousNodeIds(const Graph& graph) {
 	graph.forNodes(addToMap);
 	return nodeIdMap;
 }
-
 std::unordered_map<node,node> getRandomContinuousNodeIds(const Graph& graph) {
 	std::unordered_map<node,node> nodeIdMap;
 	std::vector<node> nodes;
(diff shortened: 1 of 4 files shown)


networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +77,39 @@ Graph restoreGraph(std::vector<node>& invertedIdMap, const Graph& G) {
...
+		compSizes.begin(), compSizes.end(),
+		[](const std::pair<index, count> &x, const std::pair<index, count> &y) {
+			return x.second < y.second;
+	});
+
+	count nodesToRemove = G.numberOfNodes() - largestCC->second;
+	S.forNodesWhile(
+		[&](){ return nodesToRemove; },
+		[&](node u) {
+			if (cc.componentOfNode(u) != largestCC->first) {
```
avdgrinten: Indentation seems to be off here.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +77,39 @@ Graph restoreGraph(std::vector<node>& invertedIdMap, const Graph& G) {
...
+
+	ConnectedComponents cc(S);
+	cc.run();
+
+	auto compSizes = cc.getComponentSizes();
+	auto largestCC = std::max_element(
+		compSizes.begin(), compSizes.end(),
+		[](const std::pair<index, count> &x, const std::pair<index, count> &y) {
+			return x.second < y.second;
+	});
```
avdgrinten: The indentation of the `}` brace does not match the `{` brace.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +77,39 @@ Graph restoreGraph(std::vector<node>& invertedIdMap, const Graph& G) {
...
+
+	count nodesToRemove = G.numberOfNodes() - largestCC->second;
+	S.forNodesWhile(
```
avdgrinten: I don't think it is supported to remove nodes while iterating over them. Instead, we should iterate over the component and call `removeNode`. This would also be faster.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +77,38 @@ Graph restoreGraph(std::vector<node>& invertedIdMap, const Graph& G) {
...
+		if (c == largestCC->first)
+			continue;
+		for (node u : components[c]) {
+			S.removeNode(u);
+		}
+	}
+
+	if (compactGraph) {
+		auto continuousNodeIds = getContinuousNodeIds(S);
+		S = getCompactedGraph(S, continuousNodeIds);
```
avdgrinten: If `compactGraph` is set, this method copies the Graph twice. We could probably get away with a single copy by deleting vertices and remapping the Graph at the same time.

avdgrinten: This can now be optimized using the getRemappedGraph() function from #324.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -73,6 +75,55 @@ Graph restoreGraph(const std::vector<node>& invertedIdMap, const Graph& G) {
 	return Goriginal;
 }
 
+Graph extractLargestConnectedComponent(const Graph &G, bool compactGraph) {
+	if (!G.numberOfNodes())
+		return Graph(G);
```
avdgrinten: This can be just `return G;`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -73,6 +75,55 @@ Graph restoreGraph(const std::vector<node>& invertedIdMap, const Graph& G) {
...
+	std::vector<bool> toRemove(G.upperNodeIdBound(), false);
+	count toRemoveCount = 0;
+
+	index nextId = 0;
+	G.forNodes([&](node u) {
+		if (cc.componentOfNode(u) == largestCC->first) {
+			continuousNodeIds[u] = nextId++;
+		}
+		else {
+			toRemove[u] = true;
```
avdgrinten: `toRemove` seems unnecessary. Why don't we just check `componentOfNode` again in the second loop?

PR #318, opened 2019-05-13
AllSimplePaths: Removed unused vector and std::move that prevents RVO

angriman:
As pointed out by @michitux in #308


avdgrinten:
LGTM (and is quite trivial). Merging.

PR #321, opened 2019-05-21
standardize warning when missing desired modules

jeffgreenca:
When reduced functionality will occur due to missing modules, emit a
consistent warning using the Python standard library.

- Defines networkit.exceptions.ReducedFunctionality
- use warnings.warn() instead of print()

This allows user control of the warning (via PYTHONWARNINGS,
stderr redirection, etc).


networkit/exceptions.py
```diff
@@ -0,0 +1,3 @@
+class ReducedFunctionality(Warning):
```
avdgrinten: I would prefer to rename this class to `ReducedFunctionalityWarning`. `exceptions.ReducedFunctionality` otherwise suggests that this is a (more critical) exception. This would also be in line with the `Warning` subclasses of the Python standard library.

jeffgreenca: Makes sense to me.

avdgrinten:
I am merging this PR. Thank you again.

PR #325, opened 2019-06-04
Add custom binary format to NetworKit

cndolo:


avdgrinten:
Adding @michitux as a reviewer as he wrote the Thrill binary readers. This might also be interesting to @manpen.

Let me write a bit about the context of this PR. Here, we introduce a new binary graph format. While many formats exist already, this one specifically aims for high performance. In particular, the goals are as follows:

- The format allows for good read performance. It can support parallel reading.
- Compactness. While the format is not as compact as compressed formats, it is still significantly more compact than text files.
- Extensibility. Adding more features will not render existing files invalid.

I do not consider the fact that this format does not exist outside of NetworKit to be problematic. Our main use case will be converting graphs to this format and then profiting from the improved I/O performance. This is especially helpful when benchmarking multiple algorithms on the same graph.

The description of the format is part of the PR (see networkit/cpp/io/NetworkitBinaryGraph.md). However, let me highlight a few design decisions here:

- Sizes of adjacency lists are stored independently of the lists themselves; hence, those lists can be preallocated.
- To achieve parallel readability, the format stores the graph data in multiple chunks. Vertices in different chunks can be processed in parallel. The number of chunks (and their sizes) can be configured at write time.
- All entries of edge lists are stored in a varint encoding, namely PrefixVarint. In contrast to Thrill, this encoding stores the size of the varint in the first byte; this allows for faster decoding due to less branch mispredictions. It is equally compact as Thrill's encoding.
- The file header stores a feature mask so that future feature additions can be done without breaking existing files.

In this PR, support for weights, parallel reading and checksums is still missing. For weights, the structure definitions are already present but support in the writer and reader is incomplete. I want to merge the basic functionality first, before adding more advanced features.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+	if(!value) {
+		buffer[0] =1;
+		return 1;
+	}
+
+	if (value > (uint64_t(1) << 63)) {
+		n = 8;
+		buffer[0] = 0;
+	} else {
+		n = (64 - __builtin_clzl(value)) / 7;
```
avdgrinten: `__builtin_clz` is GCC/Clang-only and breaks on Windows. We should use the clz function from TLX (which is already a depending of NetworKit). See https://github.com/tlx/tlx/blob/master/tlx/math/clz.hpp.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+namespace NetworKit {
+
+size_t NetworkitBinaryReader::decode(const uint8_t* data, uint64_t& value) {
+	int n;
+	if(!data[0]) {
+		n = 8;
+		value = 0;
+	} else {
+		n = __builtin_ctz(data[0]);
```
avdgrinten: Here, we can use `ffs` from TLX.

avdgrinten: However, `ffs` (in https://github.com/tlx/tlx/blob/master/tlx/math/ffs.hpp), returns the number of trailing zeros + 1, so we have to take `ffs(data[0]) - 1`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+		G.forNodes([&](node u) {
+			size_t adj_size = 0;
+			G.forNeighborsOf(u,[&](node v) {
+				uint8_t tmp [10];
+				adj_size += encode(v, tmp);
+			});
+			adj_size += prev;
+			encArrSize.push_back(adj_size);
+			prev = adj_size;
+		});
```
michitux: This could also be computed in the loop above, avoiding an iteration over the whole graph.

Futher: why is this computing the offset for every node? The chunk offsets are completely independent of the edges, therefore the chunk offsets could also be determined before, then only the sizes for each chunk would need to be stored.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+			uint64_t offset;
+			if(!i) {
+				offset = 0;
+			} else {
+				offset = encArrSize.at(begin-1);
+			}
+			chunksMap.push_back(std::make_pair(begin, offset));
+		}
+
+        // Write header
```
michitux: Indentation is wrong (also below).

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+			G.forNeighborsOf(u,[&](node v){
+				uint8_t tmp [10];
+				uint64_t node_size = encode(v, tmp);
+				outfile.write(reinterpret_cast<char*>(tmp), node_size);
+			});
+		});
+
+		INFO("Written graph to ", path);
+	} else {
+		throw std::runtime_error("G is empty.");
```
michitux: Why is it not possible to store empty graphs? As far as I understood the format, this should be valid.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+		int k = 0;
+		G.forNodes([&](node u) {
+			if (ctr >= nodesPerChunk) {
+				split.push_back(std::make_pair(k,(u+1)));
+				ctr = 1;
+				k++;
+			} else {
+				ctr++;
+			}
+		});
```
michitux: As the file format doesn't support deleted nodes, this loop could be simplified by simply increasing `u` directly by `nodesPerChunk`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+NetworkitBinaryWriter::NetworkitBinaryWriter(uint64_t chunks): chunks(chunks) {}
+
+uint64_t NetworkitBinaryWriter::splitNodes(uint64_t nodes, uint64_t chunks) {
+
+	nodesPerChunk = nodes / chunks;
+	return nodesPerChunk;
+}
+
+size_t NetworkitBinaryWriter::encode(uint64_t value, uint8_t* buffer) {
```
michitux: As far as I can see, `encode` is always used with a temporary buffer and never used to write in a larger buffer. Therefore, it would be much cleaner imho if `encode` used an internal buffer and returned a pair (or tuple) of the buffer and the size.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+				begin = 0;
+			} else {
+				begin = split[i-1].second;
+			}
+
+			uint64_t offset;
+			if(!i) {
+				offset = 0;
+			} else {
+				offset = encArrSize.at(begin-1);
```
michitux: Why is this using `at` but above `operator[]` is used?

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+			}
```
michitux: This logic seems overly complicated. Why don't you just put both into a single if statement? To further simplify this, `begin` and `offset` could also be initialized outside the loop and then set after pushing into `chunksMap` by using `begin = split[i].second`.

networkit/cpp/io/NetworkitBinaryGraph.md
```diff
@@ -0,0 +1,57 @@
...
+- offsetAdjLists: Offset of the adjacency lists in the file
+- offsetWeights: Offset of the weights in the file
+
+All offsets are relative to the beginning of the file.
+
+Base data
+------------
+```
+uint64_t sizeSum[nodes]: The sum of sizes of adjacency vertices with indices <= i
+uint64_t firstVertex[chunks]: The index of the first vertex of each chunks
```
michitux: To simplify reading the file, it would be good to store one more element in both arrays that indicates the end (so you can e.g. use a single `memcpy` to copy the whole chunk).

networkit/cpp/io/NetworkitBinaryGraph.md
```diff
@@ -0,0 +1,57 @@
...
+- magic: A constant value used to identify the file format version.
+    - The current version is '*nkbg000*'
+- checksum: Currently not used
+- features: Contains the graph information bitwise
+    - Bit 0 : directed or undirected
+    - Bit 1-2 : the weight format of the graph. Below are the possible formats:
+        - 0 = no weights
+        - 1 = Weights are unsigned integers
+        - 2 = Weights are signed integers
+        - 3 = Weights are floats/doubles
```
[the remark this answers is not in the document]
avdgrinten: The plan is to store integral edge weights as varints. But yes, we need to distinguish floats/doubles.

networkit/cpp/io/NetworkitBinaryGraph.md
```diff
@@ -0,0 +1,57 @@
...
+    uint64_t nodes;
+    uint64_t chunks;
+    uint64_t offsetBaseData;
+    uint64_t offsetAdjLists;
+    uint64_t offsetWeights;
+};
+```
+- magic: A constant value used to identify the file format version.
+    - The current version is '*nkbg000*'
+- checksum: Currently not used
```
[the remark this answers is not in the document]
avdgrinten: We plan to add the checksum in a future PR. It should protect against accidental truncation and/or corruption (but not against malicious modification).

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,147 @@
...
+			uint64_t curr = vertex+i;
+			if(curr > 0) {
+				nbrs = prefSum.at(curr) - prefSum.at(curr-1);
+			} else {
+				nbrs = prefSum.at(0);
+			}
+			for (uint64_t j = 0; j < nbrs; j++) {
+				uint64_t add;
+				size_t rd = NetworkitBinaryReader::decode(reinterpret_cast<const uint8_t*>(adj_it + off), add);
+				if(!G.isDirected()) {
```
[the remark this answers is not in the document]
avdgrinten: In a future patch, we will add parallel reading. For that, I plan to support adding half edges directly in the `Graph`` class (with UB if the user doesn't add both halves in the undirected case).

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	for(int i = 0; i < n; i++) {
+		buffer[i+1] = value & 0xFF;
+	   	value >>= 8;
+	}
+	return n + 1;
+}
+
+uint64_t NetworkitBinaryWriter::encodeZigzag(int value) {
```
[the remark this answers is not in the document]
cndolo: Currently, it is. The zigzag encoding/decoding functions will be used in the future when weights are supported, as we intend to encode both signed and unsigned integral weights as varints.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,147 @@
...
+		value = data[0] >> (n+1);
+	}
+
+	for(int i = 0; i < n; i++) {
+		value |= data[i+1] << (8 - (n + 1) + i * 8);
+	}
+	return n+1;
+}
+
+int NetworkitBinaryReader::decodeZigzag(uint64_t value) {
```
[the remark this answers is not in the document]
cndolo: See above comment.

networkit/cpp/io/NetworkitBinaryGraph.md
avdgrinten: Can you elaborate on that? It's not clear to me what you would like to achieve.

avdgrinten:
Thank you for the review. There are quite a few things that I missed. I responded to some format-related comments above. @cndolo will probably take care of the technical comments.

Regarding deleted nodes: I must admit that I did not think about those. It would be nice if the format supported deleted nodes so that no remapping if node IDs is necessary. I think we can sacrifice one bit of the adjacency size prefix sum for this purpose.

Regarding the double encoding: This is fine from my perspective. I do not care so much about writer performance. I suspect that relying on the fact the adjacency data can be allocated at a fixed position is quite fragile -- for example, the same does not hold for weight information.

networkit/cpp/io/NetworkitBinaryWriter.cpp
avdgrinten: I am not sure if I agree here. Returning a pointer to a static buffer does not seem to be such a great interface. We could add a symbolic constant `maxEncodedSize` or so to make sure we never run into buffer overflows.

include/networkit/graph/Graph.hpp
```diff
@@ -683,6 +683,14 @@ class Graph final {
     /** Default copy assignment operator */
     Graph &operator=(const Graph &other) = default;
 
+	/**
+	 * Reserves memory in the node's edge containers.
+	 *
+	 * @param u the node memory should be reserved for 
+	 * @param size the amount of memory to reserve
+	 * */
+	void preallocate(node u, size_t size);
```
manpen: I would like to use this function in a parallel context. Can we please add the guarantee to the documentation that it is thread-safe to use as long as two concurrent invocations to not access the same node u. The current implementation allows this without modification.

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef NETWORKIT_BINARY_GRAPH_
+#define NETWORKIT_BINARY_GRAPH_
+
+struct Header {
+	char magic[8];
```
manpen: Regarding @michitux's comment about endianness: Would it make sense to declare the magic word as uint64_t and specify that the writer uses its native endianness? Then, a later revision of the reader could support different coding standards. This would achieve maximal performance if the file remains on one system. (also the majority of data will likely be stored in varints where endianess is well-defined)

include/networkit/graph/Graph.hpp
avdgrinten: Agreed.

include/networkit/io/NetworkitBinaryGraph.hpp
avdgrinten: Regarding the endianness: The vast majority of all systems are probably little endian today (AFAIK this applies even to ARM, which supports both in theory). Hence, I would prefer to specify the format as little endian. Big endian machines can probably convert the few fixed-size numbers without huge performance issues.

networkit/cpp/io/NetworkitBinaryWriter.cpp
michitux: I did not mean returning a pointer to the buffer, I meant just returning a buffer by value. As never more than 10 bytes are written, returning a buffer of size 10 should be enough. Even without RVO, this should be no problem.

networkit/cpp/io/NetworkitBinaryGraph.md
michitux: If you want to know how many vertices are in a chunk, with this structure you always need to check if the current chunk is the last chunk. If there was an additional value, for getting the size of chunk `i`, you could simply subtract `firstVertex[i]` from `firstVertex[i+1]`.

The same is true for `sizeSum`. The description "The sum of sizes of adjacency vertices with indices <= i" is also not clear. From the code I understand that this is a prefix sum of the node's degrees. Here, again, if I want to get the degree of node `i`, I need to check first if this is the first node, if yes, it is the degree, otherwise I need to subtract the previous entry from it. Here, an additional entry would only help if the definition was changed to be the exclusive prefix sum, i.e., if it was "indices < i".

Btw., it seems the number of edges is only written in the last entry of this array?

avdgrinten: Yes, the number of edges is only written in the last entry. We could also store it in the header if a simple way to access it is considered to be important.

For computing the size of a chunk: to make firstVertex[i + 1] - firstVertex[i] always valid, we would have to store a zero in the first index (and shift indices by 1). I think this is better handled in the reader. But you're right that it will reduce the number of branch mispredictions and is probably a worthwhile optimization.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+	// Read base data
+	std::vector<uint64_t> prefSum;
+	std::vector<uint64_t> firstVert;
+	const char* base_it = mmfile.cbegin() + header.offsetBaseData;
+	for(uint64_t i = 0; i < nodes; i++) {
+		uint64_t sum = 0;
+		memcpy(&sum, base_it, sizeof(uint64_t));
+		base_it += sizeof(uint64_t);
+		prefSum.push_back(sum);
```
avdgrinten: We can immediately delete the nodes here if `sum & DELETED_BIT`. We should also only push `sum & SIZE_MAX` into the `prefSum`.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,150 @@
...
+			n = nodes - firstVert[c];
+		}
+
+		for(uint64_t i = 0; i < n; i++) {
+			uint64_t nbrs;
+			uint64_t curr = vertex+i;
+			uint64_t deleted;
+			if(curr > 0) {
+				nbrs = prefSum[curr] - prefSum[curr-1];
+				deleted = (nbrs & DELETED_BIT);
```
avdgrinten: This will not work. If (exactly) one of the `prefSum` entries has the `DELETED_BIT` bit set, `nbrs` is not computed correctly. To correctly handle this, mask out the `DELETED_BIT` when constructing `prefSum`.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+	for(int i = 0; i < n; i++) {
+		value |= data[i+1] << (8 - (n + 1) + i * 8);
+	}
+	return n+1;
+}
+
+int64_t NetworkitBinaryReader::decodeZigzag(uint64_t value) {
+	int64_t orig = (value >> 1) ^ (-(value & 1));
+	return orig;
```
avdgrinten: The temporary variable is unnecessary.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+	for(int i = 0; i < n; i++) {
+		buffer[i+1] = value & 0xFF;
+	   	value >>= 8;
+	}
+	return n + 1;
+}
+
+uint64_t NetworkitBinaryWriter::encodeZigzag(int64_t value) {
+	uint64_t mapped = (value << 1) ^ (value >> 31);
```
avdgrinten: The temporary variable is unnecessary.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <networkit/io/NetworkitBinaryWriter.hpp>
+
+namespace NetworKit {
+
+NetworkitBinaryWriter::NetworkitBinaryWriter(uint64_t chunks): chunks(chunks) {}
+
+uint64_t NetworkitBinaryWriter::splitNodes(uint64_t nodes, uint64_t chunks) {
+
+	nodesPerChunk = nodes / chunks;
+	return nodesPerChunk;
```
avdgrinten: This is quite confusing: the function both sets a member variable and returns a value. Make it either only compute the number of nodes per chunk, or do not write `nodesPerChunk` inside this function.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+	uint64_t lastNode = 0;
+	std::vector<size_t> encArrSize;	//  pref sum of size encoded adj arrays
+	for(uint64_t c = 0; c < chunks; c++) {
+		uint64_t adj_size = 0;
+		for(uint64_t n = lastNode; n < (lastNode + nodesPerChunk); n++) {
+			G.forNeighborsOf(n,[&](node v) {
+				uint8_t tmp [10];
+				adj_size += encode(v, tmp);
+			});
+		}
```
avdgrinten: If I understand the logic of the whole reader correctly, the last chunk can be bigger than the other chunks. AFAICS this loop does not work if that is the case.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+		INFO("reducing chunks to ", chunks, " chunks");
+	}
+	nodesPerChunk = splitNodes(nodes, chunks);
+
+	// compute prefix sum
+	count sum = 0;
+	size_t prev = 0;
+	std::vector<count> nodePrefSum;
+	G.forNodes([&](node u) {
+		size_t adj_size = 0;
```
avdgrinten: This variable is not used.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+		sum += G.degree(u);
+		nodePrefSum.push_back(sum);
+	});
```
avdgrinten: AFAICS we do not need the `nodePrefSum` array. We can compute it on-the-fly while writing the file.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    // Write base data
+	G.forNodes([&] (node u) {
+		uint64_t sum = nodePrefSum[u];
+		assert(!(sum & ~SIZE_MASK));
+		if(!G.hasNode(u)) {
+			sum |= DELETED_BIT;
+		}
+		outfile.write(reinterpret_cast<char*>(&sum), sizeof(uint64_t));
+	});
```
avdgrinten: `nodePrefSum` can be computed on-the-fly in this loop.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+	int k = 0;
+	G.forNodes([&](node u) {
+		if (ctr >= nodesPerChunk) {
+			split.push_back(std::make_pair(k,(u+1)));
+			ctr = 1;
+			k++;
+		} else {
+			ctr++;
+		}
+	});
```
avdgrinten: AFAICS this logic does not work if the last chunk is bigger than `nodesPerChunk`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+		outfile.write(reinterpret_cast<char*>(&header.offsetWeights), sizeof(uint64_t));
+	};
+
+	count prefSum = 0;
+	nodes = G.numberOfNodes();
+	if (nodes < chunks) {
+		chunks = nodes;
+		INFO("reducing chunks to ", chunks, " chunks");
+	}
+	nodesPerChunk = splitNodes(nodes, chunks);
```
avdgrinten: Suggestion: Instead of storing `nodesPerChunk`, compute a `firstInChunk` array here. That will simplify the logic of the rest of the writer. `nodesPerChunk` will then not be needed anymore.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,150 @@
...
+		uint64_t off = 0;
+		memcpy(&off, adj_it + c * sizeof(uint64_t), sizeof(uint64_t));
+		off += chunks * sizeof(uint64_t);
+
+		uint64_t n;
+		if(c != (chunks -1)) {
+			n = firstVert[c+1] - vertex;
+		} else {
+			n = nodes - firstVert[c];
+		}
```
avdgrinten: Why do we need a `if` here? I think we should *not* store `0` in the first `firstVertex` entry of the file (that information is already known). However, we can store `0` in the `firstVertex` *vector* here to avoid an `if`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
avdgrinten: See below for a suggestion to remove `splitNodes` entirely.

networkit/cpp/graph/Graph.cpp
```diff
@@ -169,6 +169,20 @@ Graph::Graph(const Graph &G, bool weighted, bool directed)
...
+		if(weighted) {
+			inEdgeWeights[u].reserve(size);
+		}
+	}
+	outEdges[u].reserve(size);
+	if(weighted) {
+		outEdgeWeights[u].reserve(size);
+	}
+}
+
```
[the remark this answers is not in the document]
avdgrinten: This should indeed be fixed. I think we need a `preallocateDirected` and `preallocateUndirected` method to handle this correctly. The directed variant should take both an `inSize` and an `outSize`.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,150 @@
...
+		memcpy(&header.offsetWeights, it, sizeof(uint64_t));
+		it += sizeof(uint64_t);
+	};
+
+	auto checkHeader = [&] () {
+		if(memcmp("nkbg000", header.magic, 8)) {
+			throw std::runtime_error("Reader expected another magic value");
+		} else {
+			directed = (header.features & DIR_MASK);
+			weighted = (header.features & WGHT_MASK) >> WGHT_SHIFT;
```
avdgrinten: With the redefinition of the weight constants, we cannot use the feature field to check whether the graph is weighted anymore. We need to check whether the `offsetWeights` is zero.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,151 @@
...
+				off+=rd;
+			}
+		}
+	};
+
+	// create graph
+	for(uint64_t c = 0; c < chunks; c++) {
+		constructGraph(c);
+	}
+	G.shrinkToFit();
```
avdgrinten: `shrinkToFit` should not be necessary as we preallocate anyway.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,151 @@
...
+	for(uint64_t ch = 1; ch < chunks; ch++) {
+		uint64_t first = 0;
+		memcpy(&first, baseIt, sizeof(uint64_t));
+		baseIt += sizeof(uint64_t);
+		firstVert.push_back(first);
+	}
+	firstVert.push_back(nodes);
+
+	// Read adjacency lists.
+	const char* adjIt = mmfile.cbegin() + header.offsetAdjLists;
```
avdgrinten: Pointer `*`s should be at the right.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,151 @@
...
+	DEBUG("# chunks here = ", chunks);
+	weighted = header.offsetWeights;
+	Graph G(nodes, weighted, directed);
+
+	// Read base data.
+	std::vector<uint64_t> prefSum;
+	const char* baseIt = mmfile.cbegin() + header.offsetBaseData;
+	for(uint64_t i = 0; i < nodes; i++) {
+		uint64_t sum = 0;
+		memcpy(&sum, baseIt, sizeof(uint64_t));
```
avdgrinten: `sum` is always written, so the initialization is not necessary.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -0,0 +1,151 @@
...
+		} else {
+			G.removeNode(i);
+		}
+	}
+
+	std::vector<uint64_t> firstVert;
+	firstVert.push_back(0);
+	for(uint64_t ch = 1; ch < chunks; ch++) {
+		uint64_t first = 0;
+		memcpy(&first, baseIt, sizeof(uint64_t));
```
avdgrinten: Same here (`first` is always written to).

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,162 @@
...
+		firstNode += (nodes/chunks);
+		firstInChunk.push_back(firstNode);
+	}
+	firstInChunk.push_back(nodes);
+
+	// Compute encoded size of arrays and store in vector.
+	uint64_t prevSize = 0; 
+	std::vector<size_t> adjOffsets;	//  pref sum of size encoded adj arrays
+	for(uint64_t c = 0; c < chunks; c++) {
+		uint64_t adj_size = 0;
```
avdgrinten: Variable is not used. But `adjSize` would probably be a better name for `prevSize`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,162 @@
...
+		for(uint64_t n = firstInChunk[c]; n < firstInChunk[c+1]; n++) {
+			G.forNeighborsOf(n,[&](node v) {
+				uint8_t tmp [10];
+				prevSize += encode(v, tmp);
+			});
+		}
+		adjOffsets.push_back(prevSize);
+	}
+
+    // Write header.
```
avdgrinten: Indentation is off.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,162 @@
...
+	setHeader();
+	writeHeader();
+
+    // Write base data.
```
avdgrinten: Same here.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,162 @@
...
+		}
+		outfile.write(reinterpret_cast<char*>(&sum), sizeof(uint64_t));
+	});
+
+	assert(!firstInChunk[0]);
+	for (uint64_t c = 1; c < chunks; c++) {
+		outfile.write(reinterpret_cast<char*>(&firstInChunk[c]), sizeof(uint64_t));
+	}
+
+    // Write adjacency data.
```
avdgrinten: Indentation is off.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -0,0 +1,162 @@
...
+	for (uint64_t c = 1; c < chunks; c++) {
```
avdgrinten: What about an assert `!adjOffsets[0]` similar to the one above?

avdgrinten:
I am merging this PR. Thank you for your work @cndolo.

PR #326, opened 2019-06-06
Add missing OpenMP dependency for non-monolith tests

manpen:
Non-monolithic build fails using default OSX compiler. While this is clearly a bug on our part, I like to blame the accessory manufacturer (the one selling monitor stands for 1k USD) for not shipping OpenMP ;)


avdgrinten:
LGTM.

PR #327, opened 2019-06-06
More efficient memory allocation for APSP

angriman:
This improves the memory footprint of APSP by avoid dynamic allocations when running BFS/Dijkstra
(all the necessary space is allocated once). The new implementation is now 1.5x to 4x faster than the original one.

### DIFF
--- a/include/networkit/distance/APSP.hpp
+++ b/include/networkit/distance/APSP.hpp
@@ -8,8 +8,8 @@
 #ifndef APSP_H_
 #define APSP_H_
 
-#include <networkit/graph/Graph.hpp>
 #include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
 
 namespace NetworKit {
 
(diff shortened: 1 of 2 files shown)


avdgrinten:
LGTM.

networkit/cpp/distance/APSP.cpp
```diff
@@ -15,21 +16,27 @@ namespace NetworKit {
...
+    distances.resize(n, distanceVector);
+
+    count nThreads = omp_get_max_threads();
+    sssps.reserve(nThreads);
+    for (count i = 0; i < nThreads; ++i) {
+        if (G.isWeighted())
+            sssps.emplace_back(new Dijkstra(G, 0, false));
+        else
+            sssps.emplace_back(new BFS(G, 0, false));
     }
```
michitux: As a minor comment, it might make sense to a) resize sssps to the desired size b) open the parallel region and then c) allocate the sssps in paralle, i.e., allocate with each thread the entry for `omp_get_thread_num()`. This would ensure that at least the SSSP-object is allocated in the correct numa region if threads are pinned. However, as this is not considered in most of the code in NetworKit anyway, this might not make much sense here.

include/networkit/distance/APSP.hpp
```diff
@@ -66,6 +69,7 @@ class APSP : public Algorithm {
   protected:
     const Graph &G;
     std::vector<std::vector<edgeweight>> distances;
+    std::vector<std::unique_ptr<SSSP>> sssps;
```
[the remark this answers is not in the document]
angriman: There is not a specific reason for that, I'll change this accordingly.

networkit/cpp/distance/APSP.cpp
avdgrinten: That's true in principle. OTOH I don't know how well the glibc memory allocator handles NUMA. If manipulation of the underlying data structures is protected by a global lock, it might not be so much faster.

But it's a low-effort optimization, so yes, why not do it.

include/networkit/distance/APSP.hpp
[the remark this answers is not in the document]
avdgrinten: If the algorithm is run multiple times, storing the SSSP objects in a member variable might improve performance considerably.

If it only runs a single time, this shouldn't make a difference (unless the user keeps the APSP object around for a long time).

networkit/cpp/distance/APSP.cpp
angriman: I implemented this optimization, I'll push it as soon as we resolve https://github.com/kit-parco/networkit/pull/327#discussion_r291573908

Issue #329, opened 2019-06-07
Reformat indentation globally


avdgrinten:
@manpen mentioned in a comment (https://github.com/kit-parco/networkit/issues/173#issuecomment-499815102) that people sometimes reformat the code due to inconsistencies in coding style.

I would like to get the indentation fixed everywhere. I think we agreed on 4 space indentation at the NetworKit dev meeting last year, so changing to that style should be uncontroversial. We should apply such this change globally.

While it would be fine to provide a clang-format file with sane defaults (that could be used for new files or by new contributors), I do not want to reformat the code base globally. Hence, I would like to ban reformatting existing code for no good reason. If contributors use clang-format or any other tool locally, they should only apply those tools to the new chunks of code that they write but not globally to entire files (and, of course, the existing style in each file still needs to be respected).

manpen:
Solved in PR #411

PR #330, opened 2019-06-07
RFC: README: Add contribution guidelines

avdgrinten:
To be applied after the resolution of #329. Please discuss what we should add to the guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,36 @@
+
+# Contributions
+
+We encourage contributions to the NetworKit source code. If you encounter any questions regarding NetworKit development please contact the [mailing list][list] or open issues in this GitHub repository.
+
+## Coding Style
+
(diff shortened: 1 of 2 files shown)


angriman:
What about fixing a maximum number of characters per line?

avdgrinten:
I think we should have a soft (i.e., not automatically enforced) limit of 100 characters.

manpen:
There's already a [DevGuide](https://networkit.github.io/dev-docs/DevGuide.html) which should either be the only place or a strict superset (i.e., the `README.md` contains a summary).
So I suggest to add these rules (esp. name conventions) also to the DevGuides and place a link the guide in the Readme.

avdgrinten:
What I dislike about the DevGuide is that 90% of it consists of common Git practices. It contains a lot of noise and is sometimes even inaccurate. I would like to collect essential contribution guidelines (and there should not be so many) in the README.md or in a CONTRIBUTING.md. The DevGuide can link to that document.

PR #331, opened 2019-06-07
Fixed memory leaks in KadabraBetweenness

angriman:
Using unique pointers to avoid memory leaks in KadabraBetweenness. 

Because this is not the first time that code with memory leaks is merged in NetworKit, I'd suggest to add a travis build with memory sanitizers. The same build could also include address sanitizers, which are very helpful to detect memory errors that are not detected by simply selecting the `Debug` build type. What do you think about this?


michitux:
:+1:  for using sanitizers.

Using unique pointers seems valid, but the code creates the impression that a lot more refactoring could be done.

avdgrinten:
What kind of refactoring? The run method looks reasonably clean to me, considering that the algorithm is somewhat complicated.

PR #334, opened 2019-06-12
Refactor SSSP getDistances()

angriman:
The `getDistances()` method of `SSSP` allows the distances vector to be moved out without a clear reason apart from avoiding to copy the vector. This can be done more safely by returning a const ref. The only class using this feature is `AffectedNodes`, which just reads the data.

### DIFF
--- a/networkit/cpp/distance/SSSP.cpp
+++ b/networkit/cpp/distance/SSSP.cpp
@@ -17,7 +17,11 @@ SSSP::SSSP(const Graph &G, node source, bool storePaths,
       storeNodesSortedByDistance(storeNodesSortedByDistance) {}
 
 std::vector<edgeweight> SSSP::getDistances(bool moveOut) {
-    return (moveOut) ? std::move(distances) : distances;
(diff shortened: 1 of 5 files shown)


avdgrinten:
Note that this changes the ABI of NetworKit. However, I doubt that anyone ever used the `moveOut` parameter (if that assumption is wrong, please complain!). I consider `moveOut`to be a huge hack.

networkit/cpp/distance/SSSP.cpp
```diff
@@ -17,7 +17,11 @@ SSSP::SSSP(const Graph &G, node source, bool storePaths,
       storeNodesSortedByDistance(storeNodesSortedByDistance) {}
 
 std::vector<edgeweight> SSSP::getDistances(bool moveOut) {
-    return (moveOut) ? std::move(distances) : distances;
```
avdgrinten: This never actually worked to begin with: the common type of `std::vector<edgeweight> &&` and `std::vector<edgeweight> &` is again `std::vector<edgeweight> &`...

PR #336, opened 2019-06-14
New feature: Bidirectional graph exploration

angriman:
This includes two new classes: `BidirectionalBFS` and `BidirectionalDijkstra`, which explore the graph faster than `BFS` and `Dijkstra` if the target node is already known.


networkit/_NetworKit.pyx
```diff
@@ -1333,6 +1333,111 @@ cdef class Graph:
 			nnodes.insert(node)
 		return Graph().setThis(self._this.subgraphFromNodes(nnodes))
 
+
+cdef extern from "<networkit/distance/STSP.hpp>":
```
avdgrinten: Indentation is off here. Why does this use double TABs?

include/networkit/distance/BidirectionalBFS.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+ * @ingroup distance
+ * Implements a bidirectional breadth-first search on a graph from two given source and target
+ * nodes. Explores the graph from both the source and target nodes until the two explorations meet.
+ */
+class BidirectionalBFS final : public STSP {
+
+public:
+    using STSP::STSP;
```
avdgrinten: Why is this `using` declaration necessary?

include/networkit/distance/STSP.hpp
```diff
@@ -0,0 +1,111 @@
...
+    static constexpr uint8_t ballMask = uint8_t(1) << 7;
+    std::vector<node> pred;
+    std::vector<node> path;
+
+    void buildPath();
+
+    void checkStoreDistances() const {
+        if (!storePred)
+            WARN("Predecessors not stored.");
+    }
```
avdgrinten: The name does not match the functionality. `checkStorePredecessors`?

include/networkit/distance/STSP.hpp
```diff
@@ -0,0 +1,111 @@
...
+    const Graph &G;
+    node source, target;
+    const bool storePred;
+    std::vector<uint8_t> visited;
+    uint8_t ts = 0;
+    static constexpr uint8_t ballMask = uint8_t(1) << 7;
+    std::vector<node> pred;
+    std::vector<node> path;
+
+    void buildPath();
```
avdgrinten: What is the intended use case for that function? Derived classes call it to reconstruct a path? Add a comment to make that clear.

include/networkit/distance/STSP.hpp
```diff
@@ -0,0 +1,111 @@
...
+        return pred;
+    }
+
+protected:
+    const Graph &G;
+    node source, target;
+    const bool storePred;
+    std::vector<uint8_t> visited;
+    uint8_t ts = 0;
+    static constexpr uint8_t ballMask = uint8_t(1) << 7;
```
avdgrinten: It is necessary to have `visited` in this class? This makes it impossible to use any implementation strategy other than bidirectional exploration.

networkit/cpp/distance/BidirectionalBFS.cpp
```diff
@@ -0,0 +1,108 @@
...
+    // Mark source as fisited by the ball we grow from the source
+    visited[source] = ts;
+    // Mark the target as visited the ball we grow from the target
+    visited[target] = ts + ballMask;
+
+    bool stop = false;
+    node s, t;
+
+    // Expands a ball of a level
+    auto expand = [&](std::queue<node> &q, std::queue<node> &q_, uint8_t idx) {
```
avdgrinten: `idx` = index of the ball (in {0, 1})? Please add a comment.

networkit/cpp/distance/BidirectionalBFS.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    init();
+    if (storePred) {
+        pred[source] = source;
+        pred[target] = target;
+    }
+
+    // Two queues per BFS: sQueue(next) for the exploration from the source node, tQueue(next) for
+    // the exploration from the target node.
+    std::queue<node> sQueue, sQueueNext, tQueue, tQueueNext;
```
avdgrinten: The BFS class allocates its objects locally, while the Dijkstra class stores them in the class. I would suggest to also store them in the class for BFS.

networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -32,6 +36,81 @@ TEST_F(DistanceGTest, testVertexDiameterPedantically) {
 	EXPECT_EQ(1000, vd);
 }
 
+TEST_F(DistanceGTest, testBidirectionalBFS) {
+		Aux::Random::setSeed(42, false);
+		Graph G = ErdosRenyiGenerator(500, 0.02, false).generate();
+		Graph G1 = ErdosRenyiGenerator(500, 0.05, true).generate();
+    auto testGraph = [&](const Graph &G) {
```
avdgrinten: Indentation is inconsistent.

include/networkit/distance/BidirectionalBFS.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    /*
+     * Returns the distance (i.e., number of hops) from the source to the target node.
+     *
+     * @return count Number of hops from the source to the target node.
+     */
+    count getDistance() {
+        assureFinished();
+        return stDist;
+    }
```
avdgrinten: What I don't like about this design is that STSP cannot be queried for the distance. I understand that this was done because the BFS distance is an integer. Suggestion: Add a `double getDistance()` method to STSP which works both for Dijkstra and for BFS. Keep this method but rename it to `count getHops()`.

include/networkit/distance/BidirectionalBFS.hpp
angriman: To use the base class constructor

include/networkit/distance/STSP.hpp
angriman: It is not, but then also `ts` and `ballMask` should be moved to the derived classes.
This choice was made to avoid to initialize and reset the `visited` vector two times, but if we want a more general implementation then I agree that all those members should be moved to the derived classes.

avdgrinten: Yes. Why does moving the vector to the derived classes imply that it needs to be initialized multiple times?

include/networkit/distance/BidirectionalBFS.hpp
avdgrinten: Alright. IMHO this is easy to overlook. We should make the intent explicit with a comment `// Inherit the constructors of STSP.`.

angriman: Sure I can add it

include/networkit/distance/STSP.hpp
angriman: Every time that the `run()` method is called we need to increase `ts` and, eventually, fill `visited` with zeros. Moving `visited` to the derived classes implies to rewrite the code that does this operation for each derived class if they use a bidirectional graph exploration as `BidirectionalBFS` and `BidirectionalDijkstra`. However, because we cannot assume that every derived class of `STSP` will use such strategy, I am fine to do so.

PR #337, opened 2019-06-14
Removed useless virtual methods, and misleading docstring

angriman:
Removed `virtual` from `run()` methods, removed `@return` docstring from a method returning `void`.


avdgrinten:
While we're at it: Should the classes be final? I don't think there is any meaningful way to inherit from them. Also: Are the `friend` declarations needed? Why?

angriman:
Makes sense to me, a lot of NetworKit classes implementing algorithms such as Dijkstra or BFS are unlikely to be extended.

PR #342, opened 2019-06-18
Fix warnings

angriman:
Compiling NetworKit with `-DNETWORKIT_WARNINGS=ON` triggers a lot of warnings, mainly due to trivial reasons (e.g., unused variables/parameters, useless assertions, etc.).
This fixes many of those warnings and enables warnings on all travis builds.


include/networkit/generators/quadtree/QuadNodeCartesianEuclid.hpp
```diff
@@ -429,7 +429,12 @@ class QuadNodeCartesianEuclid {
 			}
 		}
 		count finalNeighbours = result.size();
-		if (probLB == 1) assert(finalNeighbours == incomingNeighbours + ownsize);
+		if (probLB == 1){
+			assert(finalNeighbours == incomingNeighbours + ownsize);
+			(void)finalNeighbours;
```
manpen: we have `tlx::unused` for this -- it's much more verbose

networkit/cpp/sparsification/SimmelianScore.cpp
```diff
@@ -77,10 +77,10 @@ Redundancy SimmelianScore::getOverlap(	const node& ego,
  * egoNeighbors and alterNeighborsUnmatched to overlap.
  */
 void SimmelianScore::matchNeighbors(
-	const node& ego,
+	const node &,
```
manpen: Passing a const ref to an integer/bool seems just wrong -- esp. when it's not used.

include/networkit/generators/MocnikGenerator.hpp
```diff
@@ -62,18 +62,18 @@ class MocnikGenerator: public StaticGraphGenerator {
...
 	 */
-	const int toIndex(LayerState &s, const std::vector<double> &v);
+	int toIndex(LayerState &s, const std::vector<double> &v);
 
 	/**
 	 * Determine, for the given multi-dimensional index, the index of the
 	 * corresponding grid cell
 	 */
-	const int toIndex(LayerState &s, const std::vector<int> &v);
+	int toIndex(LayerState &s, const std::vector<int> &v);
```
manpen: (out of scope of this PR?) Move const to end of line and make it const this? Alsso s can be a const ref

include/networkit/generators/MocnikGenerator.hpp
```diff
@@ -62,18 +62,18 @@ class MocnikGenerator: public StaticGraphGenerator {
 	/**
 	 * Add the node with number j to the corresponing grid cell
 	 */
-	const void addNode(LayerState &s, const int &j);
+	void addNode(LayerState &s, const int &j);
```
manpen: const ref to integer

include/networkit/generators/quadtree/QuadNodeCartesianEuclid.hpp
manpen: I won't fight over it, but the first time I came across this construct, it took me a few moments to understand what's going on. And using `tlx::unused` clearly states intent.

networkit/cpp/centrality/DynBetweennessOneNode.cpp
```diff
@@ -91,7 +91,7 @@ void DynBetweennessOneNode::update(GraphEvent event) {
 		// queue with all visited nodes
 		// if u has a new shortest path going through v, it updates the distance of u
 		// and inserts u in the priority queue (or updates its priority, if already in Q)
-		auto updateQueue = [&](node u, edgeweight priority) {
+		auto updateQueue = [&](node u, edgeweight) {
```
manpen: `updateQueue` is not passed to any function. Hence, removing the two places of invocation to not pass the second parameter seems better.

networkit/cpp/sparsification/SimmelianScore.cpp
avdgrinten: Agreed.

include/networkit/generators/MocnikGenerator.hpp
avdgrinten: Out of scope.

networkit/cpp/generators/MocnikGenerator.cpp
```diff
@@ -292,11 +292,11 @@ Graph MocnikGenerator::generate() {
 	assert(dim > 0);
 	for (auto &n : ns) {
```
manpen: Rewrite as `assert(std::all_of(ns.cbegin(), ns.cend(), [] (count n) {return n >1;}));` and `assert(std::all_of(ks.cbegin(), ks.cend(), [] (double k) {return k >1.0;}));`

include/networkit/graph/Graph.hpp
```diff
@@ -9,6 +9,13 @@
 #ifndef GRAPH_H_
 #define GRAPH_H_
 
+#if __cplusplus >= 201402L
```
manpen: Why not use `TLX_DEPRECATED` ?

PR #343, opened 2019-06-18
Fix unit test parser

manpen:
When starting unit test binaries without a command line parameter, GoogleTest presents the following warning:

```
IMPORTANT NOTICE - DO NOT IGNORE:
This test program did NOT call testing::InitGoogleTest() before calling RUN_ALL_TESTS(). This is INVALID. Soon Google Test will start to enforce the valid usage. Please fix it ASAP, or IT WILL START TO FAIL.
```

This is message is misleading, as `RUN_ALL_TESTS` was indeed call, but with erroneous `argc` and `argv`; this is fixed now. While I was at it, I also remove now obsolete checks and command line parameters and fixed some includes. I also fixed the indentation (which was a mix of Tabs, 4 space and 2 spaces).

CI Test AppleClang 10.13 will probably be failing again due to an Homebrew issue.


networkit/cpp/Unittests-X.cpp
```diff
@@ -1,81 +1,58 @@
...
-{TESTS,		0,	"t",	"tests",		OptionParser::Arg::None,	"  --tests \t Run unit tests"},
-{RUNNABLE,	0,	"r",	"run",			OptionParser::Arg::None,	"  --run \t Run unit tests which don't use assertions"},
-{DEBUG,		0,	"d",	"debug",		OptionParser::Arg::None,	"  --debug \t Run tests to debug some algorithms"},
-{BENCHMARKS,0,	"b",	"benchmarks",	OptionParser::Arg::None,	"  --benchmarks \t Run benchmarks"},
-{FILTER,	0,	"f",	"gtest_filter",	OptionParser::Arg::Required,"  --gtest_filter=<FILTER_PATTERN> \t Run tests that match the filter pattern" },
-{UNKNOWN,	0,	"",		"",				OptionParser::Arg::None,	"\nExamples:\n TODO" },
-{0,			0,	0,		0,				0,							0}
+enum  optionIndex { UNKNOWN, HELP, LOGLEVEL, THREADS, TESTS, RUNNABLE, DEBUG, BENCHMARKS };
+const OptionParser::Descriptor usage[] = {
+    {UNKNOWN,	0,	"" ,	"",				OptionParser::Arg::None,	"Options:" },
```
avdgrinten: If we touch this, it might be a good idea to fix the alignment by replacing tabs by spaces.

manpen: Yes, indeed.

networkit/cpp/Unittests-X.cpp
```diff
@@ -75,13 +75,14 @@ int main(int argc, char **argv) {
 	// feenableexcept(FE_ALL_EXCEPT);
 #endif
 
-	std::string program_name = argv[0];
-	// PARSE OPTIONS
-	argc-=(argc>0); argv+=(argc>0); // skip program name argv[0] if present
+	std::string program_name(argc ? argv[0] : "");
```
avdgrinten: Is there really any OS where the program name is *not* in argv/argc?

networkit/cpp/Unittests-X.cpp
```diff
@@ -75,13 +75,14 @@ int main(int argc, char **argv) {
...
 
-	OptionParser::Stats  stats(usage, argc, argv);
+	// PARSE OPTIONS
+	const auto parse_argc = std::max(0, argc - 1);
+	const auto parse_argv = argv + (argc>0); // skip program name argv[0] if present
```
avdgrinten: If I understand this change correctly, we need these local variables here as we still need the original argv for gtest, right?

networkit/cpp/Unittests-X.cpp
```diff
@@ -1,81 +1,58 @@
...
-// Version     :
-// Copyright   : 2012, Christian Staudt, Henning Meyerhenke
-// Description : Calling unit tests and benchmarks
-//============================================================================
+/*
+ * Unittests-X.cpp
+ *
+ *  Created on: 2012
+ *      Author: Christian Staudt <christian.staudt@kit.edu>,
+ *              Henning Meyerhenke <henning.meyerhenke@kit.edu>,
```
angriman: I guess that these email addresses are quite outdated, can we do something about this?

manpen: My suggestion would be the following: Leave them in for the moment and in near future reformat all headers to have consistent layout and information. For instances, some headers state `cls` or `henning`. Then we can keep the mail addresses at one central place where all authors are listed.

networkit/cpp/Unittests-X.cpp
manpen: Yes, this was the underlying bug, which triggered this PR. I first generated `stats` and `option` via an lambda, to scope those vars. But this seemed to make it much harder to read.

networkit/cpp/Unittests-X.cpp
manpen: I also wondered that. The standard says that *if* `argc>0` then `argv[0]` is the program name. So we cannot assume `argc > 0` (even if I'm not aware of any OS that doesn't pass the name).  But in the end, it seems really not worth to introduce UB to get rid of three simple checks.

networkit/cpp/Unittests-X.cpp
avdgrinten: Sounds reasonable.

avdgrinten:
It does not make sense to do it in this patch series but in the long term we might want to get rid of this option parser library (which is embedded in the NetworKit source) and replace it by tlx' implementation.

manpen:
I agree. Maybe I've time on the weekend to have a look at it -- then we do not need another PR.

manpen:
Done.

avdgrinten:
Thank you. Does this have any impact on the command line syntax? Or do all (or at least all commonly used options still work?

manpen:
The only change I'm aware of is that TLX does not support `--longParam=123` but rather uses `--longParam 123` syntax. This affects parameters `theads` and `loglevel` (hence the changed description).  There can also be some minor changes, as previously the parameter passed to `GTest` were not processed properly. So, there's a chance, that if e.g., a `gtest_filter` passed as the first argument is now working even if it did not previously.

hmeyerhenke:
As always: If this changes an interface (in this case the command line), then make sure this is highlighted in the release notes. Is there already a document that tracks such changes until the next release? If not, I think there should be one.

Issue #340, opened 2019-06-18
extractLargestConnectedComponent introduced circular dep


manpen:
Sorry I missed this earlier, but: #317 added `extractLargestConnectedComponent` to GraphTools in the `graph` module. The implementation, however, depends on the `components` module. This is currently not reflected in CMake (I have a PR ready, but see below), and hence leads to linking errors when building non-monolithic tests (which is currently not done in Travis and I will have a PR for this as well).

From a dependency management point of view, it would make more sense to move `extractLargestConnectedComponent` to the `components` module because:
- We introduced in a circular dependency between those modules. As we do not support static non-monolith builds, this is not issue per-se. But it might cause problems in the future.
- Every non-monolithic tests now also needs to pull in `components`, even if its not required.

avdgrinten:
Circular dependencies cannot work (without hacks), even in the shared library case (which library would could build first?). This needs to be fixed indeed. Let's add a non-monolithic compilation to Travis to avoid similar regressions in the future. I am looking forward to your PR.

manpen:
Fixed in #341 and #344

PR #344, opened 2019-06-18
Move extractLargestConnectedComponent to component

manpen:
Fixes issue #340 by moving `extractLargestComponent` from `GraphTools` to `ConnectedComponent`. As a bonus, there are now also Python Test for the interface and const methods in `ConnectedComponent`.

CI Homebrew issue still causing one test to fail.


avdgrinten:
Looks good to me. Thank you for the PR.

PR #345, opened 2019-06-19
Support for coverage

angriman:
Support for coverage using coveralls


angriman:
Reopened in #346 to test functionality

PR #349, opened 2019-06-19
Catch exceptions by reference

angriman:
Fix warnings generated by catching exceptions by value.

### DIFF
--- a/networkit/cpp/io/GMLGraphReader.cpp
+++ b/networkit/cpp/io/GMLGraphReader.cpp
@@ -69,8 +69,7 @@ Graph GMLGraphReader::read(const std::string& path) {
 						nodeMap.insert(std::make_pair(pair.second,u));
 						DEBUG("added node: ",u,", ",pair.second);
 					}	
-				} catch (std::exception e) {
-					//throw std::runtime_error("something went wrong when parsing a node");
+				} catch (const std::exception &e) {
 					return false;
 				}
 				std::getline(graphFile, line);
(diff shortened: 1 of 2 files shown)


networkit/cpp/io/GMLGraphReader.cpp
```diff
@@ -69,7 +69,7 @@ Graph GMLGraphReader::read(const std::string& path) {
 						nodeMap.insert(std::make_pair(pair.second,u));
 						DEBUG("added node: ",u,", ",pair.second);
 					}	
-				} catch (std::exception e) {
+				} catch (std::exception &e) {
 					//throw std::runtime_error("something went wrong when parsing a node");
```
avdgrinten: Those comments should be removed.

networkit/cpp/io/GMLGraphReader.cpp
```diff
@@ -69,7 +69,7 @@ Graph GMLGraphReader::read(const std::string& path) {
 						nodeMap.insert(std::make_pair(pair.second,u));
 						DEBUG("added node: ",u,", ",pair.second);
 					}	
-				} catch (std::exception e) {
+				} catch (std::exception &e) {
```
avdgrinten: We should catch by `const` reference, unless we need to modify the exception (which would be pretty strange).

networkit/cpp/io/METISParser.cpp
```diff
@@ -76,7 +76,7 @@ static std::vector<std::pair<node,double>> parseWeightedLine(std::string line, c
 			std::tie(weight, it) = Aux::Parsing::strTo<double,decltype(it),Aux::Checkers::Enforcer>(it,end);
 			adjacencies.push_back(std::make_pair(v,weight));
 			content << v << " " << weight << "\t";
-		} catch (std::exception e) {
+		} catch (std::exception &e) {
```
avdgrinten: This should be `const`.

avdgrinten:
LGTM now. We can merge it after the CI runs.

PR #350, opened 2019-06-19
Fix travis Apple build

angriman:
Fix travis build with osx 10.13 and Apple Clang


avdgrinten:
Thank you for fixing this!

PR #351, opened 2019-06-20
Fix IOGTests

angriman:
Tests for the `NetworkitBinaryReader/Writer` are not executed due to wrong naming.
This fixes the test names so they are executed with the `ctest` command.


angriman:
Should we merge this?

avdgrinten:
Yes. Merging.

Issue #339, opened 2019-06-17
Error in algorithm DynBetweenness


gilcu3:
Algorithm DynBetweenness returns wrong results. Minimal example:
Directed Graph with nodes {0, 1, 2} and edges {(0, 1), (1, 2)}. When edge (2, 0) is inserted, the aforementioned algorithm returns that the betweenness of node 0 is 0, and it should be 1.

angriman:
Hi, dynamic algorithms do not update the ranking automatically, you should use the `update` method as reported [here](https://networkit.github.io/dev-docs/python_api/centrality.html?highlight=dynbetw#networkit.centrality.DynBetweenness.update). The `update` method gets a `GraphEvent` as input, for example:

```
g = Graph(3, False, True)
g.addEdge(0, 1)
g.addEdge(1, 2)

db = centrality.DynBetweenness(g)
db.run()
print(db.ranking()) # [(1, 1.0), (0, 0.0), (2, 0.0)]

g.addEdge(2,0) # Add new edge (2, 0)
# Update the ranking
db.update(dynamic.GraphEvent(dynamic.GraphEvent.EDGE_ADDITION, 2, 0, 1))
print(db.ranking()) # [(0, 1.0), (1, 1.0), (2, 1.0)]
```

To do multiple updates at once, use `updateBatch`, more about it [here](https://networkit.github.io/dev-docs/python_api/centrality.html?highlight=dynbetw#networkit.centrality.DynBetweenness.updateBatch).

gilcu3:
Thank you for your response, I forgot to say, I am using the library from c++, and the master branch. This is my code:
```
auto d = DynBetweenness(g);
d.run();
auto e = GraphEvent(GraphEvent::EDGE_ADDITION, edge.first, edge.second);
d.update(e);
d.scores();
```

Which branch did you test?

angriman:
I tested the Dev branch, but the implementation of `DynBetweenness` should be the same in both branches.

As you wrote, you should insert the edge (2, 0) to the graph (i.e., run `g.addEdge(2,0)`) before doing the update. By design, all NetworKit classes implementing algorithms (like `DynBetweenness`) are not allowed to make changes to the input graphs; otherwise, if you had another dynamic algorithm class working on the same graph `g`, they would have added the same edge twice.

A working example in C++ would be:

```
#include <networkit/centrality/DynBetweenness.hpp>
#include <networkit/dynamics/GraphEvent.hpp>
[...]

PR #352, opened 2019-06-20
fixing division by zero error in DynBetweenness

gilcu3:
Fixes error found in [link](https://github.com/kit-parco/networkit/issues/339#issuecomment-503653021)


gilcu3:
Added the tests by which the original bug was found

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -16,6 +16,9 @@
 #include <networkit/graph/Sampling.hpp>
 #include <networkit/generators/DorogovtsevMendesGenerator.hpp>
 #include <networkit/generators/ErdosRenyiGenerator.hpp>
+#include <networkit/centrality/DynBetweenness.hpp>
+#include <networkit/auxiliary/Random.hpp>
+#include <random>
```
angriman:
```suggestion
```
`<random>` is not needed, you can just use `<networkit/auxiliary/Random.hpp>` to extract random nodes.

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
 	DEBUG("Number of samples: ", bc1.numberOfSamples());
 }
 
+TEST_F(DynBetweennessGTest, runDynVsStaticCaseInsertDirected){
+	srandom(0);
```
angriman:
```suggestion
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+	Aux::Random::setSeed(0, false);
+
+	auto genEdgeInsert = [](const Graph& g){
+		int n = g.numberOfNodes();
```
angriman:
```suggestion
		count n = g.numberOfNodes();
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+		while(true){
+			int u = random() % n, v = random() % n;
+			if(u == v or g.hasEdge(u, v))continue;
+			return std::make_pair(u, v);
+		}
+		//return {-1, -1};
```
angriman:
```suggestion
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+	srandom(0);
+	Aux::Random::setSeed(0, false);
+
+	auto genEdgeInsert = [](const Graph& g){
+		int n = g.numberOfNodes();
+		while(true){
+			int u = random() % n, v = random() % n;
+			if(u == v or g.hasEdge(u, v))continue;
+			return std::make_pair(u, v);
+		}
```
angriman: This loop should be implemented as follows:
```
node u = g.randomNode();
node v = g.randomNode();
while (u == v || g.hasEdge(u, v))
    v = g.randomNode();
return std::make_pair(u, v);
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+			brandes.run();
+			g.forNodes([&](node & v){
+				EXPECT_NEAR(brandes.score(v), ibet.score(v), 1e-8);
+			});
+		}
+}
+
+TEST_F(DynBetweennessGTest, runDynVsStaticCaseInsertUndirected){
+	//srandom(0);
+	//Aux::Random::setSeed(0, false);
```
angriman:
```suggestion
	Aux::Random::setSeed(0, false);
```
The test should be be reproducible.

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+			auto brandes = Betweenness(g);
+			brandes.run();
+			g.forNodes([&](node & v){
+				EXPECT_NEAR(brandes.score(v), ibet.score(v), 1e-8);
+			});
+		}
+}
+
+TEST_F(DynBetweennessGTest, runDynVsStaticCaseInsertUndirected){
+	//srandom(0);
```
angriman:
```suggestion
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+
+	auto genEdgeInsert = [](const Graph& g){
+		int n = g.numberOfNodes();
+		while(true){
+			int u = random() % n, v = random() % n;
+			if(u == v or g.hasEdge(u, v))continue;
+			return std::make_pair(u, v);
+		}
+		//return {-1, -1};
+	};
```
angriman: Same comments as in the `genEdgeInsert` above.

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+			ibet.update(ne);
+			assert(g.hasEdge(e.first, e.second));
+
+			auto brandes = Betweenness(g);
+			brandes.run();
+			g.forNodes([&](node & v){
+				EXPECT_NEAR(brandes.score(v), ibet.score(v), 1e-8);
+			});
+		}
+}
```
angriman: Same comments as above

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+			auto e = genEdgeInsert(g);
+			auto ibet = DynBetweenness(g);
+
+			ibet.run();
+
+			auto ne = GraphEvent(GraphEvent::EDGE_ADDITION, e.first, e.second);
+			g.addEdge(e.first, e.second);
+			ibet.update(ne);
+
+			assert(g.hasEdge(e.first, e.second));
```
angriman:
```suggestion
			EXPECT_TRUE(g.hasEdge(e.first, e.second));
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+		int n = g.numberOfNodes();
+		while(true){
+			int u = random() % n, v = random() % n;
+			if(u == v or g.hasEdge(u, v))continue;
+			return std::make_pair(u, v);
+		}
+		//return {-1, -1};
+	};
+
+	for(int n = 2; n <= 25; n++)
```
angriman:
```suggestion
	for(count n = 2; n <= 25; n++)
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
```diff
@@ -134,5 +137,80 @@ TEST_F(DynBetweennessGTest, runApproxBetweenness) {
...
+		for(int t = 0; t < 100; t++){
```
angriman:
```suggestion
		for(count t = 0; t < 100; t++){
```

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
gilcu3: If the loop is implemented that way, and node u is full, that while would never end. Following your suggestion I will put it like:
```
node u = g.randomNode();
node v = g.randomNode();
while (u == v || g.hasEdge(u, v)){
    u = g.randomNode();
    v = g.randomNode();
}
return std::make_pair(u, v);
```
do you agree?

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
gilcu3: What do you mean here?

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
angriman: You are right, I agree.

networkit/cpp/centrality/test/DynBetweennessGTest.cpp
angriman: Sorry my comment was not clear. As you already did, I meant to apply the same changes that I suggested for the for-loop in `runDynVsStaticCaseInsertDirected`.

gilcu3:
guidance is welcomed, is it possible to.do it from the web?

angriman:
No problem: Your PR has three commits (9757a59, 867dc8b, and ea5f732), and we want to create a single commit out of those three commits. To do so we need to use the rebase tool to rewrite the commit history. Type:
```
git rebase -i 9757a59~
```
This should open your default text editor with this content:
```
pick 9757a59 fixing division by zero error in DynBetweenness
pick 867dc8b added test for DynBetweenness against static algorithm Betweenness
pick ea5f732 changes suggested by @angriman
...
```

Change the second two `pick` into `fixup`, like this:
```
pick 9757a59 fixing division by zero error in DynBetweenness
fixup 867dc8b added test for DynBetweenness against static algorithm Betweenness
fixup ea5f732 changes suggested by @angriman
...
```

Save and close. Then, force-push to your remote branch (usually it is named `origin`):
```
git push -f origin Dev
```
And you are done. More about this process [here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

avdgrinten:
It's not really clear to me why this is the correct fix algorithmically. What does it mean if `sigmaOld[x][y]` is zero? Why is the dependency of `w` not affected in this case?

gilcu3:
if sigmaOld[x][y] is zero, then there was no shortest path from x to y, and so this pair had no contribution to the dependency of w

avdgrinten:
Okay, makes sense. I am merging this PR.

PR #354, opened 2019-06-21
Support for directed graphs in NetworkitBinaryGraph file format

cndolo:
1. The file format needed to be changed, as we need to save two prefix sums per directed graphs; one for the in edges and one for the out edges.
2. Directed graphs are now supported
3. We can now read in parallel

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -994,17 +994,17 @@ class Graph final {
      * @param weight Optional edge weight.
      */
     void addEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
-    
-	/**
-     * Insert an edge between the nodes @a u and @a v. Unline the addEdge function, this function does not not add any information to v. If the graph is
-     * weighted you can optionally set a weight for this edge. The default
-     * weight is 1.0. Note: Multi-edges are not supported and will NOT be
-     * handled consistently by the graph data structure.
-     * @param u Endpoint of edge.
-     * @param v Endpoint of edge.
-     * @param weight Optional edge weight.
-     */
-	void addPartialEdge(Unsafe, node u, node v, uint64_t index = 0, edgeweight ew = defaultEdgeWeight);
+
+    /**
+    * Insert an edge between the nodes @a u and @a v. Unline the addEdge function, this function does not not add any information to v. If the graph is
+    * weighted you can optionally set a weight for this edge. The default
+    * weight is 1.0. Note: Multi-edges are not supported and will NOT be
+    * handled consistently by the graph data structure.
+    * @param u Endpoint of edge.
+    * @param v Endpoint of edge.
+    * @param weight Optional edge weight.
+    */
+    void addPartialEdge(Unsafe, node u, node v, uint64_t index = 0, edgeweight ew = defaultEdgeWeight);
 
     /**
      * Insert an in edge between the nodes @a u and @a v in a directed graph. If the graph is
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -57,6 +57,8 @@ struct Edge {
 inline bool operator==(const Edge &e1, const Edge &e2) {
     return e1.u == e2.u && e1.v == e2.v;
 }
+struct Unsafe {};
+static constexpr Unsafe unsafe {};
 } // namespace NetworKit
 
 namespace std {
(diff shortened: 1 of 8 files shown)


include/networkit/graph/Graph.hpp
```diff
@@ -973,6 +994,40 @@ class Graph final {
      * @param weight Optional edge weight.
      */
     void addEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
+    
+	/**
```
manpen: Indent is off.

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -23,10 +23,8 @@ enum WEIGHT_FORMAT {
 enum MASKS {
```
manpen: Please define `MASKS` and `WEIGHT_FORMAT` as `enum class` reduce pollution of the namespace.

networkit/cpp/graph/Graph.cpp
```diff
@@ -179,6 +179,19 @@ void Graph::preallocateUndirected(node u, size_t size) {
 	}
 }
 
+void Graph::preallocateDirected(node u, size_t outSize, size_t inSize) {
```
manpen: Assert that (also for `preallocateUndirected`)
- the graph is directed
- `u` is in range

Probably we do not want to assert that `outSize` and `inSize` are within range, as a user might know that the graph will be growing in future?

manpen:
While this is technically a comment to the last PR (#325), it seems only plausible to fix it here.

In header files please only include other headers that are really necessary and move header files only required by the implementation to the CPP file. e.g., `NetworkitBinaryWriter.hpp`:

```cpp
#include <networkit/io/GraphWriter.hpp>
#include <networkit/auxiliary/Enforce.hpp> // not needed in header
#include <networkit/io/NetworkitBinaryGraph.hpp> // not needed in header
#include <tlx/math/clz.hpp> // not needed in header

#include <fstream> // not needed in header
#include <cstring>// not needed in header
```

Strictly speaking `<cstdint>` is required for `uint64_t`; but I think it's fair to assume it comes with the `Graph`.

include/networkit/io/NetworkitBinaryGraph.hpp
avdgrinten: An `enum class` does not work well here: we need the numerical values of those constants (and enum classes need casts to access that). But you are right that this pollutes the namespace. My suggestion: move this to a `namespace nkbg` and change the values to be `static constexpr uint8_t DELETED_BIT = 1;` and similar (with a constant-specific type instead of `uint8_t`). This also has the advantage that they will have the proper type.

manpen:
While this is technically a comment to the last PR (#325), it seems only plausible to fix it here.

Please mark `NetworkitBinaryWriter::write` as `override`.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -102,46 +98,77 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+	memcpy(&adjListSize, adjIt + (chunks -1) * sizeof(uint64_t), sizeof(uint64_t));
+	uint64_t transposeListSize;
+	memcpy(&transposeListSize, transpIt + (chunks -1) * sizeof(uint64_t), sizeof(uint64_t));
+
+	if(!directed) {
+		assert(adjListSize == transposeListSize);
+	} 
+	G.setEdgeCount(unsafe, adjListSize);
+	
+	uint64_t selfLoops = 0;
```
manpen: Make atomic as concurrent write are possible. Alternatively return number of local self-loops from `constructGraph` and use omp reduction.

avdgrinten: Oh, that's correct. Using a `std::atomic<count>` variable here, with a single `.fetch_add(<number of self loops in chunk>, std::memory_order_relaxed);` looks like the best option here. That will be much faster than a OMP reduction (and also less complex).

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -102,46 +98,77 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+			uint64_t inNbrs;
+			transpOff += NetworkitBinaryReader::decode(reinterpret_cast<const uint8_t*>(transpIt + transpOff), inNbrs);
 			if(!directed) {
-				G.preallocateUndirected(curr, nbrs);
+				G.preallocateUndirected(curr, outNbrs+inNbrs);
+			} else  {
+				G.preallocateDirected(curr, outNbrs, inNbrs);
+			}
+			//Read adjacency lists.
+			uint64_t add;
```
manpen: Move `add` into for-loop as it's value is only relevant in this context.

avdgrinten:
LGTM now. Thank you @manpen for the review and thanks to @cndolo for the PR.

PR #359, opened 2019-06-27
Refactor CSRMatrix

angriman:
This adds a template parameter to the `CSRMatrix` class allowing to store generic data types. The class is renamed to `CSRGeneralMatrix`. Additional minor changes:
- Remove ad-hoc implementation of binary search in favor of `std::lower_bound`
- Add `noexcept` operator where applicable
- Replace `push_back` with `emplace_back` where applicable (and addition of two new constructors for `Triplet`)


include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1220 @@
+/*
+ * CSRGeneralMatrix.hpp
+ *
+ *  Created on: May 6, 2015
+ *      Author: Michael Wegner (michael.wegner@student.kit.edu)
```
avdgrinten: Add yourself as an author (in addition to Michael Wegner).

include/networkit/algebraic/CSRMatrix.hpp
```diff
@@ -1,612 +0,0 @@
-/*
- * CSRMatrix.h
- *
- *  Created on: May 6, 2015
- *      Author: Michael Wegner (michael.wegner@student.kit.edu)
- */
```
avdgrinten: We should not delete the entire header file. This requires quite a bit of refactoring in NetworKit and it might also require users to adopt. Instead, keep `CSRMatrix.hpp` header, let it include the `CSRGeneralMatrix.hpp` header and move the typedef to the `CSRMatrix.hpp` header.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1216 @@
...
+	 * If @a j is not present, the index that is immediately left of the place where @a j would be
+	 * is returned. If
+	 * @param left
+	 * @param right
+	 * @param j
+	 * @return The position of column @a j in columnIdx or the element immediately to the left of the place where @a j
+	 * would be.
+	 */
+	index binarySearchColumns(index left, index right, index j) const {
+	assert(sorted());
```
avdgrinten: Indentation is broken here (and in large parts of the remainder of the file). Please fix it and make it consistent.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1216 @@
...
+	} else if (columnIdx[mid] > j && mid > 0) {
+		return binarySearchColumns(left, mid-1, j);
+	} else {
+		return binarySearchColumns(mid+1, right, j);
+	}
+	}
+
+public:
+	/** Default constructor */
+	CSRGeneralMatrix() : rowIdx(0), columnIdx(0), nonZeros(0), nRows(0), nCols(0), isSorted(true), zero(0) {}
```
avdgrinten: Move the the initializer list to its own line:
```cpp
CSRGeneralMatrix()
: rowIdx(0), columnIdx(0), nonZeros(0), nRows(0), nCols(0), isSorted(true), zero(0) {}
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1313 @@
...
+     * false.
+     */
+    CSRGeneralMatrix(const count nRows, const count nCols,
+                     const std::vector<Triplet> &triplets,
+                     const ValueType zero = 0, bool isSorted = false)
+        : nRows(nRows), nCols(nCols), isSorted(isSorted), zero(zero) {
+        count nnz = triplets.size();
+        rowIdx = std::vector<index>(nRows + 1, 0);
+        columnIdx = std::vector<index>(nnz);
+        nonZeros = std::vector<double>(nnz);
```
avdgrinten: Actually, looking at the code, there are still a lot of places, where `double` is used. This does not look intentional?

angriman: My bad, it is not intentional, I'll change them to `ValueType`.

> LGTM now. I did not review the algorithm; I assume that they did not change except for changing `double` -> `ValueType` (?).

Exactly, apart from that there are no further changes.

avdgrinten:
Why does the code break with an earlier version of Clang? We should not increase the version of Clang without a good reason. People who want run NetworKit on clusters that they do not control might not have the option to upgrade Clang.

angriman:
The build fails with clang 3.8 because of this error:
```
/home/travis/build/networkit/networkit/include/networkit/algebraic/CSRGeneralMatrix.hpp:1231:10: error: 
      definition with same mangled name as another definition
    void forNonZeroElementsInRow(index row, L handle) const {
         ^
/home/travis/build/networkit/networkit/include/networkit/algebraic/CSRGeneralMatrix.hpp:1231:10: note: 
      previous definition is here
```

I searched for this error and did not find much apart [this](https://bugs.llvm.org/show_bug.cgi?id=33924) page. It could be a bug in clang 3.8 but I am not sure about that. Do you have some other solutions in mind?

angriman:
Moving `forNonZeroElementsInRow` somewhere else in the class did not solve the issue, I still get the same error. Shall we switch to clang 3.9?

angriman:
Now that we raised the minimum supported clang version to 3.9 this PR can be reviewed.

fabratu (dismissed):
LGTM.

angriman:
We also need approval from @avdgrinten.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1251 @@
...
+     */
+    void quicksort(index left, index right) {
+        if (left >= right)
+            return;
+        index pivotIdx = partition(left, right);
+        if (pivotIdx != 0) {
+            quicksort(left, pivotIdx - 1);
+        }
+        quicksort(pivotIdx + 1, right);
+    }
```
avdgrinten, line 58: We do we use an ad-hoc quicksort instead of using the standard library?

avdgrinten (commented):
I have one comment below.

As a side note: a few algorithms are *really* inefficient here, for example setValue(). There is no way to implement setValue() well for CSR, so why does the class provide it at all?

angriman:
I agree that there are several inefficiencies in this class that should be fixed, I already fixed a trivial one in the last commit. However, I think that fixing those inefficiencies is out of the scope of this PR and it should be done on a separate PR.

hmeyerhenke:
Not sure that setValue() is really obsolete. But the docs should be clear about the inherent inefficiency. Other -- unnecessary -- inefficiencies should be addressed. This is rather old code that has not been touched for quite some time, I believe.

hmeyerhenke:
But I leave it up to you guys if they are addressed in this PR or separately.

avdgrinten:
It's not necessarily the case that the code is inefficient due to the implementation but it just provides tons of operations that you really would not use if you want your code to be fast (setting values, getting individual values, extracting columns, etc.). These operations are simply not fast because CSR cannot implement them in a fast way.

hmeyerhenke:
I know @avdgrinten. Such operations should not be the preferred way to do things, I completely agree. But there may be cases where they are simply necessary. We can make a decision, though, that we do not want to support such cases. But that's a decision I'd like to be involved in and should follow from a comprehensive discussion of use cases.

angriman:
It might be a good idea to open a new issue/discussion about that.

angriman:
Are there any other issues that needs to be addressed in this PR?

include/networkit/algebraic/CSRGeneralMatrix.hpp
angriman: A possible reason for this is because we need to sort two vectors at once (columnIdx and nonZero). We could replace this ad-hoc quicksort by computing the permutation of indices using `std::sort` and then apply the permutation to the two vectors as done in #734 (but to do so we might want to merge the `applyPermutation` function in #734 first).
Do you know if there is a better way to do this?

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -0,0 +1,1226 @@
...
+     * element are the same as well as
+     * all entries, otherwise returns true.
+     * @param other
+     */
+    bool operator!=(const CSRGeneralMatrix &other) const { return !((*this) == other); }
+
+    /**
+     * @return Number of rows.
+     */
+    inline count numberOfRows() const noexcept { return nRows; }
```
avdgrinten, line 282: `inline` is redundant if a method is defined within `class { ... }` or `struct { ... }`.

include/networkit/algebraic/Vector.hpp
```diff
@@ -309,7 +309,7 @@ Matrix Vector::outerProduct(const Vector &v1, const Vector &v2) {
         for (index j = 0; j < v2.getDimension(); ++j) {
             double result = v1[i] * v2[j];
             if (fabs(result) >= FLOAT_EPSILON) {
-                triplets.push_back({i, j, result});
+                triplets.emplace_back(i, j, result);
```
avdgrinten, line 312: Unrelated change?

include/networkit/numerics/LAMG/Lamg.hpp
```diff
@@ -148,7 +148,7 @@ void Lamg<Matrix>::setup(const Matrix& laplacianMatrix) {
 
             for (node u : components[compIdx]) {
                 G.forNeighborsOf(u, [&](node v, edgeweight w) {
-                    triplets.push_back({graph2Components[u], graph2Components[v], w});
+                    triplets.emplace_back(graph2Components[u], graph2Components[v], w);
```
avdgrinten, line 151: Unrelated change?

include/networkit/numerics/LAMG/MultiLevelSetup.hpp
```diff
@@ -344,7 +344,7 @@ void MultiLevelSetup<Matrix>::eliminationOperators(const Matrix& matrix, const s
             if (fSet[k] == j) {
                 q[k] = 1.0 / w;
             } else {
-                triples.push_back({k, coarseIndex[j], w});
+                triples.emplace_back(k, coarseIndex[j], w);
```
avdgrinten, line 347: Unrelated change?

networkit/cpp/algebraic/DynamicMatrix.cpp
```diff
@@ -246,7 +246,7 @@ DynamicMatrix DynamicMatrix::extract(const std::vector<index>& rowIndices, const
         (*this).forNonZeroElementsInRow(rowIndices[i], [&](index k, double value) {
             if (!columnMapping[k].empty()) {
                 for (index j : columnMapping[k]) {
-                    triplets.push_back({i, j, value});
+                    triplets.emplace_back(i, j, value);
```
avdgrinten, line 249: Unrelated change?

networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -105,10 +105,10 @@ CSRMatrix MatrixMarketReader::read(std::istream& in) {
                 throw std::runtime_error("invalid index: " + std::to_string(j));
             }
 
-            triplets.push_back({i,j,val});
+            triplets.emplace_back(i,j,val);
```
avdgrinten, line 108: Unrelated change?

networkit/cpp/numerics/LAMG/MultiLevelSetup.cpp
```diff
@@ -117,7 +117,7 @@ void MultiLevelSetup<CSRMatrix>::galerkinOperator(const CSRMatrix& P, const CSRM
         }
 
         spa.gather([&](index i, index j, double value) {
-            triplets.push_back({i,j,value});
+            triplets.emplace_back(i,j,value);
```
avdgrinten, line 120: Unrelated change?

networkit/cpp/viz/PivotMDS.cpp
```diff
@@ -35,7 +35,7 @@ void PivotMDS::run() {
         PQ.insert(0, pivots[j]);
         while (!PQ.empty()) {
             node v = PQ.extractMin().second;
-            triplets.push_back({v, j, dist[v]});
+            triplets.emplace_back(v, j, dist[v]);
```
avdgrinten, line 38: Unrelated change?

include/networkit/algebraic/Vector.hpp
angriman: If I do not change this then the clang-tidy build fails due to `modernize-use-emplace`.

avdgrinten: Why doesn't clang-tidy fail right now?

angriman:
I addressed the comments, the `emplace_back` were not really necessary, so I removed them.

avdgrinten (approved):
LGTM now.

PR #362, opened 2019-07-01
logging: Replace NOLOGGING

avdgrinten:
As mentioned in https://github.com/networkit/networkit/pull/361#issuecomment-507181713, the current behavior of the `-NOLOGGING` processor definition (and the corresponding CMake option `-DNETWORKIT_LOGGING=OFF`) is problematic.

This PR fixes the issue by splitting the responsibilities of `-DNOLOGGING` into two parts:
- NETWORKIT_QUIET_LOGGING sets the default log level to a new QUIET setting. This setting disables logging at run time. However, the logging calls are still compiled in and other log levels can explicitly be set at run time.
- NETWORKIT_RELEASE_LOGGING disables the compilation of logging calls for TRACE and DEBUG levels. This option is automatically enabled in release mode but can also be manually enabled/disabled using the CMake option of the same name.

Library users who want to hide the logging functionaltiy of NetworKIT should now set `-DNETWORKIT_QUIET_LOGGING` (instead of `-DNETWORKIT_LOGGING=OFF`) during configuration.


CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ endif("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
 # BUILD OPTIONS
 option(NETWORKIT_BUILD_CORE "Build NetworKit core library" ON)
 option(NETWORKIT_BUILD_TESTS "Build NetworKit C++ tests" OFF)
-option(NETWORKIT_LOGGING "Build with logging support" ON)
```
manpen: Should we first mark `NETWORKIT_LOGGING` as DEPREACTED and issue an error message hinting to the new options if it is ever set to OFF?

networkit/cpp/auxiliary/Log.cpp
```diff
@@ -41,6 +43,8 @@ std::string getLogLevel() {
 		return "ERROR";
 	} else if (current == LogLevel::fatal) {
 		return "FATAL";
+	} else if (current == LogLevel::quiet) {
```
manpen: It seems that this if-block should be a switch-case

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(NETWORKIT_COVERAGE "Build with support for coverage" OFF)
 set(NETWORKIT_PYTHON "" CACHE STRING "Directory containing Python.h. Implies MONOLITH=TRUE")
 set(NETWORKIT_PYTHON_SOABI "" CACHE STRING "Platform specific file extension. Implies MONOLITH=TRUE")
 set(NETWORKIT_WITH_SANITIZERS "" CACHE STRING "Uses sanitizers during the compilation")
+set(NETWORKIT_RELEASE_LOGGING "AUTO" CACHE STRING "Do not compile log messages at levels TRACE or DEBUG")
```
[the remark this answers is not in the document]
avdgrinten: Good point.

networkit/cpp/auxiliary/test/AuxGTest.cpp
```diff
@@ -37,7 +37,7 @@ class AuxGTest: public testing::Test{};
 
 TEST_F(AuxGTest, produceRandomIntegers) {
 	Aux::Random::setSeed(1, false);
-#if (LOG_LEVEL == LOG_LEVEL_TRACE)
+#ifndef NETWORKIT_RELEASE_LOGGING
```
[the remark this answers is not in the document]
avdgrinten: AFAICS the main use of the `#ifdef` is to prevent unused variable warnings when `TRACE` or `DEBUG` is compiled out. A if with constant argument will not help in this case as the dead code elimination will only be done in the compiler backend, after the unused variable warning has already been issued.

That being said, a lot of the code using the `#if LOGLEVEL == [...]` idiom is quite questionable. I decided to only do syntactical changes to `LOG_LEVEL` in this PR but we can definitely take another look at those cases after the PR is merged.

CMakeLists.txt
avdgrinten: I am not sure if it's worthwhile to do this. The option is not used in any standard build (e.g., `setup.py`) and it is quite easy to adapt to the new option. If you think that it is worthwhile, I suggest to fail the build if the option is specified so that users are forced to adapt their builds.

networkit/cpp/auxiliary/Log.cpp
avdgrinten: Makes sense.

avdgrinten:
I rebased the patch series onto the current `Dev` branch. Comments have been addressed in the last 3 commits.

avdgrinten:
The `push` failure has nothing to do with this PR. The `pr` build succeeded.

avdgrinten:
Thank you for the review.

PR #365, opened 2019-07-01
Add InGroup annotations to generators

manpen:
The `@ingroup` attribute was missing for a number of generators. Hence they were not listed in the documentation in the Generator module.

I also added two ToDo requesting code clean up out-of-scope for this PR.


avdgrinten:
This is an uncontroversial change, I am merging it.

Issue #213, opened 2018-08-09
Google Benchmark


manpen:
I'm currently in the process of trying to optimise the GraphBuilder and hence are going through some benchmarking code (see PR coming soon I hope). The situation here is, that we have several helper functions such as `measure`, `measureOnce` etc. that are spread around the code base and basically invent the wheel all over again each time.

What's even worse is that each time, the developer has to consider how to properly setup its data structures, average over runtimes, find the right number of iterations for statistically meaningful results and so on. While this is not rocket science, it leads to bugs, inconsistencies and bike-shad discussions.

https://github.com/google/benchmark/ is a library with very similar concepts and syntax to google-test that deals with all these issues. Would you accept a PR to add this as a git submodule in the CMake branch?

avdgrinten:
Would the plan be to use this in future code or to rewrite existing code using Google Benchmark? I would need to take a closer look at it but in principle I am open to accept such PRs.

avdgrinten:
As a side node: A better benchmarking suite for NetworKit which can be run automatically has been on my mind for quite some time. This would require scripts to generate or download large graphs and some scripts to print the benchmarking results. If we had such a suite, it would become much easier to detect performance regressions and/or to optimize the core code (e.g. the graph class).

manpen:
I guess the ultimate goal would be to check (important) and migrate (if sensible) all benchmarks and if possible have some scaling experiments in them; e.g. I was surprised to find that the GraphBuilder's benchmark did not yield any meaningful results (PR follows). However, this is more than a small job. 

What I can do is add some infrastructure similarly to `networkit_add_benchmark` in CMake and to migrate benchmarks which I touch anyway. The remainder should then be investigated separately.

PR #366, opened 2019-07-01
Fix a bunch of "unused" warnings

manpen:
Fix remaining "unused" warnings.


angriman:
Thank you for fixing this.

I see that in `AuxRandomBenchmark.cpp` there are several comments like `// TODO: replace with google benchmark infrastructure`. Do we actually want to use the [google/benchmark](https://github.com/google/benchmark) infrastructure in the future?

avdgrinten:
I am not convinced about the use of `volatile` to prevent optimization. In GCC and Clang, one can use inline assembly constraints to force the compiler to keep a variable around and to commit changes to memory:

```c
asm volatile ("" : : "p"(&x) : "memory");
```
(Demo: https://godbolt.org/z/blMf01)

This should be more robust than a volatile write. However, I do not know if MSVC has comparable syntax. Maybe it's worthwhile to check what other projects (Google Benchmark?) do.

In any case, I think that this `doNotOptimize()` function should not be specific to this test. Instead, it should be moved to a global header (with the comment that it should only be used for benchmarking).

manpen:
@avdgrinten This is exactly the implementation of Google Benchmark. In issue #213, I said that I would add Google Benchmark and CMake infrastructure, but did not come around to do so. My suggestion would be to leave the PR as is (it's the same implementation we had before, only without warnings) and, to add Google Benchmark soon.

avdgrinten:
I think the question whether Google Benchmark is the correct solution for NetworKit is not clearly answered yet. Google Benchmark seems to be aimed at microbenchmarks (100s to millions of iterations / second), whereas NetworKit's algorithms often require multiple seconds to run (on interesting instances).

PR #367, opened 2019-07-01
Modernize Globals.hpp

manpen:
This PR does the following:
- Move `PI` from global namespace into `NetworKit`
- Remove case distinction for the definition of `PI` and always use constant
- Use "using" rather than "typedef"
- Define constexpr debug to be used rather than `NDEBUG`. It also has the advantage that we can shadowing to selectively put module into debug mode.
- Remove now unused `#include <cmath>` from Globals and add it to headers where it's actually used.

### DIFF
--- a/include/networkit/Globals.hpp
+++ b/include/networkit/Globals.hpp
@@ -9,7 +9,6 @@
 #define GLOBALS_H_
 
 #include <cstdint>
-#include <cmath>
 #include <limits>
 
 #include <networkit/ext/ttmath/ttmath.hpp>
--- a/include/networkit/Globals.hpp
+++ b/include/networkit/Globals.hpp
@@ -14,34 +14,29 @@
 
 #include <networkit/ext/ttmath/ttmath.hpp>
 
-
 namespace NetworKit {
-	/** Typedefs **/
-	typedef uint64_t index; // more expressive name for an index into an array
-#ifdef _MSC_VER	
-	typedef int64_t omp_index;
-#else
-	typedef index omp_index;
-#endif
-	typedef uint64_t count; // more expressive name for an integer quantity
-	typedef ttmath::Big<TTMATH_BITS(64),TTMATH_BITS(64)> bigfloat;	// big floating point number
-	typedef index node; // node indices are 0-based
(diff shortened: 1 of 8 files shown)


avdgrinten:
I don't quite understand what die use case for `debug` is. Can you elaborate on that?

manpen:
There are a lot of blocks as following:
```cpp
#ifndef NDEBUG
// some expensive check
#endif
```

In many cases we can directly write the `#ifndef` as a CPP condition allowing for easier code.

Another advantage is that we can selectively enable debugging: simply add a `static constexpr debug = true;` to the class / method / block and do not pay for all other debug constructs. TLX's logger actually has quite nice semantics, where we can write ``LOG << "Hello World"`` which is only included if `debug` is true in that scope.

avdgrinten:
If we want to replace users if `#ifdef NDEBUG` by some non-preprocessor solution, we should do that in the same PR. Otherwise, the new definition will just linger in the file but never be used. That being said, I'm not sure whether we should tie sanity checks to `NDEBUG`. If we're talking about assertion-style checks, why not just use `assert()`?

michitux:
I think tying assertion-style sanity checks to `NDEBUG` makes perfectly sense as `assert` is disabled by the same mechanism. However, selectively setting `debug` to true locally breaks this as then the code around the `assert` is run but not the `assert` itself. Putting the expensive check in a lambda does not always work as in some cases you need to collect information for the `assert` in multiple places at different times in the execution. Unless we replace `assert`, I think we should stick to the good old `#ifndef NDEBUG`. I am also not sure we should mix logging controlled by `debug` as used in TLX with asserts. While you might want to have enabled asserts in all places, you might not want to have debug logging enabled in the same places. 

To summarize my opinion: :+1: for globally defining and setting `debug` to some value for the TLX logging mechanism if we want to use it in NetworKit (which is a separate discussion, so this should be moved to a later PR). :-1: for using `debug` to replace `#ifndef NDEBUG`.

manpen:
Thank you for the feedback. Since `debug` seems to be quite controversial, let's follow @michitux suggestion and approach this in a separate PR -- I removed the commit in question from this PR.

PR #370, opened 2019-07-04
Fix removeNode

angriman:
This fixes a bug in the `Graph::removeNode` method, which sometimes fails to remove all incident edges to the node to be removed. The new tests included in this PR failed with the old `removeNode` method.


manpen:
Thanks for fixing this! I just stumbled across the same error and it saved me some nasty debugging :)
To me the underlying problem seems to be that we do not have well-defined rules how to interact with `for*` functions -- what changes to graph are allowed in the callback and what not. It might be worth to spend some time coming up with clear rules, esp. if we want to change the `Graph` storage in the future.

PR #371, opened 2019-07-04
Refactor betweenness

angriman:
As already done with `Closeness` and `APSP`, this improves the memory efficiency of `Betweenness` by allocating the necessary memory once per each thread.

### DIFF
--- a/networkit/cpp/centrality/Betweenness.cpp
+++ b/networkit/cpp/centrality/Betweenness.cpp
@@ -41,17 +41,26 @@ void Betweenness::run() {
 	}
 	DEBUG("edge score per thread: ", edgeScorePerThread.size());
 
+	std::vector<std::vector<double>> dependencies(maxThreads, std::vector<double>(z));
+	std::vector<std::unique_ptr<SSSP>> sssps;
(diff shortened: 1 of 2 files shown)


networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -41,17 +41,26 @@ void Betweenness::run() {
...
+		std::fill(dependency.begin(), dependency.end(), 0);
 
 		// run SSSP algorithm and keep track of everything
-		std::unique_ptr<SSSP> sssp;
-		if (G.isWeighted()) {
-			sssp.reset(new Dijkstra(G, s, true, true));
-		} else {
-			sssp.reset(new BFS(G, s, true, true));
-		}
+		std::unique_ptr<SSSP> &sssp = sssps[omp_get_thread_num()];
```
avdgrinten: Its more idiomatic to do `auto sssp = ssps[omp_get_thread_num()].get();`. This will result in a plain `SSSP *` instead of a reference to a `std::unique_ptr` (which is an unnecessary double indirection).

PR #373, opened 2019-07-06
`Graph::subgraphFromNodes` with neighbors

manpen:
This PR adds a new feature to `Graph::subgraphFromNodes` which originally created an induced subgraph on the set `nodes` of nodes provided. In some applications it however also makes sense to include their neighbours -- hence the new feature which can be enabled by setting the parameter `includeOutNeighbors` or `includeInNeighbors`. They default to false, so the standard behaviour does not change.

All edges in added must have at least one endpoint in `nodes`, the other one is either in `nodes` or in the set of neighbors included.

This PR also adds:
 - CPP Unit tests from `subgraphFromNodes`
 - Python Unit tests for `subgraphFromNodes` (with an invitation to add more Python test for Graph)
 - Shebang line in all Python tests and exectuable flags, so we can easily execute those tests from the command line (if you do not like it in this PR, I can drop the commit and open another one).


networkit/cpp/graph/Graph.cpp
```diff
@@ -1113,19 +1113,44 @@ void Graph::merge(const Graph &G) {
...
+		}
+
+		return neighbors;
+	}();
+
+	auto isRelevantNode = [&] (const node u) {
+		if (nodes.find(u) != nodes.end()) return 2;
+		if (!neighbors.empty() && neighbors.find(u) != neighbors.end()) return 1;
+		return 0;
+	};
```
angriman: It might be worth to add a short comment explaining how the "relevance" of a node works.

angriman:
I just have a minor comment, apart from that the PR looks good to me.

manpen:
Thanks for the review. I added the comments (only change in the force-push). 
@avdgrinten: I completely agree, and TBH, I first looked into GraphTools when I was searching this feature. There are additional methods such as `merge` (?), `to(Unw|W)eighted`, `transpose`, BFS that I do not really see as part of `Graph`. However, at the same time, these are typically methods used by user code and I do not really think it's worth breaking them.

PR #375, opened 2019-07-09
Refactor/graph writers

angriman:
All classes that write graphs inherit from `GraphWriter`. Exposed `NetworkitBinaryReader/Writer` to Python, together with the `GraphWriter` hierarchy.

### DIFF
--- a/include/networkit/io/DotGraphWriter.hpp
+++ b/include/networkit/io/DotGraphWriter.hpp
@@ -30,7 +30,7 @@ public:
 	 * @param[in]	graph	The graph object
 	 * @param[in]	path	The file path to be written to
 	 */
-	void write(const Graph &G, const std::string &path) const override;
+	void write(const Graph &G, const std::string &path) override;
 
 };
 
--- a/include/networkit/io/DotGraphWriter.hpp
+++ b/include/networkit/io/DotGraphWriter.hpp
@@ -1,12 +1,15 @@
 /*
- * DotGraphWriter.h
+ * DotGraphWriter.hpp
+ *
+ *  Created on: Jun 5, 2013
+ *      Author: forigem
  */
 
 #ifndef DOTGRAPHWRITER_H
 #define DOTGRAPHWRITER_H
 
-#include <fstream>
 #include <networkit/graph/Graph.hpp>
+#include <networkit/io/GraphWriter.hpp>
 
 namespace NetworKit {
 
@@ -15,19 +18,19 @@ namespace NetworKit {
  * @ingroup io
  *
  * This class turns a graph into a very basic GraphViz file as documented in the official manual [1].
- * If a more thorough support is desired, please contact the developers over networkit@ira.uni-karlsruhe.de.
+ * If a more thorough support is desired, please contact the developers: https://github.com/networkit/networkit
  *
- * [1] http://www.graphviz.org/Documentation/dotguide.pdf
- */ 
-class DotGraphWriter {
+ * [1] https://graphviz.gitlab.io/_pages/pdf/dotguide.pdf
+ */
+class DotGraphWriter final : public GraphWriter {
 public:
 	/**
 	 * Write a graph as a GraphViz/file.
-	 * 
+	 *
 	 * @param[in]	graph	The graph object
 	 * @param[in]	path	The file path to be written to
 	 */
-	virtual void write(Graph& graph, std::string path) const;
+	void write(const Graph &G, const std::string &path) const override;
 
 };
 
--- a/include/networkit/io/EdgeListWriter.hpp
+++ b/include/networkit/io/EdgeListWriter.hpp
@@ -37,7 +37,7 @@ public:
 	 * @param[in]	G		the graph
 	 * @param[in]	path	the output file path
 	 */
-	void write(const Graph &G, const std::string &path) const override;
+	void write(const Graph &G, const std::string &path) override;
 
 protected:
 
--- a/include/networkit/io/EdgeListWriter.hpp
+++ b/include/networkit/io/EdgeListWriter.hpp
@@ -1,5 +1,5 @@
 /*
- * EdgeListWriter.h
+ * EdgeListWriter.hpp
  *
  *  Created on: 18.06.2013
  *      Author: cls
@@ -8,11 +8,7 @@
 #ifndef EDGELISTWRITER_H_
 #define EDGELISTWRITER_H_
 
-#include <fstream>
-#include <iostream>
-#include <string>
-
-#include <networkit/io/GraphReader.hpp>
+#include <networkit/io/GraphWriter.hpp>
 
 namespace NetworKit {
 
@@ -23,7 +19,7 @@ namespace NetworKit {
  * the user.
  *
  */
-class EdgeListWriter {
+class EdgeListWriter final : public GraphWriter {
 
 public:
 
@@ -41,7 +37,7 @@ public:
 	 * @param[in]	G		the graph
 	 * @param[in]	path	the output file path
 	 */
-	void write(const Graph& G, std::string path);
+	void write(const Graph &G, const std::string &path) const override;
 
 protected:
 
--- a/include/networkit/io/METISGraphWriter.hpp
+++ b/include/networkit/io/METISGraphWriter.hpp
@@ -1,5 +1,5 @@
 /*
- * METISGraphWriter.h
+ * METISGraphWriter.hpp
  *
  *  Created on: 30.01.2013
  *      Author: Christian Staudt (christian.staudt@kit.edu)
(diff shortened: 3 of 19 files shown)


include/networkit/io/DotGraphWriter.hpp
```diff
@@ -1,12 +1,17 @@
...
+ * DotGraphWriter.hpp
+ *
+ *  Created on: Jun 5, 2013
+ *      Author: forigem
  */
 
 #ifndef DOTGRAPHWRITER_H
 #define DOTGRAPHWRITER_H
 
 #include <fstream>
```
manpen: Include not required in header.

include/networkit/io/SNAPGraphWriter.hpp
```diff
@@ -44,13 +44,10 @@ namespace NetworKit {
  *
  * Good job
```
manpen: I agree, it's a good job. but why put it in the comment?

include/networkit/io/EdgeListWriter.hpp
```diff
@@ -13,6 +13,7 @@
 #include <string>
```
manpen: Includes of fstream and iostream ont required

networkit/_NetworKit.pyx
```diff
@@ -3134,6 +3134,47 @@ cdef class GraphReader:
...
+		Write the graph to a file.
+
+		Parameters
+		----------
+		G     : networkit.Graph
+			The graph to write
+		paths : str
+			The output path
+		"""
+		cdef string c_path = stdstring(path)
```
manpen: Optional: assert if path is empty? I mean we're check that Graph's not None ...

include/networkit/io/DotGraphWriter.hpp
```diff
@@ -18,16 +23,16 @@ namespace NetworKit {
  * If a more thorough support is desired, please contact the developers over networkit@ira.uni-karlsruhe.de.
```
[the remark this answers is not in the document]
angriman: Good idea, that mailing is outdated.

include/networkit/io/DotGraphWriter.hpp
```diff
@@ -18,16 +23,16 @@ namespace NetworKit {
...
+class DotGraphWriter final : public GraphWriter {
 public:
 	/**
 	 * Write a graph as a GraphViz/file.
-	 * 
+	 *
 	 * @param[in]	graph	The graph object
 	 * @param[in]	path	The file path to be written to
 	 */
-	virtual void write(Graph& graph, std::string path) const;
```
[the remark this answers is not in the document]
angriman: Good point, as you wrote those classes could easily be adapted to implement a const `write` method.

networkit/_NetworKit.pyx
angriman: I agree, we should also check for that.

include/networkit/io/DotGraphWriter.hpp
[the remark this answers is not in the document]
avdgrinten: I disagree here. Why `write()` is essentially the `GraphWriter`s `run` method. If you want to write in parallel, construct two writers. Making the virtual method `const` heavily constraints what is efficiently possible in a `GraphWriter`.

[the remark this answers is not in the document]
avdgrinten: To elaborate: The base class for `GraphWriter`s should absolutely remain non-`const`. I don't see what advantage `const` offers for writers.

More broadly, I do not think that we should support parallel usage of *any* algorithm (or I/O interface) in NetworKit *unless* there is a good reason to do so. `const` is an unnecessary pessimization here: it forces implementers to maintain local copies of algorithm state instead of storing the state in the algorithm's object. Often, this is much less efficient, especially if an algorithm's object is reused. Reuse of algorithm objects is arguably much more common than parallel use of the same algorithm's object.

Also note that `const` does affect the API: turning a `const` method to non-`const` is a breaking change. We should only apply `const` to functions that will not write to the algorithm's state, forever, regardless of implementation strategy. I am not sure if *any* of the `GraphWriter`s should have `const write()` methods.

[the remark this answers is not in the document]
manpen: While I see your point and are kind of convinced that the base class should remain non-const, I still have two points:
- Currently there ARE writers that are const and the breaking of API you mentioned would occur in this PR
- AFAICS we currently have no examples where the state of the write should change (I do not think that `nodes` should be a member of `NetworKitBinaryWriter` but rather a local variable; it's only used in `write`. In `GraphToolBinaryWriter` no state is changed)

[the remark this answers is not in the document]
avdgrinten: I agree that those `count` variables should be local. However, there might be writers that have a legitimate reason to construct large data structures (e.g., `std::vector`s or `std::unordered_map`s to remap node IDs?) in memory. Those structures should probably be stored in the object to avoid reallocating them whenever `write()` is called.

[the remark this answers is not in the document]
manpen: Yes, I agree that for future writers it might be an issue, and -- given that -- I agree that the base class should remain non-const. I'm just saying that if we do this, we should make a conscious decision to break the interface (also as you already pointed out with probably no serious implications).

include/networkit/io/METISGraphWriter.hpp
```diff
@@ -8,24 +8,22 @@
...
 
 public:
 
 	METISGraphWriter() = default;
 
-	virtual void write(const Graph& G, const std::string& path) override;
+	virtual void write(const Graph &G, const std::string &path) override;
 
-	virtual void write(const Graph& G, bool weighted, std::string path);
+	virtual void write(const Graph &G, bool weighted, const std::string &path);
```
manpen: This does not have to be `virtual` -- the writer is declared `final`

include/networkit/io/EdgeListWriter.hpp
```diff
@@ -8,11 +8,8 @@
 #ifndef EDGELISTWRITER_H_
 #define EDGELISTWRITER_H_
 
-#include <fstream>
-#include <iostream>
-#include <string>
-
 #include <networkit/io/GraphReader.hpp>
```
manpen: Sorry for bringing this up again: I believe `#include <networkit/io/GraphReader.hpp>` is not required.
Also while `fstream` was not required in the header, the implemenation needs it. So please include it in the .cpp file.

angriman: I removed `fstream` because it is already included in `GraphWriter.hpp`. However, `GraphWriter.hpp` does not need `fstream`, only its derived classes use it. Should I move `fstream` to the derived classes and remove it from `GraphWriter.hpp`?

manpen: Ah, I missed that. Sorry. Never mind then. While I try to not rely on indirect includes, it's quite common in networkit, so I guess it's fine.

angriman: I think that there is no valid reason for including `fstream` in `GraphWriter.hpp`, and I agree that we should not rely on indirect includes. I'll move `fstream` to the derived classes.

include/networkit/io/DotGraphWriter.hpp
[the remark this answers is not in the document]
angriman: Now the base class `write` is non-const. Are there any other issues or can we merge this?

manpen:
Thank you :)

manpen:
I just wanted to leave a big thank you to @cndolo and @angriman. I used this PR (based on #325) to cache my rather large ASCII-based inputs with NKB. In one case, reading the file dropped from ~5min to less than 10s; that's really cool!

avdgrinten:
That nice to hear @manpen! Indeed, fast caching is one of the main motivations of the NKBG format.

Unfortunately, #380 bumps the version number of the format once more, without adding compatibility fixes for the current version (as I didn't expect that there were any users already). Is that a problem for your use case? If yes, we should add those compatibility fixes (those fixes should only consist of a few lines anyway - basically we would need to add code to ignore the last two fields of the file header).

My general plan is provide backwards-compatibility once the format appears in a released version of NetworKit.

manpen:
For me personally that's not an issue and I would even vote to not include backward compatibility just yet in favour of an easier code -- after all, it the `Dev` branch and thinks are bound to break ;)

avdgrinten:
Great, then our compatibility guarantee for the NKBG format is: once it appears in a release, we provide backwards compatibility indefinitely.

avdgrinten:
@angriman This PR has conflicts against `Dev`.

angriman:
@avdgrinten conflicts have been fixed.

PR #376, opened 2019-07-10
Revoke execute permissions

angriman:
`Licence.txt` and `README.md` files should not be executable.


angriman:
> Please also include `.gitignore`.

Good point.

manpen:
LGTM. Seems to be neither controversial nor urgent.

PR #377, opened 2019-07-10
Better documentation for Closeness

angriman:
`Closeness` takes a third parameter which can be either `bool` or `ClosenessVariant`. Some users did not find the documentation clear enough, this explains more clearly how this class works.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -7095,18 +7095,31 @@ cdef extern from "<networkit/centrality/Closeness.hpp>":
 
 cdef class Closeness(Centrality):
 	"""
-		Closeness(G, normalized=True, variant=ClosenessVariant::standard/checkConnectdedness=True)
+		Closeness(G, normalized, bool checkConnectdedness)
+		Closeness(G, normalized, networkit.centrality.ClosenessVariant variant)
 
 		Constructs the Closeness class for the given Graph `G`. If the Closeness scores should not be normalized,
-  		set `normalized` to False. The run() method takes O(nm) time, where n is the number
-	 	 of nodes and m is the number of edges of the graph. NOTICE: the graph has to be connected.
+		set `normalized` to False. The run() method takes O(nm) time, where n is the number
+		of nodes and m is the number of edges of the graph.
 
-	 	Parameters
-	 	----------
-	 	G : networkit.Graph
-	 		The graph.
-	 	normalized : bool, optional
-	 		Set this parameter to False if scores should not be normalized into an interval of [0,1]. Normalization only for unweighted graphs.
+		Parameters
+		----------
+		G : networkit.Graph
+			The graph.
+		normalized : bool
+			Set this parameter to False if scores should not be normalized into an interval of [0,1].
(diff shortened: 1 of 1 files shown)


networkit/_NetworKit.pyx
```diff
@@ -7095,18 +7095,29 @@ cdef extern from "<networkit/centrality/Closeness.hpp>":
...
+			The graph.
+		normalized : bool
+			Set this parameter to False if scores should not be normalized into an interval of [0,1].
+			Normalization only works for unweighted graphs.
+		checkConnectdedness : bool
+			Set this parameter to True to also check if the graph is connected before computing closeness.
+			Set this parameter to False to not check if the graph is connected (note: the standard definition
+			of closeness works for connected graphs, choose this if the input graph is known to be connected).
+		ClosenessVariant : networkit.centrality.ClosenessVariant
+		Set this parameter to networkit.centrality.ClosenessVariant.Standard to use the standard
```
manpen: Indent is off which makes it harder to read.

networkit/_NetworKit.pyx
```diff
@@ -7095,18 +7095,29 @@ cdef extern from "<networkit/centrality/Closeness.hpp>":
...
+		definition of closeness, that is defined for connected graphs only.
+		Set this parameter to networkit.centrality.ClosenessVariant.Generalized to use the generalized
+		definition of closeness, that is defined for also non-connected graphs.
```
manpen: How do `networkit.centrality.ClosenessVariant.Generalized` and `checkConnectedness` interact? I'd guess the latter is ignored, but maybe add that?

networkit/_NetworKit.pyx
```diff
@@ -7095,18 +7095,29 @@ cdef extern from "<networkit/centrality/Closeness.hpp>":
 
 cdef class Closeness(Centrality):
 	"""
-		Closeness(G, normalized=True, variant=ClosenessVariant::standard/checkConnectdedness=True)
+		Closeness(G, normalized=True, checkConnectdedness=True)
```
manpen: This notation is highly misleading, as it suggests that the third parameter is a keyword parameter, while it is in fact positional. Invoking `algo = nk.centrality.Closeness(G, normalized=True, checkConnectdedness=True)` is a bug.
Also `normalized` has to be passed, while documentation indicates that it has a default value.

manpen: Maybe something like?
```
Closeness(G, normalized, bool checkConnectdedness)
Closeness(G, normalized, networkit.centrality.ClosenessVariant variant)
```

networkit/_NetworKit.pyx
angriman: Exactly, choosing `standard` makes Closeness also compute the connected components, with `generalized` instead they are not computed.

networkit/_NetworKit.pyx
angriman: That's true, thanks for the suggestion.

angriman:
The comments have been addressed, shall we merge this?

manpen:
Sorry it took that long ;)

PR #379, opened 2019-07-14
Refactor: top-closeness

angriman:
This brings some improvements to `TopCloseness`:

- Instead of `Aux::PrioQueue`s (i.e., rb-trees), now it uses heaps.
- Update tlx to use `build_heap`: this allows to build `Q` in `O(n)` instead of `O(n*logn)`
- When `sec_heu` is set to `false`, the algorithm does not allocate `O(n)` memory at every iteration, but pre-allocates it beforehand.
- Uses a `unique_ptr<StronglyConnectedComponents>`  to avoid copying the `components` vector. This saves `O(n)` memory.
- Tracking of "trailing" nodes is simpler and more efficient.
- Simpler GTests.

### DIFF
--- a/networkit/cpp/centrality/TopCloseness.cpp
+++ b/networkit/cpp/centrality/TopCloseness.cpp
@@ -11,7 +11,6 @@
 
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/components/ConnectedComponents.hpp>
-#include <networkit/components/StronglyConnectedComponents.hpp>
 #include <networkit/centrality/TopCloseness.hpp>
 #include <networkit/graph/BFS.hpp>
 
@@ -49,6 +48,7 @@ void TopCloseness::init() {
 
 void TopCloseness::computeReachable() {
   if (G.isDirected()) {
+    sccsPtr = std::unique_ptr<StronglyConnectedComponents>(new StronglyConnectedComponents(G));
     computeReachableNodesDir();
   } else {
     computeReachableNodesUndir();
@@ -58,9 +58,8 @@ void TopCloseness::computeReachable() {
 void TopCloseness::computeReachableNodesDir() {
   reachL = std::vector<count>(n);
   reachU = std::vector<count>(n);
-  component = std::vector<count>(n);
   DEBUG("Before running SCCs");
-  StronglyConnectedComponents sccs(G);
+  auto &sccs = *(sccsPtr.get());
   sccs.run();
 
   count N = sccs.numberOfComponents();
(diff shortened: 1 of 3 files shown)


hmeyerhenke:
General comment: This was on the "suboptimal" list for a long time, thanks for taking care of it. (This is not a review, of course.)

networkit/cpp/centrality/TopCloseness.cpp
```diff
@@ -1,649 +1,626 @@
...
+    reachL = std::vector<count>(n);
+
+    ConnectedComponents comps(G);
+    comps.run();
+    std::map<index, count> sizes = comps.getComponentSizes();
+    G.forNodes([&](node v) {
+        index cv = comps.componentOfNode(v);
+        reachL[v] = sizes[cv];
+    });
+    reachU = reachL;
```
fabratu, line 171: Don't know how much could be saved here in terms of memory, but since `reachU` and `reachL` are the same for undirected graphs, maybe it would be a good idea to make these vectors references/pointers?

angriman:
@fabratu `unique_ptr` owns the object and ensures that memory is freed once it goes out of scope, it's considered safer to use instead of C pointers (and also a good practice).

About the "trailing nodes", the changes have been applied starting from here https://github.com/networkit/networkit/pull/379/files#diff-e0d92f18e66cf4b8b711d189a85ed957R540 (TopCloseness.cpp, line 540)

networkit/cpp/centrality/TopCloseness.cpp
angriman: That's a good point, I'll address that.

avdgrinten:
I don't think `shared_ptr` is a good solution here. There is no shared ownership involves. Why don't you just keep the reference logic but only operate on one vector in the undirected case?

angriman:
Maybe I got it wrong, but I wanted to set the shared ownership here: https://github.com/angriman/networkit/blob/refactor/centrality/top-closeness/networkit/cpp/centrality/TopCloseness.cpp#L165
I wanted to use `shared_ptr` to avoid changing those parts of the code that use both the upper and lower bounds even if the graph is undirected.

fabratu:
For the undirected case, I also think there is active sharing (as mentioned set in line 165) after calling`init()` for an TopCloseness-Object. Empty vectors would also be way to safe memory - but since `unique_ptr` is used in the same class, using `shared_ptr` shows resemblance in style.

Apart from that - I could not spot any problems in the new logic overall.

Issue #364, opened 2019-07-01
`Graph::addPartialOutEdge` ignores index.


manpen:
The `index` parameter in `Graph::addPartialOutEdge` is not used. This seems to me like a bug (since the other two related functions use it).

```cpp
void Graph::addPartialOutEdge(Unsafe, node u, node v, uint64_t index, edgeweight ew) {
	assert(u < z);
	assert(exists[u]);
	assert(v < z);
	assert(exists[v]);

	outEdges[u].push_back(v);

	if (weighted) {
		outEdgeWeights[u].push_back(ew);
	}
}
```

Is this intended behavior? 
On a related note: the parameters are not mentioned in the Doxygen documentation.

manpen:
Thanks!

PR #380, opened 2019-07-15
Add support for weights in the NetworkitBinaryGraph format. 

cndolo:
This PR adds support for weights in the NetworkitBinaryGraph format. 

Unsigned, signed, float and double weights are supported, and can be passed to the writer as a constructor parameter. If none is passed, the writer tries to detect the type of the weights.

The `addPartial*Edge` functions of the graph were slightly modified too i.e., the order of the function parameters `uint64_t index = 0, edgeweight ew = defaultEdgeWeight` was exchanged. This was necessary because the NetworkitBinaryGraph format does not support indexes, and therefore, the default `index` value should be used.

### DIFF
--- a/include/networkit/io/NetworkitBinaryGraph.hpp
+++ b/include/networkit/io/NetworkitBinaryGraph.hpp
@@ -16,11 +16,11 @@ struct Header {
 };
 
 enum WEIGHT_FORMAT {
-	none = 0,
-	unsignedFormat = 1,
-	signedFormat = 2,
-	doubleFormat = 3,
-	floatFormat = 4
+	NONE = 0,
+	VARINT = 1,
+	SIGNED_VARINT = 2,
+	DOUBLE = 3,
+	FLOAT = 4
 };
 
 static constexpr uint8_t DELETED_BIT = 0x1; // bit 0
--- a/include/networkit/io/NetworkitBinaryGraph.hpp
+++ b/include/networkit/io/NetworkitBinaryGraph.hpp
@@ -1,5 +1,5 @@
-#ifndef NETWORKIT_BINARY_GRAPH_
-#define NETWORKIT_BINARY_GRAPH_
+#ifndef NETWORKIT_BINARY_GRAPH_HPP
+#define NETWORKIT_BINARY_GRAPH_HPP
 
 namespace nkbg {
 struct Header {
@@ -11,19 +11,21 @@ struct Header {
 	uint64_t offsetBaseData;
 	uint64_t offsetAdjLists;
 	uint64_t offsetAdjTranspose;
-	uint64_t offsetWeights;
+	uint64_t offsetWeightLists;
+	uint64_t offsetWeightTranspose;
 };
 
 enum WEIGHT_FORMAT {
-	VARINT = 0,
-	SIGNED_VARINT = 1,
-	DOUBLE = 2,
-	FLOAT = 3
+	none = 0,
+	unsignedFormat = 1,
+	signedFormat = 2,
+	doubleFormat = 3,
+	floatFormat = 4
 };
 
 static constexpr uint8_t DELETED_BIT = 0x1; // bit 0
 static constexpr uint64_t DIR_MASK = 0x1; // bit 0
-static constexpr uint64_t WGHT_MASK = 0x6; //bit 1-2
+static constexpr uint64_t WGHT_MASK = 0xE; //bit 1-3
 static constexpr uint64_t WGHT_SHIFT = 0x1;
 
 }
--- a/include/networkit/io/NetworkitBinaryWriter.hpp
+++ b/include/networkit/io/NetworkitBinaryWriter.hpp
@@ -9,7 +9,6 @@
 
 #include <networkit/graph/Graph.hpp>
 #include <networkit/io/GraphWriter.hpp>
-#include <networkit/io/NetworkitBinaryGraph.hpp>
 
 namespace NetworKit {
 
@@ -30,7 +29,6 @@ enum class NetworkitBinaryWeights {
 class NetworkitBinaryWriter : public GraphWriter {
 
 public:
-
 	NetworkitBinaryWriter(uint64_t chunks = 32, NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
 
 	void write(const Graph& G, const std::string& path) override;
--- a/include/networkit/io/NetworkitBinaryWriter.hpp
+++ b/include/networkit/io/NetworkitBinaryWriter.hpp
@@ -9,6 +9,7 @@
 
 #include <networkit/graph/Graph.hpp>
 #include <networkit/io/GraphWriter.hpp>
+#include <networkit/io/NetworkitBinaryGraph.hpp>
 
 namespace NetworKit {
 
@@ -17,11 +18,20 @@ namespace NetworKit {
  *
  * Writes a graph in the custom Networkit format documented in cpp/io/NetworkitGraph.md
  */
+enum class NetworkitBinaryWeights {
+	none,
+	unsignedFormat,
+	signedFormat,
+	doubleFormat,
+	floatFormat,
+	autoDetect
+};
 
 class NetworkitBinaryWriter : public GraphWriter {
 
 public:
-	NetworkitBinaryWriter(uint64_t chunks = 32);
+
+	NetworkitBinaryWriter(uint64_t chunks = 32, NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
 
 	void write(const Graph& G, const std::string& path) override;
 
@@ -32,6 +42,7 @@ private:
 
 	count nodes;
 	count chunks;
+	NetworkitBinaryWeights weightsType;
 };
 } /* namespace */
 #endif
--- a/networkit/cpp/io/NetworkitBinaryReader.cpp
+++ b/networkit/cpp/io/NetworkitBinaryReader.cpp
@@ -79,7 +79,7 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
 	DEBUG("# nodes here = ", nodes);
 	chunks = header.chunks;
 	DEBUG("# chunks here = ", chunks);
-	if(weightFormat == nkbg::WEIGHT_FORMAT::none) {
+	if(weightFormat == nkbg::WEIGHT_FORMAT::NONE) {
 		weighted = false;
 	} else {
 		weighted = true;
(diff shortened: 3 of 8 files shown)


include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -11,19 +11,21 @@ struct Header {
...
 enum WEIGHT_FORMAT {
-	VARINT = 0,
-	SIGNED_VARINT = 1,
-	DOUBLE = 2,
-	FLOAT = 3
+	none = 0,
+	unsignedFormat = 1,
+	signedFormat = 2,
+	doubleFormat = 3,
+	floatFormat = 4
```
avdgrinten: For consistency with the remainder of the file, we should keep the old ALL_UPPER_CASE names.

manpen:
Just two comments:
- While I think changing the parameter order is very critical (due to implicit type conversion it will fail silently), I want to point out that the new order is also consistent with the `for*`-callbacks. So given that it's a quite young function it seems okay to change it.
- Does this PR address issue #364? It seems to me that this is not the case. If the issue is in fact not a bug, please comment on it and close it.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -9,6 +9,7 @@
 
 #include <networkit/graph/Graph.hpp>
 #include <networkit/io/GraphWriter.hpp>
+#include <networkit/io/NetworkitBinaryGraph.hpp>
```
avdgrinten: This include was moved here (from `NetworKitBinaryWriter.cpp`); however, as far as I can see, it is not used in this header.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -17,11 +18,20 @@ namespace NetworKit {
...
  * Writes a graph in the custom Networkit format documented in cpp/io/NetworkitGraph.md
  */
+enum class NetworkitBinaryWeights {
+	none,
+	unsignedFormat,
+	signedFormat,
+	doubleFormat,
+	floatFormat,
+	autoDetect
+};
```
avdgrinten: Here, the lowerCamelCase names make sense. :+1:

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -17,11 +18,20 @@ namespace NetworKit {
...
 
 class NetworkitBinaryWriter : public GraphWriter {
 
 public:
-	NetworkitBinaryWriter(uint64_t chunks = 32);
+
+	NetworkitBinaryWriter(uint64_t chunks = 32, NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
```
avdgrinten: Unnecessary blank line.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -32,6 +42,7 @@ class NetworkitBinaryWriter : public GraphWriter {
 
 	count nodes;
 	count chunks;
+	NetworkitBinaryWeights weightsType;
```
avdgrinten: This can be stored in a local variable instead of a member variable (see the discussion in https://github.com/networkit/networkit/pull/375#discussion_r302413286).

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -148,24 +157,81 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+				double weight;
 				off += NetworkitBinaryReader::decode(reinterpret_cast<const uint8_t*>(adjIt + off), add);
+				switch(weightFormat) {
+					case nkbg::WEIGHT_FORMAT::unsignedFormat:
+						{
+							uint64_t unsignedWeight;
+							wghtOff += NetworkitBinaryReader::decode(reinterpret_cast<const uint8_t*>(adjWghtIt + wghtOff), unsignedWeight);
+							weight = unsignedWeight;
+						}
+					break;
```
avdgrinten: The `break` is part of the `case`, indent it accordingly. Furthermore, the double indentation is unnecessary. Indent the `case` as follows:
```c
switch (baz) {
    case foo: {
        bar();
        break;
    }
    [...]
}
```

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -148,24 +157,81 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+					case nkbg::WEIGHT_FORMAT::doubleFormat:
+						{
+							memcpy(&weight, adjWghtIt + wghtOff, sizeof(double));
+							wghtOff += sizeof(double);
+						}
```
avdgrinten: Braces are only needed in `case`s that define variables.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -148,24 +157,81 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+					break;
+				}
 				if(!directed) {
-					G.addPartialEdge(unsafe, curr, add);
+					G.addPartialEdge(unsafe, curr, add, weight);
 				} else {
-					G.addPartialOutEdge(unsafe, curr, add);
-				}
-				if(curr == add) {
-					selfLoops++;
```
avdgrinten: Why was this `selfLoop` handling removed?

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -173,6 +281,7 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string& path) {
 		outfile.write(reinterpret_cast<char*>(tmp), nbrsSize);
 		G.forNeighborsOf(u,[&](node v){
 			uint64_t nodeSize;
+			uint64_t weightSize;
```
avdgrinten: Is this variable used?

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -210,6 +319,141 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string& path) {
...
+							outfile.write(reinterpret_cast<char*>(tmp), weightSize);
+						}
+					break;
+					case nkbg::WEIGHT_FORMAT::floatFormat:
+						{
+							float weight = w;
+							outfile.write(reinterpret_cast<char*>(&weight), sizeof(float));
+						}
+					break;
+				}
```
avdgrinten: This code is duplicated 4 times, right? Could we move it into a lambda function?

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -88,33 +140,79 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string& path) {
...
+						if(weightFormat == nkbg::WEIGHT_FORMAT::unsignedFormat) {
+							transpWeightSize += encode(w,tmp);
+						} else if(weightFormat == nkbg::WEIGHT_FORMAT::doubleFormat) {
+							transpWeightSize += sizeof(double);
+						} else if(weightFormat == nkbg::WEIGHT_FORMAT::signedFormat) {
+							uint64_t weight = encodeZigzag(w);
+							transpWeightSize += encode(weight,tmp);
+						} else if(weightFormat == nkbg::WEIGHT_FORMAT::floatFormat) {
+							transpWeightSize += sizeof(float);
+						}
```
avdgrinten: Could we move this code into a lambda function to remove code duplication?

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -866,4 +866,24 @@ TEST_F(IOGTest, testNetworkitBinaryWiki) {
...
+TEST_F(IOGTest, NetworkitBinaryAutoDetect) {
+
+	Graph G(10, true, false);
+	int64_t weight = -1;
+	for(count n = 0; n < G.numberOfNodes(); n++) {
+		if(n != G.numberOfNodes()-1)
+			G.addEdge(n, n+1, weight++);
+	}
+	NetworkitBinaryWriter writer(32, NetworkitBinaryWeights::autoDetect);
+	writer.write(G, "../output/binary_auto");
```
avdgrinten: The relative file paths in this test should be adjusted.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -866,4 +866,24 @@ TEST_F(IOGTest, testNetworkitBinaryWiki) {
...
 	});
 }
+TEST_F(IOGTest, NetworkitBinaryAutoDetect) {
+
+	Graph G(10, true, false);
+	int64_t weight = -1;
+	for(count n = 0; n < G.numberOfNodes(); n++) {
+		if(n != G.numberOfNodes()-1)
+			G.addEdge(n, n+1, weight++);
+	}
```
avdgrinten: This tests negative weights, correct? My suggestion is to rename the test to `testNetworkitBinarySignedWeights`. Maybe we should also add a test for floating point weights?

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -866,4 +866,24 @@ TEST_F(IOGTest, testNetworkitBinaryWiki) {
 		});
 	});
 }
+TEST_F(IOGTest, NetworkitBinaryAutoDetect) {
```
avdgrinten: Names of tests need to be prefixed with `test`.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -88,33 +168,85 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string& path) {
...
+					adjWeightSize += encode(weight,tmp);
+				}
+				break;
+			case nkbg::WEIGHT_FORMAT::FLOAT:
+				adjWeightSize += sizeof(float);
+				break;
+		}
+	};
+
+	auto computeTransWeightsOffsets = [&] (edgeweight w) {
```
avdgrinten: What about letting the lambda return the size of the encoded weight? The `+=` would then be moved out of the lambda and the lambda could be combined with `computeAdjWeightsOffsets()` such that only a single lambda for adjacency lists + transpose lists is necessary.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -88,33 +168,85 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string& path) {
...
+		switch(weightFormat) {
+			case nkbg::WEIGHT_FORMAT::VARINT:
+				adjWeightSize += encode(w,tmp);
+				break;
+			case nkbg::WEIGHT_FORMAT::DOUBLE:
+				adjWeightSize += sizeof(double);
+				break;
+			case nkbg::WEIGHT_FORMAT::SIGNED_VARINT:
+				{
+					uint64_t weight = encodeZigzag(w);
```
avdgrinten: Double indentation is not necessary here. Just put the `{` brace on the same line as the `case` and indent this case in the same way as the `case`s without braces.

PR #386, opened 2019-07-24
Faster Graph::transpose

manpen:
The current  `Graph.transpose` function uses a GraphBuilder to copy the input graph into a transposed instance. By doing so it loses edge indices.

The new variant use the new awesome `preallocateDirected` and `addPartial*Edge` interface to directly create the new graph -- in parallel, without random I/O, and preserving edge ids. In one instance with ~110k nodes and ~60M edges to runtime decreased from `356s` to just under `6s`.

The PR also removes the `Graph.cpp` -> `GraphBuilder.hpp` dependency and adds a unit test for the Python interface.

### DIFF
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -11,7 +11,6 @@
 #include <sstream>
 
 #include <networkit/graph/Graph.hpp>
-#include <networkit/graph/GraphBuilder.hpp>
 
 namespace NetworKit {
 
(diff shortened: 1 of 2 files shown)


hmeyerhenke:
Sounds great, Manuel!

networkit/test/test_graph.py
```diff
@@ -62,5 +63,46 @@ def testSubgraphFromNodes(self):
 		self.assertEqual(res.numberOfNodes(), 4)
 		self.assertEqual(res.numberOfEdges(), 4)
 
+	def test_graph_transpose(self):
```
angriman: For reproducibility, I think that it would better to fix a random seed for both the `random` module and NetworKit (with `networkit.setSeed`).

angriman:
Apart from a minor comment in the Python tests, LGTM.

networkit/cpp/graph/Graph.cpp
```diff
@@ -1041,21 +1040,43 @@ std::vector<node> Graph::neighbors(node u) const {
...
 
-	GraphBuilder gB(z, weighted, true);
+	Graph GTranspose(z, weighted, true);
 
-	this->forEdges([&](node u, node v) { gB.addHalfEdge(v, u, weight(u, v)); });
+	// prepare edge id storage if input has indexed edges
+	GTranspose.edgesIndexed = edgesIndexed;
+	GTranspose.omega = omega;
+	GTranspose.inEdgeIds.resize(z);
+	GTranspose.outEdgeIds.resize(z);
```
avdgrinten: Shouldn't the resize only happen in the case that edge IDs exist?

manpen: Yes, you're right!

networkit/test/test_graph.py
manpen: Yes.

manpen:
Thanks for your feedback again. I'm now back from vacation and implemented your comments.

angriman:
Thank you, there is a conflict the the `test_graph.py`, maybe this is because of the new `removeMultiEdges` test.

Issue #378, opened 2019-07-11
Allowed operations in graph callbacks


angriman:
... we do not have well-defined rules how to interact with `for*` functions -- what changes to graph are allowed in the callback and what not. It might be worth to spend some time coming up with clear rules, esp. if we want to change the `Graph` storage in the future.

_Originally posted by @manpen in https://github.com/networkit/networkit/pull/370#issuecomment-510158849_

Ro avoid to stumble on similar issues in the future, I agree that we should decide and document which kind of operations are safe to do in those kind of callbacks.

PR #391, opened 2019-07-29
New Graph method: remove multi-edges

angriman:
This adds a new method that removes all the multi-edges from a NetworKit graph.
The same code is also used to remove all the self-loops, as the previous implementation was using `removeNode` within a `forEdges` loop, which shall not be allowed according to #378.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1067,9 +1067,27 @@ class Graph final {
     void removeAllEdges();
 
     /**
+     * Removes edges adjacent to a node according to a specific criterion.
+     *
+     * @param u The node whose adjacent edges shall be removed.
+     * @param condition A function that takes a node as an input and returns a
+     * bool. If true the edge (u, v) is removed.
+     * @param edgesIn Whether in-going or out-going edges shall be removed.
+     * @return count The number of removed edges.
+     */
+    template <typename Condition>
+    count removeAdjacentEdges(node u, Condition condition, bool edgesIn = false);
+
+
+    /**
      * Removes all self-loops in the graph.
      */
     void removeSelfLoops();
+
+    /**
+     * Removes all multi-edges in the graph.
+     */
+    void removeMultiEdges();
 
     /**
      * Changes the edges {@a s1, @a t1} into {@a s1, @a t2} and the edge {@a
@@ -2094,6 +2112,36 @@ template <typename L> void Graph::DFSEdgesFrom(node r, L handle) const {
         });
     } while (!s.empty());
 }
+
+/* EDGE MODIFIERS */
+
+template <typename Condition>
+count Graph::removeAdjacentEdges(node u, Condition condition, bool edgesIn) {
+	count removedEdges = 0;
+	auto &edges_ = outEdges[u];
+	for (index vi = 0; vi < edges_.size();) {
+		if (condition(edges_[vi])) {
+			++removedEdges;
+			edges_[vi] = edges_.back();
+			edges_.pop_back();
+			if (isWeighted()) {
+				auto &weights_ = edgesIn ? inEdgeWeights[u] : outEdgeWeights[u];
+				weights_[vi] = weights_.back();
+				weights_.pop_back();
+			}
+			if (hasEdgeIds()) {
+				auto &edgeIds_ = edgesIn ? inEdgeIds[u] : outEdgeIds[u];
+				edgeIds_[vi] = edgeIds_.back();
+				edgeIds_.pop_back();
+			}
+		} else {
+			++vi;
+		}
+	}
+
+	return removedEdges;
+}
+
 
 } /* namespace NetworKit */
 
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -784,7 +784,7 @@ class Graph final {
 	 * @param edges the edge count of a graph
 	 */
 	void setEdgeCount(Unsafe, count edges) { m = edges; }
-    
+
 	void setNumberOfSelfLoops(Unsafe, count loops) { storedNumberOfSelfLoops = loops; }
 	/**
      * Returns a string representation of the graph.
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -829,11 +829,41 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
 }
 
 void Graph::removeSelfLoops() {
-	this->forEdges([&](node u, node v, edgeweight) {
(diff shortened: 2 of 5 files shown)


include/networkit/graph/Graph.hpp
```diff
@@ -640,6 +640,18 @@ class Graph final {
...
+    /**
+     * Removes edges adjacent to a node according to a specific criterion.
+     *
+     * @param u The node whose adjacent edges shall be removed.
+     * @param condition A function that takes a node as an input and returns a
+     * bool. If true the edge (u, v) is removed.
+     * @param edgesIn Whether in-going or out-going edges shall be removed.
+     * @return count The number of removed edges.
+     */
+    template <typename Condition>
```
manpen: Implementation is not available to user-code as provided in `Graph.cpp` file.

networkit/cpp/graph/Graph.cpp
```diff
@@ -828,12 +828,65 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
...
+			edges_.pop_back();
+			if (isWeighted()) {
+				weights_[vi] = weights_.back();
+				weights_.pop_back();
+			}
+			if (hasEdgeIds()) {
+				edgeIds_[vi] = edgeIds_.back();
+				edgeIds_.pop_back();
+			}
+		} else
```
manpen: Consider using braces here.

networkit/cpp/graph/Graph.cpp
```diff
@@ -828,12 +828,65 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
...
+	count nEdgesRemoved = 0;
+
+	forNodes([&](node u) {
+		auto isSelfLoop = [&u](node v) { return u == v; };
+		nEdgesRemoved += removeAdjacentEdges(u, isSelfLoop);
+		if (isDirected())
+			nEdgesRemoved += removeAdjacentEdges(u, isSelfLoop, true);
 	});
+
+	m -= nEdgesRemoved;
```
manpen: `assert(nEdgesRemoved == numberOfSelfLoops());`?

networkit/cpp/graph/Graph.cpp
```diff
@@ -828,12 +828,65 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
...
+	storedNumberOfSelfLoops = 0;
+}
+
+void Graph::removeMultiEdges() {
+	count nEdgesRemoved = 0;
+	std::unordered_set<node> nodes;
+
+    forNodes([&](node u) {
```
manpen: Indention is off.

networkit/cpp/graph/Graph.cpp
```diff
@@ -828,12 +828,65 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
 	this->m -= removedEdges;
 }
 
+template <typename Condition>
+count Graph::removeAdjacentEdges(node u, Condition condition, bool edgesIn) {
+	count removedEdges = 0;
+	std::vector<node> &edges_ = outEdges[u];
```
manpen: why not use `auto&`?

networkit/cpp/graph/Graph.cpp
manpen: It seems to me that the lambda is thread-safe with the exeception of `nEdgesRemoved`. Parallelize?

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2381,5 +2381,21 @@ TEST_P(GraphGTest, testSubgraphFromNodesDirected) {
...
+
+	// Adding multiedges at random
+	for (count i = 0; i < nMultiEdges; ++i) {
+		auto e = G.randomEdge();
+		G.addEdge(e.first, e.second);
+	}
+
+	EXPECT_EQ(G.numberOfEdges(), m + nMultiEdges);
+	G.removeMultiEdges();
+	EXPECT_EQ(G.numberOfEdges(), m);
```
manpen: Please check whether edge sets are identical or at least add a TODO (I've an algorithm for that, but can create PR only in three weeks)

include/networkit/graph/Graph.hpp
angriman: Now it is public.

networkit/cpp/graph/Graph.cpp
angriman: Good point, I added the assertion.

networkit/cpp/graph/Graph.cpp
angriman: Now it is ok.

include/networkit/graph/Graph.hpp
manpen: But template-instatiation does not work for user code, as the implementation is in a different compilation uint (`Graph.cpp`); if you have a templated method, the implementation has to be provided in the header file (e.g., `forNodes`).

manpen:
As a retaliation of https://github.com/networkit/networkit/pull/386#discussion_r308214215, please also consider to add a Python unit test for `Graph.removeMultiEdges()` ;)

include/networkit/graph/Graph.hpp
angriman: True, my bad I forgot that. I moved the method to the header.

networkit/cpp/graph/Graph.cpp
angriman: Good point.

networkit/cpp/graph/Graph.cpp
angriman: That's right, now it is parallelized.

networkit/cpp/graph/test/GraphGTest.cpp
angriman: Right, I've added those checks in the tests.

angriman:
Right, I've added a Python unit test for `Graph.removeMultiEdges()`.

angriman:
> As a general remark: we keep on adding functions to the Graph class. Maybe it's better to have it in Graph-Tools?

I would agree on that. The main issue is that `GraphTools` is not a friend of `Graph`, and modifiers like this cannot be implemented (unless we create another copy of the graph).

avdgrinten:
We could add more `unsafe` methods if we want to.

angriman:
> We could add more `unsafe` methods if we want to.

Makes sense. I think that the purpose of `removeMultiEdges` is very close to `removeSelfLoops`, so I'd be in favor of keeping it within the `Graph` class for now. 
On the other hand, I see that several methods in `Graph` could be moved somewhere else (e.g., in `GraphTools`), and we might do so in a separate PR.

PR #396, opened 2019-08-05
Added rpath support for setup.py and consequetive cmake build system.

fabratu:
This adds the possibility to add rpath to both setup.py and cmake. Use case: consecutive packages (python -> C++ external-core) in build systems like spack, which make heavy usage of rpath-manipulation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ set(NETWORKIT_PYTHON "" CACHE STRING "Directory containing Python.h. Implies MON
 set(NETWORKIT_PYTHON_SOABI "" CACHE STRING "Platform specific file extension. Implies MONOLITH=TRUE")
 set(NETWORKIT_WITH_SANITIZERS "" CACHE STRING "Uses sanitizers during the compilation")
 set(NETWORKIT_RELEASE_LOGGING "AUTO" CACHE STRING "Do not compile log messages at levels TRACE or DEBUG (AUTO|ON|OFF)")
-set(NETWORKIT_RPATH "" CACHE STRING "Build specific rpath references.")
+set(NETWORKIT_PYTHON_RPATH "" CACHE STRING "Build specific rpath references.")
 
 set (NETWORKIT_CXX_STANDARD "11" CACHE STRING "CXX Version to compile with. Currently fixed to 11")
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -140,7 +140,7 @@ def cythonizeFile(filepath):
 			exit(1)
 		print("_NetworKit.pyx cythonized", flush=True)
 
-def buildNetworKit(install_prefix, externalCore=False, withTests=False):
+def buildNetworKit(install_prefix, externalCore=False, withTests=False, rpath=False):
```
avdgrinten: `rpath=None` makes more sense than `rpath=False` here.

CMakeLists.txt
```diff
@@ -225,10 +226,10 @@ if(NETWORKIT_PYTHON)
 	# DSOs on Apple OSes use different conventions for RPATH.
 	if(APPLE)
 		set_target_properties(_NetworKit PROPERTIES
-				INSTALL_RPATH "@loader_path")
+				INSTALL_RPATH "@loader_path;${NETWORKIT_RPATH}")
```
avdgrinten: I think the rpath given as a command line parameter should overwrite the entire rpath. If the user wants to keep `@loader_path`/`$ORIGIN`, they can pass an appropriate rpath.

CMakeLists.txt
```diff
@@ -225,10 +226,10 @@ if(NETWORKIT_PYTHON)
...
 	else()
 		set_target_properties(_NetworKit PROPERTIES
-				INSTALL_RPATH "$ORIGIN")
+				INSTALL_RPATH "$ORIGIN;${NETWORKIT_RPATH}")
```
avdgrinten: Same here.

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(NETWORKIT_PYTHON "" CACHE STRING "Directory containing Python.h. Implies MON
 set(NETWORKIT_PYTHON_SOABI "" CACHE STRING "Platform specific file extension. Implies MONOLITH=TRUE")
 set(NETWORKIT_WITH_SANITIZERS "" CACHE STRING "Uses sanitizers during the compilation")
 set(NETWORKIT_RELEASE_LOGGING "AUTO" CACHE STRING "Do not compile log messages at levels TRACE or DEBUG (AUTO|ON|OFF)")
+set(NETWORKIT_RPATH "" CACHE STRING "Build specific rpath references.")
```
avdgrinten: This only affects the rpath of the Python module. Should we call the option `NETWORKIT_PYTHON_RPATH`?

PR #399, opened 2019-08-14
Jupyter Notebook for the centrality module

cndolo:
This PR adds a new Jupyter-notebook for the centrality module. It covers most of the algorithms on
- Betweenness Centrality,
- Closeness Centrality,
- the PageRank Centrality algorithms,
- and the Degree Centrality algorithm.

All dynamic and group centrality algorithms were intentionally excluded from this notebook.
I made minor changes to the documentation too.

### DIFF
--- a/notebooks/Centrality.ipynb
+++ b/notebooks/Centrality.ipynb
@@ -1,0 +1,887 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Centrality Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook covers some of the centrality algorithms implemented in NetworKit. [Centrality](http://en.wikipedia.org/wiki/Centrality) measures the relative importance of a node within a graph. Code for centrality analysis in NetworKit is grouped into the [centrality](https://networkit.github.io/dev-docs/python_api/centrality.html) module."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As always, the first step is to start by importing NetworKit."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "import networkit as nk"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Every algorithm within the `centrality` module in NetworKit has a `run()` function that must be called in order for the algorithm to be executed after initialization."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


notebooks/Centrality.ipynb
```diff
@@ -0,0 +1,798 @@
+{
```
angriman: Thank you for the new notebook. Some issues to address:
- First sentence: remove "the" before "some"
- After you import networkit, change the text "Every centrality algorithm" to "Every algorithm within the centrality module"
- Betweenness Centrality: "... lies on paths" -> "... lies on shortest paths"
- Betweenness: replace the first sentence with "The fastest algorithm for the exact computation of the betweenness centrality is Brandes' algorithm, and it is implemented in the NetworKit Betweenness class"
- The graph you choose for running Brandes' algorithm is quite large, it took ~15 sec on my machine. I'd suggest to use a smaller graph.
- KadabraBetweenness: in the text above the initialization, I think you should change "additive error of 0.5" to "additive error of 0.05". Also, the probability is 0.8, not 0.2
- Closeness Centrality: it is calculated as the inverse of the average shortest path length ...
- Closeness: The NetworKit Closeness class computes the exact closeness centrality of all the nodes of a graph. Its constructor ...
- Closeness: try to add some more details about the difference between the standard and generalized closeness before you use them. For example you could write something like https://github.com/networkit/networkit/pull/377 (it is not yet merged, but clarifies how this parameter works). Also, it might be clearer for the user to use nk.centrality.ClosenessVariant.Standard/Generalized instead of 0/1.
- ApproxCloseness: In the sentence "`epsilon` can be any value between 0 can $\infty$-1" did you mean "`epsilon` can be any value between 0 and 1"?
- Top(Harmonic)Closeness: remark that those algorithms compute the exact top-k nodes with highest harmonic closeness centrality.
- Katz Centrality: the chosen example results in a score of zero for the first top 10 nodes. Try to use a larger alpha (e.g., 1e-3) to get a more illustrative ranking.
- Spanning Edge Centrality: the scores are almost all zeros. Try a different graph and just print the top 10 nodes.

cndolo: Thanks for the review, I will make the changes soon.

Concerning  `epsilon` in the `ApproxCloseness` constructor: the documentation in the C++ header file states that `epsilon is a value in [0, $\infty$).

angriman: I see, my bad. In that case just change "0 can $\infty$-1" to "0 and $\infty$-1".

cndolo:
I have made the above changes. However, I wasn't able to find a graph with illustrative results for the `SpanningEdgeCentrality`, so I just left it as it was.

notebooks/Centrality.ipynb
angriman: Just very minor adjustments:
- In some comments a space is missing after the `#`
- normalised -> normalized, initialise -> initialize

cndolo: I have made the changes.

angriman: Thank you, it looks fine to me

PR #400, opened 2019-08-16
Fix number of parameters in Graph/neighbors

angriman:
This fixes a bug in the `Graph.neighbors()` method. The current implementation returns the error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "networkit/_NetworKit.pyx", line 922, in networkit._NetworKit.Graph.neighbors
    self.forEdgesOf(u, lambda v : neighborList.append(v))
  File "networkit/_NetworKit.pyx", line 1019, in networkit._NetworKit.Graph.forEdgesOf
    self._this.forEdgesOf[EdgeCallBackWrapper](u, dereference(wrapper))
RuntimeError: An Exception occurred, aborting execution of iterator: lambda() takes exactly one argument (4 given)
```
because the lambda does not have enough parameters (1 instead of 4), and cannot be used by external users.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -925,7 +925,7 @@ cdef class Graph:
 	 		List of neighbors of `u`.
 		"""
 		neighborList = []
-		self.forEdgesOf(u, lambda v : neighborList.append(v))
+		self.forEdgesOf(u, lambda u, v, w, eid : neighborList.append(v))
 		return neighborList
 
 	def inNeighbors(self, u):
--- a/networkit/test/test_graph.py
+++ b/networkit/test/test_graph.py
@@ -92,6 +92,34 @@ class TestGraph(unittest.TestCase):
 		m = G.numberOfEdges()
 		testGraph(G)
 
+	def testNeighbors(self):
+		# Directed
+		G = nk.Graph(4, False, True)
+
+		G.addEdge(0, 1)
+		G.addEdge(0, 2)
+		G.addEdge(3, 1)
+		G.addEdge(3, 2)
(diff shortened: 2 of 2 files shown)


manpen:
Two comments:
1.) Can you please add a small unit test. This should have been spotted much earlier.
2.) While this fixes this issue, I think it would be nice to have similar semantics for the CPP and Python interfaces (i.e. overloads for different signatures). While we cannot distinguish between overloads for `edgeweight` and `edgeindex`, we could easily support the following signatures: (u), (u,v), (u,v,w,eid). A solution could be as follows:

```Python
import inspect

[...]

cdef cppclass EdgeCallBackWrapper:
	void* callback
	object fu

	__init__(object callback):
		n = len(inspect.signature(callback).parameters.items())
		if n == 1:
				this.fu = lambda u, v, w, eid: callback(v)
		elif n == 2:
				this.fu = lambda u, v, w, eid: callback(u, v)
		elif n == 4:
				this.fu = callback
		else:
    				throw_runtime_error(stdstring("Unsupported signature"))

		this.callback = <void*>this.fu
[...]
```

what do you think?

angriman:
Thank you. I added a small Python test.

I agree with overloading different signatures, this would simply a lot the usage of that function.
Also, it might be better to write some Python iterators to iterate over nodes/neighbors more efficiently.

avdgrinten:
I am not sure about the overloading. If we chose to implement this, it should be done in a different PR.

manpen:
Adding overloads and iterators may be too much for the PR, agreed. Thanks for adding the test, though!

PR #402, opened 2019-08-21
macOS 10.14 with gcc 9

angriman:
This PR updates to gcc 9 the macOS 10.14 travis build.


avdgrinten:
What's the rationale? Why do we update GCC 8 -> GCC 9 here but not for all other builds?

PR #403, opened 2019-08-21
New algorithm A*

angriman:
This PR brings an implementation of the A* path-finding algorithm.


include/networkit/distance/AStar.hpp
```diff
@@ -0,0 +1,65 @@
...
+        astar.run();
+        hasRun = true;
+    }
+
+    edgeweight getDistance() const override { return astar.getDistance(); }
+
+    std::vector<node> getPath() const { return astar.getPath(); }
+
+private:
+    const std::vector<double> &distanceHeu;
```
manpen: `distanceHeu` is unused

include/networkit/distance/AStar.hpp
```diff
@@ -0,0 +1,65 @@
...
+     * node to the target.
+     * @param source The source node.
+     * @param target The target node.
+     * @param storePred If true, the algorithm will also store the predecessors
+     * and reconstruct a shortest path from @a source and @a target.
+     */
+    AStar(const Graph &G, const std::vector<double> &distanceHeu, node source, node target,
+                 bool storePred = true)
+        : STSP(G, source, target, storePred), distanceHeu(distanceHeu),
+          astar(AStarGeneral<Heuristic>(G, Heuristic(distanceHeu), source, target, storePred)) {}
```
manpen: Consider checking that `G.numberOfNodes() == distanceHeu.size()` and throw otherwise.

include/networkit/distance/AStar.hpp
```diff
@@ -0,0 +1,65 @@
...
+    edgeweight getDistance() const override { return astar.getDistance(); }
+
+    std::vector<node> getPath() const { return astar.getPath(); }
+
+private:
+    const std::vector<double> &distanceHeu;
+    struct Heuristic {
+    public:
+        Heuristic(const std::vector<double> &distanceHeu) : distanceHeu(distanceHeu) {}
+        bool operator()(node u) { return distanceHeu[u]; }
```
manpen: Why does the Function return bool? Should be `double`!
Functor should be const (and noexcept)

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+     * Executes the AStar algorithm.
+     */
+    void run() override {
+        if (source == target) {
+            hasRun = true;
+            return;
+        }
+
+        init();
+        const edgeweight infdist = std::numeric_limits<edgeweight>::max();
```
manpen: can be constexpr; type is specified redudantly. consider `constexpr auto infdist= ...`

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param G The graph.
+     * @param heu Heuristic function that takes a node as input and returns a
+     * lower bound of the distance of that node to the target node.
+     * @param source The source node.
+     * @param target The target node.
+     * @param storePred If true, the algorithm will also store the predecessors
+     * and reconstruct a shortest path from @a source and @a target.
+     */
+    AStarGeneral(const Graph &G, Heuristic heu_, node source, node target, bool storePred = true)
```
manpen: Name `heu_` does not match comment. Rename it `heu`

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+        heap.push(source);
+
+        node top = none;
+        do {
+            top = heap.extract_top();
+            if (top == target)
+                break;
+
+            G.forNeighborsOf(top, [&](node u, edgeweight w) {
+                double newDist = distance[top] + w;
```
manpen: const double

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+private:
+    // Distance from the source node, priorities used for the heap
+    std::vector<double> distance, prio;
+    // Lower bound of the distance to the target node
+    Heuristic heu;
+
+    struct Compare {
+    public:
+        Compare(const std::vector<double> &prio_) : prio(prio_) {}
```
manpen: Rename `prio` to prio?

networkit/_NetworKit.pyx
```diff
@@ -1447,6 +1447,53 @@ cdef class BidirectionalDijkstra(STSP):
...
+		The source node.
+	target : node
+		The target node.
+	storePred : bool
+		If true, the algorithm will also store the predecessors
+		and reconstruct a shortest path from @a source and @a target.
+	"""
+
+	cdef vector[double] heu
+	def __cinit__(self, Graph G, vector[double] &heu_, node source, node target, bool_t storePred):
```
manpen: Rename heu_ to heu? Otherwise adapt docstring

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        init();
+        const edgeweight infdist = std::numeric_limits<edgeweight>::max();
+        const count n = G.upperNodeIdBound();
+
+        distance.resize(n, infdist);
+        distance[source] = 0.;
+
+        std::fill(prio.begin(), prio.end(), infdist);
```
manpen: Why do you intialise distance and prio differently? Do you support multiple `run` invocations? If so, I think distance should be filled. But I might be mistanken. If not, resize suffices for prio.

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -0,0 +1,115 @@
...
+private:
+    // Distance from the source node, priorities used for the heap
+    std::vector<double> distance, prio;
+    // Lower bound of the distance to the target node
+    Heuristic heu;
+
+    struct Compare {
+    public:
+        Compare(const std::vector<double> &prio_) : prio(prio_) {}
+        bool operator()(node u, node v) { return prio[u] < prio[v]; }
```
manpen: make it const

include/networkit/distance/AStar.hpp
```diff
@@ -0,0 +1,65 @@
...
+     */
+    void run() override {
+        astar.run();
+        hasRun = true;
+    }
+
+    edgeweight getDistance() const override { return astar.getDistance(); }
+
+    std::vector<node> getPath() const { return astar.getPath(); }
+
```
manpen: `std::vector<node> getPredecessors() const` is inherited from `STSP` but is not forwarded as getDistance. Also observe that neither getDistance nor getPredecessor are virtual in STSP, so we might get unexcepted behaivor ...

angriman: Do you have some suggestions? Also, `getDistance` is virtual in STSP, while `getPredecessor` is not.

include/networkit/distance/AStarGeneral.hpp
angriman: You are right, we support multiple `run` invocations and  we need to re-fill the `distance` vector. I was using another vector to mark nodes as visited, but after removing that I forgot to do this change.

include/networkit/distance/AStar.hpp
manpen: While you fixed the bug in my comment and used `upperNodeIdBound` rather than `numberOfNodes`, the exception msg still says number of nodes. This might be misleading.

include/networkit/distance/AStar.hpp
manpen: Should we make it virtual in STSP and mark all implementations final? Then virtual dispatch only occurs, if one uses generic STSP types?

include/networkit/distance/AStar.hpp
angriman: I think that `upperNodeIdBound ` makes more sense if the graph has deleted nodes. However, I agree that this should be reported in the error message.

include/networkit/distance/AStar.hpp
manpen: Just so we're not in a deadlock waiting on each other. This issue is the only thing stopping me from approving the PR. Should we just make `STSP::getPath` virtual and mark `AStar::getPath` as override (it's the only method overriding; all other implementations rely on the generic impl).

angriman: `AStar::getPath` does not do anything different than `STSP::getPath`, so I think that the best solution here is to just delete `AStar::getPath`.

angriman: That's true. I made `STSP::getPath`virtual and marked `AStar::getPath` as override.

manpen:
LGTM. Thank you!

PR #404, opened 2019-08-23
Jupyter Notebook for the components module

cndolo:
This PR adds a new Jupyter-notebook for the `components` module; most algorithms in the module are covered. I made minor changes to the documentation too.

### DIFF
--- a/notebooks/Components.ipynb
+++ b/notebooks/Components.ipynb
@@ -1,0 +1,301 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Components"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this notebook we present some algorithms to analyze the connectivity properties of a graph.\n",
+    "NetworKit's components classes are grouped in the [networkit.components](https://networkit.github.io/dev-docs/python_api/components.html) module. They all have a `run` method that must be called after the chosen algorithm has been initialised. The first step is to import Networkit."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import networkit as nk"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Connected Components"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


hmeyerhenke:
Note to all: these notebooks displaying the use of certain functionality should be easily accessible from the NetworKit homepage as well. We may want to have an own category on the website for this -- or a link to the respective github folder(s).

notebooks/Components.ipynb
```diff
@@ -0,0 +1,270 @@
+{
```
angriman: Thank you for the notebook. Some issues to be addressed:
- The introduction is a bit misleading because the notebook covers more than connected components. You could mention that in this notebook we present some algorithms to analyze the connectivity properties of a graph.
- It would be better to add spaces in the headings (e.g., ConnectedComponents -> Connected Components)
- The connected components example is a bit trivial as the graph is already connected. Use a disconnected 
graph instead (or remove some edges so that the graph becomes disconnected). Also, we recently added the feature `extractLargestConnectedComponent`. We could add an example of that function (@avdgrinten what do you think?).
- BiConnectedComponent(G) -> BiconnectedComponent(G)
- The biconnected components class expects an undirected graph
- `getVector()` gives you a vector of indexes that represent the component of each node. There is no need to wrap it into another list. Also, the loop below does not what you intend to do, you could correct it with something like this:
```
# Get sorted list of components
partition = scc.getPartition()
indexes = sorted(set(partition.getVector()))

# Print every member of each component
for cmp in indexes:
    print("Members of component {} : {}".format(cmp, partition.getMembers(cmp)))
```
Anyway, this example is also quite naive as we have a single node per component. Use a different graph or add some more nodes/edges to it so we have a more illustrative example

cndolo: I have made the changes; please have a look and let me know if I should change anything else.

angriman: Thank you for the changes. In the Connected Components section it might look a bit confusing if you remove some edges before running the algorithm without any explanation. I'd suggest to use the `hep-th.graph` graph, which is already disconnected, and avoid to print the full map of component sizes a because it would be too large.

cndolo: I have used the `hep-th.graph` graph, and only printed the first component containing the first node as it contained a reasonable (maybe too few?) amount of nodes. The following components are large, and only the last ones are of reasonable size again.

angriman: I think it is fine if you print the nodes in the first component. Instead of `cc.getComponents()[:1]` use `cc.getComponents()[0]`.

cndolo: I have made the change according to your suggestion.

angriman: Thank you, I have no further comments.

Issue #398, opened 2019-08-14
Complexity of BiconnectedComponents


gilcu3:
As far I can see, your implementation of Tarjan's algorithm for finding biconnected components is not linear as expected. The culprit is line https://github.com/networkit/networkit/blob/cd5c91511a0fe045011ef8a27d56220641203554/networkit/cpp/components/BiconnectedComponents.cpp#L63, in which neighbours will be visited several times, as the neighbours visited already must be remember somehow to avoid redundant computations. A test case for which this behavior is present is not hard to construct.

avdgrinten:
This is indeed problematic, thank you for the report. As a fix, we could store the neighbor iterator on the stack, in addition to the node that we are expanding.

gilcu3:
> This is indeed problematic, thank you for the report. As a fix, we could store the neighbor iterator on the stack, in addition to the node that we are expanding.

Yes exactly the standard solution :)

angriman:
Thank you @gilcu3 for reporting this issue. I've addressed it in #405.

PR #405, opened 2019-08-26
Fix performance issue in BiconnectedComponents

angriman:
This PR fixes a performance issue of `BiconnectedComponents` raised in #398.


avdgrinten:
Shouldn't `if(allVisited)` now be changed to `if(iter == G.neighborRange(u).end())`? AFAICS, we want to check whether the descendants of the node have been fully visited or not.

angriman:
That's true, thank you. Using the iterator makes that variable useless.

avdgrinten:
LGTM.

PR #406, opened 2019-09-13
Python wrapper for the C++ NeighborRange iterator

cndolo:
The `NeighborRange` iterator is wrapped, and we introduce an equivalent iterator in Python - `iterNeighbors(node u)`. Currently, it is only available for undirected, unweighted graphs.

The current iterators, `forEdges*` or `forNodes*`, require a callback function and are not necessarily intuitive to use for all users. The `iterNeighbors(node u)` iterator is simple to use and pythonic .

Let me briefly explain the changes made in `Graph.hpp`.
 - Addition of a default no parameter constructor: Cython needed a default contructor.
 - `Const node u` -> `node u`: When using the new default constructor, it should be possible to assign a value to `node u` later. This is not possible is `node u` is declared as `const`.
 - `Graph& -> Graph*`: Again, this is because of the new constructor.
 - NeighborIterator from `private` to `public`: This was necessary in order to access the NeighborIterator from Cython.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -444,7 +444,7 @@ class Graph final {
     auto callBFSHandle(F &f, node u, count) const -> decltype(f(u)) {
         return f(u);
     }
-
+public:
     /**
      * Class to iterate over the in/out neighbors of a node.
      */
@@ -453,25 +453,30 @@ class Graph final {
       public:
         // The value type of the neighbors (i.e. nodes). Returned by
         // operator*().
-        typedef node value_type;
+        using value_type = node;
 
         // Reference to the value_type, required by STL.
-        typedef value_type &reference;
+        using reference = value_type &;
 
         // Pointer to the value_type, required by STL.
-        typedef value_type *pointer;
+        using pointer = value_type *;
 
         // STL iterator category.
-        typedef std::forward_iterator_tag iterator_category;
+        using iterator_category = std::forward_iterator_tag;
 
         // Signed integer type of the result of subtracting two pointers,
         // required by STL.
-        typedef ptrdiff_t difference_type;
+        using difference_type = ptrdiff_t;
 
         // Own type.
-        typedef NeighborIterator self;
+        using self = NeighborIterator;
 
         NeighborIterator(std::vector<node>::const_iterator nodesIter) : nIter(nodesIter) {}
+
+        /**
+         * @brief WARNING: This contructor is required for Python and should not be used as the iterator is not initialised.
+         */
+        NeighborIterator() {}
 
         NeighborIterator operator++() {
             auto prev = *this;
@@ -518,23 +523,23 @@ class Graph final {
       public:
         // The value type of the neighbors (i.e. nodes). Returned by
         // operator*().
-        typedef std::pair<node, edgeweight> value_type;
+        using value_type = std::pair<node, edgeweight>;
 
         // Reference to the value_type, required by STL.
-        typedef value_type &reference;
+        using reference = value_type &;
 
         // Pointer to the value_type, required by STL.
-        typedef value_type *pointer;
+        using pointer = value_type *;
 
         // STL iterator category.
-        typedef std::forward_iterator_tag iterator_category;
+        using iterator_category = std::forward_iterator_tag;
 
         // Signed integer type of the result of subtracting two pointers,
         // required by STL.
-        typedef ptrdiff_t difference_type;
+        using difference_type = ptrdiff_t;
 
         // Own type.
-        typedef NeighborWeightIterator self;
+        using self = NeighborWeightIterator;
 
         NeighborWeightIterator(
             std::vector<node>::const_iterator nodesIter,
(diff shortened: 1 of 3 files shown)


include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +592,29 @@ class Graph final {
...
         }
 
       private:
-        const Graph &G;
-        const node u;
+        const Graph *G;
+        node u;
     };
 
+	typedef NeighborRange<false> UndirectedNeighborRange;
```
manpen: While it's not in our style guide, I feel `typedefs` are superseded by `using. So please rewrite  `using UndirectedNeighborRange = NeighborRange<false>`. While you're at it: could you please also refactor the typedefs in the iterators themselves?

include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +592,29 @@ class Graph final {
      */
     template <bool InEdges = false> class NeighborRange {
       public:
-        NeighborRange(const Graph &G, node u) : G(G), u(u){};
+        NeighborRange(const Graph &G, node u) : G(&G), u(u){};
+
+        NeighborRange() : G(nullptr) {};
 
         NeighborIterator begin() const {
             if (InEdges)
```
manpen: Check that `G` is not nullptr. At the very least use `assert()`.

include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +592,29 @@ class Graph final {
...
-                return NeighborIterator(G.inEdges[u].begin());
-            return NeighborIterator(G.outEdges[u].begin());
+                return NeighborIterator(G->inEdges[u].begin());
+            return NeighborIterator(G->outEdges[u].begin());
         }
 
         NeighborIterator end() const {
             if (InEdges)
```
manpen: Check that `G` is not nullptr.

include/networkit/graph/Graph.hpp
manpen: On a more general node: This iterator is also used to iterate over directed out neighbors. So the name is somewhat misleading. In anycase, please also adapt the signature of `Graph::neighborRange`.

networkit/_NetworKit.pyx
```diff
@@ -1347,6 +1365,23 @@ cdef class Graph:
 			nnodes.insert(node)
 		return Graph().setThis(self._this.subgraphFromNodes(nnodes, includeOutNeighbors, includeInNeighbors))
 
+	def iterNeighbors(self, u):
```
manpen: Please add support for  directed edges. If for some reason that should be not possible, please do not silently fail.

networkit/_NetworKit.pyx
```diff
@@ -1347,6 +1365,23 @@ cdef class Graph:
...
+		a for loop.
+		Parameters
+		----------
+		u : Node
+
+		Returns
+		-------
+		List of a node u's neighbours
+		"""
+		if(self._this.isDirected() == False):
```
manpen: if not self._ ... :

include/networkit/graph/Graph.hpp
```diff
@@ -444,7 +444,7 @@ class Graph final {
     auto callBFSHandle(F &f, node u, count) const -> decltype(f(u)) {
         return f(u);
     }
-
+public:
```
manpen: Please remove public in line 649

include/networkit/graph/Graph.hpp
```diff
@@ -473,6 +473,8 @@ class Graph final {
 
         NeighborIterator(std::vector<node>::const_iterator nodesIter) : nIter(nodesIter) {}
 
+        NeighborIterator() {}
```
manpen: Please add a comment that this c'stor is required for Python (and why) and should not be used directly.

include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +592,29 @@ class Graph final {
      */
     template <bool InEdges = false> class NeighborRange {
       public:
-        NeighborRange(const Graph &G, node u) : G(G), u(u){};
+        NeighborRange(const Graph &G, node u) : G(&G), u(u){};
```
manpen: Consider adding `assert(G->hasNode(u));`

cndolo:
Thank you for your review @manpen; I have made the changes.

include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +595,34 @@ class Graph final {
      */
     template <bool InEdges = false> class NeighborRange {
       public:
-        NeighborRange(const Graph &G, node u) : G(G), u(u){};
+        NeighborRange(const Graph &G, node u) : G(&G), u(u){
+			assert(exists[u]);
```
manpen: This has to be `assert(G.hasNode(u));`

include/networkit/graph/Graph.hpp
```diff
@@ -590,25 +595,34 @@ class Graph final {
...
+        };
+
+        NeighborRange() : G(nullptr) {};
 
         NeighborIterator begin() const {
+            assert(!(G==nullptr));
```
manpen: Should be either `assert(G)` or `assert(G != nullptr)`; I, personally, prefer the second.

include/networkit/graph/Graph.hpp
manpen: Please make that a doxygen comment.

include/networkit/graph/Graph.hpp
cndolo: Thanks for this - this is the reason the CI was failing. I have changed it, as well as the other issues.

manpen:
Thank you. LGTM

avdgrinten:
Thanks to @cndolo for the work and to @manpen for the review.

PR #407, opened 2019-09-14
Fix Graph::randomEdges and Accelerate Graph::randomEdge

manpen:
This PR includes:
- Fix: Previously `Graph::randomEdges()` used `std::random_device` to seed its own `std::default_random_engine`. This obviously leads to non-reproducible tests. We now use the generator obtained via `Aux::Random::getURNG()`.
- Improvement: Previously `Graph::randomEdge()` used `Graph::randomEdges(1)` to obtain a single edge uniformly at random. This is highly suboptimal, since `Graph::randomEdges` uses `std::discrete_distribution` which builds a costly Theta(n) data structure -- this effort cannot be amortized when only sampling a single edge. This PR hence uses a single linear scan over the graph. In (maybe unrepresentative) measurements, the new implementation is more than an order of mag. faster.
- Adds CPP and Python unit tests

### DIFF
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -987,7 +987,7 @@ std::vector<std::pair<node, node>> Graph::randomEdges(count nr) const {
 	}
 	std::vector<std::pair<node, node>> edges;
 
-	std::default_random_engine gen{std::random_device{}()};
+	auto& gen = Aux::Random::getURNG();
 	std::vector<count> outDeg(upperNodeIdBound());
 	for (count i = 0; i < upperNodeIdBound(); ++i) {
 		outDeg[i] = outEdges[i].size();
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -919,16 +919,65 @@ std::pair<node, node> Graph::randomEdge(bool uniformDistribution) const {
 	}
 
 	if (uniformDistribution) {
-		return randomEdges(1)[0];
+		/*
+		 * The simple idea here is to interpret all neighborhoods next to each other, resulting
+		 * in a virtual vector of size m. Then we draw a random index and return the edge.
+		 * For undirected edges, the vector has size 2m; but the idea remains. There is one minor
+		 * complication for undirected edges with self-loops: each edge {u,v} with u != v is stored
+		 * twice (once in the neighborhood of u, once in v) but a loop (u, u) is only stored once.
+		 * To equalize the probabilities we reject edges {u,v} with u > v and try again. This leads
+		 * to less than two expected trails in and is only done for undirected graphs with self-loops.
+		 */
+
+		while (true) {
+			const auto upper = directed
+				? numberOfEdges()
+				: 2 * numberOfEdges() - numberOfSelfLoops();
+			auto idx = Aux::Random::index(upper);
+
+			node u, v;
+
+			if (idx > upper / 2) {
+				// assuming degrees are somewhat distributed uniformly, it's better to start with
+				// larger nodes for large indices. In this case we have to mirror the index:
+				idx  = (upper-1) - idx;
+
+				for(u = upperNodeIdBound() - 1; idx >= degreeOut(u); --u) {
+					assert(0 <= u);
+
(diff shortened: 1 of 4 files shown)


manpen:
The failing CI test seems to be a hickup of AppVeyor. There were no errors on my AppVeyor build of the same commit.

networkit/cpp/graph/Graph.cpp
```diff
@@ -919,16 +919,48 @@ std::pair<node, node> Graph::randomEdge(bool uniformDistribution) const {
...
+		 * to less than two expected trails in and is only done for undirected graphs with self-loops.
+		 */
+
+		while(true) {
+			node u = 0;
+
+			const auto upper = directed ? numberOfEdges() : 2 * numberOfEdges() - numberOfSelfLoops();
+			auto idx = Aux::Random::index(upper);
+
+			while (idx >= degreeOut(u)) {
```
angriman: Would that be possible to initialize another node to `upperNodeIdBound() - 1`, another index to `numberOfEdges() - idx`, and iterate also from above? This could reduce the number of iterations of this loop.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -898,10 +898,37 @@ TEST_P(GraphGTest, testHasEdge) {
 
 TEST_P(GraphGTest, testRandomEdge) {
```
angriman: Should we set random seeds as in the Python tests?

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -898,10 +898,37 @@ TEST_P(GraphGTest, testHasEdge) {
...
+		count id = (e.first * e.second) % 5;
+		drawCounts[id]++;
+	}
+	for (node id = 0; id < m; id++) {
+		double p = drawCounts[id] / (double)samples;
+		ASSERT_NEAR(1.0 / m, p, maxAbsoluteError);
+	}
+}
+
+TEST_P(GraphGTest, testRandomEdges) {
```
angriman: Random seeds also here?

networkit/cpp/graph/Graph.cpp
```diff
@@ -919,16 +919,48 @@ std::pair<node, node> Graph::randomEdge(bool uniformDistribution) const {
...
+		 * The simple idea here is to interpret all neighborhoods next to each other, resulting
+		 * in a virtual vector of size m. Then we draw a random index and return the edge.
+		 * For undirected edges, the vector has size 2m; but the idea remains. There is one minor
+		 * complication for undirected edges with self-loops: each edge {u,v} with u != v is stored
+		 * twice (once in the neighborhood of u, once in v) but a loop (u, u) is only stored once.
+		 * To equalize the probabilities we reject edges {u,v} with u > v and try again. This leads
+		 * to less than two expected trails in and is only done for undirected graphs with self-loops.
+		 */
+
+		while(true) {
```
angriman: Space after while?

networkit/cpp/graph/Graph.cpp
manpen: If we assume that degrees are randomly distributed, your suggestion works. However, I often deal with graph in which this is not the case and then your suggestion would be pessimization. Hence, as the outcome is unclear and we just got a speed-up of one order of magnitude, I would lean to not implement it for clarity's sake.

If you feel strongly about it, I obviously will do it.

manpen: What about the following compromise: if idx < m / 2 start from begin, if idx > m/2 start from end?

angriman: Then I think that it is better to leave it unchanged.

angriman: If we do not have any prior information about the degree distribution, I guess that this is the best that we can do. So I agree with this compromise.

networkit/cpp/graph/Graph.cpp
avdgrinten: As a side note: in general, I don't think this is important -- such spacing issues do not usually be demanded in reviews.

manpen:
Thank your review and merging!

PR #408, opened 2019-09-14
Fix compiler warnings and treat warnings as errors in CI

manpen:
We've come a long way in improving code quality and removing warnings. Unfortunately, we sometimes merge code to `Dev` which causes compiler warnings, or more recent compilers add additional warnings which remain unfixed for some time. 

This PR fixes exactly this type of warnings. Additionally it introduces the CMake flag `NETWORKIT_WARNINGS_AS_ERRORS` (default: OFF) and enables it during CI. As the name suggests, this treats compiler warnings as errors (with the exception of deprecation warnings).


avdgrinten:
This is a great idea. Indeed, it took a lot of effort to remove all warnings. Thank you for the patch, i will review it shortly.

manpen:
Due to merge conflicts, I need to rebase the PR and now had to fix warnings that were introduced in the mean time. I hope that CI now runs through.

angriman:
Did the rebase invalidate @avdgrinten approval?

manpen:
I believe so. @avdgrinten approved it on the 27th

PR #409, opened 2019-09-15
Fix VarInt and Zigzag coding of NKBG

manpen:
UBSAN was complaining about both encodings used in NKBG. As all related routines were private methods to the Reader and Writer, I first moved the implementation into the `nkgb` namespace and added unit tests -- checking that `x == decode(encode(x)`). Unfortunately both failed.

While zigzagEncode was very similar to [a GitHub gitst](https://gist.github.com/mfuerstenau/ba870a29e16536fdbaba), our implementation uses 64bit rather than 32bit integers leading to different constants. Additionally in C++ a left shift of signed ints is UB, while a right shift is implementation specific. I believe that my fix avoids both issues. Also the expression is more complicated GCC and CLANG yield the intended assembly of three instructions.

The implementation of the (de)serialization of varints had corner cases which lead to negative shifts. Also I do not understand why the number of bytes required was derived by dividing the number of bits by seven. As I did not find a documentation of the varint format, I assumed that the following format was intended:
The first byte stores the number of following bytes unary as the number of least significant zeros. This pattern is followed by a 1 (if n < 8). The potentially remaining bits are used to store the least significant bits of the value. The subsequent bytes then encode 8 bits of the value. I reimplemented both (de|en)coding routings to emit these structure.

Be advised that both fixes potentially break compatibility with existing files. I consider this a minor issue since the previous implementation was unable to consistently recover its own files. Should we bump up the magic code?

Two lessons learned:
- When ever we are in the comfortable situation to have both forward and inverse projects, we should add a unit test to check that their composition yields identity.
- We should pay attention to UBSAN messages. **Hence, this PR also treats ASAN and UBSAN warnings as errors during CI**.

(As a sidenote I also moved the `nkgb` namespace into the `NetworKit`)

### DIFF
--- a/include/networkit/io/NetworkitBinaryGraph.hpp
+++ b/include/networkit/io/NetworkitBinaryGraph.hpp
@@ -98,6 +98,19 @@ inline size_t varIntDecode(const uint8_t *data, uint64_t &value) noexcept {
 	return n + 1;
 }
 
+/**
+ * Encodes a singed value, s.t. the sign bit is in the LSB bit, while the absolute value is
+ * stored in the remaining upper 63 bits. This allows more effective compression using varint.
+ */
+inline uint64_t zigzagEncode(int64_t value) noexcept {
+	return (value << 1) ^ (value >> 31);
+}
+
+//! Reverses zigzagEncode.
+inline int64_t zigzagDecode(uint64_t value) noexcept {
+	return (value >> 1) ^ (-(value & 1));
+}
+
 } // namespace nkbg
 } // namespace NetworKit
 
--- a/include/networkit/io/NetworkitBinaryGraph.hpp
+++ b/include/networkit/io/NetworkitBinaryGraph.hpp
@@ -1,7 +1,21 @@
+/*
+ * NetworkitBinaryGraph.hpp
+ *
+ *      Author: Charmaine Ndolo <charmaine.ndolo@b-tu.de>
+ */
+
 #ifndef NETWORKIT_BINARY_GRAPH_HPP
 #define NETWORKIT_BINARY_GRAPH_HPP
 
+#include <cstdint>
+#include <cstddef>
+
+#include <tlx/math/clz.hpp>
+#include <tlx/math/ffs.hpp>
+
+namespace NetworKit {
 namespace nkbg {
+
 struct Header {
 	char magic[8];
 	uint64_t checksum;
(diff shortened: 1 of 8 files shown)


avdgrinten:
The intended varint encoding is exactly what you described. Thank you for the PR, I will submit a review shortly.

avdgrinten:
Regarding the division by 7 to calculate the number of storage bytes from the number of bits in the input, that was correct: in our encoding, a n-byte variant stores exactly 7n bits of information. That is because each additional byte reduces the amount of information in the first byte by one bit.

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -28,5 +42,77 @@ static constexpr uint64_t DIR_MASK = 0x1; // bit 0
...
+	}
+
+	const auto bits = 64 - tlx::clz(value);
+	auto dataBytes = bits / 8;
+
+	if (dataBytes < 8) {
+		// check whether the remaining value bits fit into remaining bits of buffer[0]
+		auto remain = bits % 8;
+		dataBytes += (remain + dataBytes >= 8);
+	}
```
avdgrinten: The computation of `dataBytes` can be simplified by dividing `bits` by 7, see my top-level comment.

manpen:
Yes, that makes sense, however we still need a special case for words with 64 bits. At least the computation is now branch-free ;)

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -28,5 +42,72 @@ static constexpr uint64_t DIR_MASK = 0x1; // bit 0
...
 
+/// Serializes value into a buffer and returns the number of bytes written.
+inline size_t varIntEncode(uint64_t value, uint8_t *buffer) noexcept {
+	if (!value) {
+		buffer[0] = 1;
+		return 1;
+	}
+
+	const auto bits = 64 - tlx::clz(value);
+	auto dataBytes = ((bits - 1) / 7) - (bits == 64);
```
avdgrinten: I think this is quite hard to read. The problem is that the <= 56 bit cases and the > 56 bit case (first byte is zero) are fundamentally different: when we go from an encoded size of 8 to an encoded size of 9, we suddenly gain 8 bits of information instead of 7 bits (because the first byte is already fully consumed at that point; we go from 0 (first byte which is already `0b10000000`) + 7x8 = 56 bits to 0 (first byte which is now 0) + 8x8 = 64 bits).

Why not branch on `if(bits <= 56)` first? This would allow us to calculate `dataBytes = (bits - 1) / 7` in the then-case and to just set `dataBytes = 8` in the else-case. This was presumably the intention of the old code, even though the if-condition was indeed broken.

avdgrinten:
Looks good now, is it ready to be merged?

manpen:
Thank you for your feedback. I'm not planning to do more changes; so yes, in my opinion it's ready.

avdgrinten:
Thank you for the patch and also for the source code comments explaining the expected behavior.

PR #411, opened 2019-09-24
Indent C++ with Spaces

manpen:
This is a rather large PR with only minor changes: It replaces Tab-indent C++ with spaces. Additionally (and this might be a little bit controversial) it provisions a infrastructure to check code quality. So in the future PRs that use tab-indents will fail during CI.

# How to deal with pending PRs
I suggest to merge this PR rather quickly. All PR which touch currently Tab-indent files will then have a merge conflict. To overcome this, the authors of the PR should run `extrafiles/tooling/CppIndentation.py -w` to fix indention and rebase.

# CI checks and fixes
IMHO we should have checks (and if possible) fixes for common structures, such as:
- Space indention
- Common header comments
- Consistent include guards
- (Auto-Format)
- Clang Tidy rules ...

So I added a small Python Lib `nktooling` which helps to find NetworKit sources.

It also suggests a strategy how to implement checks and fixes: We ask `nkt` for a list of files and open each file with the `FileRewriter` wrapper. It allows to iterate over lines and expects the user code to push back (un)changes lines. Afterwards a commit command can be issued to update the file (atomically). By default, however, `nkt` is in a read-only mode, and hence all fixes are simply ignored and we only get warnings about non-compliant files. All interactions with command-line args are encapsulated by `nkt`.

```python
import nktooling as nkt

nkt.setup()

noncompliantFiles = 0

for file in nkt.getCXXFiles():
	rw = nkt.FileRewriter(file)
	for line in rw.lines():
		# compute fix 
		rw.write(fixedLine)

	if not rw.isIdentical():
		noncompliantFiles += 1

	rw.commit()

if noncompliantFiles > 0:
	nkt.failIfReadonly(__file__)
```


extrafiles/tooling/nktooling/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+	Helper function indented to give a summary after a test in read-only mode.
+	In read-only mode, fixes could not be applied, hence, we fail with a negative
+	return code. This can be used during CI to check whether repository is compliant.
+	"""
+	if not isReadonly():
+		return
+
+	print("Changes are necessary.")
+	print("Use '%s -w' to apply changes." % prgname)
+	sys.exit(-1)
```
avdgrinten: Traditionally, UNIX programs exit with 1 on error.

extrafiles/tooling/nktooling/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+		If the module is currently readonly (which is the default, see setup) no change
+		is carried out.
+		This function may only be called if not reads are taking place currently (e.g.,
+		via lines())
+		"""
+		assert(not self.reading)
+		if isReadonly() and not force:
+			return
+
+		with tempfile.TemporaryDirectory() as dir:
```
avdgrinten: We create an entirely new directory for each file here. Why don't we just create a temporary file?

manpen: For me this looked like the cleanest approach (I do really think that the cost of creating a directory is insignificant compared to a Python parsing source code ;) to make sure
0.) we get a unique path
1.) the file is automatically cleaned up
2.) we can move the file out of it.

I'm aware only of one other variant: `NamedTemporaryFile`; however here we do [not get the guarantee](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) that the file can be (i.e., moved) while still being opened. Do you know a better alternative?

extrafiles/tooling/nktooling/__init__.py
manpen: Technically `-1` is at least 32 times stronger than `1` ;) But you're right. I'll fix it.

extrafiles/tooling/nktooling/__init__.py
avdgrinten: Hm, I was not aware of this limitation. In this case the temporary directory indeed looks like a good solution.

manpen:
Since this PR will conflict with merging of #408 and #414, I propose to merge the former two; then I will rebase this again and suggest to merge afterwards, s.t. no new conflicts will surface.

avdgrinten:
I merged the other two PRs.

manpen:
I just rebased and force-pushed. There's one tiny change to the code: The temporary directory for the `FileRewriter` is now created in `networkit`'s root path. The reason for this is that sometimes `/tmp` is mounted as a different filesystem. Then a move is not atomic anymore.

manpen:
The PR now also fixed the unused warning in `Graph.cpp:860` currently breaking our CI.

avdgrinten:
Alright. I will wait for the CI, merge this PR and rebase my other PRs afterwards.

PR #413, opened 2019-09-26
Handle multiple self-loops removal

angriman:
This PR fixes a bug in `removeMultiEdges`: If a node has multiple self-loops, and the graph is undirected, `removeMultiEdges` counts self-loop removals twice, and therefore the edge count is updated wrongly.

I dislike that to do this fix the return value of `removeAdjacentEdges` is changed because we need to distinguish between removed non-self-loop edges and removed self-loops. If we want to avoid this, I am open to suggestions.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1086,11 +1086,11 @@ public:
      * @param condition A function that takes a node as an input and returns a
      * bool. If true the edge (u, v) is removed.
      * @param edgesIn Whether in-going or out-going edges shall be removed.
-     * @return count The number of removed edges.
+     * @return std::pair<count, count> The number of removed edges (first) and the number of removed
+     * self-loops (second).
      */
     template <typename Condition>
-    count removeAdjacentEdges(node u, Condition condition, bool edgesIn = false);
-
+    std::pair<count, count> removeAdjacentEdges(node u, Condition condition, bool edgesIn = false);
 
     /**
      * Removes all self-loops in the graph.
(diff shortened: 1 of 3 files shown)


networkit/cpp/graph/Graph.cpp
```diff
@@ -828,41 +828,42 @@ void Graph::removeEdgesFromIsolatedSet(const std::vector<node> &nodesInSet) {
...
 }
 
 void Graph::removeMultiEdges() {
-	count nEdgesRemoved = 0;
+	count removedEdges = 0;
+	count removedSelfLoops = 0;
 	std::unordered_set<node> nodes;
 
 	forNodes([&](node u) {
 		auto isMultiedge = [&nodes](node v) { return !nodes.insert(v).second; };
```
manpen: add `nodes.reserve(degree(u))` ? not sure whether this helps

include/networkit/graph/Graph.hpp
```diff
@@ -2129,12 +2129,17 @@ template <typename L> void Graph::DFSEdgesFrom(node r, L handle) const {
 /* EDGE MODIFIERS */
 
 template <typename Condition>
-count Graph::removeAdjacentEdges(node u, Condition condition, bool edgesIn) {
+std::pair<count, count> Graph::removeAdjacentEdges(node u, Condition condition, bool edgesIn) {
 	count removedEdges = 0;
+	count removedSelfLoops = 0;
 	auto &edges_ = outEdges[u];
```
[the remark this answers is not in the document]
angriman: In the directed case we should call `removeAdjacentEdges` twice as you wrote. However, the current implementation is wrong, if `edgesIn = true` we should iterate over `inEdges`. I'll fix this and add some comments.

networkit/cpp/graph/Graph.cpp
angriman: Probably it is a bit more efficient, I'll do that.

angriman:
Thanks for the quick review. I addressed all the comments.

manpen:
While there's a small case of mixed tab/space indentation (line 2141 in Graph.cpp), it will be fixed as soon as I rebase #411, the changes look good to me. 

As it's a bug-fix I would fast track it, but let's leave it open until Monday, just in case somebody spots something?!

PR #414, opened 2019-09-28
Clean up include dependencies

manpen:
This PR moves `networkit/ext/ttmath` into `extlibs/ttmath`. The reason for doing so is that `ttmath` is not actively maintained by us and used as an external library. Moving it out of networkit's file tree simplifies tools and checks that target only the actively maintained files.

As `ttmath` is only rarely used (primarily by the `distance` module), it does not really make sense to include it in `Globals.h`; this is esp. true given that `ttmath` pulls iostream etc. I rather added the wrapper file `auxiliary/Multiprecision.hpp` which defines `bigfloat` and pulls in `ttmath`. This header is only loaded by files actually requiring it.

Unfortunately `iostream` was still included in every build due to a number of unnecessay `#include`s in header files (e.g., `Graph.hpp` pulled `Log.hpp` which effectively pulled `iostreams`). These deps are now removed. As a result, I had to add a number of includes which were previously indirectedly loaded (this accounts for most of the changes in the PR).

As a result the compilation time of the networkit core module went down from 38s to 30s on my machine. There's almost no effect for the tests, as GTest has a lot deps itself.


angriman:
Would that be better to remove the ttmath library entirely? It looks like that the only usage is to count paths with the `bigfloat` type. Is `double` really not enough?

avdgrinten:
`double` is not really an alternative. However, `bigfloat` is not the correct type to store numbers of paths either. If anything, we want a big integer. One thing to consider is that in an exact representation and in the worst-case, the size of the numbers grows linearly in the graph size. Thus, storing them for large graphs is not really feasible - it requires quadratic amounts of memory.

manpen:
I presume @avdgrinten's argument is the very reason a floating points are used. Observe that `ttmath` uses stack-allocation only (in contrast to, e.g., `gmp`). Hence we have a fixed size determined at compile-time. `bigfloat` is configured to use a 64-bit mantissa and a 64-bit exponent, giving it a (lossy range) ~10^18 digits -- compared to ~300 digits for `double`.

Coming back to the original suggesting of removing `ttmath` altogether: I do not have a strong opinion about that and consider this out of scope of this PR. This PR gives the structure to allow removal by simply modifying all modules using `auxiliary/Multiprecision.hpp`; but it does much more.

PR #421, opened 2019-10-05
Graph: deprecate and move methods to GraphTools

angriman:
This PR deprecates several methods of `Graph`, and copies some of them to `GraphTools`.
In particular:
- `getId`,  `typ`, `set/getName`, `toString`, `nodes`, `edges`, `time/timeStep`, and `volume`  have been deprecated but **not** copied to `GraphTools` because we should drop those features in the future (or move them somewhere outside `Graph`, if there is some good reason to do so).
Since `volume` is a particular case of weighted degree, ad additional parameter has been added to `weightedDegree` to also support volume calculation.
- `copyNodes`, `subgraphFromNodes`, `transpose`, `BFS/DFSfrom`, `toUnweighted/Undirected`, `append`, `merge`, `maxDegree`, `maxWeightedDegree`, `randomNode/Edge/Edges`, `removeEdgesFromIsolatedSet`, `size`, `density` have been deprecated **and** copied to `GraphTools`. All NetworKit classes using them now use the implementation in `GraphTools`.

Further changes are:
- Parametrized `GraphToolGTest` to deal with all combinations of directed/weighted graphs.
- The version of `BFSfrom` that accepts multiple source nodes now takes iterators as input parameters, so it can be invoked using an arbitrary container.
- New function `GraphTools::toWeighted`.
- New `unsafe` function `Graph::setOmega`. This is needed by `GraphTools::transpose` because `addPartialIn/OutEdge` does not affect `omega` (i.e., `upperEdgeIdBound`).
- Refactor `merge` implementation so it can handle merge operations of a graph with a larger one.
- C++ (where missing) and Python tests for all the functions moved to `GraphTools`.
- Clang-format of heavily modified files

### DIFF
--- a/include/networkit/graph/BFS.hpp
+++ b/include/networkit/graph/BFS.hpp
@@ -1,0 +1,109 @@
+#ifndef GRAPH_BFS_H
+#define GRAPH_BFS_H
+
+#include <queue>
+#include <vector>
+
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+namespace Traversal {
+
+/**
+ * Calls the given BFS handle with distance parameter
+ */
+template <class F>
+auto callBFSHandle(F &f, node u, count dist) -> decltype(f(u, dist)) {
+    return f(u, dist);
+}
+
+/**
+ * Calls the given BFS handle without distance parameter
+ */
+template <class F>
+auto callBFSHandle(F &f, node u, count) -> decltype(f(u)) {
+    return f(u);
+}
+
+/**
+ * Iterate over nodes in breadth-first search order starting from the nodes within the given range.
+ *
+ * @param G The input graph.
+ * @param first The first element of the range.
+ * @param last The end of the range.
+ * @param handle Takes a node as input parameter.
+ */
+template <class InputIt, typename L>
+void BFSfrom(const Graph &G, InputIt first, InputIt last, L handle) {
+    std::vector<bool> marked(G.upperNodeIdBound());
+    std::queue<node> q, qNext;
+    count dist = 0;
+    // enqueue start nodes
+    for (; first != last; ++first) {
+        q.push(*first);
+        marked[*first] = true;
+    }
+    do {
+        const auto u = q.front();
+        q.pop();
+        // apply function
+        callBFSHandle(handle, u, dist);
+        G.forNeighborsOf(u, [&](node v) {
+            if (!marked[v]) {
+                qNext.push(v);
+                marked[v] = true;
+            }
+        });
+        if (q.empty() && !qNext.empty()) {
+            q.swap(qNext);
+            ++dist;
+        }
+    } while (!q.empty());
+}
+
+/**
+ * Iterate over nodes in breadth-first search order starting from the given source node.
+ *
+ * @param G The input graph.
+ * @param source The source node.
+ * @param handle Takes a node as input parameter.
+ */
+template <typename L>
+void BFSfrom(const Graph &G, node source, L handle) {
+    std::vector<node> startNodes({source});
+    BFSfrom(G, startNodes.begin(), startNodes.end(), handle);
+}
+
+/**
+ * Iterate over edges in breadth-first search order starting from the given source node.
+ *
+ * @param G The input graph.
+ * @param source The source node.
+ * @param handle Takes a node as input parameter.
+ */
+template <typename L>
+void BFSEdgesFrom(const Graph &G, node source, L handle) {
+    std::vector<bool> marked(G.upperNodeIdBound());
+    std::queue<node> q;
+    q.push(source); // enqueue root
+    marked[source] = true;
+    do {
+        const auto u = q.front();
+        q.pop();
+        // apply function
+        G.forNeighborsOf(u, [&](node, node v, edgeweight w, edgeid eid) {
+            if (!marked[v]) {
+                handle(u, v, w, eid);
+                q.push(v);
+                marked[v] = true;
+            }
+        });
+    } while (!q.empty());
+}
+
+} // namespace Traversal
+
+} // namespace NetworKit
+
+#endif
--- a/include/networkit/graph/BFS.hpp
+++ b/include/networkit/graph/BFS.hpp
@@ -1,5 +1,7 @@
-#ifndef GRAPH_BFS_H
-#define GRAPH_BFS_H
+// networkit-format
+
+#ifndef NETWORKIT_GRAPH_BFS_HPP_
+#define NETWORKIT_GRAPH_BFS_HPP_
 
 #include <queue>
 #include <vector>
@@ -106,4 +108,4 @@ void BFSEdgesFrom(const Graph &G, node source, L handle) {
 
 } // namespace NetworKit
 
-#endif
+#endif // NETWORKIT_GRAPH_BFS_HPP_
--- a/include/networkit/graph/DFS.hpp
+++ b/include/networkit/graph/DFS.hpp
@@ -1,5 +1,7 @@
-#ifndef GRAPH_DFS_H
-#define GRAPH_DFS_H
+// networkit-format
+
+#ifndef NETWORKIT_GRAPH_DFS_HPP_
+#define NETWORKIT_GRAPH_DFS_HPP_
 
 #include <stack>
 #include <vector>
@@ -67,4 +69,4 @@ void DFSEdgesFrom(const Graph &G, node source, L handle) {
 } // namespace Traversal
 
 } // namespace NetworKit
-#endif
+#endif // NETWORKIT_GRAPH_DFS_HPP_
--- a/include/networkit/graph/DFS.hpp
+++ b/include/networkit/graph/DFS.hpp
@@ -1,0 +1,70 @@
+#ifndef GRAPH_DFS_H
+#define GRAPH_DFS_H
+
+#include <stack>
+#include <vector>
+
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+namespace Traversal {
+
+/**
+ * Iterate over nodes in depth-first search order starting from the given source node.
+ *
+ * @param G The input graph.
+ * @param source The source node.
+ * @param handle Takes a node as input parameter.
+ */
+template <typename L>
+void DFSfrom(const Graph &G, node source, L handle) {
+    std::vector<bool> marked(G.upperNodeIdBound());
+    std::stack<node> s;
+    s.push(source); // enqueue root
+    marked[source] = true;
+    do {
+        const auto u = s.top();
+        s.pop();
+        // apply function
+        handle(u);
+        G.forNeighborsOf(u, [&](node v) {
+            if (!marked[v]) {
+                s.push(v);
+                marked[v] = true;
+            }
+        });
+    } while (!s.empty());
+}
+
+/**
+ * Iterate over edges in depth-first search order starting from the given source node.
+ *
+ * @param G The input graph.
+ * @param source The source node.
+ * @param handle Takes a node as input parameter.
+ */
+template <typename L>
+void DFSEdgesFrom(const Graph &G, node source, L handle) {
+    std::vector<bool> marked(G.upperNodeIdBound());
+    std::stack<node> s;
+    s.push(source); // enqueue root
+    marked[source] = true;
+    do {
+        const auto u = s.top();
+        s.pop();
+        // apply function
+        G.forNeighborsOf(u, [&](node, node v, edgeweight w, edgeid eid) {
+            if (!marked[v]) {
+                handle(u, v, w, eid);
+                s.push(v);
+                marked[v] = true;
+            }
+        });
+    } while (!s.empty());
+}
+
+} // namespace Traversal
+
+} // namespace NetworKit
+#endif
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -845,6 +845,38 @@ public:
     void removeNode(node v);
 
     /**
+     * Removes out-going edges from node @u. If the graph is weighted and/or has edge ids, weights and/or edge ids will also be removed.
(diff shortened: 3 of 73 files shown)


hmeyerhenke:
Please remember that deprecation is a delicate thing for users and external developers. Make such big changes only for very good reasons. Also consider to make big changes obvious by changing the primary version number in the release.

avdgrinten:
This is "only" deprecation and not removal: the functionality will remain in place for at least one release.

hmeyerhenke:
@avdgrinten: Sure, but eventually this will break someone's code. Regarding the version number: the major number should of course change when the break actually happens (removal).

angriman:
I fixed the last warning that caused the build to break, hopefully the CI will run successfully now.
About graph functions using coordinates: they have been deprecated long ago and I think that they could have been removed in this PR. However, this requires to modify some other NetworKit classes that still use them, and would have been too much for a single PR.
Also, I think that @manpen was already working on that, or am I wrong?

angriman:
This PR can be reviewed, for some reason travis did not update the check status even if the build succeeded.

manpen:
Yes, I already looked at how to remove coordinate related methods, and while it's a little bit tedious to do, I really think it's worth it: The storage of coordinates is not related to Graph's logic. In fact it's not really supported, as coordinates often get lost e.g. when (implicitly) copying a graph. 
However, if we are going to remove more methods (as suggested by this PR), I would postpone this change until after we published version 6.0.

include/networkit/graph/Graph.hpp
```diff
@@ -782,34 +782,47 @@ class Graph final {
...
      * @param edges the edge count of a graph
      */
     void setEdgeCount(Unsafe, count edges) { m = edges; }
 
+    /**
+     * Set upper bound to the edge count of the graph to edges.
+     *
+     * @param newOmega the new upper bound to the edge count of the graph.
+     */
+    void setOmega(Unsafe, edgeid newOmega) noexcept { omega = newOmega; }
```
manpen: Rename to `setUpperEdgeIdBound`. Nobody should know what `Omega` is and we already have a getter named `upperEdgeIdBound`.

include/networkit/graph/GraphTools.hpp
```diff
@@ -97,6 +184,146 @@ Graph getRemappedGraph(const Graph& graph, count numNodes,
...
+
+/**
+ * Iterate over the nodes of the input graph in breadth-first search order starting from @a source
+ * until the connected component of @a source has been visited.
+ *
+ * @param G The input graph.
+ * @param source Source node of the breadth-first search.
+ * @param handle Callback function, takes one parameter <code>(node)</code>.
+ */
+template <typename L> void BFSfrom(const Graph &G, node source, L handle) {
```
manpen: Should we put BFS and DFS into dedicated headers (e.g. `graph/DFS.hpp` and `graph/BFS.hpp`)? To me it seems that they are build-blocks with a well-known name and somehow do not fit into the 'heap of algorithms that didn't fit in somewhere else'

manpen: Also, if we really try to avoid the callback idiom in the future, we might have to add BFS/DFS iterators, which will unnecessarily blow up GraphTools.

angriman: That makes sense, we might want to use a different name since `distance/BFS.hpp` already exist.

angriman: So I moved BFS and DFS in separated headers, and defined a `GraphTraversal` namespace.

avdgrinten:
> However, if we are going to remove more methods (as suggested by this PR), I would postpone this change until after we published version 6.0.

I would like to remove the coordinate functions now (in the next release) and deprecate the other functions at the same time. Further postponing the removal of the coordinate functions is not really beneficial IMHO.

include/networkit/graph/BFS.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+namespace NetworKit {
+
+namespace GraphTraversal {
+
+/**
+ * Calls the given BFS handle with distance parameter
+ */
+template <class F>
+auto callBFSHandle(F &f, node u, count dist) -> decltype(f(u, dist)) {
```
manpen: I just implemented a new `GraphTools` algortihm that needs access to the same `edgeLambda` dispatcher functionality used by Graph itself. Hence I factored out all those dispatcher functions into an own header. I suggest that we keep it here as is, but refactor it as soon as this PR is merged.

include/networkit/graph/BFS.hpp
```diff
@@ -0,0 +1,108 @@
...
+    std::queue<node> q;
+    q.push(source); // enqueue root
+    marked[source] = true;
+    do {
+        const auto u = q.front();
+        q.pop();
+        // apply function
+        G.forNeighborsOf(u, [&](node, node v, edgeweight w, edgeid eid) {
+            if (!marked[v]) {
+                handle(u, v, w, eid);
```
manpen: Here's another case that would benefit from a central dispatcher header. Keep it, and let's support all overloads with the an dedicated PR.

include/networkit/graph/DFS.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Iterate over nodes in depth-first search order starting from the given source node.
+ *
+ * @param G The input graph.
+ * @param source The source node.
+ * @param handle Takes a node as input parameter.
+ */
+template <typename L>
+void DFSfrom(const Graph &G, node source, L handle) {
```
manpen: If we keep the functionality as is (see below), this is only a special case of `DFSEdgesFrom` and we should remove one implementation.

However, it's somewhat strange that BFS supports to report the depth of a node (yes, I know it's unweighted SSSP), while DFS does not. I cannot think of an application right now, but it seems strange to have different callback signatures for these dual algorithms; esp. when giving the depth comes basically for free.

include/networkit/graph/DFS.hpp
```diff
@@ -0,0 +1,69 @@
+#ifndef GRAPH_DFS_H
+#define GRAPH_DFS_H
+
+#include <queue>
```
manpen: Should be `<stack>` please include `<vector>`

include/networkit/graph/BFS.hpp
```diff
@@ -0,0 +1,108 @@
+#ifndef GRAPH_BFS_H
+#define GRAPH_BFS_H
+
+#include <stack>
```
manpen: Should be `<queue>`; please include `<vector>`

include/networkit/graph/Graph.hpp
```diff
@@ -989,7 +1004,7 @@ class Graph final {
      * @param v Node.
      * @return The volume of the @a v.
      */
-    edgeweight volume(node v) const;
+    edgeweight TLX_DEPRECATED(volume(node v) const);
```
manpen: I must admit that I didn't know the definition of an node's volume (but in my defence and more to the point, it doesn't appear in [Wikipedia's Glossary of graph theory terms](https://en.wikipedia.org/wiki/Glossary_of_graph_theory_terms)).  If we move the function anyhow, can we please rename it to something more verbose, e.g. `totalEdgeWeights`, `edgeWeightSum`, ... ?

avdgrinten:
I think the reason @angriman decided not to call GraphTools from Graph is that this introduces a circular dependency (?).

manpen:
I'd just add `from _NetworKit import GraphTools` to `networkit/__init__.py`. If there's an issue with that, can you please elaborate?

include/networkit/graph/Graph.hpp
manpen: I just realized that there's already a `Graph::computeWeightedDegree`. Can't we just add the flag `countSelfloopTwice`?

avdgrinten:
In Python, that works. In C++, however, calling GraphTools from Graph introduces a circular dependency when doing a modularized build of NetworKit (and circular dependencies of `.so` files do not really work).

manpen:
I'm sorry. I misunderstood your initial comment. It was targeted at my request to remove duplicate code, right? In this case: I believe it's not an issue, since while `Graph.cpp` and `GraphTools.cpp` are build in different compilation units they are always linked into the same library -- even in non-monolithic builds.

manpen:
Regarding the last commit: In my opinion methods such as `degree`, `maxDegree` etc. should be part of the Graph object. I'd suggest to have an easy guideline that explains (esp. to users) what type of functionality can be expected at which point. 

One attempt:
With the exception of `volume` (see my comments above), currently `GraphTools` contains functions that create a (modified) copy of the (sub)graph. Contains `Graph` functions to query the current network, and to alter it in-place.

angriman:
I moved `maxDegree` because it is something that can be computed externally with the same efficiency, and does not require any of the internal `Graph` APIs. However, I do not have a strong opinion about this and I am also fine with keeping it into `Graph`.

include/networkit/graph/Graph.hpp
angriman: Yes, now `volume` returns `weightedDegree` with self-loops counted twice.

angriman:
Thanks for the feedback, I did the following changes:
- added warnings on deprecated methods that state if and where they are replaced
- changed the original Graph implementation with the GraphTools implementation (where possible)
- removed unit tests of deprecated methods in `GraphGTest`
- moved `GraphTools` into NetworKit's Python namespace (because all those names are lowercase, I opted for `nk.graphtools` rather than `nk.GraphTools`)
- renamed `GraphTraversal` to `Traversal`

angriman:
Do you think there are other methods that we should move to `GraphTools`? For example, `randomNode/Edge/Neighbor` and `density`.

avdgrinten:
Let's merge the removal of coordinates (#434) first. Afterwards, we rebase this PR. It should still go into the upcoming NetworKit 5.1.

avdgrinten:
> Regarding the last commit: In my opinion methods such as `degree`, `maxDegree` etc. should be part of the Graph object. I'd suggest to have an easy guideline that explains (esp. to users) what type of functionality can be expected at which point.
> 
> One attempt:
> With the exception of `volume` (see my comments above), currently `GraphTools` contains functions that create a (modified) copy of the (sub)graph. Contains `Graph` functions to query the current network, and to alter it in-place.

I am not sure if the in-place/copy distinction is a good one. I would like to keep the "basic"/minimal function (e.g. `degree` etc.) in `Graph` but move the "auxiliary" functions (e.g. `maxDegree` or `volume`) to `GraphTools`.

avdgrinten:
If we disregard the `forSomething` iterators, my suggestion is to also deprecate `maxDegree`/`maxDegreeIn` (rename to `maxInDegree` when moving to GraphTools while we're at it), `maxWeightedDegree`/`maxWeightedDegreeIn`, `randomNode`, `randomNeighbor`, `removeEdgesFromIsolatedSet` (might require new `Unsafe` methods for an efficient implementation), `removeAllEdges`, `removeSelfLoops`, `removeMultiEdges`, `swapEdge`, `randomEdge`, `randomEdges`, `size`, `density`.

I think it would be justifiable to keep `isIsolated`, `weightedDegree`/`weightedDegreeIn`, `totalEdgeWeight`.

`getIthNeighbor` should be reconsidered in the future.

angriman:
> If we disregard the `forSomething` iterators, my suggestion is to also deprecate `maxDegree`/`maxDegreeIn` (rename to `maxInDegree` when moving to GraphTools while we're at it), `maxWeightedDegree`/`maxWeightedDegreeIn`, `randomNode`, `randomNeighbor`, `removeEdgesFromIsolatedSet` (might require new `Unsafe` methods for an efficient implementation), `removeAllEdges`, `removeSelfLoops`, `removeMultiEdges`, `swapEdge`, `randomEdge`, `randomEdges`, `size`, `density`.
> 
> I think it would be justifiable to keep `isIsolated`, `weightedDegree`/`weightedDegreeIn`, `totalEdgeWeight`.
> 
> `getIthNeighbor` should be reconsidered in the future.

@manpen do you agree?

networkit/cpp/graph/Graph.cpp
```diff
@@ -496,22 +498,36 @@ edgeweight Graph::computeMaxWeightedDegree(const bool inDegree) const {
...
+            forNeighborsOf(u, sumWeights);
         }
         return sum;
     }
-    return defaultEdgeWeight * (inDegree ? degreeIn(v) : degreeOut(v));
+
+    count sum = inDegree ? degreeIn(u) : degreeOut(u);
+    auto countSelfLoops = [&](node v) { sum += (u == v); };
+
+    if (countSelfLoopsTwice) {
```
michitux: It might be a good idea to add a (constant time) check if there are any self-loops (in the whole graph) here.

angriman: That's a good idea, thank you. I've included it in the PR.

angriman:
The coding style compliance fails but if I run `check_code` locally it does not complain. How can this happen?

manpen:
`check_code.sh` also does not find a bug on my local machine. I'll look into it now. My current guess: there different clang-format versions used.

avdgrinten:
As a side note: we could make `check_code.sh` output a diff after `-w` to handle situations like this one in the future.

manpen:
Yes, this is exactly what I'm currently implementing ;)

manpen:
It is indeed an issue with the clang-format version; I'll open a different PR with an improvement. As a quick fix please replace to following lines (282/283) in `.travis.yml`
```
-         packages:
-            clang-format-7
```

with 

```
        apt:
          sources: ['llvm-toolchain-bionic-8']
          packages: ['clang-format-8']
```

angriman:
Thank you @manpen for the patch, it fixed the issue.

include/networkit/graph/DFS.hpp
angriman: That's true, we might want to refactor this in a separate PR, the current one is already quite big.

PR #423, opened 2019-10-07
Fix: move out parameters

angriman:
Some NetworKit classes provide a `moveOut` parameter that, if set true, should move the content of a vector outside the class. However, as @avdgrinten commented in https://github.com/networkit/networkit/pull/334#discussion_r292924992, move-out parameters cannot work as expected.

This PR deprecates all methods using `moveOut`, and provides alternative methods that return a const ref. to the vector.

### DIFF
--- a/networkit/cpp/algebraic/algorithms/test/AlgebraicTriangleCountingGTest.cpp
+++ b/networkit/cpp/algebraic/algorithms/test/AlgebraicTriangleCountingGTest.cpp
@@ -156,7 +156,7 @@ TEST(AlgebraicTriangleCountingGTest, testLocalClusteringCoefficient) {
     LocalClusteringCoefficient lcc(graph);
     timer.start();
     lcc.run();
-    std::vector<double> lccValues = lcc.scores(true);
+    std::vector<double> lccValues = lcc.scores();
     timer.stop();
 
     INFO("Graph theoretic local clustering coefficient took ", timer.elapsedTag());
--- a/networkit/cpp/algebraic/algorithms/test/AlgebraicTriangleCountingGTest.cpp
+++ b/networkit/cpp/algebraic/algorithms/test/AlgebraicTriangleCountingGTest.cpp
@@ -27,7 +27,7 @@ TEST(AlgebraicTriangleCountingGTest, testToyGraphOne) {
     AlgebraicTriangleCounting<CSRMatrix> atc(graph);
     atc.run();
 
-    std::vector<count> nodeScores = atc.getScores(true);
+    std::vector<count> nodeScores = atc.getScores();
 
     EXPECT_EQ(1u, nodeScores[0]) << "wrong triangle count";
     EXPECT_EQ(1u, nodeScores[1]) << "wrong triangle count";
@@ -50,8 +50,7 @@ TEST(AlgebraicTriangleCountingGTest, testToyGraphTwo) {
     atc.run();
 
     EXPECT_TRUE(atc.hasFinished());
-    std::vector<count> nodeScores = atc.getScores(true);
-    EXPECT_FALSE(atc.hasFinished());
+    std::vector<count> nodeScores = atc.getScores();
 
     EXPECT_EQ(1u, nodeScores[0]) << "wrong triangle count";
     EXPECT_EQ(1u, nodeScores[1]) << "wrong triangle count";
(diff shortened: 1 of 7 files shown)


manpen:
I presume that the original indent of `moveout` was to acquire the score vector very fast, e.g. to make it outlive the algorithm. Users would accept that the algorithm itself is in an undefined state afterwards.
If we return a **const** reference, this cannot be done. So IF we want to allow this pattern, the there also should be non-const ref accessors.

avdgrinten:
I'm not sure if that functionality is really needed. I'm not against a moveOut version in principle, but we should definitely also have a `const &` accessor. Right now we are pessimizing the common case (by returning a copy). I don't think a non-const `&` accessor is a good idea either. If we really want a moveOut version, I would suggest something like `std::vector<double> moveScores();` (or `consumeScores()`, `takeScores()`, the name is up for bikeshedding). @angriman do we have any internal users that would benefit from such a function?

What certainly does not work is the current *implementation*:
```cpp
return moveOut? std::move(nodeScores) : nodeScores;
```
This just always copies (maybe due to a misunderstanding of how `std::move` works by the original author (?)).

angriman:
There is no internal usage for this function, apart from some tests that were not even testing that the vector was actually moved out (they were removed in this PR). However, there might exist some use cases for this functionality. As @avdgrinten suggests we could provide additional `moveScores` methods.

avdgrinten:
If there are no internal users, I suggest that we wait with adding such functionality until actual internal/external users appear.

manpen:
That's a classical *chicken or the egg* scenario. I'm coming across this right now while removing the coordinates-functionality from `Graph` and suggest we handle it consistently throughout the code base.

 The `DibapGraphReader` currently returns a graph with (optional) coordinates. While returning a `std::pair<Graph, std::vector<coord>>` would be an option, it does not play nice with the class hierarchy and the interface defined by `GraphReader`. Hence I added the const-ref accessor `const std::vector<Point<coord>>& getCoordinates();`.

To avoid copying the info out, we've two options:
``` c++
std::vector<Point<coord>>& getCoordinates(); // non-const ref
std::vector<Point<coord>>&& moveCoordinates(); // "destructive" rvalue ref
```

I'd prefer the rvalue-variant as it clearly say: This function is meant to extra values -- it's not inviting users to keep ownership with the reader, but manipulating the values as they please.

avdgrinten:
Why not just a `std::vector<Point<coord>> moveCoordinates()` that really moves the coordinates out? I'd consider this to be cleaner as it does not allow arbitrary modification of the algorithm's internal vector.

manpen:
I do not see an option in which `std::vector<Point<coord>>&& moveCoordinates();` would allow to modify the internal vector (other than clearing it). I suggested returning an rvalue ref to avoid the construction of a temporary. However, on second thought (and some experiments on compiler explorer) RVO seems to work here consistently.

So, as far as I can see the only differences between 
```c++
std::vector<Point<coord>>&& moveCoordinates() {return std::move(coordinates);} // 1
std::vector<Point<coord>>   moveCoordinates() {return std::move(coordinates);} // 2
```
is that (2) will definitely leave `coordinates` in an empty state, while (1) only might (e.g., if the return value is discarded, `coordinates` should remain unchanged). Therefore I consider (2) is in fact the cleaner solution.

avdgrinten:
With (1), you can turn the rvalue reference into a normal reference and manipulate the internal vector.

networkit/cpp/algebraic/algorithms/test/AlgebraicTriangleCountingGTest.cpp
```diff
@@ -27,7 +27,7 @@ TEST(AlgebraicTriangleCountingGTest, testToyGraphOne) {
     AlgebraicTriangleCounting<CSRMatrix> atc(graph);
     atc.run();
 
-    std::vector<count> nodeScores = atc.getScores(true);
+    std::vector<count> nodeScores = atc.getScores();
```
manpen: Theoretically one could use a `const auto& nodeScores =`. But since the test instance is quite small that's not strictly necessary.

manpen:
@avdgrinten I was not aware of this behaviour and assumed this would move out the vector at the same time; but you're right. Thanks for pointing that out! Then returning an rvalue ref is out of the question.

PR #424, opened 2019-10-07
logging: Always log to `stderr`

avdgrinten:
Logging to `stdout` is annoying, especially if `stdout` is piped to a consumer that programmatically extracts information from the output.


angriman:
Looks reasonable to me, I think we can merge this as soon as @avdgrinten rebases on Dev.

avdgrinten:
This PR is rebased. One test fails due to an unrelated Travis issue (macOS fails to install GCC 8).

PR #426, opened 2019-10-07
NetworKit binary graphs: Fix weight auto detection for unweighted graphs

avdgrinten:
The current auto detection code chooses the positive varint storage format for unweighted graphs. Fix the detection to not store weights at all.

### DIFF
--- a/networkit/cpp/io/NetworkitBinaryWriter.cpp
+++ b/networkit/cpp/io/NetworkitBinaryWriter.cpp
@@ -24,8 +24,10 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
     Aux::enforceOpened(outfile);
     nkbg::WEIGHT_FORMAT weightFormat;
 
-    auto detectWeightsType = [&] () {
-        weightFormat = nkbg::WEIGHT_FORMAT::VARINT;
+    auto detectWeightsType = [&] () -> nkbg::WEIGHT_FORMAT {
+        if(!G.isWeighted())
+            return nkbg::WEIGHT_FORMAT::NONE;
+
         bool isUnsigned = true;
         bool fitsIntoInt64 = true;
         bool fitsIntoFloat = true;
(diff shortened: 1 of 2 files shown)


networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -24,8 +24,10 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
...
 
-	auto detectWeightsType = [&] () {
-		weightFormat = nkbg::WEIGHT_FORMAT::VARINT;
+	auto detectWeightsType = [&] () -> nkbg::WEIGHT_FORMAT {
+		if(!G.isWeighted())
+			return nkbg::WEIGHT_FORMAT::NONE;
+
 		bool isUnsigned = true;
 		bool fitsIntoInt64 = true;
 		bool fitsIntoFloat = true;
```
manpen: While you didn't touch this code: Shouldn't this be `G.forEdges` rather than `G.forNodes .. G.forNeighborsOf` ? It seems to be the faster and more idiomatic choice.

manpen: Also we can make the loop practically branch-free: https://godbolt.org/z/S8WwPd

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -826,6 +828,8 @@ TEST_F(IOGTest, testNetworkitBinaryKonect) {
 
 	NetworkitBinaryReader reader;
 	Graph G2 = reader.read("output/binary_konect");
+	EXPECT_EQ(G2.isDirected(), true);
+	EXPECT_EQ(G2.isWeighted(), true);
 	ASSERT_EQ(G2.numberOfEdges(), G.numberOfEdges());
 	ASSERT_EQ(G2.numberOfNodes(), G.numberOfNodes());
 	G.forNodes([&](node u){
```
manpen: IMHO tests should not only check whether the edges exist, but also whether their weights match.

networkit/cpp/io/NetworkitBinaryWriter.cpp
avdgrinten: I agree about the `G.forEdges`.

About the branch-free version: It's not immediately clear that the branch-free code is faster. You trade a control flow dependency for a data flow dependency.

avdgrinten:
I finally updated this and fixed the comments (after finding out about the `-X ignore-all-space` merging strategy :smiley:).

networkit/cpp/io/NetworkitBinaryWriter.cpp
fabratu: Though that is clearly not enough, but I did some benchmarks with the writer in both loop variations for synthetic networks (order of 10 million nodes, random negative and positive float-values). For this, the results indicate a slight advantage in the "original" version with branches

fabratu:
LGTM.

Apart from the discussion about the loop (added a comment there) every change request seems to be resolved. @manpen Can you approve/deny?

avdgrinten:
This is an important change, it should go into the next release.

PR #428, opened 2019-10-08
Added CMake as a requirement

paluskapter:
Added CMake to the requirements.txt as it is needed, and listed in the [README](https://github.com/networkit/networkit#requirements).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+cmake
 seaborn
 sklearn


manpen:
Thank you for your PR. When we switched to CMake, we discussed to not automatically install `cmake` and `cython` as these may "leak" into system's binary directory. I still think it's better to not install it automatically, however, would recommend adding installation hints in `setup.py`; e.g. change 
"ERROR: NetworKit compilation requires cmake." to "ERROR: NetworKit compilation requires cmake. You may install it with `pip3 install cmake`.".

Is there a specific scenario in which this is an issue for you?

paluskapter:
I wanted to use the tool in a cloud environment (Databricks AWS).
When I add networkit as a library to install, it fails, as CMake is not in the requirements.txt.
So CMake must be installed manually every time (can be handled with an init script, but it's still very hacky).
That's why I believe it might be a good idea to include it.

manpen:
@avdgrinten, @angriman: Do you have an opinion on this matter?

I still don't like the fact that if `pip` is run as root and installs `cmake`, pip's version can shadow the system's one. This can also lead to issues where a modern (pip's)  `cmake` binary uses old (system's) `cmake` libraries.

If `cmake` is added as an dependency so should `ninja` and `cython`.

@paluskapter We could also provide a `full_requirements.txt` file that includes such packages. Would that be compatible with your workflow?

avdgrinten:
I agree with your assessment, Manuel. Installing `cmake` and `ninja` binaries in a global location `/usr/bin/local` is almost certainly unexpected and could to lead to subtle problems on user's machines.

Note that `requirements.txt` dependencies are not installed during `pip3 install networkit`. The installation requirements are really specified in `setup.py` (or, in our case, `version.py`). We could definitely supply a `full-requirements.txt` that also includes cmake and ninja.

paluskapter:
I can understand you concerns, closing this PR.

avdgrinten:
@paluskapter Before we dismiss this issue entirely, would the `full-requirements.txt` help in your case?

paluskapter:
Unfortunately not really.

PR #429, opened 2019-10-10
Fix BarabasiAlbertGenerator

manpen:
The default implementation of the Barabasi Albert Generator (due to Batagelj and Brandes) was buggy, inefficient and unnecessarily complicated (rather than using push_back on a vector, the index was (wrongly) computed each time). This PR tries to resolve it:

- Originally the index computation miscalculated the edges from the initial seed graph. This led to reads from uninitialized memory or to overwriting the parts of the seed graph
- The original implementation maintained an vector `M` (required for constant time sampling), a set of produced edges and a graph G. The new implementation exploits the fact that only the $k$ most recent additions can lead to parallel edges and hence does not require the set.
- The original implementation removed duplicated edges, changing the degree distribution. Thus, technically, the generator did not produce BA graphs.

- The new implementation also counts degrees which requires a little bit more memory, but accelerates the generator by ~20%. This should be changed as soon as we have a fast graph builder.

### DIFF
--- a/include/networkit/generators/BarabasiAlbertGenerator.hpp
+++ b/include/networkit/generators/BarabasiAlbertGenerator.hpp
@@ -4,6 +4,7 @@
  *  Created on: May 28, 2013
  *      Author: forigem
  */
+// networkit-format
 
 #ifndef BarabasiAlbertGenerator_H_
 #define BarabasiAlbertGenerator_H_
@@ -16,12 +17,12 @@ namespace NetworKit {
  * @ingroup generators
  * Generates a scale-free graph using the Barabasi-Albert preferential attachment model.
  */
-class BarabasiAlbertGenerator: public StaticGraphGenerator {
+class BarabasiAlbertGenerator : public StaticGraphGenerator {
 private:
     Graph initGraph;
-    count k{0}; //!< Attachments made per node
+    count k{0};    //!< Attachments made per node
     count nMax{0}; //!< The maximal number of nodes attached
-    count n0{0}; //!< The number of initial connected nodes
+    count n0{0};   //!< The number of initial connected nodes
     bool batagelj; //!< Specifies whether to use batagelj's method or the original one
 
     /**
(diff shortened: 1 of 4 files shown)


include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -35,24 +35,29 @@ class BarabasiAlbertGenerator: public StaticGraphGenerator {
...
+     * @param k     Number of attachments per node
+     * @param nMax  Number of nodes in the graph
+     * @param n0    Number of connected nodes to begin with
+     * @param turbo Specifies whether to use Batagelj and Brandes's method rather than the naive one; default: true
      */
-    BarabasiAlbertGenerator(count k, count nMax, count n0 = 0, bool batagelj=true);
+    BarabasiAlbertGenerator(count k, count nMax, count n0 = 0, bool turbo=true);
 
-    BarabasiAlbertGenerator(count k, count nMax, const Graph& initGraph, bool batagelj=true);
+    BarabasiAlbertGenerator(count k, count nMax, const Graph& initGraph, bool turbo=true);
```
angriman: I think it is fine to keep `batagelj`, it clarifies that we are using that algorithm. Is there a specific reason for this change?

networkit/cpp/generators/BarabasiAlbertGenerator.cpp
```diff
@@ -84,20 +84,18 @@ Graph BarabasiAlbertGenerator::generateOriginal() {
         while (targets.size() - 1 < k) {
             uint64_t random = (uint64_t) Aux::Random::integer(degreeSum);
```
angriman: `Aux::Random::integer` creates a new integer distribution every time. It is more efficient to create the generator and the random distribution before (as you already did in the faster function below). I think it would be nice to have this optimization also here.

networkit/cpp/generators/BarabasiAlbertGenerator.cpp
```diff
@@ -84,20 +84,18 @@ Graph BarabasiAlbertGenerator::generateOriginal() {
...
                     targets.insert(v);
                 }
                 random -= G.degree(v);
-                //if (j >= G.numberOfNodes() && found==false) throw std::runtime_error("Last node, but still nothing happened.");
             });
+
+            handler.assureRunning();
         }
 
         targets.erase(u);
```
angriman: Is it necessary to add edges in a specific order? Otherwise, would an `unordered_set` help?

include/networkit/generators/BarabasiAlbertGenerator.hpp
manpen: My intention here was to make the implications of this flag more explicit. However, I also see your point and reverted the changes (in doubt leave it as is).

manpen:
Thank you for your review. I agree with all comments and implemented the proposed improvements of the naive BA generator -- primarily because it does not complicate the program logic. We should, however, avoid anything complicated, as nobody should use the code anyhow... while the naive implementation could be O(m \log n) (using appropriate weighted sampling) it seems to be O(m * n) due to the ineffective linear search.

manpen:
There was a little bug in processing the default value for the initial seed graph size. This caused a segfault in the new Generator Notebook. It's fixed now. I also added a tiny amendment to test_notebooks: it now outputs the name of the notebook under test BEFORE it gets executed. This helps in narrowing down segfaults, as previously the offending notebook was not reported. I hope it's okay, that this is included in this PR.

The force-push is rather excessive as I rebased the branch to the current Dev.

PR #430, opened 2019-10-10
Improved Tooling

manpen:
This PR adds two checks enforcing style guides. It enables them in CI and makes them available for easy "pre-commit" execution:

- `CppClangFormat` applies clang-format to all files that contain the string `networkit-format`. This allows a to opt-in gradually. I suggest the convention to put this string directly below the file header:
```C++
/*
 * GlobalCurveballImpl.hpp
...
 */
// networkit-format
```
- The PR opts-in the whole `randomization` module (as it will undergo some modernization anyway)

- `CppIncludeGuards` ensures that all header files contain an include guard (surprisingly there were three header files which did not ... so I guess, it's worth having this check). As our historically grown and now inconsistent naming of the guard macro annoyed me more than they should have, the script also enforces the following names in NETWORKIT_MODULENAME_HEADERFILE_HPP_
As a little bonus the script will also decorate all long-ranged `#endif` which started with an `#if[n]def X` as `#endif // X`.
This PR touches so many files, as I executed this fix.

- The PR adds the file `/check_code.sh` which carries out all checks in read-only mode. If errors are found, `/check_code.sh -w` will fix them. This script is also executed during CI and will fail if errors are detected.

- The PR removes the folders `include/networkit/test` and `networkit/cpp/test` as the files were not included, not build and did not contain useful stuff. They are removed here, because they contain broken code (e.g. missing guards), which I did not want to fix only to remove them later.


extrafiles/tooling/CppIncludeGuards.py
```diff
@@ -0,0 +1,116 @@
...
+
+numScannedFiled = 0
+numNoncompliantFiles = 0
+numIllegalFiles = 0
+
+files = nkt.getCXXFiles(includeSources = False)
+
+msgType = "[ERROR]" if nkt.isReadonly() else "[FIXED]"
+
+processPragmaOnce = False
```
avdgrinten: What is the intention here? We should not allow `#pragma once` in some files and include guards in others. Since we already have include guards, I see little reason in switching to `#pragma once`.

extrafiles/tooling/CppIncludeGuards.py
```diff
@@ -0,0 +1,116 @@
...
+
+TabWidth = 4
+
+numScannedFiled = 0
+numNoncompliantFiles = 0
+numIllegalFiles = 0
+
+files = nkt.getCXXFiles(includeSources = False)
+
+msgType = "[ERROR]" if nkt.isReadonly() else "[FIXED]"
```
avdgrinten: Should this formatting consistently be provided by `nkt`?

extrafiles/tooling/CppIncludeGuards.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+"""
+This tool iterates over all C++ files and replaces tab-based
+indentation with spaces (tabwidth = 4). If used in read-only
+mode it returns a negative exit code if a change is necessary.
+"""
```
avdgrinten: This description seems to be off.

extrafiles/tooling/CppIncludeGuards.py
```diff
@@ -0,0 +1,116 @@
...
+import nktooling as nkt
+import re
+import os, sys
+
+nkt.setup()
+
+TabWidth = 4
```
avdgrinten: Unused.

extrafiles/tooling/CppIncludeGuards.py
manpen: `#pragma once` supposedly speed-up preprocessor runs as file does not have to be parsed several times. Hence, I added this feature to test it. I could however not measure any significant impact. 

To make the intent more explicit I renamed the variable to `includePragmaOnce`; it it's set to false, pragmas added by the user are removed for consistency reasons.

manpen:
Thank you for your feedback. I addressed all comments. Your assumption that the last two commits result from running the checks/fixes is correct (and the last commit also adds `// networkit-format` to the appropriate files). I would have preferred to solve it differently too, however, the CI would fail, if these changes were not included immediately.

avdgrinten:
Thank you very much for this PR.

PR #431, opened 2019-10-11
Jupyter-notebook for randomization module

cndolo:
This PR adds a new jupyter-notebook for the `randomization` module. There are no links to the mentioned functions because I didn't find the randomization module in the [Python documentation](https://networkit.github.io/dev-docs/python_api/modules.html).

### DIFF
--- a/notebooks/Randomization.ipynb
+++ b/notebooks/Randomization.ipynb
@@ -1,0 +1,313 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# NetworKit Randomization Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The randomization module implements algorithms to perturb existing graphs. This is commonly used to obtain a null-model for hypothesis testing in network analysis (see below for an example). Intuitively one tests whether an observation in original also appears in *similar* graphs. By doing so one can quantify the statistical significance of the observation. The ensemble of *similar* networks is commonly chosen as the set of (simple) graphs that have the same degree sequence. \n",
+    "\n",
+    "The Curveball algorithms creates a random sample with the same degree sequence, by repeatedly selecting two random nodes and randomly exchanging their neighbourhoods (this is referred to as a trade). This corresponds to a random walk on the space of all graphs in the ensemble. The more trades are carried out, the less correlated input and output are. While there a no practical lower bounds, one typically choose 10 to 100 times the number n of nodes.\n",
+    "\n",
+    "GlobalCurveball is a related algorithm that is typically faster than Curveball. Also it's implementation is more versatile (e.g., it support directed and undirected graphs). In each step it carries out n/2 trades involving all nodes. A typical choice of the number of GlobalTrades is 10 to 100. **We recommend the usage of GlobalCurveball over Curveball for performance reasons**."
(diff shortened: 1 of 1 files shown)


notebooks/Randomization.ipynb
```diff
@@ -0,0 +1,208 @@
...
+   "metadata": {},
+   "source": [
+    "# NetworKit Randomization Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Randomization is the process of altering a graph and generating a random graph from the given input graph. It is done by changing node neighborhoods while preserving the node degrees of the original graph. Such algorithms in NetworKit can be found in the `randomization` module. Most algorithms in the `randomization` module implement a `run` function that must be called after the chosen algorithm has been initialiaszed."
```
manpen: The randomization process does not necessarily preserve degree sequences. It just so happens that we're only implementing such algorithms. Hence, I would chose a rather broader introduction a long the lines of

```
The randomization module implements algorithms to perturb existing graphs. This is commonly used to obtain a null-model for hypothesis testing in network analysis (see below for an example). Intuitively one tests whether an observation in original also appears in *similar* graphs. By doing so one can quantify the statistical significance of the observation. The ensemble of *similar* networks is commonly chosen as the set of (simple) graphs that have the same degree sequence. 

The Curveball algorithms creates a random sample with the same degree sequence, by repeatedly selecting two random nodes and randomly exchanging their neighbourhoods (this is referred to as a trade). This corresponds to a random walk on the space of all graphs in the ensemble. The more trades are carried out, the less correlated input and output are. While there a no practical lower bounds, one typically choose 10 to 100 times the number n of nodes.

GlobalCurveball is a related algorithm that is typically faster than Curveball. Also it's implementation is more versatile (e.g., it support directed and undirected graphs). In each step it carries out n/2 trades involving all nodes. A typical choice of the number of GlobalTrades is 10 to 100. **We recommend the usage of GlobalCurveball over Curveball for performance reasons**.
```

notebooks/Randomization.ipynb
```diff
@@ -0,0 +1,208 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import networkit as nk"
+   ]
+  },
+  {
+   "cell_type": "markdown",
```
manpen: As we recommend `GlobalCurveball` it may be interesting to introduce it first.

manpen:
I'd might also be interesting to point out that `Curveball.run` may be called multiple times; unfortunately this is not true for `GlobalCurveball.run` for technical reasons.

manpen:
`If the graph is directed, either allowSelfLoops or degreePreservingShufflePreprocessing should be set.`
->
To accelerate the perturbation process we recommend to set `degreePreservingShufflePreprocessing` to true. This jump-starts the randomization process and yields faster convergence of the algorithm (see `"Smaller Universes for Uniform Sampling of 0,1-matrices with fixed row and column sums" *  by Annabell Berger, Corrie Jacobien Carstens [https://arxiv.org/abs/1803.02624]`). 
**For directed simple graphs the algorithm will not yield a uniform sample unless the preprocessing is enabled**.

manpen:
Also `DegreePreservingShuffle` is available as a standalone module. It will relabel nodes while keeping their degrees but wont change the topology of the graph.

PR #432, opened 2019-10-12
Fix Remove Edge

manpen:
Most of the functions to remove edges did not take edge ids into account. This results in ever growing graphs and -- even worse -- changing edge ids.

### DIFF
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -766,10 +766,16 @@ void Graph::removeEdge(node u, node v) {
         throw std::runtime_error(strm.str());
     }
 
+    const auto isLoop = (u == v);
     m--; // decrease number of edges
+    if (isLoop) storedNumberOfSelfLoops--;
+
     erase<node>(u, vi, outEdges);
     if (weighted) {
         erase<edgeweight>(u, vi, outEdgeWeights);
+    }
+    if (edgesIndexed) {
+        erase<edgeid>(u, vi, outEdgeIds);
     }
 
     if (directed) {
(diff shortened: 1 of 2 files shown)


networkit/cpp/graph/Graph.cpp
```diff
@@ -771,6 +771,9 @@ void Graph::removeEdge(node u, node v) {
     if (weighted) {
         erase<edgeweight>(u, vi, outEdgeWeights);
     }
+    if (edgesIndexed) {
+        erase(u, vi, outEdgeIds);
+    }
```
angriman: I see that it is not necessary to specify the template parameter of `erase`. I think that it would be better to remove it when used, at least within this function, otherwise it is a bit confusing for the reader.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -40,7 +40,8 @@ class GraphGTest : public testing::TestWithParam<std::tuple<bool, bool>> {
 
     bool isWeighted() const;
     bool isDirected() const;
-    Graph createGraph(count n = 0) const;
+    Graph createGraph(node n = 0) const;
+    Graph createGraph(node n, count m) const;
```
angriman: About the types of `n` and `m`: if we are using `node` for `n`, shouldn't we use `edgeid` for `m`? A simpler choice could be `count` for both.

manpen:
Thank you for your comments. I fixed both.

angriman:
Thank you. Did you forget to update the `erase` above [my comment](https://github.com/networkit/networkit/pull/432#discussion_r336734547)?

manpen:
You got a 2 for 1 deal: I fixed it at another spot, but not where you commented ;) Now both are fixed. Sorry for the extra round.

Issue #419, opened 2019-10-02
Implicitly add non-exisitng incident nodes upon edge insertion


JunpengCode:
The following code is wrong: segmentation fault,
```python
import networkit as nk
if __name__ == '__main__':
    G= nk.graph.Graph()
    G.addEdge(0,1)
    print(G.numberOfNodes())
```
It is due to that the node index is not exist.  The following code is correct,
```python
import networkit as nk
if __name__ == '__main__':
    G= nk.graph.Graph()
    G.addNode()
    G.addNode()
    G.addNode()
    G.addEdge(0,1)
    print(G.numberOfNodes())
```

I consider that the function may be added into the networkit, In the networkx library, These are added automatically when edges are added.

manpen:
I would vote against doing this implicitly. IMHO adding edges to non-existing nodes should be illegal as you're using unallocated memory. You would not expect the following code to work either:

```Python
x = [0]
x[1] = 1
```

Having an error message for the Python interface (which I can add as soon as we agree on this matter) should help to uncover potential logic bugs. And if you really want to add the nodes you can do it manually. A compromise would be flag that automatically increases the set of nodes, e.g.,

```Python
G = nk.Graph()
G.addEdge(0, 1, addNodes = True)
```

JunpengCode:
I think that  adding the flag that automatically increases the set of nodes is reasonable.

avdgrinten:
I agree with Manuel. The number of implicit functionality in the C++ core should be kept to a minimum. If this feature is *really* desired, we could add a helper function in addition to `addEdge`.

JunpengCode:
I consider that  adding this feature makes networkit more better. For example, the file format is following:
1 2 3 4 5
2 3 4 5 6
3 4 5 6 7

I may be selected the col 1,3, and 5 as node1, node2 and weight.  This avoids adding more read functions.  And it is also more flexible.

JunpengCode:
For example:

```python
def readUndirectedGraph(file):
    G = nk.graph.Graph()
    f = open(file)
    try:
        for line in f.readlines():
            cell = line.split()
            G.add_edge(cell[0],cell[1])
    except:
        print("Unexpected error:", sys.exc_info())
        input('press enter key to exit')
    return G
```

JunpengCode:
```python
def readUndirectedGraph(file):
    G = nk.graph.Graph()
    f = open(file)
    try:
        for line in f.readlines():
            cell = line.split()
[...]

PR #433, opened 2019-10-12
Fix Graph::addNode, Add Graph::addNodes, Guard Python Graph.addEdge

manpen:
This PR includes three related changes:

- The logic of `Graph::addNode` was broken (e.g. it pushed to `inEdgeWeights` for undirected graphs) and inefficent (rather than using `outEdges.emplace_back()`, first an empty vector was created and then pushed). Both is fixed now.
- To add multiple nodes (required by next point), the method Graph::addNodes(k) was added which performs resizes rather than repeated emplace_backs.
- As discussed in #419, there are no boundary checks for Python's Graph.addEdge(u, v). Thus adding an edge to a non-existing node corrupts memory and (in the best-case) leads to a SegFault. While this is common for C++ interfaces, boundary checks are expected in Python interfaces. Hence, now an exception is thrown. Additionally `addEdge` has now the optional flag `addMissing=False` which will automatically add necessary nodes.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -820,6 +820,13 @@ public:
     node addNode();
 
     /**
+     * Add numberOfNewNodes new nodes.
+     * @param  numberOfNewNodes Number of new nodes.
+     * @return The index of the last node added.
+     */
+    node addNodes(count numberOfNewNodes);
+
+    /**
      * DEPRECATED: Coordinates should be handled outside the Graph class
      * like general node attributes.
      *
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -305,6 +305,7 @@ cdef extern from "<networkit/graph/Graph.hpp>":
 		bool_t isIsolated(node u) except +
 		_Graph copyNodes() except +
 		node addNode() except +
+		node addNodes(node) except +
 		void removeNode(node u) except +
 		bool_t hasNode(node u) except +
 		void restoreNode(node u) except +
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -707,6 +707,23 @@ cdef class Graph:
 	 	"""
 		return self._this.addNode()
 
+	def addNodes(self, numberOfNewNodes):
+		""" Add numberOfNewNodes many new nodes to the graph and return
+		the id of the last node added.
+
+		Parameters
+		----------
+		numberOfNewNodes : node
+			Number of nodes to be added.
+
+		Returns
+		-------
+		node
+			The id of the last node added.
+		"""
+		assert(numberOfNewNodes >= 0)
+		return self._this.addNodes(numberOfNewNodes)
+
 	def removeNode(self, u):
 		""" Remove a node `v` and all incident edges from the graph.
 
@@ -715,7 +732,7 @@ cdef class Graph:
 	 	Parameters
 	 	----------
 	 	u : node
-	 		Node.
+	 		Id of node to be removed.
 		"""
 		self._this.removeNode(u)
 
@@ -735,7 +752,7 @@ cdef class Graph:
 		Parameters
 		----------
 		u : node
-			Node
+			Id of node queried.
 
 		Returns
 		-------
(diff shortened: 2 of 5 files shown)


include/networkit/graph/Graph.hpp
```diff
@@ -819,6 +819,12 @@ class Graph final {
      */
     node addNode();
 
+    /**
+     * Adds n nodes.
+     * @return The index of the last node added.
+     */
+    node addNodes(node numberOfNewNodes=1);
```
angriman: Type of `numberOfNewNodes`: should we use `count`? Also, do not forget to document `numberOfNewNodes` (`@param numberOfNewNodes ...` is missing).

networkit/_NetworKit.pyx
```diff
@@ -778,6 +792,20 @@ cdef class Graph:
 		w : edgeweight, optional
 			Edge weight.
 		"""
```
angriman: Don't forget to document `addMissing` under `Parameters`.

networkit/_NetworKit.pyx
```diff
@@ -706,6 +707,18 @@ cdef class Graph:
 	 	"""
 		return self._this.addNode()
 
+	def addNodes(self, numberOfNewNodes):
```
angriman: In C++ `numberOfNewNodes` is 1 by default, in Python we should have the same behavior.

networkit/_NetworKit.pyx
```diff
@@ -706,6 +707,18 @@ cdef class Graph:
...
+		""" Add numberOfNewNodes many new nodes to the graph and return
+		the id of the last node added.
+
```
angriman: `Parameters` are missing.

include/networkit/graph/Graph.hpp
manpen: I prefer the original name, as its clear without doubt that the number of new nodes (rather than the total number of nodes; does not seem plausible, but in doubt, one has to check) has to be provided. Backed up by [Kate G.'s talk on CppCon](https://www.youtube.com/watch?v=MBRoCdtZOYg) I'm currently leaning towards longer names ;)

networkit/_NetworKit.pyx
manpen: Actually I did the opposite: removed the default in C++'s implementation. If you only want to add a single node every time, you should directly use `addNode`

manpen:
Thank you, @angriman, for your review. I pushed the changes (sorry for the multiple force-pushes -- I messed up a rebase ;))

include/networkit/graph/Graph.hpp
avdgrinten: I think Eugenio was referring to the type here.

avdgrinten: Should we also have a `resizeNodes` function?

angriman: That's indeed a good talk ;) however, my point was about the type, not the name. I am totally fine with the chosen name.

include/networkit/graph/Graph.hpp
```diff
@@ -819,6 +819,13 @@ class Graph final {
      */
     node addNode();
 
+    /**
+     * Add numberOfNewNodes new nodes.
+     * @param  numberOfNewNodes Number of new nodes
```
angriman:
```suggestion
     * @param  numberOfNewNodes Number of new nodes.
```

networkit/_NetworKit.pyx
```diff
@@ -706,6 +707,23 @@ cdef class Graph:
...
+		Parameters
+		----------
+			numberOfNewNodes : node
+				Number of nodes to be added.
```
angriman: Wrong indentation, remove a tab on both lines?

include/networkit/graph/Graph.hpp
manpen: That makes more sense. Thank's for clarifying. I changed it. Somewhat related: If we progress with 32bit ids (#427), we should add a check to `addNode` and `addNodes` not to overflow.

Regarding `resizeNodes()`: Shrinking is a little more involved as growing, since we have to remove edges to non-existings nodes. I have a PR that introduces an `removeEdgeIf` method that could be used, but are waiting until #421 due to some conflicts. Hence I suggest to post-pone adding `resizeNodes()`.

angriman: Thank you. Yes if we make progress with 32 bit ids we should add those checks.

angriman:
This should also be merged before the next release. @manpen can you resolve the conflicts?

PR #434, opened 2019-10-13
Remove Graph::*coordinates

manpen:
This PR removes the coordinates interface from `Graph`. These methods have been deprecated a long time ago. The interface is dropped without replacement. Similarly, the unused `graph/Coordinates.hpp` header was dropped.  Coordinates should be maintained by the application in a vector.

Previously coordinates were stored as 32bit floating points. This causes a lot of headaches due to implicit casts (just take a look at MSVC warnings); as we're only storing very little coordinates, there's only a small overhead by using doubles. However, there's a type alias in `Globals.hpp` to switch back (please don't, or first fix all implicit casts !).

There are now two ways to store a point:
- The new type alias `using coord2d = std::pair<coord, coord>` for 2D points.
- The old and slow class `Point<T>` to store points where the number of dimensions is unknown at compile time. It seems that this Point class is used very scarcely and it might also be dropped (it's usage in the random hyperbolic generators can and should replaced with `coord2d`)

There were only few classes Graph's coordinate interface:
- `PostscriptWriter` which is not exposed to Python. Here, the `write` method now expects an additional vector `std::vector<coord2d>` along with the graph.
- `DibapGraphReader` which is also not exposed to Python. The implementation of the reader supports arbitraty dimensions. Hence it uses `Point<coord>`s (also the old implementation only dealt with the 2d case). This reader should undergo an extensive modernization, as it will leak resource on errors.
- `(Dynamic)PubWebGenerator`: These are also only 2d and hence use `coord2d`.

Since significant portions of all four classes had to be implemented, I also used modern C++ constructs, reduced namespace pollution and clang-format.

This PR also add (minimalistic) Python unit tests for the PubWeb generator family. Since I've limited experience with dynamic generators and did not find a documentation on the semantic, please review this part carefully.


avdgrinten:
This might lead to some bikeshedding but: what is our policy about lower-case type names? `node`, `count` etc. are lowercase. Similarly, it makes sense for `coord` to be lower-case, too. But does it really make sense for `coord2d`? Can be define a general policy that allows to answer this question in the future?

Related: should the type be called `coord` or `coordinate`?

avdgrinten:
This PR should be reviewed and merged soon.

manpen:
Regarding `coord` ./. `coordinate` -- while preparing this PR I felt that `coord` is pretty self-explanatory and that the shorting would be appropriate since it's used a lot. Seening now, that the PR contains ~50 occurrences of `coord`, it'd also support the full version.

Regarding the capitalization of `coord(inate)2d`: I'm absolutely indifferent about it. My rational for using lower case was that it is semantically used as a primitive type with no added features what so ever (this is done to use Cython's type conversion facilities ...). But I won't fight for that in any way ;)

If we formulate this as an explicit rule: 
```
All types directly under NetworKit's namespace are PascalCased. The only exception are frequently used type-aliases (i.e., defined in `Globals.hpp`) to semantically primitive types.
```

If we want to strengthen this statement, we remove 'semantically' -- a primitive type is well-defined. This would then mean that we should name it `Coordinate2D`.

avdgrinten:
Is the Cython integration also possible  if we use `std::array<coord, 2>`?

include/networkit/viz/Point.hpp
```diff
@@ -30,18 +30,18 @@ namespace NetworKit {
  *
  * Points in any dimension of templated type.
  */
-template<class T>
+template<class T = coord>
 class Point {
```
avdgrinten: Should this class be moved to the toplevel NetworKit namespace to avoid importing the viz module from other modules?

include/networkit/io/DibapGraphReader.hpp
```diff
@@ -8,26 +8,13 @@
 #ifndef DIBAPGRAPHREADER_H_
 #define DIBAPGRAPHREADER_H_
 
-#if !defined _WIN32 && !defined _WIN64 && !defined WIN32 && !defined WIN64
+#ifndef DNETWORKIT_WINDOWS
```
avdgrinten: This macro is called `NETWORKIT_WINDOWS`, isn't it?

networkit/cpp/io/DibapGraphReader.cpp
```diff
@@ -5,17 +5,34 @@
  *      Author: Henning
  */
 
-#if !defined _WIN32 && !defined _WIN64 && !defined WIN32 && !defined WIN64
+#ifndef DNETWORKIT_WINDOWS
```
avdgrinten: AFAICS the only reason that this is not available on Windows is that we need `htonX`? This is not something that should be changed in this PR but in the future we could migrate to an internal (or TLX?) byteswapping function.

avdgrinten:
The PR looks good. It can be merged once we resolve the bikeshedding issues.

include/networkit/viz/Point.hpp
manpen: I don't think it's necessary, since it's a header-only dependency and as such can be hidden from CMake.

include/networkit/io/DibapGraphReader.hpp
manpen: Yes!

avdgrinten:
From a purely C++ perspective, I would suggest the following: rename `coord2d` to `Point2D` and implement it as an alias of `std::array<coord, 2d>` (or as a class with an interface similar to `Point`). I don't know how ergonomic this would be in Python.

I would still like to rename `coord` to `coordinate` as that frees up the shorter name for local variables.

manpen:
Since this PR is blocking #421, a short update: On the weekend, I made significant progress here and hope to have the change by tomorrow.

avdgrinten:
The last commit tries to sneak in the GIRGs generator ;).

manpen:
Yes, thank you for the remark. Indeed it was a cherry-picking + squash gone wrong. But `Girgs` are awesome, we should have them anyway.

After our discussion I changed the following:
- `coord` -> `coordinate`
- `coord2d` is now `Point2D` -- a full type defined in `viz/Point.hpp`.
- The previous `Point<T>` class only supported runtime-defined dimensions. Now, both compile-time and runtime-defined dimensions are supported with the same logic by adding a second template parameter `Point<T, Dims=0>`. If `Dims` is zero the number of dimensions is defined at runtime and data is stored on heap (only change: we use `tlx::simple_vector` as it requires one less pointer). If `Dims` is positive, it defines the fixed number of dimensions at compile time.

Additional stuff:
- The PR fixes the Travis GCC 8 bug (GCC8 is installed with OSX 10.14, so homebrew fails. Simple solution: don't install the compiler)
- DSGWriter was missing a header dependency preventing builds on MSVC.

Missing stuff:
- Random Hyperbolic Graphs still use there own `Point2DWithIndex` (previously `Point2D` now renamed to resolve the name clash with the new `Point2D` type), which should be migrated to `Point2D` in a future PR. 
- The Point class is missing unit tests. We should really change that, but I currently do have the capacity to do it on short notice.

networkit/cpp/io/DibapGraphReader.cpp
```diff
@@ -5,17 +5,36 @@
...
+// networkit-format
+
+#ifndef NETWORKIT_WINDOWS
 
-#include <networkit/io/DibapGraphReader.hpp>
 #include <networkit/auxiliary/Log.hpp>
+#include <networkit/io/DibapGraphReader.hpp>
 
+#include <cstdio>
+#include <netinet/in.h>
```
angriman: Is this include necessary?

hmeyerhenke: I think it is necessary for the little/big endian computations iirc.

.travis.yml
```diff
@@ -43,12 +43,6 @@ matrix:
       os: osx
       osx_image: xcode10.2
       sudo: false
-      addons:
-        homebrew:
-          packages:
-            - gcc@8
-      before_install:
-        - brew link gcc@8
```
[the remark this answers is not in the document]
manpen: The commit is in here to fix the (still broken) CI -- I assumed the PR would be more convincing with the "green circle" next to it ;) Judging by the Travis output, GCC 8.3 is indeed used:

```
$CXX --version

g++-8 (Homebrew GCC 8.3.0) 8.3.0
``` 

and 

``` 
cmake -DNETWORKIT_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release -DNETWORKIT_CXX_STANDARD=$CXX_STANDARD -DNETWORKIT_WARNINGS=ON -DNETWORKIT_WARNINGS_AS_ERRORS=ON ..

-- The CXX compiler identification is GNU 8.3.0
```

manpen:
Thank you two for your comments, I addressed all,  rebased the branch, and removed the commit regarding Travis since the CI is back to normal w/o modification.

angriman:
Thank you for addressing the comments. I think this PR can be approved as soon as the conflicts have been resolved.

avdgrinten:
Why is the "Fix conflicts" commit so huge? Is that a merge commit? We should rebase the patch series on `Dev` instead of adding a merge commit.

angriman:
In that case I need to force-push on @manpen's branch and this would add myself as co-author of all his commits. I am not sure if we really want this so I pushed a merge commit. Should I rebase and force-push?

manpen:
Sorry guys. I will look into this evening.

Am 27. November 2019 13:34:16 MEZ schrieb Eugenio Angriman <notifications@github.com>:

avdgrinten:
If you're fine with Eugenio being the committer, he could also just take the patches and rebase them.

manpen:
While I'm fine with  @angriman  being a contributor, I rebased the PR yesterday, hope to avoid additional work for you.

PR #435, opened 2019-10-15
Fix graphio.readGraph / graphio.writeGraph

manpen:
As pointed out in the mailing list, Python's graphio interfaces `graphio.readGraph` and `graphio.writeGraph` require the user to provide arguments to the EdgeList reader/writer. Error messages get lost in an unnecessary `try ... except` block which is removed by this PR.

We also fix two bugs in the logic regarding `EdgeList`: While the `EdgeList` interface has to *optional* parameter `continuous`, it was required by the `readGraph` interface. This is not the case anymore. Additionally, the required `separator` and `firstNode` parameters have to be passed as positional arguments. Hence this PR add `*kargs` to readGraph/writeGraph.

It also adds a very simple unittest that tries for each format to write and read a graph again. This surprisingly failed for three Formats (SNAP, GEXF and MAT). I fixed MAT here.

### DIFF
--- a/networkit/graphio.py
+++ b/networkit/graphio.py
@@ -24,67 +24,40 @@ except ImportError:
 	print("module 'scipy' not available - some functionality will be restricted")
 import fnmatch
 
-try:
-	from enum import Enum
+from enum import Enum
 
-	class __AutoNumber(Enum):
-		def __new__(cls):
-			value = len(cls.__members__) + 1
-			obj = object.__new__(cls)
-			obj._value_ = value
-			return obj
-
-
-	class Format(__AutoNumber):
-		""" Simple enumeration class to list supported file types. Currently supported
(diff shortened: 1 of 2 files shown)


networkit/graphio.py
```diff
@@ -24,67 +24,40 @@
...
-
-
+from enum import Enum
+
+class __AutoNumber(Enum):
+	def __new__(cls):
+		value = len(cls.__members__) + 1
+		obj = object.__new__(cls)
+		obj._value_ = value
+		return obj
```
avdgrinten: Can't we just get rid of that class? Maintaining the numbers is less mental overhead than thinking about this mechanism.

avdgrinten:
I'm merging this PR since the only objection is a "nice to have".

PR #438, opened 2019-10-22
Jupyter-notebook for distance module

cndolo:
 * Fix for Python constructors for Bidirectional BFS, Bidirectional Dijkstra and AStar,
 * Expose ReverseBFS to Python,
 * Some documentation fixes and
 * Jupyter-notebook for the `distance` module


notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The distance from source to target node\n",
+    "print(biBFS.getHops())\n",
+    "print(biBFS.getPath())"
```
angriman: Mention that `getPath` does not include the source at the beginning, and the target at the end while BFS does. Otherwise, one might get confused if the length of the paths returned by BFS and BidirectionalBFS are different.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+   "source": [
+    "## Breadth-First Search (BFS) "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "BFS is an algorithm for traversing a graph which starts from the source node `u`, and explores all of the u's neighbors nodes at the present depth before moving on to the nodes at the next depth level.\n",
+    "\n",
```
angriman: Mention that BFS finds the shortest paths from a source to all the reachable nodes of an unweighted graph.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+   "metadata": {},
+   "source": [
+    "## Dijkstra "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Dijkstra's algorithm finds the shortest path from a source node a target node. It algorithm creates a tree of shortest paths from the source to all other nodes in the graph.\n",
```
angriman:
```suggestion
    "Dijkstra's algorithm finds the shortest path from a source node a target node. It algorithm creates a tree of shortest paths from the source to all other nodes in the graph.\n",
```
Same as BFS, but Dijkstra does it also for weighted graphs. The descriptions of BFS and Dijkstra should be similar, the difference is that BFS only works with unweighted graphs.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+   "source": [
+    "## Bidirectional Dijkstra"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Bidirectional Dijkstra algorithm explores the graph from both the source and target nodes until the two explorations meet.  This version of Dijkstra is more efficient than the convential Dijkstra when the target node is known.\n",
+    "\n",
```
angriman: Again, mention that the result of `getPath` does not include the source and the target node, Dijkstra includes them.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Run\n",
+    "ctd.run()\n",
+    "# Alternatively \n",
+    "# ctd.runSinglePair(source,target)"
```
angriman: You can run this in another cell, and mention that if one wants to compute the commute time distance between two nodes, then they should use `runSinglePair`.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+   "metadata": {},
+   "source": [
+    "## Diameter "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This algorithm gives an estimation of the diameter of a given graph. The algorithm is based on the ExactSumSweep algorithm presented in Michele Borassi, Pierluigi Crescenzi, Michel Habib, Walter A. Kosters, Andrea Marino, Frank W. Takes[1](http://www.sciencedirect.com/science/article/pii/S0304397515001644).\n",
```
angriman:
```suggestion
    "This algorithm gives an estimation of the diameter of a given graph. The algorithm is based on the ExactSumSweep algorithm presented in Michele Borassi, Pierluigi Crescenzi, Michel Habib, Walter A. Kosters, Andrea Marino, Frank W. Takes: http://www.sciencedirect.com/science/article/pii/S0304397515001644.\n",
```

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+    "newGraph.numberOfNodes()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Initialize algorithm\n",
```
angriman:
```suggestion
    "# Initialize algorithm to compute the exact diameter of the input graph\n",
```

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+    "diam.run()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get diameter of graph\n",
```
angriman: Mention why the returned diameter is a pair (lower bound, upper bound), and that in the exact case the diameter is the first value of the pair.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,812 @@
...
+    "## Reverse BFS"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This class does a reverse breadth-first search (following the incoming edges of a node) on a directed graph from a given source node.\n",
+    "\n",
+    "The [ReverseBFS(G, source, storePaths=True, storeNodesSortedByDistance=False, target=none)](https://networkit.github.io/dev-docs/python_api/distance.html?highlight=bfs#networkit.distance.BFS) constructor expects a graph and a source node as mandatory parameters. If the paths should be stored, set `storedPaths` to true. If `storeNodesSortedByDistance` is set, a vector of nodes ordered in increasing distance from the source is stored. `target` is the target node."
```
angriman: Swap source and target, it looks like that there is not target -> source path and the returned values are not interesting.

cndolo:
I have made the changes; I added a new commit which exposes the `distance` function of the `AdamicAdarDistance` class to Python. This was necessary in order to get the distance between 2 nodes like you suggested.

cndolo:
Please review this PR again @angriman - I have made the changes we recently discussed.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "If one wants to compute the commute time distance between two nodes, then they should use [runSinglePair(u, v)](https://networkit.github.io/dev-docs/python_api/distance.html?highlight=runsingle#networkit.distance.CommuteTimeDistance.runSinglePair) method."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ctd.runSinglePair(source,target)"
```
angriman:
```suggestion
    "ctd.runSinglePair(source, target)"
```

This output is different than the one returned by ctd.distance(), probably a bug?

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "newGraph.numberOfNodes()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Initialize algorithm to compute the exact diameter of the input graph\n",
```
angriman:
```suggestion
    "# Initialize algorithm to compute the exact diameter of the input graph\n",
```

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,800 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Initialize algorithm\n",
+    "rbfs = nk.distance.ReverseBFS(G, target, True, False, source)"
```
angriman:
```suggestion
    "rbfs = nk.distance.ReverseBFS(G, source, True, False, target)"
```

Now that the graph is undirected, there is no need for this swap anymore.

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,800 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Read a graph\n",
+    "G = nk.readGraph(\"../input/foodweb-baydry.konect\", nk.Format.KONECT)\n",
+    "GDir = G\n",
```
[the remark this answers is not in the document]
cndolo: We are keeping the directing graph because the runtime error `RuntimeError: Error, AllSimplePaths class has been implemented for directed graphs only.` if I pass an undirected graph. Is this not expected?

notebooks/Distance.ipynb
```diff
@@ -0,0 +1,800 @@
...
+    "G = G.toUndirected()\n",
```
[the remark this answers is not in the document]
cndolo: Ok, this affects several others notebooks as well.

notebooks/Distance.ipynb
cndolo: It seems so. I tried with different source-target combinations of the same graph, and got different results there too.

notebooks/Distance.ipynb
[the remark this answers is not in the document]
angriman: I did not remember that `AllSimplePaths` only supports directed graphs. In that case, let's keep this as it is.

notebooks/Distance.ipynb
angriman: This seems to be a bug in the CommuteTimeDistance class, resolving because unrelated to this notebook.

PR #440, opened 2019-10-23
Add standalone edge switching

manpen:
We already have an EdgeSwitchingGenerator that implements FDSM -- it takes a degree sequence, produces a biased graph from it, and then uses edge switching to obtain a uniform sample from all graphs with the prescribed degree sequence.

This PR migrates the (non-optimized) edge switching implementations into its own algorithm `randomization.EdgeSwitching`. The previously mentioned generator now also uses the stand-alone module.


avdgrinten:
The strategy of copying the graph in the constructor is somewhat ugly. Why is the other constructor `Unsafe`? Why don't you just always modify a given graph in-place? The caller can always make a copy if that is desired.

manpen:
The main reason is consistency: none of the other randomization algorithms can perform in-place updates as they build alternative graph data structure to have an efficient implementation. I fear that a user may expect this behaviour and encounter hard to debug bugs.

Two technical reasons:
- The algorithm also uses `DegreePreservingShuffle` which does not yet work in-place (but can be fixed)
- If we ever want to make `EdgeSwitching` efficient, we need constant time `hasEdge` queries. Unless this will be supported by `Graph` directly, we will eventually have to build our own data structure anyhow. This then effectively is a copy ...

avdgrinten:
I see. The consistency argument makes sense. What about having an `InPlaceEdgeSwitching` class with the in-place behavior and an `EdgeSwitching` class that uses the in-place class to modify the graph?

Obviously, this only solves the first issue. However, IMHO it also makes the interface much nicer (and the constant-time/fast `hasEdge` might be fixed in a future update).

manpen:
As suggested, I split `EdgeSwitching` into `EdgeSwitching` and `EdgeSwitchingInPlace` with the obvious semantics.

networkit/_NetworKit.pyx
```diff
@@ -4,7 +4,7 @@
 #includes
 # needed for collections.Iterable
 from networkit.exceptions import ReducedFunctionalityWarning
-import collections
+import collections.Iterable
```
fabratu, line 7: Shouldn't it be `from collections.abc import Iterable` (or similar)?
 https://docs.python.org/3/library/collections.abc.html

manpen:
I forgot about this old PR --- that is, until I needed the functionality. In the recent force push I rebased the PR, slightly optimized the implementation and additionally exposed `EdgeSwitchingInPlace` to Python (as I needed it in an application).

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+     */
+    void run() override { run(10 * graph->numberOfEdges()); }
+
+    /**
+     * Carry out numberOfSwitches many edge swaps.
+     * @note This function can be called several times. The graph's topology may not be changed
+     * externally between two invocations (this is relevant only for graphs not owned by the
+     * algorithm)
+     */
+    void run(count numberOfSwitches);
```
avdgrinten, line 68: I would prefer an interface that accepts the number of switches in the constructor. Such an interface is more in line with the other algorithms that we have. We can expose a public `numberOfSwitches` member variable if this needs to be changed dynamically.

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Returns twice the total number of non-rejected swaps carried out during calls to run().
+     * It hence has the same semantics as Curveball::getNumberOfAffectedEdges().
+     * @note If run() is called multiple time, the sum over all invocations is returned.
+     */
+    count getNumberOfAffectedEdges() const noexcept { return 2 * numberOfSwapsPerformed; }
+
+    // required by Algorithm
+    std::string toString() const override { return "EdgeSwitchingInPlace"; }
```
avdgrinten, line 81: This will be deprecated in #631.

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+    bool isParallel() const override { return false; }
+
+protected:
+    // Necessary only for EdgeSwitch which will update the perturbedGraph pointer after calling
+    // its own constructor
+    EdgeSwitchingInPlace() = default;
+
+    Graph *graph = nullptr;
+    std::discrete_distribution<node>
+        degreeDistribution; ///< Return node x with probability proportional to degree(x)
```
avdgrinten, line 91: Strange spacing. Is this done by clang-format?

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    ~EdgeSwitching() override = default;
+
+    /**
+     * Move graph owned by the algorithm out.
+     * @warning Do not call run() after calling moveGraph()
+     */
+    Graph moveGraph() { return std::move(ownedGraph); }
+
+    std::string toString() const override { return "EdgeSwitching"; }
```
avdgrinten, line 130: See above.

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+    count getNumberOfAffectedEdges() const noexcept { return 2 * numberOfSwapsPerformed; }
+
+    // required by Algorithm
+    std::string toString() const override { return "EdgeSwitchingInPlace"; }
+    bool isParallel() const override { return false; }
+
+protected:
+    // Necessary only for EdgeSwitch which will update the perturbedGraph pointer after calling
+    // its own constructor
+    EdgeSwitchingInPlace() = default;
```
avdgrinten, line 87: Huh? Why does EdgeSwitching need this? Can't it just pass a reference to an internal Graph?

include/networkit/randomization/EdgeSwitching.hpp
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @note In general, simple edge switching does not yield a uniform distribution on simple DIRECTED
+ * graphs because the orientation of directed triangles cannot be changed. Using
+ * DegreePreservingShuffle as a preprocessing step overcomes this limitation. The preprocessing can
+ * also jump-start the perturbation process, yielding to potentially faster mixing.
+ *
+ * It's recommended to implement the preprocessing step in the calling code or to use copying
+ * EdgeSwitching implementation.
+ */
+class EdgeSwitching : public EdgeSwitchingInPlace {
```
avdgrinten, line 117: It is strange that this inherits from EdgeSwitchingInPlace. We should use composition, not inheritance here.

avdgrinten (commented):
Looks mostly good, some comments regarding the interface below.

manpen:
Thank you for your comments! I will address them as soon as possible.

include/networkit/randomization/EdgeSwitching.hpp
manpen: Yes; could be fixed by moving comment above the definition, this way looks clearer to me.

include/networkit/randomization/EdgeSwitching.hpp
manpen: Thank you for this comment. I did it to avoid redundant wrapper methods, but agree that composition looks much clearer and gets rid of the ugly graph-pointer/constructor hack.

include/networkit/randomization/EdgeSwitching.hpp
manpen: As I do not have a strong opinion either way, I implemented it as suggest.

manpen:
Thanks again for your review! I took care of all suggestions

avdgrinten (approved):
LGTM, thanks again for the PR and sorry for the delay in reviewing.

PR #446, opened 2019-11-04
Reduce NetworKit's dependencies

cndolo:
In this PR we try to reduce NetworKit's Python dependencies by making most packages optional.
If the package is not available when it is needed, a runtime error is thrown. 
The runtime error, `MissingDependencyError` specifies which particular package was needed.


avdgrinten:
One issue here is that notebooks (e.g., UserGuide) use optional dependencies (such as matplotlib). We need to work around that somehow.

cndolo:
> One issue here is that notebooks (e.g., UserGuide) use optional dependencies (such as matplotlib). We need to work around that somehow.

Additionally, almost all (if not all) the Python tests use optional dependencies.

.travis.yml
```diff
@@ -305,6 +305,8 @@ script:
  - export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}$(pwd)/core_build
  - NETWORKIT_PARALLEL_JOBS=2 python3 ./setup.py build_ext --inplace --networkit-external-core
  - NETWORKIT_PARALLEL_JOBS=2 pip3 install -e .
+ - python3 -c "import networkit"
```
avdgrinten: Please add a comment on why we add this line; it is not immediately obvious.

benchmark/Benchmark.py
```diff
@@ -8,40 +8,51 @@
...
 from util import *
+from networkit.support import MissingDependencyError
 import nk
 
 try:
 	import networkx
 	import nx
 except ImportError as ex:
-	print("networkx not available")
+	networkx = None
```
avdgrinten: What about the `nx` variable here? We should either consistently define all module symbols as `None` if the packages are not available, or consistently only set one symbol per package to `None`.

benchmark/Benchmark.py
```diff
@@ -8,40 +8,51 @@
...
 
 try:
 	import igraph
 	import ig
 except ImportError as ex:
-	print("igraph not available")
+	igraph = None
+	ig = None
```
avdgrinten: For example, here we set both `igraph` and `ig` to `None`.

benchmark/Benchmark.py
```diff
@@ -8,40 +8,51 @@
...
 import sys
 import warnings
 import math
 import os
 import numpy
-import matplotlib.pyplot as plt
-import seaborn
+try:
+	import matplotlib.pyplot as plt
+except ImportError as ex:
```
avdgrinten: The exception `ex` is not needed, right? Do not give it a name. (This applies to the other `except` statements, too.)

benchmark/Benchmark.py
avdgrinten: An alternative would be to have a Boolean `have_networkx` variable that we set to `False` if the import failed. Maybe that is the cleanest solution.

networkit/graphio.py
```diff
@@ -10,19 +10,17 @@
...
-except ImportError:
-	print("module 'numpy' not available - some functionality will be restricted")
-try:
-	import scipy.io
-except ImportError:
-	print("module 'scipy' not available - some functionality will be restricted")
-import fnmatch
+	import fnmatch
+except ImportError as importError:
+	fnmatch = None
```
avdgrinten: `fnmatch` is part of the Python standard library, so it must always be available.

networkit/profiling/stat.py
```diff
@@ -514,9 +519,9 @@ def funcHexBinning(sample_1, sample_2):
...
 			plt.close(fig)
 			return result
-		results["Binning"] = funcHexBinning(
-			sample_1,
-			sample_2
-		) 
+			results["Binning"] = funcHexBinning(
+				sample_1,
+				sample_2
+			)
```
avdgrinten: This looks like it changes the logic. Is that intended?

cndolo:
Kindly review the new changes as per what we discussed.

PR #449, opened 2019-11-12
Fix deprecation warnings with GCC 9.2

michitux:
This removes implementations of a copy constructor and an assignment
operator in viz/Octree and Point because if one implementation is
provided, the other (assignment operator/copy constructor) must be
provided, too. As far as I can see, none of the implementations did
anything non-default, so I reverted both to default implementations.

### DIFF
--- a/include/networkit/viz/Octree.hpp
+++ b/include/networkit/viz/Octree.hpp
@@ -39,7 +39,7 @@ public:
     /**
      *
      */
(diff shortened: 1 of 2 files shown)


michitux:
The continuous integration failures are completely unrelated to the changes (GCC installation problems on macOS, Github timeout on Windows).

avdgrinten:
Thank you for the PR. We could even delete the declarations entirely, but `= default` is also fine.

PR #451, opened 2019-11-18
Publish wheels

normanrz:
We need to pre-compile wheels for linux and mac, because we can't compile the library on our target machines. Therefore, I implemented building and publishing of the wheels as Github Actions (free for open-source). Essentially, this could replace the current travis builds.

Currently, we are publishing the wheels to our custom registry, but could easily be switched to the official pypi.org (with your credentials).

Please let me know if you find this PR useful. Thanks!

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -76,8 +76,11 @@ jobs:
           -w/app \
           quay.io/pypa/manylinux2010_x86_64 \
           bash -c "
-            \$PYBIN/python setup.py build_ext install
-            \$PYBIN/python setup.py sdist bdist_wheel -p manylinux2010_x86_64"
+            for PYBIN in /opt/python/*/bin; do
+              echo --> \$PYBIN
+              \$PYBIN/python setup.py build_ext install
+              \$PYBIN/python setup.py sdist bdist_wheel -p manylinux2010_x86_64
+            done"
     - name: Try to import on different python versions
       run: |
         docker run \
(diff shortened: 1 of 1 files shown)


avdgrinten:
Thank you for the PR. In general, it would be great to publish wheels but we don't have a lot of expertise in that area. One concern of mine is that NetworKit depends on libstdc++ and the version of libstdc++ might vary depending on the Linux distribution. Can Python wheels deal with this issue?

If not, we might want to build NetworKit using the `-static-libstdc++` option of GCC.

normanrz:
We build the bindings using the `manylinux2010` docker image which is made for creating wheels that can be installed on most linux distributions. The docker image includes fairly old versions of the system libraries with the rationale that newer libraries would be ABI-compatible. You can read more at https://www.python.org/dev/peps/pep-0571/. So, in theory the wheels should work on most linux machines. It worked well on the couple of different distributions that we tested on (Ubuntu, SLES).

Of course, the performance is best when you compile directly for the target platform. Users still have the option to do that, if they choose to.

This PR got a bit out of hand, because we added other features that we need. I'm happy to clean that up, if you want to move forward.

ulupo:
Publishing pre-compiled wheels might be a great way to enlarge the user base, in my opinion.

We opted for doing so in making [giotto-learn](https://github.com/giotto-ai/giotto-learn), and set up our own Azure pipelines (also free) using a manylinux2010 docker image as @normanrz describes, see the [configuration file](https://github.com/giotto-ai/giotto-learn/blob/master/azure-pipelines.yml).  With some effort, we could also support Mac OS and Windows systems. We haven't tried Github Actions but it looks like it has similar features.

hmeyerhenke:
Thanks, let's take a closer look in January...

fabratu:
I am closing the PR, since the ideas from here are adapted in #813. Thanks for the good and hard work!

PR #452, opened 2019-11-26
Graph::addEdge: fix directed in-edge getting the wrong id

michitux:
Note that "id" is a member variable that contains the id of the graph.


avdgrinten:
LGTM. The naming of the member variable is probably not a good idea, but we want to remove graph IDs anyway.

PR #454, opened 2019-11-28
Add Diff to Tooling

manpen:
This PR introduces more verbose output in code-style tooling:
- By default, all checkers now output diffs (can be disabled with `-q`)
- There's an optional verbose mode which outputs additional infos, e.g., the version reported by the clang-format binary

Additionally, `clang-format-8` is required to avoid incompatibilities as encountered by @angriman in #421. 
Finally, `check_code` now runs all tests and does not fail early. The rational is that before you were only shown the changes of the first failing test. If you then executes `-w`, previously not shown changes were performed.


extrafiles/tooling/nktooling/__init__.py
```diff
@@ -117,22 +149,53 @@ def isIdentical(self):
...
 		Writes are carried out atomically, i.e. we first write into a temporary file
 		and then replace the original.
-		If the module is currently readonly (which is the default, see setup) no change
-		is carried out.
+		If the module is currently readonly (which is the default, see setup),
+		the original file is not replaced.
 		This function may only be called if not reads are taking place currently (e.g.,
 		via lines())
+
+		If outputDiff is None, its value is copied from doShowDiff(). If it is True,
```
fabratu, line 162: A (small) typo in the code-documentation: It should be `doReportDiff()` instead of `doShowDiff()`.

manpen:
Thanks for approving and sorry for not responding earlier. I will fix the typo in the next couple of days as soon as I find the time. Same goes for #527.

manpen:
Thanks again for your review. I fixed a couple of typos in the comments, and did not change anything else.

PR #455, opened 2019-11-28
Deprecate Sampling

angriman:
`Sampling.hpp` defines three methods that are already implemented within GraphTools, and never used in NetworKit (after we merge #421). I think we should deprecate them now, and remove `Sampling` in future releases.


fabratu:
@angriman Please resolve the conflicts.

LGTM - especially since a variant of it is also deprecated in `Graph.hpp` and we should reduce the number of implementations (if they all do the same).

PR #456, opened 2019-11-30
Refactor: Strongly Connected Components

angriman:
This PR brings a cleaner implementation of Tarjan's algorithm for strongly connected components.
It also removes the deprecated method `getIthNeighbor` in Graph that was designed specifically for StronglyConnectedComponents to return the (deg(v) - i)-th neighbor of a node v (using neighbor iterators this operation is not needed anymore).

Furthermore, it includes more extensive C++ tests, and basic Python tests.

### DIFF
--- a/include/networkit/components/StronglyConnectedComponents.hpp
+++ b/include/networkit/components/StronglyConnectedComponents.hpp
@@ -1,70 +1,145 @@
 /*
- * StronglyConnectedComponents.h
+ * StronglyConnectedComponents.hpp
  *
  *  Created on: 01.06.2014
- *      Author: Klara Reichard (klara.reichard@gmail.com), Marvin Ritter (marvin.ritter@gmail.com)
+ *      Authors: Klara Reichard <klara.reichard@gmail.com>
+ *               Marvin Ritter <marvin.ritter@gmail.com>
+ *               Obada Mahdi <omahdi@gmail.com>
+ *               Eugenio Angriman <angrimae@hu-berlin.de>
  */
+
+// networkit-format
 
 #ifndef NETWORKIT_COMPONENTS_STRONGLY_CONNECTED_COMPONENTS_HPP_
 #define NETWORKIT_COMPONENTS_STRONGLY_CONNECTED_COMPONENTS_HPP_
 
+#include <map>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
 #include <networkit/graph/Graph.hpp>
 #include <networkit/structures/Partition.hpp>
+
+#include <tlx/define/deprecated.hpp>
 
 namespace NetworKit {
 
 /**
  * @ingroup components
- * Determines the strongly connected components of an directed graph.
+ *
  */
-class StronglyConnectedComponents {
+class StronglyConnectedComponents final : public Algorithm {
+
 public:
-    StronglyConnectedComponents(const Graph& G, bool iterativeAlgo=true);
+    /**
+     * Determines the strongly connected components of a directed graph using Tarjan's algorithm.
+     *
+     * @param G Graph A directed graph.
+     */
+    StronglyConnectedComponents(const Graph &G);
 
     /**
-     * This method determines the connected components for the graph g
-     * (by default: iteratively).
+     * Determines the strongly connected components of a directed graph using Tarjan's algorithm.
+     *
+     * @param G Graph A directed graph.
+     * @param iterativeAlgo bool Whether to use the iterative algorithm or the recursive one.
+     *
+     * This constructor has been deprecated because the recursive algorithm has been deprecated.
      */
-    void run();
+    TLX_DEPRECATED(StronglyConnectedComponents(const Graph &G, bool iterativeAlgo));
+
+    /**
+     * Runs the algorithm.
+     */
+    void run() override;
 
     /**
      * This method determines the connected components for the graph g
      * (iterative implementation).
+     *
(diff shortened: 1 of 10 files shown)


networkit/cpp/centrality/TopCloseness.cpp
```diff
@@ -148,8 +148,8 @@ void TopCloseness::computeReachableNodesDir() {
   }
 
   for (count v = 0; v < n; v++) {
-    reachL[v] = reachL_scc[sccs.componentOfNode(v) - 1];
-    reachU[v] = reachU_scc[sccs.componentOfNode(v) - 1];
```
fabratu, line 152: It is more a question than direct feedback: Does omitting `-1` has side-effects in other parts of the code. `TopHarmonicCloseness.cpp` still refers to sccs with `-1`. In `CutClustering.cpp` connected components (without strongly) are used without the index-shift. At least there is some inconsistency in how `componentOfNode` is used for different types of components.

include/networkit/components/StronglyConnectedComponents.hpp
```diff
@@ -1,70 +1,143 @@
...
 
+        return result;
+    }
 
 private:
-    const Graph& G;
+    const Graph &G;
     bool iterativeAlgo;
-    Partition component;
+    std::vector<index> component;
```
fabratu, line 137: Maybe not for this PR - but if changing the datatype is the preferred solution for `component` - I think we should refactor all other component-types (or refactor the header-inheritance since they are all quite similar).

include/networkit/components/StronglyConnectedComponents.hpp
```diff
@@ -1,70 +1,143 @@
...
+    }
+
+    /**
+     * Returns the component of the input node.
+     *
+     * @param[in] u The input node.
+     * @return count The index of the component of the input node.
      */
-    void runRecursively();
+    index componentOfNode(node u) const {
```
fabratu, line 88: Remark: Some getter-functions have a verb-based suffix, some have not. Is there any agreement on what should be used? Probably outside the scope of the PR, but if there is - maybe we should align function-names on refactoring?

networkit/cpp/centrality/TopCloseness.cpp
angriman: I checked other parts of the codebase, `TopCloseness` was the only one to shift the component ids for `StronglyConnecteedComponents`. This was done because the lowest component id from `StronglyConnectedComponents` was 1. Now it is 0 (same as `ConnectedComponents`), so the id shift is not necessary anymore.

include/networkit/components/StronglyConnectedComponents.hpp
angriman: Probably, I find `std::vector` to be a simpler choice here, and it does not restrict the features of the class.

include/networkit/components/StronglyConnectedComponents.hpp
avdgrinten: In general, I'd say that verbs are preferred but sometimes there is just no good verb (getComponentOfNode doesn't really add anything). Renaming stuff certainly has a high cost that we should not pay here (we'd need to deprecate the old method, introduce a new one, users need to update their code etc.).

networkit/cpp/centrality/TopCloseness.cpp
fabratu: Ok, I understand. However on branch `Dev` there is at least one reference  for  `-1` in `TopHarmonicCloseness` (but maybe this is intended then). 
https://github.com/networkit/networkit/blob/Dev/networkit/cpp/centrality/TopHarmonicCloseness.cpp#L467

include/networkit/components/StronglyConnectedComponents.hpp
fabratu: Good points.

networkit/cpp/centrality/TopCloseness.cpp
angriman: Yes I addressed that in this PR.

fabratu: Oh boy, somehow I missed that. Thanks for clarification

fabratu:
LGTM.

include/networkit/components/StronglyConnectedComponents.hpp
```diff
@@ -1,70 +1,143 @@
...
-     * This method returns the the component in which node query is situated.
+     * Return a map with the component indexes as keys, and their size as values.
      *
-     * @param[in]	query	the node whose component is asked for
+     * @return std::map<index, count> Map with components indexes as keys, and their size as values.
      */
-    count componentOfNode(node u);
+    std::map<node, count> getComponentSizes() const {
+        assureFinished();
+        std::map<node, count> map;
```
michitux, line 111: As component ids are continues, wouldn't it be a lot more efficient to use `std::vector<count>` here? Or is this just to be compatible with the `Partition` interface? But even then, it should be more efficient to first aggregate the values in a `std::vector<count>` and only then convert this to a `std::map<node, count>`.

angriman: I am using a `std::map<node, count>` to be compatible with both `Partition` and `ConnectedComponents`, we might change those APIs or provide new ones in a separate PR. Apart from that it's true that using a `std::vector<count>` to aggregate values would be better, I implemented this change.

PR #457, opened 2019-12-02
Ged walk centrality

angriman:
This PR brings the implementation of the algorithms for GedWalk centrality based on the paper [Group Centrality Maximization for Large-scale Graphs](https://arxiv.org/abs/1910.13874) (ALENEX 2020).

Updating tlx is needed because the algorithms use some features introduced in a newer tlx version.

### DIFF
--- a/include/networkit/centrality/GedWalk.hpp
+++ b/include/networkit/centrality/GedWalk.hpp
@@ -1,0 +1,220 @@
+/* GedWalk.hpp
+ *
+ *  Created on: 18.6.2018
+ *     Authors: Eugenio Angriman <angrimae@hu-berlin.de>
+ *              Alexander van der Grinten <avdgrinten@hu-berlin.de>
+ */
+
+// networkit-format
+
+#ifndef NETWORKIT_CENTRALITY_GED_WALK_HPP_
+#define NETWORKIT_CENTRALITY_GED_WALK_HPP_
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+#include <tlx/container/d_ary_addressable_int_heap.hpp>
+
+namespace NetworKit {
+
+class GedWalk final : public Algorithm {
+
+    using walks = double;
+
+public:
+    enum GreedyStrategy { lazy, stochastic };
+
+    enum BoundStrategy { no, spectral, geometric, adaptiveGeometric };
+
+    double stocEpsilon = 0.1;
+
+private:
+    const Graph *G;
+    const count k;
+    const double epsilon;
+    double alpha;
+    const BoundStrategy boundStrategy;
+    GreedyStrategy greedyStrategy;
+    double spectralDelta;
+    count degOutMax;
+    count degInMax;
+    std::vector<double> alphas;
+
+    // Size of pathsIn / pathsOut and related vectors.
+    count allocatedLevels = 15;
+
+    double sigmaMax;
+
+    // Similar to gainScore, gainBound and gainW, but for the entire group.
+    // These values are always maintained exactly.
+    double groupScore; // Score of the group.
+    walks groupW;      // Number of walks on the last level.
+    walks groupBound;  // Upper bound on the score of the group.
+
+    walks graphW;
+
+    count nLevels = 2;
+    std::vector<node> group;
+    std::vector<unsigned char> inGroup;
+    std::vector<unsigned char> isExact;
+    std::vector<unsigned char> nodesToPick;
+    std::vector<std::vector<walks>> pathsIn, pathsOut;
+    std::vector<std::vector<walks>> pathsHit, pathsMiss;
+
+    // Unless isExact[x], these values might be upper bounds on the true values.
+    std::vector<double> gainScore; // Marginal score of the group.
+    std::vector<walks> gainW;      // Number of marginal walks on the last level.
+    std::vector<double> gainBound; // Upper bound on the marginal score.
+
+    struct CompareScore {
+    public:
+        CompareScore(const std::vector<double> &score) : score(&score) {}
+        bool operator()(node x, node y) const { return (*score)[x] > (*score)[y]; }
+
+    private:
+        const std::vector<double> *score;
+    };
+
+    struct EvaluationResult {
+        double score;
+        walks w;
+    };
+
+    tlx::d_ary_addressable_int_heap<node, 2, CompareScore> scoreQ{CompareScore(gainScore)};
+    tlx::d_ary_addressable_int_heap<node, 2, CompareScore> boundQ{CompareScore(gainBound)};
+
+    void init();
+    double computeGamma();
+    void estimateGains();
+    void computeMarginalGain(node z);
+    void maximizeGain();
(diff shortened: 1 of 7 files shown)


include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include <networkit/graph/Graph.hpp>
+
+#include <tlx/container/d_ary_addressable_int_heap.hpp>
+
+namespace NetworKit {
+
+class GedWalk final : public Algorithm {
+
+    using walks = double;
+    using uchar = unsigned char;
```
avdgrinten: Non-idiomatic typedef. Please remove this.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+    };
+
+    struct EvaluationResult {
+        double score;
+        walks w;
+    };
+
+    tlx::d_ary_addressable_int_heap<node, 2, CompareScore> scoreQ{CompareScore(gainScore)};
+    tlx::d_ary_addressable_int_heap<node, 2, CompareScore> boundQ{CompareScore(gainBound)};
+    CSRMatrix gMat;
```
avdgrinten: I think it makes more sense to use a graph here. Yes, a CSR will give slightly better performance, but it requires an entire copy of the graph.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+    std::vector<walks> gainW;      // Number of marginal walks on the last level.
+    std::vector<double> gainBound; // Upper bound on the marginal score.
+
+    struct CompareScore {
+    public:
+        CompareScore(const std::vector<double> &score) : score(score) {}
+        bool operator()(node x, node y) const { return score[x] > score[y]; }
+
+    private:
+        const std::vector<double> &score;
```
avdgrinten: Member variables should not be references.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+    EvaluationResult evaluateGraph();
+    EvaluationResult evaluateGroup();
+
+    template <typename T>
+    void parallelFill(std::vector<T> &vec, const T &val) {
+#pragma omp parallel for
+        for (omp_index i = 0; i < static_cast<omp_index>(vec.size()); ++i) {
+            vec[i] = val;
+        }
+    }
```
avdgrinten: Does this actually improve performance? I would expect that there is too little work done in this loop to make parallelization worthwhile (over auto-vectorization / lowering to `memset()` which should happen anyway).

It adds some maintainance cost as there is mental overhead to remember what this function does (compared to the open-coded `for` loop that ones expects for this purpose), so maybe it would be better to open-code it.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Return the GedWalk score of the input group.
+     *
+     * @param first/last The range that contains the input group.
+     * @param scoreEpsilon The precision of the score to be computed.
+     */
+    template <class InputIt>
+    double scoreOfGroup(InputIt first, InputIt last, double scoreEpsilon = .1);
```
avdgrinten: Is this named consistently with the other group centralities?

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+     * Run the algorithm.
+     */
+    void run() override;
+
+    /**
+     * Return the approximate GedWalk score of the computed group.
+     *
+     * @return The approximate score of the computed group.
+     */
+    double getApxScore() const {
```
avdgrinten: This name is confusing.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,231 @@
...
+     */
+    template <class InputIt>
+    double scoreOfGroup(InputIt first, InputIt last, double scoreEpsilon = .1);
+
+    /**
+     * Return the computed group.
+     *
+     * @return The computed group.
+     */
+    std::vector<node> groupMaxGedWalk() const {
```
avdgrinten: Is this named consistently with the other group centralities?

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -244,8 +245,8 @@ TEST_F(CentralityGTest, testKatzDynamicAddition) {
     DEBUG("finish kc");
     node u, v;
     do {
-        u = GraphTools::randomNode(G);
-        v = GraphTools::randomNode(G);
+        u = G.randomNode();
+        v = G.randomNode();
```
avdgrinten: Random unrelated changes to the PR. Please remove this.

include/networkit/centrality/GedWalk.hpp
angriman: Yes, but we might want to change this in the future if we want to build an abstract GroupCentrality class.

include/networkit/centrality/GedWalk.hpp
angriman: Yes

include/networkit/centrality/GedWalk.hpp
angriman: I removed it.

include/networkit/centrality/GedWalk.hpp
angriman: Alright now GedWalk does not use an additional CSR matrix anymore.

include/networkit/centrality/GedWalk.hpp
angriman: Removed it in favor of std::fill

include/networkit/centrality/GedWalk.hpp
angriman: Renamed to `getApproximateScore`

networkit/cpp/centrality/test/CentralityGTest.cpp
angriman: That was a mistake, I removed this change.

angriman:
Thanks @avdgrinten for the review, I've addressed all the comments.

include/networkit/centrality/GedWalk.hpp
```diff
@@ -0,0 +1,220 @@
...
+        return group;
+    }
+
+    const std::vector<double> &boundOfMarginalGains() const { return gainBound; }
+};
+
+template <class InputIt>
+double GedWalk::scoreOfGroup(InputIt first, InputIt last, double scoreEpsilon) {
+    if (boundStrategy == BoundStrategy::spectral) {
+        assert(alpha < 1 / static_cast<double>(sigmaMax));
```
fabratu, line 160: Only a very small remark: Depending on the value of `sigmaMax` and `degInMax`, the assertion can benefit (both in precision and clocktime) if compare `alpha * static_cast<double>(something)` to `1`.

PR #458, opened 2019-12-03
Remove deprecated features

angriman:
This PR removes the methods that were deprecated in #421. `Graph::nodes/edges` are handled in #462 because they are replaced by new iterators for nodes and edges, which should be provided before `Graph::nodes/edges` are removed.


avdgrinten:
This should wait for one release cycle.

avdgrinten:
The new iterators *should* be split to a new PR so that we can merge them in the current release cycle.

angriman:
The iterators commits have been moved to #462.

avdgrinten:
Did you verify that everything that is removed in this PR has been deprecated for two major releases?

angriman:
The recursive algorithm for SCCs has been deprecated in NetworKit 6.0, all the remaining removed features have been deprecated in NetworKit <= 5.0.

networkit/__init__.py
```diff
@@ -134,7 +134,7 @@ def getComponentPartition(G):
 			cc = components.ConnectedComponents(G).run()
 		return cc.getPartition()
 
-	print("Network Properties for:\t\t{}".format(G.getName()))
+	print("Network Properties:")
```
fabratu, line 137: Is G.getName() dropped completely?

angriman: Yes, I think that line is still file, the remaining network properties will be listed below.

include/networkit/generators/DynamicGraphSource.hpp
```diff
@@ -74,7 +76,7 @@ class DynamicGraphSource {
     /**
      * Continue generating until the number of time steps reaches this upper limit.
      */
-    virtual void generateTimeSteps(count t);
+    virtual void TLX_DEPRECATED(generateTimeSteps(count t));
```
avdgrinten, line 79: Why is this deprecated now?

input/example2.dgs
```diff
@@ -3,7 +3,6 @@ authorGraph 0 0
 an 1
 an 2
 ae 1-2 1 2 weight=1
-st 1
```
avdgrinten, line 6: Huh?

networkit/GEXFIO.py
```diff
@@ -105,15 +105,10 @@ def read(self, fpath):
 
 		#5.2 Sorting the eventStream by time and adding timeStep between events that happen in different times
 		self.eventStream.sort(key=lambda x:x[1])
-		for i in range(1, len(self.eventStream)):
-			if self.eventStream[i][1] != self.eventStream[i-1][1]:
-				self.eventStream.append((GraphEvent(GraphEvent.TIME_STEP, 0, 0, 0), self.eventStream[i-1][1]))
-		self.eventStream.sort(key=lambda x:x[1])
```
avdgrinten, line 111: Why is this removed?

networkit/cpp/dynamics/DGSStreamParser.cpp
```diff
@@ -58,9 +58,7 @@ std::vector<GraphEvent> DGSStreamParser::getStream() {
             std::string tag = split[0];
 
             // parse commands
-            if (tag.compare("st") == 0) { // clock
-                stream.emplace_back(GraphEvent::TIME_STEP);
-            } else if (tag.compare("an") == 0) { // add node
+            if (tag.compare("an") == 0) { // add node
```
avdgrinten, line 61: Why are these changes done?

networkit/cpp/dynamics/DGSStreamParser.cpp
```diff
@@ -120,9 +118,7 @@ std::vector<GraphEvent> DGSStreamParser::getStream() {
...
```
avdgrinten, line 121: Why are these changes done?

avdgrinten (commented):
There are several changes related to time steps. We should have a replacement available for the functionality that is removed here.

angriman:
Those changes are done because timestamps are not supported anymore as already discussed. What replacement are you referring to?

avdgrinten:
That they are not supported in `Graph` anymore does not mean that we have to drop support entirely. What are the timestamps good for / what is the use case?

angriman:
Time steps are not used by any algorithm in NetworKit so far but I see the point of not dropping support entirely as they might be useful for dynamic graphs. However, I don't see an easy way to do that as time steps are graph events and thus inherently tied to `Graph`. If we move time steps from `Graph` how should they be managed?

avdgrinten:
If they are not used in any algorithm, it might be fine to remove the feature. However, support in graph readers etc. should be kept.

hmeyerhenke:
Time stamps were meant to be used in dynamic algorithms, yes. I think in the end we realized the dynamic algorithms realized as part of NetworKit in a different way. Not sure if that is true for user code, though (but if nobody complained about deprecation...). Keeping support in graph readers still makes sense, I guess.

include/networkit/generators/DynamicGraphSource.hpp
angriman: I forgot to deprecate this before, the implementation is this
```
void DynamicGraphSource::generateTimeSteps(count t) {
     while (G->time() < t) {
         this->generate();
     }
     G->shrinkToFit();
  }
```
which is something that we cannot support anymore if `time` is removed from `Graph`.

michitux:
I have implemented a dynamic graph generator in NetworKit (see https://github.com/michitux/networkit/tree/ckbdynamic) that uses time steps. The generator has the concept that in each time step a set of events happens. After each time step, a community detection algorithm should detect a given (also dynamic) reference community structure. I therefore use time steps to mark the point where the communities should be detectable, and the idea for plots would be that you have these time steps on one axis. None of my code relies on the graph knowing which time step it represents, but it certainly relies on time step events.

angriman:
Alright, so I removed the time steps only from `Graph`, but I kept the support for time step events.

include/networkit/generators/DynamicGraphSource.hpp
fabratu: I think given the total amount of changes, it is totally fine to miss something - but IMHO I would remove the function directly. As of now it does nothing than compacting the graph, which is misleading.

angriman: That's true, OTOH removing it might break some other people's code so probably deprecating it is the safest solution.

avdgrinten:
Sounds reasonable. Maybe timestamps would fit well into a framework for arbitrary data ("attributes") attached to graphs/nodes/edges (as people suggested support for such a feature at ND last week).

Maybe the easiest path forward is to split the time step stuff into a different PR and to merge the remainder of the PR independently.

angriman:
Agree, I dropped the time steps commit.

include/networkit/generators/DynamicGraphSource.hpp
fabratu: Even though it is not interesting for this PR, but (a more general opinion):

I would prefer breaking code over unexpected behavior (and you don't know what caused it). Or at least mention here, that the function might not do what you think.

avdgrinten (approved):
LGTM.

PR #460, opened 2019-12-13
Refactor components module

cndolo:
This PR brings some improvements to the C++ code of the components module. The changes include:
 - All classes that are not inherited from are made final.
- The `virtual` keyword is removed from any non-base classes.
- All __member__ variables are changed from references to pointers. The point of this is to make classes assignable.
- The `override` keyword is added to functions accordingly.

### DIFF
--- a/include/networkit/components/BiconnectedComponents.hpp
+++ b/include/networkit/components/BiconnectedComponents.hpp
@@ -65,7 +65,7 @@ private:
   void init();
   void newComponent(std::pair<node, node> e);
 
-  const Graph &G;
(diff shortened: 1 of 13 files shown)


avdgrinten:
For the general direction of this PR, we should also make sure that other major contributors know what we're up to: @manpen @michitux . The idea here is that most `Algorithm` implementations are currently not assignable, i.e., `ConnectedComponents x{G}; x = y;` fails. IMHO, this is contrary to modern C++ expectations. Hence the change to replace member variables that are references by pointers.

PR #463, opened 2019-12-18
Deprecate workflows.extractLargestComponent

angriman:
This PR deprecates the Python function `workflows.extractLargestComponent` as the same algorithm is already implemented in C++ in `ConnectedComponents::extractLargestConnectedComponent`.

### DIFF
--- a/networkit/workflows.py
+++ b/networkit/workflows.py
@@ -26,14 +26,9 @@ def extractLargestComponent(G):
 		Subgraph of largest component, preserving node ids of orignal graph.
 	"""
 
-	cc = nk.components.Connectednk.components(G)
-	cc.run()
-	cSizes = cc.getnk.componentsizes()
-	(largestCompo, size) = max(cSizes.items(), key=operator.itemgetter(1))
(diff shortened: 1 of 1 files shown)


networkit/workflows.py
```diff
@@ -26,14 +26,9 @@ def extractLargestComponent(G):
...
-	cc.run()
-	cSizes = cc.getnk.componentsizes()
-	(largestCompo, size) = max(cSizes.items(), key=operator.itemgetter(1))
-	logging.info("extracting component {0} containing {1} nodes".format(largestCompo, size))
-	compoNodes = [v for v in G.nodes() if cc.componentOfNode(v) is largestCompo]
-	C = G.subgraphFromNodes(compoNodes)
-	return C
+	from warnings import warn
+	warn("This function is deprecated, use extractLargestConnectedComponent in the ConnectedComponents module instead.")
+	return nk.components.testExtractLargestComponent(G)
```
fabratu, line 31: It should be `nk.components.extractLargestComponent(G)`, right? Also the implementation in C++ could benefit from some improvements (but outside the scope of this PR).

angriman: Yes, thank you.

Issue #469, opened 2020-01-02
inNeighbors() fails on directed networks.


Nchudson95:
Hello, below is a sample program that reproduces the error.
```
import networkit as nk

g = nk.graph.Graph(directed=True)
g.addNode()
g.addNode()
g.addEdge(0,1)

print(g.nodes())        # "[0, 1]"
print(g.inNeighbors(0)) # "[]"
print(g.inNeighbors(1)) # Error here!
```
On the final line of code, when node 1's in-neighbors are requested, I get the following error:
```
Traceback (most recent call last):
  File "Untitled.py", line 10, in <module>
    print(g.inNeighbors(1))
  File "networkit/_NetworKit.pyx", line 1061, in networkit._NetworKit.Graph.inNeighbors
  File "networkit/_NetworKit.pyx", line 1154, in networkit._NetworKit.Graph.forInEdgesOf
RuntimeError: An Exception occurred, aborting execution of iterator: lambda1() takes exactly one argument (4 given)
```
I've been playing around with this for a while and I've been unable to get it to work. Just thought I would bring it up so it can be addressed unilaterally soon.

angriman:
Hi, this is a bug of `inNeighbors`, thank you for bringing this to our attention. I've addressed this issue in #470.

Before we merge it, you can use iterators, here is an example:
```
import networkit as nk

g = nk.graph.Graph(2, directed=True)
g.addEdge(0, 1)

inNeighbors = []

# Iterate over in-neighbors of node 1
for u in g.iterInNeighbors(1):
    inNeighbors.append(u)

print(inNeighbors)
```

Nchudson95:
Thank you! I figured out how to use `iterInNeighbors()` to circumvent that bug, but wanted to make sure that this issue was noticed by those working on the project. Thank you!

angriman:
The patch has been merged, I am closing this issue.

PR #470, opened 2020-01-02
Fix inNeighbors

angriman:
This PR fixes the bug in the Python code of `Graph.inNeighbors` described in #469. It also includes new Python tests for `inNeighbors`.


angriman:
Looks like @cndolo does not have write access and therefore her approval is not enough for merging. @avdgrinten can you review this PR? It is a really basic bug fix.

PR #471, opened 2020-01-08
Fix extract largest cc

angriman:
This PR fixes a bug in `extractLargestConnectedComponents`: connected graphs are never compacted, even if `compactGraph` is set to true. Additional tests for this scenario are also included.

### DIFF
--- a/networkit/cpp/components/ConnectedComponents.cpp
+++ b/networkit/cpp/components/ConnectedComponents.cpp
@@ -84,44 +84,49 @@ std::map<index, count> ConnectedComponents::getComponentSizes() const {
 }
 
 Graph ConnectedComponents::extractLargestConnectedComponent(const Graph &G, bool compactGraph) {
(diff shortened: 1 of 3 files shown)


networkit/cpp/components/ConnectedComponents.cpp
```diff
@@ -93,6 +93,9 @@ Graph ConnectedComponents::extractLargestConnectedComponent(const Graph &G, bool
 
     const auto compSizes = cc.getComponentSizes();
     if (compSizes.size() == 1) {
+        if (compactGraph) {
+            GraphTools::getCompactedGraph(G, GraphTools::getContinuousNodeIds(G));
```
avdgrinten: This doesn't return. Why is this not caught by tests?

angriman: The new tests were failing but because of another reason. I fixed that now.

PR #473, opened 2020-01-10
LS-Restrict algorithm for Group Closeness 

angriman:
This PR brings the LS-restrict and the Grow-Shrink algorithms presented in the paper "Local Search for Group Closeness Maximization on Big Graphs" presented at IEEE BigData 2019.

### DIFF
--- a/include/networkit/centrality/GroupClosenessLocalSwaps.hpp
+++ b/include/networkit/centrality/GroupClosenessLocalSwaps.hpp
@@ -1,0 +1,114 @@
+/*
+ *  GroupClosenessLocalSwaps.hpp
+ *
+ *  Created on: 19.12.2019
+ *      Author: Eugenio Angriman <angrimae@hu-berlin.de>
+ */
+
+// networkit-format
+
+#ifndef NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SWAPS_HPP_
+#define NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SWAPS_HPP_
+
+#ifdef AVX2_AVAILABLE
+#include <immintrin.h>
+#include <networkit/auxiliary/AlignedAllocator.hpp>
+#endif // AVX2_AVAILABLE
+
+#include <functional>
+#include <limits>
+#include <random>
+#include <stdexcept>
+#include <unordered_map>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class GroupClosenessLocalSwaps final : public Algorithm {
+
+    static constexpr float maxInt16 = static_cast<float>(std::numeric_limits<uint16_t>::max());
+    // Number of random numbers generated at once
+    static constexpr count K = 16;
+
+public:
+    GroupClosenessLocalSwaps(const Graph &G, const std::vector<node> &group, count maxSwaps = 100);
+
+    /**
+     * Finds a group of nodes with high group closeness centrality. This is the LS-restrict
+     * algorithm presented in Angriman et al. "Local Search for Group Closeness Maximization on Big
+     * Graphs" IEEE BigData 2019. The algorithm takes as input a graph and an arbitrary group of
+     * nodes, and improves the group closeness of the given group by performing vertex exchanges.
+     *
+     * @param G A connected, undirected, and unweighted graph.
+     * @param first, last A range that contains the initial group of nodes.
+     * @param maxSwaps Maximum number of vertex exchanges allowed.
+     */
+    template <class InputIt>
+    GroupClosenessLocalSwaps(const Graph &graph, InputIt first, InputIt last, count maxSwaps = 100)
+        : G(&graph), group(first, last), maxSwaps(maxSwaps) {
+
+        if (G->isDirected())
+            throw std::runtime_error("Error, this algorithm does not support directed graphs.");
+
+        if (group.empty())
+            throw std::runtime_error("Error, empty group.");
+
+        if (G->isWeighted())
+            WARN("This algorithm does not support edge Weights, they will be ignored.");
+    }
+
+    /**
+     * Runs the algorithm.
+     */
+    void run() override;
+
+    /**
+     * Returns the computed group.
+     */
+    std::vector<node> groupMaxCloseness() const;
+
+    /**
+     * Returns the total number of vertex exchanges performed by the algorithm.
+     */
+    count numberOfSwaps() const;
+
+private:
+    const Graph *G;
+    std::vector<node> group, stack;
+    std::vector<uint32_t> distance, sumOfMins;
+    std::vector<bool> gamma, canSwap, visited;
+    std::unordered_map<node, index> idxMap;
+    std::vector<int64_t> value, valueDecrement;
+
+    count maxSwaps, totalSwaps, stackSize;
+
+    void init();
+    void bfsFromGroup();
+    bool findAndSwap();
+    node estimateHighestDecrement();
+    void initRandomVector();
+    int64_t computeFarnessDecrement(node u);
+    void resetGamma(node x, index idx);
+
+    // Returns true if the vertex at the give index is the only nearest vertex to x
+    bool gammaIsU(node x, index idx) const;
+
+#ifdef AVX2_AVAILABLE
+    union RandItem {
+        uint16_t items[K];
+        __m256i vec;
+    };
+    std::vector<RandItem, AlignedAllocator<RandItem, sizeof(RandItem)>> randVec;
+#else
+    std::vector<uint16_t> randVec;
+#endif // AVX2_AVAILABLE
+
+    std::vector<std::uniform_int_distribution<uint32_t>> intDistributions;
+    std::vector<std::reference_wrapper<std::mt19937_64>> urngs;
+};
+} // namespace NetworKit
+
+#endif // NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SWAPS_HPP_
--- a/include/networkit/centrality/GroupClosenessLocalSwaps.hpp
+++ b/include/networkit/centrality/GroupClosenessLocalSwaps.hpp
@@ -10,14 +10,12 @@
 #ifndef NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SWAPS_HPP_
 #define NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SWAPS_HPP_
 
-#ifdef AVX2_AVAILABLE
+#ifdef __AVX2__
 #include <immintrin.h>
 #include <networkit/auxiliary/AlignedAllocator.hpp>
-#endif // AVX2_AVAILABLE
+#endif // __AVX2__
 
-#include <functional>
 #include <limits>
-#include <random>
 #include <stdexcept>
 #include <unordered_map>
 #include <vector>
@@ -79,24 +77,21 @@ private:
     const Graph *G;
     std::vector<node> group, stack;
     std::vector<uint32_t> distance, sumOfMins;
-    std::vector<bool> gamma, canSwap, visited;
+    std::vector<bool> gamma, visited;
     std::unordered_map<node, index> idxMap;
-    std::vector<int64_t> value, valueDecrement;
+    std::vector<int64_t> farness, farnessDecrease;
 
     count maxSwaps, totalSwaps, stackSize;
 
     void init();
     void bfsFromGroup();
     bool findAndSwap();
-    node estimateHighestDecrement();
+    node estimateHighestDecrease();
     void initRandomVector();
-    int64_t computeFarnessDecrement(node u);
+    int64_t computeFarnessDecrease(node u);
     void resetGamma(node x, index idx);
 
-    // Returns true if the vertex at the give index is the only nearest vertex to x
-    bool gammaIsU(node x, index idx) const;
-
-#ifdef AVX2_AVAILABLE
+#ifdef __AVX2__
     union RandItem {
         uint16_t items[K];
         __m256i vec;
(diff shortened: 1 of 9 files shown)


angriman:
The new algorithms also support AVX instructions, which are used if the `NETWORKIT_NATIVE` flag is set to `ON`, and if they are available on the machine. This operation can be done easily on Linux and MacOS, but I could not find a solution on Windows. Does anyone know how to do this on Windows?

CMakeLists.txt
```diff
@@ -95,6 +95,54 @@ else()
...
+		set(AVX_512vl_TRUE ${AVX_FOUND} PARENT_SCOPE)
+	endfunction(check_avx512vl)
+
+	# Check if AVX instructions are supported on host. If yes, enable the required AVX flags.
+	function(set_avx_flags CPUINFO)
+		check_avx512f(${CPUINFO})
+		check_avx512vl(${CPUINFO})
+
+		if (AVX_512f_TRUE AND AVX_512vl_TRUE)
+			set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} -mavx -mavx512vl -mavx512f -DNETWORKIT_WITH_AVX")
```
fabratu, line 124: I wonder if the support for `AVX512` necessary here. You use 256bit-width variables (`__m256i`), which are also supported by `AVX2` ([Intrinsics Guide from Intel](__m256i)). This would lead to many more users/platforms able to use it. Does 512 vs 2 result in a speed-up?

Disclaimer:
Not an expert on this subject, but the common sense (at least on Skylake microarchitecture) is, that a speed-up in real-life applications is only expected if 512-bit width variables are used. And even then it is sublinear, since `AVX512` functions have a higher latency.

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
```diff
@@ -0,0 +1,168 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    typedef struct {
+        uint64_t state;
+        uint64_t inc;
+    } pcg32_random_t;
```
fabratu, line 152: Based on the discussion in PR #415, is PCG still needed?

angriman: No, I will remove it.

include/networkit/centrality/GroupClosenessGrowShrink.hpp
```diff
@@ -0,0 +1,665 @@
+/*
```
fabratu, line 1: Whats the rationale behind implementing `GrowClosenessGrowShrink` and `Dijkstra` as header only?

include/networkit/centrality/GroupClosenessGrowShrink.hpp
```diff
@@ -0,0 +1,665 @@
...
+            throw std::runtime_error("Error, empty group.");
+        }
+    }
+
+    // Easier constructor to wrap in Cython.
+    GroupClosenessGrowShrinkGeneral(const Graph &graph, const std::vector<node> &group,
+                                    bool extended = false, count insertions = 0)
+        : GroupClosenessGrowShrinkGeneral(graph, group.begin(), group.end(), extended, insertions) {
+    }
+
```
fabratu, line 67: Please add a default-destrctor.

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
```diff
@@ -0,0 +1,168 @@
...
+        }
+        if (G->isWeighted()) {
+            WARN("This algorithm does not support edge Weights, they will be ignored.");
+        }
+    }
+
+    // Easier constructor to wrap in Cython.
+    GroupClosenessLocalSwaps(const Graph &graph, const std::vector<node> &group)
+        : GroupClosenessLocalSwaps(graph, group.begin(), group.end()) {}
+
```
fabratu, line 54: Please add a default-destructor.

include/networkit/centrality/GroupClosenessGrowShrink.hpp
```diff
@@ -0,0 +1,665 @@
...
+            std::fill(visited.begin(), visited.end(), 0);
+        }
+    }
+
+    typedef struct {
+        uint64_t state;
+        uint64_t inc;
+    } pcg32_random_t;
+
+    std::vector<pcg32_random_t> rng;
```
fabratu, line 195: Same question about PCG here.

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
avdgrinten: Why?

include/networkit/centrality/GroupClosenessGrowShrink.hpp
angriman: Because they are templated, moving the implementation to a source file would yield compile-time errors.

avdgrinten: A templated Dijskstra is reasonable but why do we need a templated Grow-Shrink?

include/networkit/centrality/GroupClosenessGrowShrink.hpp
avdgrinten: Why?

include/networkit/centrality/GroupClosenessGrowShrink.hpp
angriman: To handle weighted/unweighted graphs. I wanted to specialize the code for the two version to avoid doing many run-time checks for `G.isWeighted()` .

include/networkit/centrality/GroupClosenessGrowShrink.hpp
```diff
@@ -0,0 +1,665 @@
...
+     * Runs the algorithm.
+     */
+    void run() override {
+        init();
+
+        if (!insertions) {
+            Diameter diam(*G, DiameterAlgo::estimatedRange, 0.1);
+            diam.run();
+            DEBUG("Diameter upper bound = ", diam.getDiameter().second);
+            insertions = std::max(1., .5
```
fabratu, line 106: Just to be sure: Is the truncation from `double` to `int` intended?

angriman: Yes, as we want the number of insertions to be an integer >= 1.

include/networkit/centrality/GroupClosenessGrowShrink.hpp
avdgrinten: As far as I can see, this does not remove the need to perform runtime checks, it only specializes the code for `count` weights instead of `double` weights. I'm not sure if that is really worthwhile (double math should only be slightly slower than integer math). If you deem it worthwhile, use the pimpl idiom to avoid putting the entire algorithm into the header.

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
```diff
@@ -0,0 +1,168 @@
...
+        uint16_t items[16];
+        __m256i vec;
+    };
+    std::vector<RandItem> randVec;
+#else
+    std::vector<uint16_t> randVec;
+#endif // NETWORKIT_WITH_AVX
+
+    void resetGamma(node x, index idx) {
+        std::fill(gamma.begin() + group.size() * x, gamma.begin() + group.size() * (x + 1), 0);
```
fabratu, line 123: Is it safe, that `gamma.begin() + group.size() * (x + 1)` will not go oob?

angriman: Yes because `gamma` has size `n * group.size()`, and `x` is always at most `n - 1`.

angriman:
@fabratu I updated the avx2 flags, now this can be reviewed.

include/networkit/generators/quadtree/QuadNodeCartesianEuclid.hpp
```diff
@@ -391,7 +391,6 @@ class QuadNodeCartesianEuclid final {
                     if (i >= lsize) break;
                     TRACE("Jumped with delta ", delta, " arrived at ", i);
                 }
-                assert(i >= 0);
```
avdgrinten, line 394: Unrelated change; please remove.

include/networkit/generators/quadtree/QuadNodePolarEuclid.hpp
```diff
@@ -503,7 +503,6 @@ class QuadNodePolarEuclid final {
                     if (i >= lsize) break;
                     TRACE("Jumped with delta ", delta, " arrived at ", i);
                 }
-                assert(i >= 0);
```
avdgrinten, line 506: Same.

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    stack.assign(n, 0);
+    farness.assign(group.size(), 0);
+    farnessDecrement.assign(group.size(), 0);
+    sumOfMins.assign(n, 0);
+
+    intDistributions.resize(omp_get_max_threads());
+    urngs.reserve(omp_get_max_threads());
+    for (index i = 0; i < static_cast<index>(omp_get_max_threads()); ++i)
+        urngs.emplace_back(Aux::Random::getURNG());
```
avdgrinten, line 36: This looks broken since the loop is not parallel and we get several identical references to the URNG of thread 0.

avdgrinten: I also wonder if this is necessary at all. `urngs` is only used in a `parallelForNodes` loop in `initRandomVector`. Accessing the thread-local variable is cheap compared to the loop. Why don't we just do the following in `initRandomVector`:
```cpp
#pragma omp parallel
{
    auto &urng = Aux::Random::getURNG();

    #pragma omp for
    for(node u = 0; u < G.upperNodeIdBound(); ++u) {
        [...]
    }
}

File: networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp [hunk 12]
```diff
@@ -0,0 +1,362 @@
...
+
+    totalSwaps = 0;
+    hasRun = false;
+}
+
+void GroupClosenessLocalSwaps::run() {
+    init();
+
+    while (findAndSwap() && ++totalSwaps < maxSwaps) {
+    }
```
avdgrinten, line 58: This loops strange. IMHO the following is cleaner:
```suggestion
    while (findAndSwap() && ++totalSwaps < maxSwaps)
        ; // Just iterate.
```

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
```diff
@@ -0,0 +1,362 @@
...
+    for (const auto &idx : idxMap) {
+        if (canSwap[idx.second]) {
+            int64_t curImprovement =
+                farnessDecrementV - farness[idx.second] + farnessDecrement[idx.second];
+            if (curImprovement > improvement) {
+                improvement = curImprovement;
+                u = idx.first;
+            }
+        }
+    }
```
avdgrinten, line 91: Why do we need this second loop? What does it do? Maybe add a comment.

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
```diff
@@ -0,0 +1,362 @@
...
+    gamma.assign(n * group.size(), false);
+    canSwap.assign(group.size(), false);
+    visited.assign(n, false);
+
+    idxMap.clear();
+    idxMap.reserve(group.size());
+
+    stack.assign(n, 0);
+    farness.assign(group.size(), 0);
+    farnessDecrement.assign(group.size(), 0);
```
avdgrinten, line 30: Decrement -> Decrease. Similarly, if that occurs elsewhere: Increment -> Increase.

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+int64_t GroupClosenessLocalSwaps::computeFarnessDecrement(node v) {
+    std::fill(visited.begin(), visited.end(), false);
+
+    std::queue<node> q;
+    q.push(v);
+    distance[v] = 0;
+    visited[v] = true;
+
+    int16_t decrement{1};
```
avdgrinten, line 201: uint16_t seems to be a mistake here (?).

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
```diff
@@ -0,0 +1,114 @@
...
+    void init();
+    void bfsFromGroup();
+    bool findAndSwap();
+    node estimateHighestDecrement();
+    void initRandomVector();
+    int64_t computeFarnessDecrement(node u);
+    void resetGamma(node x, index idx);
+
+    // Returns true if the vertex at the give index is the only nearest vertex to x
+    bool gammaIsU(node x, index idx) const;
```
avdgrinten, line 97: Why is that needed? Seems that this is not called.

avdgrinten (changes requested):

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
angriman: That's not needed, I probably forgot to remove it from a previous version. I removed it together with the `canSwap` vector, which was used just here.

include/networkit/centrality/GroupClosenessLocalSwaps.hpp
angriman: True, I removed it.

networkit/cpp/centrality/GroupClosenessLocalSwaps.cpp
angriman: Yes, that should be `int64_t`.

avdgrinten (approved):
LGTM now.

PR #477, opened 2020-01-14
Refactor: PageRank

angriman:
This PR brings two new features to PageRank:
- Possibility to set the maximum number of iterations of the algorithm
- Possibility to choose between the L1 and the L2 norms as stopping criterion of the algorithm

### DIFF
--- a/include/networkit/centrality/PageRank.hpp
+++ b/include/networkit/centrality/PageRank.hpp
@@ -26,6 +26,8 @@ namespace NetworKit {
 class PageRank final : public Centrality {
 
 public:
+    enum Norm { L1Norm, L2Norm };
+
     /**
      * Constructs the PageRank class for the Graph @a G
      *
@@ -58,6 +60,9 @@ public:
 
     // Maximum number of iterations allowed
     count maxIterations = std::numeric_limits<count>::max();
+
+    // Norm used as stopping criterion
+    Norm norm = Norm::L2Norm;
 
 private:
     double damp;
--- a/include/networkit/centrality/PageRank.hpp
+++ b/include/networkit/centrality/PageRank.hpp
@@ -10,6 +10,8 @@
 #ifndef NETWORKIT_CENTRALITY_PAGE_RANK_HPP_
 #define NETWORKIT_CENTRALITY_PAGE_RANK_HPP_
 
+#include <limits>
+
 #include <networkit/centrality/Centrality.hpp>
 
 namespace NetworKit {
@@ -44,9 +46,23 @@ public:
      */
     double maximum() override;
 
+    /**
+     * Return the number of iterations performed by the algorithm.
+     *
+     * @return Number of iterations performed by the algorithm.
+     */
+    count numberOfIterations() const {
+        assureFinished();
+        return iterations;
+    }
+
+    // Maximum number of iterations allowed
+    count maxIterations = std::numeric_limits<count>::max();
+
 private:
     double damp;
     double tol;
+    count iterations;
 };
 
 } /* namespace NetworKit */
--- a/include/networkit/centrality/PageRank.hpp
+++ b/include/networkit/centrality/PageRank.hpp
@@ -5,6 +5,8 @@
  *      Author: Henning
  */
 
+// networkit-format
+
 #ifndef NETWORKIT_CENTRALITY_PAGE_RANK_HPP_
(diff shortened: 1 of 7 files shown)


include/networkit/centrality/PageRank.hpp
```diff
@@ -19,31 +23,53 @@ namespace NetworKit {
...
+    count numberOfIterations() const {
+        assureFinished();
+        return iterations;
+    }
+
+    // Maximum number of iterations allowed
+    count maxIterations = std::numeric_limits<count>::max();
+
+    // Norm used as stopping criterion
+    Norm norm = Norm::L2Norm;
```
fabratu, line 65: `norm` and `maxIterations` are public, while the rest is private. Is this to avoid changing the function header?

networkit/_NetworKit.pyx
```diff
@@ -8666,14 +8666,26 @@ cdef class ApproxCloseness(Centrality):
 		return (<_ApproxCloseness*>(self._this)).getSquareErrorEstimates()
 
 
+cdef extern from "<networkit/centrality/PageRank.hpp>" namespace "NetworKit::PageRank":
+
+	cdef enum Norm:
+		L1Norm = 0
+		L2Norm = 1
+
+class _PageRankNorm(object):
```
fabratu, line 8675: Why is this additional `_PageRankNorm`-class introduced here?

include/networkit/centrality/PageRank.hpp
angriman: Yes, I wanted to avoid to add additional parameters in the constructor.

networkit/cpp/centrality/PageRank.cpp
```diff
@@ -5,73 +5,83 @@
...
+
+        return G.parallelSumForNodes(sumL1Norm) <= tol;
     });
 
-    while (! isConverged) {
+    bool isConverged = false;
+    while (!isConverged) {
         handler.assureRunning();
-        G.balancedParallelForNodes([&](node u) {
+        G.balancedParallelForNodes([&](const node u) {
```
fabratu, line 55: Since `auto converged ...` is moved outside the loop, maybe it is a good idea also to move the scoring-update function outside (consistency of style)?

networkit/_NetworKit.pyx
angriman: I need that in `centrality.py` so one can set the norm of `PageRank` to `networkit.centrality.Norm.L1Norm/L2Norm`.

include/networkit/centrality/PageRank.hpp
fabratu, line 65: Not to familiar in with this topic, but is `L2Norm` the recommended norm for PageRank? And maybe it is a good idea to add example about different norms (I guess it only changes the time it takes to converge?) in the appropriate notebook.

networkit/cpp/centrality/PageRank.cpp
angriman: Maybe, the problem is that we are doing that in two steps, and I this would result in more lines of code without improving the style that much.

include/networkit/centrality/PageRank.hpp
angriman: I think that both are valid choices, before we were using the `L2` norm so it makes sense to keep that as the default choice. But I agree that I should add an example in the centrality notebook.

networkit/_NetworKit.pyx
fabratu: Ah I see ... thought this could be done by accessing the public enums from the C++ code.

networkit/cpp/centrality/PageRank.cpp
```diff
@@ -5,73 +5,83 @@
...
-            });
-//			TRACE("sqrt(diff): ", sqrt(diff));
-            return (sqrt(diff) <= tol);
-        });
-
+        ++iterations;
         isConverged = converged();
-        scoreData = pr;
-    }
+        std::copy(pr.begin(), pr.end(), scoreData.begin());
```
michitux, line 69: Why is this a copy and not a swap? As far as I can see the values in `pr` are not read in the next iteration and thus replacing them by `scoreData` should be no problem.

angriman: That's true, thanks I fixed that.

PR #478, opened 2020-01-15
Refactor edgescores & sparsification modules

cndolo:
This PR brings some improvements to the C++ code of the edgescores and sparsification modules. The major changes include:
- Member variables are changed from references to pointers
- All classes that are not inherited from are made final.
- The virtual keyword is removed from any non-base classes.
- The documentation in the header files is changed from *.h to *.hpp.


cndolo:
@angriman You should be able to push commits to this branch now - I must have accidentally deactivated the "Allow edits from maintainers." box, but I have checked the box now.

PR #479, opened 2020-01-16
Added cmake option for adding externally build tlx lib

fabratu:
Enables option -DNETWORKIT_EXT_TLX=<TLX-DIR> in case an external version of tlx is used.


fabratu:
The failed build from Travis seems to be stuck on llvm8 installation rather than anything NetworKit related. Should we ignore it or is there something wrong with the PR?

angriman:
In those cases I'd suggest to just restart the failed build(s).

PR #484, opened 2020-01-24
Refactor generators module

cndolo:
This PR brings some improvements to the C++ code of the generators module.


cndolo:
The clang and LLVM build are failing because of an unused variable in the `StochasticBlockmodel` class. However, the variable, nBlocks, is clearly being used in the `generate` function; hence removing it does not  solve the issue.

michitux:
No, it is only used in an assert, which is not present in the release build. This somehow doesn't make sense. Why do we pass the number of blocks if it is the same as the size of the affinity matrix?

networkit/cpp/generators/StochasticBlockmodel.cpp
```diff
@@ -1,16 +1,16 @@
...
 #include <networkit/auxiliary/Random.hpp>
+#include <networkit/generators/StochasticBlockmodel.hpp>
 
 namespace NetworKit {
 
-StochasticBlockmodel::StochasticBlockmodel(count n, count nBlocks, const std::vector<index>& membership, const std::vector<std::vector<double> >& affinity)
+StochasticBlockmodel::StochasticBlockmodel(count n, count nBlocks, const std::vector<index>& membership, const std::vector<std::vector<double>>& affinity)
     : n(n), nBlocks(nBlocks), membership(membership), affinity(affinity) {
     if (affinity.size() != nBlocks) {
         throw std::runtime_error("affinity matrix must be of size nBlocks x nBlocks");
```
angriman: From the error message I think that the original intention was to test whether the affinity matrix has size nBlocks x nBlocks (but the number of columns is not checked). Anyway, we can avoid to store nBlocks in a private member variable.

PR #487, opened 2020-01-28
Fix travis warnings

cndolo:
This PR fixes the warnings that are thrown by travis in the .travis.yml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
     - export CXX_STANDARD=11
     - export MONOLITH=ON
 
-matrix:
+jobs:
   include:
     - name: macOS 10.14, Apple Clang
       env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -267,11 +258,11 @@ matrix:
         - rm -rf htmldocs/{.buildinfo,.doctrees}
       deploy:
         provider: pages
-        skip-cleanup: true
-        github-token: $SECRET_DEPLOY_GITHUB_PAGES
-        local-dir: core_build/htmldocs
+        cleanup: true
```
angriman: Before we were skipping the cleanup, so this should be set to `false` I suppose.

cndolo: Thanks for that! I have updated the file.

Issue #489, opened 2020-01-29
OMP: Info #270: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.


gijzelaerr:
hi!

When using networkit I get this warning printed in my console:

`OMP: Info #270: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.`

using homebrew python 3.7 on OSX 10.15.

angriman:
Thank you, that's because NetworKit is still using some deprecated openmp APIs so, we'll address this issue.

angriman:
The patch has been merged. I am closing this issue, thank you again @gijzelaerr  for spotting this, and @cndolo for the patch.

gijzelaerr:
no thank you! :)

PR #492, opened 2020-01-31
Address issue #489: Deprecate enabling nested parallelism

cndolo:
In this PR the `enableNestedParallelism` function is deprecated.

### DIFF
--- a/networkit/_NetworKit.pyx
+++ b/networkit/_NetworKit.pyx
@@ -243,7 +243,6 @@ def enableNestedParallelism():
 	""" Enable nested parallelism for OpenMP"""
 	from warnings import warn
 	warn("Nested parallelism has been deprecated.")
-	_enableNestedParallelism()
 
 cdef extern from "<networkit/auxiliary/Random.hpp>" namespace "Aux::Random":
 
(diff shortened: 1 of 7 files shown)


cndolo:
@angriman  `__init.py` calls the now deprecated function `enableNestedParallelism()`. I thought that should be removed too, and added that to the PR. What do you think?

networkit/_NetworKit.pyx
```diff
@@ -241,6 +241,8 @@ def getMaxNumberOfThreads():
 
 def enableNestedParallelism():
 	""" Enable nested parallelism for OpenMP"""
+	from warnings import warn
+	warn("Nested parallelism has been deprecated.")
 	_enableNestedParallelism()
```
angriman: You can remove this line, we do not allow nested parallelism anymore

networkit/__init__.py
```diff
@@ -161,7 +161,6 @@ def setup():
 	""" This function is run once on module import to configure initial settings """
 	setLogLevel("ERROR")  # set default loglevel for C++ code
 	setPrintLocation(True)
-	enableNestedParallelism()  # enable nested parallelism
```
angriman: This line should not be removed, just deprecate the method. We should remove it in future releases.

networkit/_NetworKit.pyx
cndolo: I have, and restored `__init.py` to its initial state.

angriman: Sorry my bad that change was ok.

cndolo: OK, it's back.

PR #496, opened 2020-02-13
Revert travis to skip-cleanup

cndolo:
PR #487 fixed travis warnings, but may have introduced an error in the documentation builds.
This PR tries to fix that by explicitly stating which version of travis should be used i.e., the most recent release.

Some relevant links:
[https://blog.travis-ci.com/2019-10-24-build-config-validation](https://blog.travis-ci.com/2019-10-24-build-config-validation)
[https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release](https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -258,7 +258,7 @@ jobs:
         - rm -rf htmldocs/{.buildinfo,.doctrees}
       deploy:
         provider: pages
-        cleanup: false
+        skip-cleanup: true
         token: $SECRET_DEPLOY_GITHUB_PAGES
         local_dir: core_build/htmldocs
         repo: networkit/dev-docs


fabratu:
Thanks for the PR.

From your first link the build config validation can either be activated by setting the version like in the PR or set it directly in Travis settings (which is already activated). So in theory this will not help us?

Additionally from what I got, the build config validation and dplv2 are not directly related, even though both change part of the config-keys. The conflicting key from PR #487 seems to be "cleanup: false", which is only introduced in dpl v2. I would suggest either change it back to "skip-cleanup: true" until dpl v2 is released or opting in to the beta[1].

Disclaimer: I may be wrong though :)

[1] https://docs.travis-ci.com/user/deployment-v2

angriman:
I guess it is safer to roll back to `skip-cleanup`, and do the update once the dpl v2 is released.

cndolo:
I have reverted the change to `skip-cleanup=false` as it was originally.

PR #501, opened 2020-02-16
Fix inconsistency in scaling z-coordinate for writeGraphToGML

deklanw:
There are appears to be a mistake in how GML is written in 3D. See https://github.com/networkit/networkit/issues/500


avdgrinten:
Thank you for the PR. This looks like a copy-and-paste mistake, indeed. LGTM.

PR #502, opened 2020-02-17
Simpler readme

angriman:
The current readme is too long and, for some parts, outdated. This PR brings a simpler readme with the most important instructions about how to install and compile NetworKit.


README.md
```diff
@@ -19,30 +19,33 @@ via the Cython toolchain. Python in turn gives us the ability to work interactiv
...
--   Pip install: Download the NetworKit Python package with pip. This is the easier way to get NetworKit but you can only use NetworKit via Python this way.
--   Build NetworKit from Source: Clone or download the source code of NetworKit and build the C++ and Python modules from source.
+    $ pip3 install [--user] networkit
 
-More detailed instructions follow after the requirements section.
-With NetworKit as a Python extension module, you get access to native high-performance code and can at the same time work interactively in the Python ecosystem. Although the standard Python interpreter works fine, we recommend [IPython] and [jupyterhub](https://github.com/jupyterhub/jupyterhub) as great environments for scientific computing.
+### Install with conda-forge
 
-Once you have installed NetworKit, please make sure to check out our [NetworKit UserGuide] for an overview of the features provided in NetworKit.
+    $ conda install networkit
```
avdgrinten: @fabratu should we add instruction on how to add the conda-forge channel?

README.md
```diff
@@ -19,30 +19,33 @@ via the Cython toolchain. Python in turn gives us the ability to work interactiv
...
-## Documentation
+### Building the Python module from source
 
-In addition to this `Readme`, the [NetworKit UserGuide] provides an introduction to the NetworKit tools, in the form of an interactive IPython Notebook. The `DevGuide` is meant for developers who would like to contribute. When using NetworKit as a Python module, refer to the docstrings of classes, methods and functions.
-
-C++ sources are also documented in [Doxygen](http://www.stack.nl/~dimitri/doxygen/) format, while the documentation for the Python sources can be generated with [Sphinx](http://sphinx-doc.org/).
-The most recent version of the [documentation can be found online](https://networkit.github.io/dev-docs/index.html). If you have both utilities installed, the documentation can be easily generated by building the `docs` target in your build directory (e.g., `make docs`).
+    git clone https://github.com/networkit/networkit networkit
+    cd networkit
+	python3 setup.py build_ext [-jX]
```
avdgrinten: Indentation is broken.

README.md
```diff
@@ -55,128 +58,75 @@ package:
...
-
-Additionally, we recommend that you familiarize yourself with NetworKit through experimenting with the interactive IPython Notebook `NetworKit_UserGuide.ipynb` located in the folder `Doc/Notebooks`. The user guide also introduces a large portion of NetworKit functionality with usage examples. To display and work with these notebooks, you have to install jupyterhub and start a local notebook server from the terminal with:
+- Cython version 0.21 or higher (e.g., `pip3 install cython`)
 
-	jupyterhub --no-ssl
+# Usage example
 
-If you run into any problems with jupyterhub, head over to the [jupyterhub documentation](https://jupyterhub.readthedocs.io/en/latest/) and make sure, you have the listed packages installed. If the notebook server starts as it is supposed to, your default browser should open a web interface or you have to open it manually. Then you can add `NetworKit_UserGuide.ipynb` from the above mentioned location or browse to the location through the web interface.
+In this example we generate a random hyperbolic graph with 100k nodes
+and we compute its communities with the PLM method:
```
avdgrinten: We should point users to our notebooks here! (I am not even sure if we want to keep the small example -- the notebooks are much more instructive.)

README.md
```diff
@@ -55,128 +58,75 @@ package:
...
+C++\-application as follows:
 
-To use the previous compiled networkit library, you have to link it while compiling your project. The include directives in your C++\-application should look like the following
+	#include <networkit/graph/Graph.hpp>
 
-	#include <NetworKit/graph/Graph.h>
+You can compile your source as follows:
 
-By default, [CMake] creates a symlink to the source tree inside your build directory. `NetworKit` in the directory `include `is a symlink to the directory `networkit/cpp`, so the directory structure from the repository is valid. To compile your application, you need to add the paths for the header files and the location of the library. There is a simple source file to demonstrate this. Feel free to compile `LibDemo.cpp` as follows:
+	g++ my_file.cpp -std=c++11 -I/path/to/install/directory/include -L/path/to/install/directory -lnetworkit
```
avdgrinten: This will only work if the library path is in `LD_LIBRARY_PATH`. Since this is probably complex to explain (and also unnecessary), should we just demonstrate how a global installation (in `/usr/local`) works?

README.md
```diff
@@ -185,20 +135,38 @@ Single tests can be executed with:
...
+
+# Documentation
+
+The most recent version of the [documentation can be found online](https://networkit.github.io/dev-docs/index.html).
+
+# Contact
+
+For questions regarding NetworKit, subscribe to our [e-mail list][list] and feel free to ask.
+
+# Known Issues
```
avdgrinten: I am not sure if we want to keep this. The GitHub issue tracker seems to be a much better tool to search for known issues.

README.md
fabratu: Yes - and for the concrete syntax I think we should go with the default readme from the feedstocks (without any extras):

conda config --add channels conda-forge

To exclude possible ambiguity we could also add:
conda install networkit -c conda-forge

angriman: @fabratu Feel free to push a commit to this branch to fix this.

README.md
fabratu: +1 for pointing to the notebooks. 

If shortening the readme is intended - than maybe also remove the example. However: Although the notebooks are much more instructive, their results don't show when viewed in the web on git. Maybe it's good for first-time users to see how an output should look like?

angriman: I think it is better to keep a short example to show basic APIs. Anyway, we should definitely add a link to the notebooks here.

angriman: @fabratu can you push another commit with the links?

fabratu: Only partially related - but since there is now more than one notebook ... is there some kind of order in how to read them? If so, should we include a prefix in the file-naming? And furthermore: does User-Guide should still be named that way or more like: Overview, Introduction?

fabratu: Moved / Edited the notebook ref.

README.md
```diff
@@ -55,166 +56,118 @@ package:
...
+    -------------------  -----------
+    # communities        4536
+    min community size      1
+    max community size   2790
+    avg. community size    22.0459
+    modularity              0.987243
+    -------------------  -----------
 
-	jupyterhub --no-ssl
+For more interactive examples you can have a look at our [notebooks]-section, especially the [Networkit UserGuide]. Note: To view the computed output from the notebooks, it is recommended to use ![JupyterHub](https://pypi.org/project/jupyterhub/) (installed for example via pip3 or conda). This requires the prior installation of NetworKit.
```
avdgrinten: IMHO, this should be put first (above the inline example). Also: JupyterHub or just plain jupyter-notebook (the notebook server)? JupyterHub is probably much harder to set up!

README.md
```diff
@@ -19,32 +19,33 @@ via the Cython toolchain. Python in turn gives us the ability to work interactiv
...
 
-For questions regarding NetworKit, subscribe to our [e-mail list][list] and feel free to ask.
+The script will call `cmake` and `ninja` (`make` as fallback) to compile
+NetworKit as a library, build the extensions and copy it to the top folder. By
+default, NetworKit will be built with the amount of available cores in
+optimized mode. It is possible the add the option `-jN` the number of threads
+used for compilation.
 
-## Requirements
+# Requirements
```
fabratu: Added a commit with small typo-fixes.

Since the order of install <-> reqs. in this PR with respect to the current Dev-state has changed, I guess this was already considered. Should we place requirements before the install instructions?

avdgrinten: Yes.

avdgrinten:
The section captions in the README should be at the second level (`##`). The only caption at the first level should be the `NetworKit` headline.

PR #504, opened 2020-02-18
Fix CMake error message

angriman:
If one is using llvm clang, this error message is wrong. This PR corrects the compiler name in the error message when libomp is not found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
 		set(OpenMP_CXX_LIB_NAMES "omp" CACHE STRING "Manually set")
 		set(OpenMP_omp_LIBRARY "${LIBOMP_PATH}" CACHE STRING "Manually set")
 	else()
-		message(FATAL_ERROR "libomp was not found, but necessary to run NetworKit with AppleClang")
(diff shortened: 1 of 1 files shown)


avdgrinten:
Merged because it's a trivial text change.

fabratu:
Maybe we should modify/rethink the whole section with clang. As it is now this method will "only" find the standard brew installation for AppleClang. This leaves out (on MacOS) installing brew in home-directories, conda and spack clang installations.

In addition when including Linux clang into this, things get more complicated. For Ubuntu 18.04 (and other distros) we can add "/usr/include" and "/usr/lib" as hint for lib/header, but other popular distros (and newer Ubuntu/Debian versions) do a similar scheme to gcc with their respective sub-folder under "/usr/lib[64]/". Also conda/spack-version should be recognized here I think, maybe even more important than on MacOS.

How about this:
- Adding HINTS as described in line 124-125
- If nothing is found try FindOpenMP from cmake regardless

The second part seems to work (early tests with CentOS 8) with the modern Linux-scheme. There is also mentioning that FindOpenMP is able to find conda-installations of clang. If we want to change more, I can have a look into different configurations on both Linux + MacOS.

avdgrinten:
I think that is true (i.e., rethinking the logic) but appropriate for another PR.

IMHO, we should first rely on CMake (findTarget) and only fall back to the ad-hoc behavior when that is not successful.

Issue #499, opened 2020-02-15
Edge count discrepancy with sparsifier


deklanw:
I'm starting with a directed graph with 8541 edges. I run a Sparsifier with a ratio of 0.5. 

```python
    print(G.numberOfEdges(), localDegGraph.numberOfEdges())
```

```8541 4234```

So far so good. 

```python
    edge_c = 0

    def count_edges(source, target, _weight, _edge_id):
        nonlocal edge_c
        edge_c += 1

    localDegGraph.forEdges(count_edges)
    print(edge_c)
```

Gives:

```3972```

What's causing the discrepancy? I also ran a layout algorithm on `localDegGraph`, saved it with `writeGraphToGML` and also got 3972 edges.

**Comment by angriman** [comment]
Probably the sparsifier does not update edge count correctly. Which sparsifier are you using?

**Comment by deklanw** [comment]
@angriman 

Thanks for response. I'm using LocalDegreeSparsifier

**Comment by angriman** [comment]
`LocalDegree` sparsifier only computes edge scores, so the edges should be removed somewhere else. How exactly do you create `localDegGraph`?

**Comment by deklanw** [comment]
```python
# Initialize the algorithm
localDegSparsifier = nk.sparsification.LocalDegreeSparsifier()

# Get sparsified graph
localDegGraph = localDegSparsifier.getSparsifiedGraphOfSize(G, targetRatio)
G.numberOfEdges(), localDegGraph.numberOfEdges()
```
As in https://github.com/networkit/networkit/blob/Dev/notebooks/Sparsification.ipynb

angriman:
Thank you. I ran this code:
```
import networkit as nk

G = nk.readGraph('./input/MIT8.edgelist', nk.Format.EdgeList, separator='\t', firstNode=0,
        continuous=False, directed=False)
G.indexEdges()

# Initialize the algorithm
localDegSparsifier = nk.sparsification.LocalDegreeSparsifier()

# Get sparsified graph
localDegGraph = localDegSparsifier.getSparsifiedGraphOfSize(G, .5)

edge_c = 0
def count_edges(source, target, _weight, _edge_id):
    global edge_c
    edge_c += 1

localDegGraph.forEdges(count_edges)
assert(localDegGraph.numberOfEdges() == edge_c)
```
but I could not replicate the issue, the edge count always match even for different ratios. Could you attach the network you are using so I can test it?

deklanw:
[...]

PR #507, opened 2020-02-23
Fix GlobalThresholdFilter

angriman:
This PR fixes two main issues:
- Templated methods of `EdgeScore` were implemented in the `.cpp` file. They have been moved to the header file.
- `GlobalThresholdFilter` now also support directed and weighted graphs. In the previous implementation the graph was transformed to an undirected graph with default edgeweights without throwing any warnings or errors. If the input graph is directed, this yields a discrepancy in the number of edges in the resulting graph (as reported in #499).

### DIFF
--- a/include/networkit/sparsification/GlobalThresholdFilter.hpp
+++ b/include/networkit/sparsification/GlobalThresholdFilter.hpp
@@ -1,9 +1,11 @@
 /*
- * GlobalThresholdFilter.h
+ * GlobalThresholdFilter.hpp
  *
  *  Created on: 23.07.2014
  *      Author: Gerd Lindner
  */
+
+// networkit-format
 
 #ifndef NETWORKIT_SPARSIFICATION_GLOBAL_THRESHOLD_FILTER_HPP_
 #define NETWORKIT_SPARSIFICATION_GLOBAL_THRESHOLD_FILTER_HPP_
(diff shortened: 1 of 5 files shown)


michitux:
Why do you move the implementation of `EdgeScore` into the header? What is the issue you are trying to solve here? The intention of the template is only to support both `double` and `count` as scores, and both are instantiated explicitly in the `.cpp` file.

avdgrinten:
If that is supposed to be the case, we need `extern template EdgeScore<double>;` in the header to avoid ODR violations (if I interpret the standard correctly here: right now the template is instantiated both implicitly *and* explicitly, in different TUs). In that case, we should also add a `static_assert` with a meaningful error message. But why do you want the class to be restricted to `double` and `count` in the first place?

michitux:
This `extern template EdgeScore<double>;` might be a good idea, but I do not think it is strictly necessary as it has only been introduced in C++11. There are no edge scores in NetworKit that use any other type. Also, the Cython interface needs to explicitly list all types, so if more types shall be supported, the Cython interface would need to be adjusted. The basic idea here is to reduce compile time. Without this, as far as I understand, the template will be instantiated every time it is used, e.g., in `_NetworKit.pyx`.

michitux:
I have searched a bit more, `extern template` is only required when the template is fully defined in the header. If it is only declared, as it is without this PR, `extern template` has no effect. See e.g. [here](https://stackoverflow.com/questions/8130602/using-extern-template-c11/59614090#59614090). Adding a `static_assert` could be a good idea, but apart from that I do not see anything that should be changed here.

avdgrinten:
Isn't the class definition complete? For a class to be complete, all members must be complete - but functions do not need to be defined.

avdgrinten:
Contrary to the example in the linked SO answer, the EdgeScore class does have function definitions for some functions (namely, inherited methods from the base classes).

fabratu:
@angriman @avdgrinten @michitux 
In the course of the (honestly very intriguing) discussion about the nature of templates/compilers, maybe this PR should be split. The second part about `GlobalThresholdFilter` fixes an active issue and should be integrated. 

To add a review for this part: Looks good to me, only a (minor) question:
- Should we add a test, that compares `numberOfEdges()` against an actual count ... or is this automatically resolved by not converting the graph?

angriman:
I see the point of having the implementation of `EdgeScore` in a separate file, I dropped that commit (which was also out of the scope of this PR).

PR #508, opened 2020-02-23
clang-tidy

angriman:
[clang-tidy](https://clang.llvm.org/extra/clang-tidy/index.html) is a very powerful tool to detect programming errors that are generally not detected by compilers, and it is easy to integrate with CMake.
I think that it would be a good idea to integrate it into our CI system: it would help us fixing existing programming errors (such as the ones fixed recently by @cndolo), and avoiding introducing new ones in future PRs.

If you agree with that, we might want to discuss about which [checks](https://clang.llvm.org/extra/clang-tidy/checks/list.html) should we use.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,87 @@
+Checks: >
+    -*,
+    bugprone-assert-side-effect,
+    bugprone-branch-clone,
+    bugprone-copy-constructor-init,
+    bugprone-dangling-handle,
+    bugprone-dynamic-static-initializers,
+    bugprone-exception-escape,
+    bugprone-fold-init-type,
+    bugprone-forward-declaration-namespace,
+    bugprone-forwarding-reference-overload,
+    bugprone-inaccurate-erase,
+    bugprone-incorrect-roundings,
+    bugprone-integer-division,
+    bugprone-misplaced-widening-cast,
+    bugprone-move-forwarding-reference,
+    bugprone-multiple-statement-macro,
+    bugprone-parent-virtual-call,
+    bugprone-swapped-arguments,
+    bugprone-unused-return-value,
+    bugprone-use-after-move,
+    bugprone-virtual-near-miss,
+    cppcoreguidelines-narrowing-conversions,
+    performance-for-range-copy,
+    performance-implicit-conversion-in-loop,
+    performance-inefficient-algorithm,
+    performance-inefficient-vector-operation,
+    performance-move-const-arg,
+    performance-noexcept-move-constructor,
+    performance-trivially-destructible,
+    performance-type-promotion-in-math-fn,
+    performance-unnecessary-copy-initialization,
+    performance-unnecessary-value-param,
+
+# Warnings whishlist
+#    modernize-deprecated-headers,
+#    modernize-use-using,
+#    modernize-avoid-c-arrays,
+#    modernize-use-emplace,
+#    modernize-use-override,
+#    readability-avoid-const-params-in-decls,
+#    readability-const-return-type,
+#    readability-container-size-empty
+
+# Turn the warnings from the checks above into errors.
+# To turn all warnings into errors, simply write '*'
+# in the first line after WarningsAsErrors
+WarningsAsErrors:  >
+    -*,
+    bugprone-assert-side-effect,
+    bugprone-copy-constructor-init,
+    bugprone-dangling-handle,
+    bugprone-dynamic-static-initializers,
+    bugprone-exception-escape,
+    bugprone-forward-declaration-namespace,
+    bugprone-forwarding-reference-overload,
+    bugprone-inaccurate-erase,
+    bugprone-misplaced-widening-cast,
+    bugprone-move-forwarding-reference,
+    bugprone-multiple-statement-macro,
+    bugprone-parent-virtual-call,
+    bugprone-swapped-arguments,
+    bugprone-unused-return-value,
+    bugprone-virtual-near-miss,
+    performance-implicit-conversion-in-loop,
+    performance-inefficient-algorithm,
+    performance-move-const-arg,
+    performance-trivially-destructible,
+    performance-unnecessary-copy-initialization
+
+FormatStyle: 'file'
+HeaderFilterRegex: '.*\/include\/networkit\/.*\.hpp'
+CheckOptions:
+  - key:   readability-identifier-naming.ClassCase
+    value: CamelCase
+  - key:   readability-identifier-naming.EnumCase
+    value: CamelCase
+  - key:   readability-identifier-naming.FunctionCase
+    value: camelBack
+  - key:   readability-identifier-naming.MemberCase
+    value: CamelCase
+  - key:   readability-identifier-naming.ParameterCase
+    value: CamelCase
+  - key:   readability-identifier-naming.UnionCase
+    value: CamelCase
+  - key:   readability-identifier-naming.VariableCase
+    value: CamelCase
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,6 +107,36 @@ jobs:
             - libomp
           update: true
 
+    - name: Linux, Clang 9 with clang-tidy
+      env:
+        - CC=clang
+        - CXX=clang++
+      os: linux
+      dist: bionic
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - sourceline: 'deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main'
+              key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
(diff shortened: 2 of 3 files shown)


hmeyerhenke:
Sounds like a good idea.

avdgrinten:
To evaluate which checks we should enable, we really need to see the effect of each check on the existing code base. Looking at the current output, it's clear that some checks cannot apply to NetworKit. For example:

```
/home/travis/build/networkit/networkit/networkit/cpp/algebraic/DenseMatrix.cpp:27:5: error: use range-based for loop instead [modernize-loop-convert,-warnings-as-errors]

    for (omp_index k = 0; k < static_cast<omp_index>(triplets.size()); ++k) {

    ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
is not feasible since it break OpenMP.

I suggest to run checks without the warnings-as-errors parameter so that we can get an overview of what is useful and what isn't.

angriman:
True, I removed warnings error and `modernize-loop-convert`.

angriman:
I think that the list of checks is ready to be reviewed. After merge we should gradually fix the warnings, and turn them into errors.

fabratu:
When building with `-DNETWORKIT_CLANG_TIDY=ON` for Ubuntu 19.10 and clang-9, make fails due to `[bugprone-branch-clone,-warnings-as-errors]`  in `CSRMatrix.cpp` (and there maybe more after that)

angriman:
I see, we might want to fix those warnings later then. I am removing them from the errors list for now.

avdgrinten:
The list of checks is much too broad right now. We have to carefully consider which checks we want to enable *on an individual (per-check) basis*. This is especially true for the checks that only affect style (and not correctness). For many of the "readability" checks, it is not clear to me that readability is truly improved. Also remember that the refactoring of coding style has a cost (possibility to introduce bugs, time spent to review PRs and the fact that existing PRs need rebasing).

fabratu:
The build is now successful.

Looked briefly through the list of warnings. However it is very long (roughly ~8k) - which makes it harder to decide on whether a feature should be included or not (at least if you are not that familiar with the code). Maybe we should only enable a view of them at a time?

avdgrinten:
I think we should start with a conservative baseline (enable only few of the style-related checks), fix the real issues uncovered by the static analysis and then re-evaluate whether we want to enable additional style-related checks.

avdgrinten:
Things that I disagree with (I have not looked at all checks, there are a lot and it's a lot of work to look them all up in the documentation):
- `readability-braces-around-statements`: It's not clear to me at all that this style is desirable; it increases the vertical space that code consumes. Since misleading indentation is already a warning in modern Clang and GCC, I also do not see any advantage of this check.
- `readability-redundant-access-specifiers`: I don't get why duplicate access specifies should be harmful. If anything, they *increase* readability since they more the access specifier closer to the declaration in large classes (e.g., classes that inherit from multiple distinct interfaces).
- `cppcoreguidelines-init-variables`: IMHO, `-Wmaybe-uninitialized` is a much better version of this.
- `readability-else-after-return`: This check enforces an asymmetric coding style even if the `if` statement is symmetric.
- `readability-function-size`: I don't see the point of this check.

avdgrinten:
Here is what LLVM enables (see https://github.com/llvm/llvm-project/blob/master/.clang-tidy): `-*,clang-diagnostic-*,llvm-*,misc-*,-misc-unused-parameters,readability-identifier-naming`.

angriman:
> Things that I disagree with (I have not looked at all checks, there are a lot and it's a lot of work to look them all up in the documentation):
> 
> * `readability-braces-around-statements`: It's not clear to me at all that this style is desirable; it increases the vertical space that code consumes. Since misleading indentation is already a warning in modern Clang and GCC, I also do not see any advantage of this check.
> * `readability-redundant-access-specifiers`: I don't get why duplicate access specifies should be harmful. If anything, they _increase_ readability since they move the access specifier closer to the declaration in large classes (e.g., classes that inherit from multiple distinct interfaces).
> * `cppcoreguidelines-init-variables`: IMHO, `-Wmaybe-uninitialized` is a much better version of this.
> * `readability-else-after-return`: This check enforces an asymmetric coding style even if the `if` statement is symmetric.
> * `readability-function-size`: I don't see the point of this check.

I removed the last four, they are indeed not necessary. I'd prefer to keep `readability-braces-around-statements`, and allowing single statements without braces. I find multiple statements without braces harder to read and more "bug-prone" if the code is not formatted correctly.

avdgrinten:
If the formatting is broken, this is already caught by `-Wmisleading-indentation` (see https://developers.redhat.com/blog/2016/02/26/gcc-6-wmisleading-indentation-vs-goto-fail/). IMHO, that check is much better than requiring braces everywhere.

angriman:
I see, then this would just be a minor style preference. I am removing this from now.

avdgrinten:
This PR still generates *way* to many warnings (19k lines of output). In this state, there is no hope to reasonably improve the code. *Enable only the checks that uncover real problems*. Then we can fix these problems and successively enable more checks.

avdgrinten:
In particular, disable the checks that only affect style, as I have suggested in a comment a few months ago. This probably also applies to most `modernize-` checks and `readability-` checks. When we get the output down to a reasonable size (maybe 1k lines of warnings), we can merge and act upon these warnings.

angriman:
I agree with that, I removed most `modernize-` and `readability-` checks, the only exceptions that I'd suggest to keep are the following.
Modernize:
- avoid-c-arrays
- deprecated-headers
- use-emplace
- use-override

Readability:
- avoid-const-params-in-decls
- const-return-type
- container-size-empty

`modernize-use-override` and `readability-avoid-const-params-in-decls` in particular are issues that @cndolo already contributed to fix in part of the codebase.

angriman:
With the new build the total amount of warnings dropped from `5318` to `1174`.

avdgrinten:
There are still 5k lines of output. Can this be further reduced (below 1k)? (Maybe some things can be moved to a "check wishlist" without enabling them now?)

Keep in mind that too many warnings are perceived as no warning at all.

angriman:
We could but I am not sure which warnings we should disable to get 1K lines of output. Anyway, those warnings are only enabled with `-DNETWORKIT_CLANG_TIDY=ON`.

avdgrinten:
Disable everything that is merely enforces a coding style instead of hinting at a bug.

avdgrinten:
If we let these warnings linger, they will not be useful. Somebody needs to act on them to make them useful. But letting someone (e.g., @cndolo, after the Cython stuff is done) go through 5000 lines of warnings is a massive waste of time.

This is *useless* without priorization. The amount of warnings *needs to be reduced* to an essential subset. (And then, this subset needs to be enabled at `-Werror`.)

angriman:
Why should one go through all 5000 lines of warnings? We already have a list of warnings in the .clang-tidy file, the majority of which can be fixed with a single auto-fix command. The output only needs to be consulted to fix warnings that cannot be auto-fixed, and to do that one can simply use a search function, without reading the whole file.

avdgrinten:
Some examples of warnings at clutter the output without hinting at a bug:
- readability-container-size-empty (use `.empty()` instead of `.size() > 0`)
- modernize-use-using (use `using x = y;` instead of `typedef y x;`)
- modernize-deprecated-headers (use `<cmath>` instead of `math.h`)

angriman:
> Some examples of warnings at clutter the output without hinting at a bug:
> 
> * readability-container-size-empty (use `.empty()` instead of `.size() > 0`)
> * modernize-use-using (use `using x = y;` instead of `typedef y x;`)
> * modernize-deprecated-headers (use `<cmath>` instead of `math.h`)

Alright I think that for now we can disable those.

avdgrinten:
Well, if the warnings can be auto-fixed, do not enable them in this PR. Instead, make a new PR (after this PR is merged) that performs the auto-fix and also enables the auto-fixed warning at `-Werror` level to avoid running into it again.

avdgrinten:
> Why should one go through all 5000 lines of warnings? [...] The output only needs to be consulted to fix warnings that cannot be auto-fixed, and to do that one can simply use a search function, without reading the whole file.

The warnings that are likely true bugs can probably not be auto-fixed. Take for example this:
```
/home/travis/build/networkit/networkit/include/networkit/centrality/GroupCloseness.hpp:140:19: warning: folding type 'unsigned long' into type 'int' might result in loss of precision [bugprone-fold-init-type]
    double dSum = std::accumulate(distance.begin(), distance.end(), 0);
```
That is a very reasonable warning but to find it, I have to scroll past 1245 lines of clutter.

angriman:
Ok I see the point, I commented the style-related warnings (all of them can be auto-fixed in a future PR). The output should be less cluttered.

angriman:
I removed all modernize-* and readability-* checks and the output has now about 2.5k lines.

PR #509, opened 2020-02-24
Restructure Cython interface

cndolo:
We are currently working on restructuring the Cython/Python side of NetworKit. The goal of this PR at this point is to see what changes to the current code will be necessary, and what final result may look like . The main aspects of this project are:
 
- Grouping the python code in to modules like the C++ code.
- Reducing the size of the `_Networkit.pyx` file significantly, if we are not able to do away with it completely.

We are deleting the [mesoscopicResponseFunction](https://github.com/networkit/networkit/blob/71c84de94443f5b260fe05b7a9605242579060ff/networkit/community.py#L163) of the `community` module as it does not work as it is.


michitux:
This looks good and I think this should allow to eliminate `_NetworKit.pyx` and make NetworKit extension modules possible which would be really great for student projects. Are there any problems with this approach?

cndolo:
@avdgrinten The graph related modules are done - maybe you would like to take a look at what has been done so far, and let me know what to change/avoid etc. going forward.

michitux:
Is there any reason why in some cases (in particular `graphtools` and `traversal`), everything is in the `.pyx` file and there is no `.pxd` file? Apart from that this looks good. It might be nicer to not to import everything in the global module namespace and possibly also move the C++ types into their own modules (such that we would not need to add the `_` prefix but could instead differentiate them based on the module), but I would suggest to change this later as I think separating the modules is more important than making this super nice and clean.

cndolo:
> Is there any reason why in some cases (in particular `graphtools` and `traversal`), everything is in the `.pyx` file and there is no `.pxd` file? 

The `.pxd` files contain extrernal declarations that will be used in other modules, whereas `.pyx` files can contain anything [1].  Hence, if we need to `cimport` anything, we need a `.pxd` file. As far as I could tell, the `Traversal` and `GraphTools` modules are not needed in other modules, so that's why there is no corresponding `.pxd`.
[1]: http://docs.cython.org/en/latest/src/userguide/sharing_declarations.html#definition-and-implementation-files

avdgrinten:
We also thought about separate modules for C++ wrappers but that requires an additional level of boilerplate (importing the C++ wrappers from the non-C++-wrappers). It's also not clear if it would really help to separate Cython and Python code (except by maybe decreasing the size of the generated `.so` files).

avdgrinten:
But if you have a convincing argument why Cython and Python code should be separated, please let us know.

cndolo:
I have marked this PR as ready for reviews.

networkit/structures.pxd
```diff
@@ -0,0 +1,78 @@
...
+from libcpp cimport bool as bool_t
+from libcpp.vector cimport vector
+from libcpp.map cimport map
+from libcpp.set cimport set
+from libcpp.string cimport string
+
+ctypedef uint64_t count
+ctypedef uint64_t index
+
+from .helpers cimport *
```
avdgrinten: Dito.

networkit/matching.pyx
```diff
@@ -0,0 +1,57 @@
+from _NetworKit import Matcher, PathGrowingMatcher
```
avdgrinten: Why are these declarations in the `.pyx` and not in the `.pxd`? Do we have a rationale for that? If not, we should probably write up a guideline. "Public" declarations go into the `.pxd` while implementation details go into `.pyx`?

Note that `structures` has all of its declarations in the `.pxd` file.

networkit/matching.pxd
```diff
@@ -0,0 +1,32 @@
+from libc.stdint cimport uint64_t
+from libcpp.vector cimport vector
+from libcpp cimport bool as bool_t
+
+ctypedef uint64_t count
+ctypedef uint64_t index
+ctypedef index node
+ctypedef double edgeweight
+
+from .helpers cimport *
```
avdgrinten: Dito.

networkit/graph.pyx
```diff
@@ -0,0 +1,1115 @@
+from cython.operator import dereference, preincrement
+from networkit.helpers import stdstring, pystring
+
+from .traversal import *
```
avdgrinten: Dito.

networkit/helpers.pxd
```diff
@@ -27,3 +27,7 @@ cdef extern from "<algorithm>" namespace "std":
 	pair[_Graph, vector[node]] move(pair[_Graph, vector[node]]) nogil
 	vector[pair[pair[node, node], double]] move(vector[pair[pair[node, node], double]]) nogil
 	vector[pair[node, node]] move(vector[pair[node, node]]) nogil
+
+#cdef extern from "<networkit/Globals.hpp>" namespace "NetworKit":
+
+#	index _none "NetworKit::none"
```
avdgrinten: ?

networkit/_NetworKit.pyx
```diff
@@ -48,15 +48,35 @@ from .base cimport Algorithm
...
+cdef extern from "<algorithm>" namespace "std":
+	void swap[T](T &a,  T &b)
+	_Graph move( _Graph t ) nogil # specialized declaration as general declaration disables template argument deduction and doesn't work
+	_Partition move( _Partition t) nogil
+	pair[_Graph, vector[node]] move(pair[_Graph, vector[node]]) nogil
+	vector[pair[pair[node, node], double]] move(vector[pair[pair[node, node], double]]) nogil
+	vector[double] move(vector[double])
+	vector[bool_t] move(vector[bool_t])
+	vector[pair[node, node]] move(vector[pair[node, node]]) nogil
+
```
avdgrinten: Why is this re-introduced into `_NetworKit.pyx`?

networkit/_NetworKit.pyx
```diff
@@ -204,737 +207,7 @@ cdef object toPoint2DVector(const vector[_Point2D]& v):
...
-		Parameters
-		----------
-		batch : list of GraphEvent.
-		"""
-		cdef vector[_GraphEvent] _batch
-		for ev in batch:
-			_batch.push_back(_GraphEvent(ev.type, ev.u, ev.v, ev.w))
-		(<_DynAPSP*>(self._this)).updateBatch(_batch)
-
+# TODO: expose all methods
```
avdgrinten: ?

networkit/centrality.pxd
```diff
@@ -1,27 +0,0 @@
...
-		vector[pair[node, double]] ranking() except +
-		double score(node) except +
-		double maximum() except +
-		double centralization() except +
-
-
-cdef class Centrality(Algorithm):
-	""" Abstract base class for centrality measures"""
-
-	cdef Graph _G
```
avdgrinten: Why is this deleted?

cndolo:
> I assume that commit [8c8570e](https://github.com/networkit/networkit/commit/8c8570e70bffbef62e036b7434c235222a89fa23) only removed stuff from `_NetworKit.pyx` and moves declarations in that file around? Are there any other changes to `_NetworKit.pyx`?

Correct. This is the initial commit where the `graph.*` files are created.

networkit/_NetworKit.pyx
cndolo: This was a mistake I made initially - I thought I could import these functions from outside of `_NetworKit.pyx` but I could not. In commit db6d351 the changes are reverted.

networkit/matching.pyx
cndolo: `Matcher` and `PathGrowingMatcher` are Python classes that were being imported from `_NetworKit.pyx`. Importing them in `matching.pxd` would have no effect in `matching.pyx`. However, in 6118db8fa91a95dcad9656a50950e3b46eba28d8  `matching` is finalised and all imports are in the `pxd` file.

networkit/centrality.pxd
cndolo: `centrality.pxd` was not necessary because no class needed a `cimport`.

networkit/_NetworKit.pyx
```diff
@@ -45,45 +45,22 @@ ctypedef double coordinate
 
 from .base cimport _Algorithm
 from .base cimport Algorithm
+from .graph cimport _Graph, Graph
+from .structures cimport _Cover, Cover, _Partition, Partition
+from .matching cimport _Matching, Matching
+from networkit.helpers cimport *
```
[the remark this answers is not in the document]
cndolo: It is resolved in a later commit.

networkit/traversal.pyx
```diff
@@ -0,0 +1,180 @@
...
+
+ctypedef uint64_t count
+ctypedef uint64_t index
+ctypedef index node
+ctypedef uint64_t edgeid
+ctypedef double edgeweight
+
+from .graph cimport _Graph, Graph
+from .helpers import *
+from .helpers cimport *
```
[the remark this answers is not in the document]
cndolo: It is resolved in a later commit.

networkit/structures.pxd
cndolo: This is resolved in a later commit.

networkit/matching.pxd
cndolo: This is resolved in a later commit.

networkit/graph.pyx
cndolo: This is resolved in a later commit.

networkit/helpers.pxd
cndolo: This was unnecessary, but is resolved in a later commit.

networkit/_NetworKit.pyx
cndolo: I believe this is the result of a rebase, however it is resolved in a later commit.

networkit/centrality.pxd
michitux: I would suggest to restore this as otherwise other extension modules cannot export further centrality classes based on this base class. While there are none in NetworKit itself, this change allows providing extensions to NetworKit that can be distributed independently of NetworKit itself.

networkit/matching.pyx
avdgrinten: So the rationale is: Python imports in .pyx, Cython imports in .pxd (and the latter become automatically visible in the .pyx)?

networkit/_NetworKit.pyx
avdgrinten: You could remove the comment using `rebase -i` (doesn't have a high priority but would be nice).

networkit/helpers.pxd
avdgrinten: This could be removed using `rebase -i`.

michitux:
Is there anything that prevents removing the remaining classes from `_NetworKit.pyx`?

networkit/centrality.pxd
cndolo: I'm not sure I agree as this increases the amounts of files we have but do not actually need. I assume we would apply this uniformly to all C++ classes, and not just to centrality.

michitux: The problem are only the Python classes that could be used as a base class as you cannot access the cdef-defined members in subclasses unless you can import their definition from a pxd file. Therefore, my argument primarily applies to all classes that are typically used as a base class and not all C++ classes. I think the number of classes this affects should be relatively low, maybe one or two per module. We can also do this later after merging this PR, this PR is already a huge improvement over the current state and should be merged as soon as possible in my opinion.

networkit/_NetworKit.pyx
cndolo: Done.

avdgrinten:
Regarding the public export of `_Centrality` and others: @michitux has a point there but this could be done gradually after this PR is merged. It could applied to all classes that (reasonably) serve as base classes.

avdgrinten:
Regarding _NetworKit.pyx: it is indeed not clear to me why the decision was made to keep these classes. Don't they fit into another module? Or is there a technical issue that prevents them from being moved? It would be nice to get rid of `_NetworKit.pyx`.

cndolo:
> Is there anything that prevents removing the remaining classes from `_NetworKit.pyx`?

Yes. When the two `dynamics` classes are moved to the dynamics Cython file, Cython throws the error: "Cannot inherit from incomplete type" coming from the graph class. I asked around and we didn't find a solution, so I left them there for now.

michitux:
The problem seem to be cyclic imports. The dynamics module then needs to import the graph class, the graph class imports base, and base imports dynamics for `DynAlgorithm`. As `DynAlgorithm` is currently not used anywhere, I suggest to simply remove it from the base module. I am currently trying to eliminate the `_NetworKit` extension module also from all other places and will then push my changes.

michitux:
I have just pushed a commit that eliminates `_NetworKit.pyx`. Further, I have started removing some traces from `_NetworKit` from setup.py. Possibly, this list of files to be cythonized could also be generated from the extension module specification. There are still a couple of issues left, though:

* [ ] The DevGuide (`docs/DevGuide.rst`) still mentions the `_NetworKit` extension and `_NetworKit.pyx`.
* [ ] Changing a `.pxd` file currently does not trigger Cython but this is definitely necessary. Further, contents of `.pxd` files that are included using `cimport` also need to trigger Cython so unfortunately this is non-trivial (e.g., if `graph.pxd` has been changed, Cython needs to be executed for `dynamics.pyx` as `dynamics.cpp` contains code that is generated from `graph.pxd` according to the comments in the file).
* [ ] `CMakeLists.txt` contains two checks for `_NetworKit`. I do not know if they can simply be removed or if they need to be adjusted for the new extension modules.

CMakeLists.txt
```diff
@@ -244,49 +244,77 @@ if (NETWORKIT_CLANG_TIDY)
 endif()
 
 if(NETWORKIT_PYTHON)
-	if(NOT EXISTS "${PROJECT_SOURCE_DIR}/networkit/_NetworKit.cpp")
-		message(FATAL_ERROR "networkit/_NetworKit.cpp is missing. Invoke Cython manually.")
-	endif()
+	 foreach (EXT base helpers structures matching graph traversal graphtools
+			 distance dynamics generators centrality community components graphio randomization
+			 linkprediction sparsification viz flow simulation coarsening scd correlation clique
+			 globals independentset engineering stats profiling/stat)
```
angriman, line 250: For better readability, please sort the modules in alphabetical order.

cndolo: I have done so as well as in `setup.py`.

fabratu:
Both if-clauses of `_Networkit` in `CMakeLists.txt` can be deleted. I also tested the restructuring against conda, spack and brew.

cndolo:
> Both if-clauses of `_Networkit` in `CMakeLists.txt` can be deleted. The restructuring also works with the conda, spack and brew packaging.

I have updated the CMakeLists.txt file.

fabratu:
> Changing a .pxd file currently does not trigger Cython but this is definitely necessary. Further, contents of .pxd files that are included using cimport also need to trigger Cython so unfortunately this is non-trivial (e.g., if graph.pxd has been changed, Cython needs to be executed for dynamics.pyx as dynamics.cpp contains code that is generated from graph.pxd according to the comments in the file).

For this part maybe it would be easiest to switch to the internal `cythonize`-function inside our custom `build_ext()`, since it designed to track dependencies (for reference: https://github.com/cython/cython/issues/3541). By exchanging line 178-179 https://github.com/cndolo/networkit/blob/restructure-cython-interface/setup.py#L178 with:

```
cythonize("networkit/*pyx", language="c++", language_level=3)
cythonize("networkit/profiling/*pyx", language="c++", language_level=3)
```

and importing at the beginning

```
from Cython.Build import cythonize
```

the compilation and installation of NetworKit from source worked fine. If for example `base.pxd` is changed, then the cythonize is triggering rebuild of 25 (out of 28) pyx-files:

```
python3 setup.py build_ext -j4
running build_ext
Warning: passing language='c++' to cythonize() is deprecated. Instead, put "# distutils: language=c++" in your .pyx or .pxd file(s)
Compiling networkit/base.pyx because it depends on networkit/base.pxd.
Compiling networkit/centrality.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/clique.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/coarsening.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/community.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/components.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/correlation.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/distance.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/dynamics.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/flow.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/generators.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/globals.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/graph.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/graphio.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/graphtools.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/helpers.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/independentset.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/linkprediction.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/matching.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/randomization.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/scd.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/simulation.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/sparsification.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/traversal.pyx because it depends on ./networkit/base.pxd.
Compiling networkit/viz.pyx because it depends on ./networkit/base.pxd.
[ 1/25] Cythonizing networkit/base.pyx
[ 2/25] Cythonizing networkit/centrality.pyx 
...
```

Don't know if thats solves the whole issue though.

avdgrinten:
> For this part maybe it would be easiest to switch to the internal `cythonize`-function inside our custom `build_ext()`, since it designed to track dependencies (for reference: [cython/cython#3541](https://github.com/cython/cython/issues/3541)). By exchanging line 178-179 https://github.com/cndolo/networkit/blob/restructure-cython-interface/setup.py#L178 with:
> 
> ```
> cythonize("networkit/*pyx", language="c++", language_level=3)
> cythonize("networkit/profiling/*pyx", language="c++", language_level=3)
> ```
Sounds like a good idea.

setup.py
```diff
@@ -349,7 +293,6 @@ def get_outputs(self):
 	platforms			= version.platforms,
 	classifiers			= version.classifiers,
 	cmdclass			= {'build_ext': build_ext},
-	ext_modules			= [Extension(m.replace('/', '.'), sources=[]) for m in extension_module_names],
```
avdgrinten: Why is this changed?

setup.py
```diff
@@ -114,69 +115,12 @@ def determineCompiler(candidates, std, flags):
...
-		print("{} cythonized" .format(filepath), flush=True)
-
 def buildNetworKit(install_prefix, externalCore=False, withTests=False, rpath=None):
 	# Cythonize file
-	for m in extension_module_names:
-		cythonizeFile("networkit/{}.pyx".format(m))
+	cythonize("networkit/*pyx", language_level=3)
+	cythonize("networkit/profiling/*pyx", language_level=3)
+        #for m in extension_module_names:
+	#	cythonizeFile("networkit/{}.pyx".format(m))
```
avdgrinten, line 123: Unnecessary.

networkit/stats.pyx
```diff
@@ -1,3 +1,5 @@
+# distutils: language=c++
```
avdgrinten, line 1: Why do we use this directive in favor of a command line switch?

fabratu: Using a command line switch is marked as deprecated (so also output from above). 

```
Warning: passing language='c++' to cythonize() is deprecated. Instead, put "# distutils: language=c++" in your .pyx or .pxd files
```
Personally I would favor the command line switch though ...

avdgrinten: Okay, if this causes a warning, we should indeed use the annotation.

setup.py
fabratu: Thought it will not be needed anymore, but it is for sdist. Will post an update.

PR #510, opened 2020-02-24
Fix travis error for macOS

fabratu:
This should fix the build-errors for travis for macOS 10.14 clang/llvm8[1]. 

[1] https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296/10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ jobs:
         homebrew:
           packages:
             - libomp
+          update: true
 
     - name: macOS 10.14, llvm 8
       env:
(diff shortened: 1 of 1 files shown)


avdgrinten:
The commit message has to be changed.

fabratu:
Besides, the mentioned deprecation of `brew bundle` is not correct. It was deprecated once (2014) and rewritten/reintroduced some time later. 

Also - this fix is more a workaround, since it increases the build-time for travis. In theory either brew bundle has to adapt (unlikely, since the breaking update removes old API) or travis. In either case we may remove this fix in the future.

avdgrinten:
Merging this since it restores the CI functionality.

PR #511, opened 2020-02-25
Refactor OpenMP library import during cmake

fabratu:
This changes how OpenMP is handled during cmake-phase. Tested for cmake 3.11, 3.13 and 3.15 on Linux (Ubuntu, CentOS) and macOS for both native and conda-environments (+ brew for macOS).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,27 +117,47 @@ if (NETWORKIT_COVERAGE)
   set(NETWORKIT_LINK_FLAGS "${NETWORKIT_LINK_FLAGS} --coverage")
 endif()
 
-# FindOpenMP.cmake does not reliably find a user installed openmp library
-# Following section manually sets the required fields for clang-like compiler
-if(("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
-	"${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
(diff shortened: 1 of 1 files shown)


angriman:
Thanks, I think you can squash those commits into one.

avdgrinten:
What if `CONDA_PREFIX` doesn't exist?

fabratu:
Worst case this folder (since env `CONDA_PREFIX` is empty) the folder does not exist. In this case cmake just advances to the next entry.

However thanks for pointing me indirectly to some other problems - fixed those and squashed. Also integrated the support for default MacPorts-folder on macOS. Think I am finished now.

avdgrinten:
Well, as far as I see, it searches `/include` if the environment var doesn't exist, right? That's probably not a huge problem, but it is still somewhat strange. Should we put this HINT behind an `if`?

avdgrinten:
Also, what is the rationale for `HINTS` vs `PATHS`?

fabratu:
Didn't know that one before, but cmake suggests putting hard-coded paths behind `PATHS` and generated-paths behind `HINTS`[1].

[1] https://cmake.org/cmake/help/latest/command/find_library.html

avdgrinten:
LGTM now. This can be squashed to a single commit.

avdgrinten:
Fast tracking this because it fixes compilation issues.

Issue #505, opened 2020-02-18
WattsStrogatz generation infinite loops if the ring lattice generated is complete


finnbar:
Calling `WattsStrogatzGenerator(5,2,p)` for nonzero `p` will cause an infinite loop. This happens because `RegularRingLatticeGenerator(5,2)` generates a complete ring lattice, and thus the rewiring process of the Watts-Strogatz generator (https://github.com/networkit/networkit/blob/Dev/networkit/cpp/generators/WattsStrogatzGenerator.cpp#L42) never finds a suitable node to rewire to.

This happens for other complete lattices, such as `WattsStrogatzGenerator(7,3,p)`. Confusingly, these are cases that RegularRingLatticeGenerator should check for (https://github.com/networkit/networkit/blob/Dev/networkit/cpp/generators/RegularRingLatticeGenerator.cpp#L12) but this does not seem to be being checked in practice.

angriman:
Thank you for bringing this to our attention. That's a bug indeed, we are working to solve it.

cndolo:
The parameter `nNeighbors` in the `WattsStrogatzGenerator` specifies the  number of neighbours per node on each side, and not the total number of neighbours. 
The code runs in to an infinite loop during the rewiring process if the number of neighbours `k` is equal to the total number of nodes - 1, i.e. `k = n-1`.

Thanks for bringing this up; I will post a fix shortly.

angriman:
The patch has been merged, I am closing this. Thanks again @finnbar for bringing this up.

PR #513, opened 2020-02-26
Fix Watts-Strogatz generator

cndolo:
The previous implementation ended up in an infinite look when certain parameters were passed to the generator (see issue #505 ). This PR changes that and adds C++ and Python tests.

### DIFF
--- a/networkit/cpp/generators/WattsStrogatzGenerator.cpp
+++ b/networkit/cpp/generators/WattsStrogatzGenerator.cpp
@@ -5,27 +5,46 @@
 *      Author: Simon Bischof
 */
 
+// networkit-format
+
+#include <limits>
+
 #include <networkit/auxiliary/Random.hpp>
+#include <networkit/auxiliary/Log.hpp>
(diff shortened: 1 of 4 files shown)


networkit/cpp/generators/WattsStrogatzGenerator.cpp
```diff
@@ -1,54 +1,65 @@
...
+    // list all valid end nodes for rewiring an edge
+    auto validEndNode = [&G](const node u, const node v) { return u != v && !G.hasEdge(u, v); };
 
-    //rewire according to Watts and Strogatz model
-    for (node u = 0; u < nNodes; u++) {
+    std::vector<node> prevNeighbors;
+    prevNeighbors.reserve(GraphTools::maxDegree(G));
+
+    // rewire according to Watts and Strogatz model
+    for (node u = 0; u < nNodes; ++u) {
```
angriman: This loop can use the new graph iterators. Please rebase on Dev and change this loop in
```suggestion
    for (const auto u : G.nodeRange()) {
```

PR #514, opened 2020-02-27
Fix link error for icons in README.md

fabratu:


avdgrinten:
Fast tracking this because it's a trivial change.

PR #515, opened 2020-03-04
Removed tlx as requirement for Python-builds with external core lib

fabratu:
Don't know if that is the best logic to solve this issue.

As of now if we use "--networkit_external_core" in setup.py, cmake still requires tlx (and compiles it) even though it not needs it. This breaks the build-process if we download a release from github and install via pip/setup.py with an external libnetworkit (like it is required for package managers like conda).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,20 +193,22 @@ else()
 endif()
 
 ################################################################################
-# Use TLX as a CMake submodule
-if(NOT NETWORKIT_EXT_TLX)
-	if(EXISTS "${PROJECT_SOURCE_DIR}/extlibs/tlx/CMakeLists.txt")
-		add_subdirectory(extlibs/tlx)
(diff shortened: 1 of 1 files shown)


fabratu:
Local builds on this branch with the same cmake-options succeed. Seems that the cythonized `_Networkit.cpp` does not include `<networkit/graph/Graph.hpp`, but maybe there is something else (already checked for empty extlibs/tlx and clean build folder).

avdgrinten:
The proper fix for this is detecting a TLX installation (e.g., via pkg-config) and using it in favor of the submodule if it is available.

fabratu:
That's true - I will look into that. Will close this PR anyway, because `_NetworKit.pyx` references `Graph.hpp`, so libtlx is needed anyway (something I missed locally) and should not be ignored when building Python-bindings.

PR #516, opened 2020-03-04
Compile numerics

angriman:
The numerics module and its tests were never compiled  because `numerics` was not included in `networkit/CMakeLists`. This PR fixes some compilation errors in the `numerics` module and includes it in the build and updates `googletest`.

Unfortunately, `LAMGGTest` segfaults when compiled with clang-3.8/3.8.1. As far as I see, this is due to a compiler bug. Since clang-3.8 has another bug that prevents us to generalize `CSRMatrix` to store generic types (see #359), I'd suggest to move to clang-3.9.


avdgrinten:
Updates of Clang versions should go into a separate PR (which we would merge first).

networkit/cpp/numerics/LAMG/MultiLevelSetup.cpp
```diff
@@ -87,7 +87,7 @@ void MultiLevelSetup<CSRMatrix>::computeAffinityMatrix(const CSRMatrix& matrix,
     for (omp_index i = 0; i < static_cast<omp_index>(matrix.numberOfRows()); ++i) {
         double nir = 1.0 / normSquared[i];
         index cIdx = rowIdx[i];
-        matrix.forNonZeroElementsInRow(i, [&](index j, double val) {
+        matrix.forNonZeroElementsInRow(i, [&](index j, double) {
```
fabratu, line 90: Only minor:In the header-file there are also functions with anonymous parameters, but their styling is slightly different like `double /* val */`. Better use one or the other.

fabratu (changes requested):

networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
```diff
@@ -122,6 +123,6 @@ INSTANTIATE_TEST_CASE_P(ErdosRenyiEnumeratorGTest, ErdosRenyiEnumeratorGTest,
                          std::make_tuple(false, 100, 0.7),  std::make_tuple(true, 100, 0.7),
                          std::make_tuple(false, 100, 1.0),  std::make_tuple(true, 100, 1.0),
                          std::make_tuple(false, 200, 0.01), std::make_tuple(true, 200, 0.01)
-                        ),);
+                        ), ); // comma required for variadic macro
```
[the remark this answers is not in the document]
angriman: No that was a mistake, I'll restore it.

include/networkit/io/LineFileReader.hpp
```diff
@@ -10,6 +10,7 @@
 #ifndef NETWORKIT_IO_LINE_FILE_READER_HPP_
 #define NETWORKIT_IO_LINE_FILE_READER_HPP_
 
+#include <fstream>
```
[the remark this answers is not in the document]
angriman: Ok, I moved the implementation to a source file.

fabratu (dismissed):
LGTM.

avdgrinten (changes requested):
I am unhappy with the structure of this PR. Why does this change stuff in io, graph, algebraic, etc. if the goal is to compile the numerics code?

angriman:
Sorry those have been introduced by mistake, the changes in `algebraic` are needed because that module is used by `numerics`.

networkit/cpp/numerics/test/LAMGGTest.cpp
```diff
@@ -79,13 +77,9 @@ TEST_F(LAMGGTest, testSmallGraphs) {
...
 
     }
-
-    delete smoother;
 }
 
-
-
-Vector LAMGGTest::randVector(count dimension, double lower, double upper) const {
+Vector LAMGGTest::randVector(count dimension, double, double) const {
```
avdgrinten, line 82: This looks suspicious. Why doesn't the function use it's arguments?

angriman: I don't know the details about that, I just removed unused constants.

hmeyerhenke: Looks like the originally wanted functionality was removed without removing the parameters.

avdgrinten: Possibly. In that case, remove the parameters as well.

angriman: I removed them.

avdgrinten (approved):
LGTM now.

fabratu:
Merging this early since this (mostly) only adds a new module and is needed for the new release.

PR #520, opened 2020-03-13
Throw MissingDependencyError when IPython is not available

cndolo:
This PR fixes a bug that caused `ModuleNotFoundError` when IPython isn't installed. Instead, a `MissingDependencyError` is thrown as IPython is not a requirement of NetworKit.

### DIFF
--- a/networkit/profiling/profiling.py
+++ b/networkit/profiling/profiling.py
@@ -13,12 +13,18 @@ import configparser
 from . import multiprocessing_helper
 from . import stat
 from . import plot
+from ..support import MissingDependencyError
 
-from IPython.core.display import *
 import collections
 import math
 import fnmatch
 import random
+try:
+	from IPython.core.display import *
+except ImportError:
+	have_ipython = False
+else:
+	have_ipython = True
 
 # colors
 colors = {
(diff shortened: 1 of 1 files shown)


networkit/profiling/profiling.py
```diff
@@ -13,12 +13,19 @@
...
 from . import plot
+from ..support import MissingDependencyError
 
-from IPython.core.display import *
 import collections
 import math
 import fnmatch
 import random
+try:
+	import IPython
```
angriman: Is this line necessary?

cndolo: You're right - it was not needed. I have removed it.

avdgrinten:
We do not throw the MissingDependencyError and do nothing with the have_ipython variable - that seems to be an oversight.

cndolo:
Indeed - I think I sorted that out.

networkit/profiling/profiling.py
```diff
@@ -422,6 +428,8 @@ def output(self, outputType, directory, style="light", color=colors["green"], pa
 
 
 	def show(self, style="light", color=colors["green"], parallel=False):
+		if not have_ipython:
+			raise MissingDependencyError("IPython")
```
angriman: I think it would be better to put this check after the documentation.

PR #522, opened 2020-03-25
Importing networkit should not modify the global logging configuration

cqql:
Hey,
at the moment `import networkit` modifies the global logging setup. In particular, it reduces the default logging level from `WARNING` to `INFO`. I noticed this when I imported networkit and suddenly other packages started logging. Importing a package should have no side effects. If this log level is somehow essential to networkit, `setup` should set it only for a package specific root logger.

Best,
Marten


avdgrinten:
Agreed. LGTM.

angriman:
I am merging this because it has already been approved last week.

PR #527, opened 2020-04-03
Use O(n logn) implemention in StaticDegreeSequenceGenerator

manpen:
`StaticDegreeSequenceGenerator::isRealizable()` uses the Erdős–Gallai theorem to test whether a degree sequence is graphical. The old implementation was a direct translation of the theorem, which involved a sum over all degree which needs to be calculated for each node --- this gave an Theta(n^2) algorithm, and prevented us from using it with n >~ 1e6.

This PR uses the same theorem but carries out straight-forward transformations to reduce the complexity to O(n log n). The main idea is to avoid calculating the same sums again and again. Hence, rather than storing a sorted sequence of degree, we store a their suffix-sums. 

The PR also adds unit tests for this function specifically. The PR has some overlap with the unit tests for HavelHakimi, but checks more cases that are relevant for `StaticDegreeSequenceGenerator::isRealizable()`.

As a "drive-by" fix, I replaced an "rand() % max" construct with an proper "std::uniform_int_distrubtion" in a unit test of HavelHakimi.

### DIFF
--- a/networkit/cpp/generators/StaticDegreeSequenceGenerator.cpp
+++ b/networkit/cpp/generators/StaticDegreeSequenceGenerator.cpp
@@ -1,3 +1,4 @@
+// networkit-format
 /*
  * StaticDegreeSequenceGenerator.cpp
  *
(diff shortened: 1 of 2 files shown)


networkit/cpp/generators/StaticDegreeSequenceGenerator.cpp
```diff
@@ -43,22 +42,54 @@ bool StaticDegreeSequenceGenerator::isRealizable() {
...
     }
 
-    /* Second inequality. */
-    // this inequality needs a sorted sequence
-    std::vector<count> sortedSeq = seq;
-    Aux::Parallel::sort(sortedSeq.begin(), sortedSeq.end(), std::greater<count>());
+    /**
+     * Second inequality
+     * We now check that for all 0 <= j < n the following inequality holds
+     *   sum(d[i] for 0 <= i <= j) !<= sum( min(j+1, d[i]) for j < i < n),
```
fabratu, line 48: If understood correctly, the inequality should be `<=` not `!<=`, right?

fabratu:
LGTM (apart from the very small remark)

networkit/cpp/generators/StaticDegreeSequenceGenerator.cpp
manpen: Yes, you are right. The exclamation mark was meant in a \stackrel{!}{\le} sense, i.e. "make sure that this holds". But I see how this is confusion in a C++ context and thus changed it.

manpen:
Thanks for your review and patience.

PR #528, opened 2020-04-03
Include TLX as a public library

manpen:
While this is a three-line change only, please check it careful.

The underlying issue it addresses, is that if a third-party application links against NetworKit and includes headers (e.g. for Graph), the application requires TLX in its search paths. 

The solution proposed here is to export this dependency via the `networkit` library (and it non-monolithic counterparts).


avdgrinten:
What exactly does this affect? Users who include NetworKit's CMakeLists.txt in their own CMake build?

manpen:
I do. The usual way I include `networkit` and other libraries is as git (and cmake) submodules. Very convenient, esp. if you need to make changes to networkit that are not supposed to be system-wide.

avdgrinten:
LGTM.

FWIF I usually install NetworKit to a directory (using `--DCMAKE_INSTALL_PREFIX=..`) and use `pkg-config` to find it.

PR #539, opened 2020-05-13
Fix APSP

angriman:
In disconnected graphs `APSP` returns distances that are less than infinity for pairs of nodes that are not reachable. This happens because`APSP` calls multiple times `BFS/Dijkstra::run()`, but neither `BFS` nor `Dijkstra` reset `distance` before running.
This PR also includes new tests that failed with the previous implementation.

### DIFF
--- a/networkit/cpp/distance/BFS.cpp
+++ b/networkit/cpp/distance/BFS.cpp
@@ -20,12 +20,15 @@ void BFS::run() {
     reachedNodes = 1;
     sumDist = 0.;
 
+    const auto infDist = std::numeric_limits<edgeweight>::max();
+    std::fill(distances.begin(), distances.end(), infDist);
+
     if (distances.size() < z) {
-        distances.resize(z, std::numeric_limits<edgeweight>::max());
+        distances.resize(z, infDist);
         visited.resize(z, ts);
     }
 
-    if (ts++ == 255) {
+    if (ts++ == std::numeric_limits<uint8_t>::max()) {
         ts = 1;
         std::fill(visited.begin(), visited.end(), 0);
     }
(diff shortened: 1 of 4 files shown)


networkit/cpp/distance/BFS.cpp
```diff
@@ -20,12 +20,15 @@ void BFS::run() {
     reachedNodes = 1;
     sumDist = 0.;
 
+    constexpr auto infDist = std::numeric_limits<edgeweight>::max();
```
avdgrinten, line 23: `constexpr` does nothing here, use `const` (or no qualifier).

networkit/cpp/distance/BFS.cpp
```diff
@@ -20,12 +20,15 @@ void BFS::run() {
...
+    std::fill(distances.begin(), distances.end(), infDist);
+
     if (distances.size() < z) {
-        distances.resize(z, std::numeric_limits<edgeweight>::max());
+        distances.resize(z, infDist);
         visited.resize(z, ts);
     }
 
-    if (ts++ == 255) {
+    if (ts++ == std::numeric_limits<uint8_t>::max()) {
```
avdgrinten, line 31: This change also does nothing (I am indifferent about it).

networkit/cpp/distance/BFS.cpp
angriman: Thanks, I using `const`.

networkit/cpp/distance/BFS.cpp
angriman: I think it's just better than having a magic number.

angriman:
Since we are now resetting the `distance` vector at every run, does it make sense to keep using `visited`? The only reason to do so might be that `visited` uses 8-bit integers, while `distance` 64-bit floats, but we need benchmark results.

PR #540, opened 2020-05-14
warning fix: clang-tidy/bugprone-incorrect-roundings

angriman:
Fix rounding by using [std::round](https://en.cppreference.com/w/cpp/numeric/math/round), which provides the same behavior but it is safer (as reported [here](https://clang.llvm.org/extra/clang-tidy/checks/bugprone-incorrect-roundings.html)).

### DIFF
--- a/networkit/cpp/distance/EffectiveDiameter.cpp
+++ b/networkit/cpp/distance/EffectiveDiameter.cpp
@@ -32,7 +32,7 @@ void EffectiveDiameter::run() {
     // the current distance of the neighborhoods
     count h = 1;
     // number of nodes that need to be connected with all other nodes
-    auto threshold = static_cast<count>(std::ceil(ratio * G->numberOfNodes()) + 0.5);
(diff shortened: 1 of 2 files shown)


avdgrinten:
LGTM.

networkit/cpp/distance/EffectiveDiameter.cpp
```diff
@@ -32,7 +32,7 @@ void EffectiveDiameter::run() {
     // the current distance of the neighborhoods
     count h = 1;
     // number of nodes that need to be connected with all other nodes
-    auto threshold = static_cast<count>(std::ceil(ratio * G->numberOfNodes()) + 0.5);
+    auto threshold = static_cast<count>(std::round(std::ceil(ratio * G->numberOfNodes())));
```
avdgrinten, line 35: Uhm, `std::ceil` returns an integer anyway, right? So what's the point of the `std::round`?

angriman: That's true, I removed it.

PR #544, opened 2020-05-15
Support user specified source nodes for APSP algorithm.

YanzheL:
Signed-off-by: YanzheL <lee.yanzhe@yanzhe.org>

### Purpose

Currently, if user want to calculate multi-source (instead of all-pairs) shortest paths for a graph, he/she can only run SSSP for multiple times sequentially, because the API of APSP and SSSP do not support multiple custom sources.

### What does this PR do

This PR enables users to specify source nodes for APSP algorithm. Since the APSP algorithm is executed in parallel with OpenMP, users can benefit from it with minimal API change.

### API Changes

Full backwards Compatible, no breaking changes.

New C++ constructor for APSP:
```c++
APSP(const Graph &G, std::vector<count>& srcs);
```

New Cython constructor for APSP:
```cython
def __cinit__(self, Graph G, vector[count] srcs = vector[count]()):
```

### Example

Python usage:
```python
import networkit as nk
srcs = [0,2,3]
apsp = nk.distance.APSP(graph, srcs)
apsp.run()
dists = apsp.getDistances()
```

### DIFF
--- a/include/networkit/distance/APSP.hpp
+++ b/include/networkit/distance/APSP.hpp
@@ -30,6 +30,14 @@ class APSP : public Algorithm {
      */
     APSP(const Graph &G);
 
+    /**
+     * Creates the APSP class for @a G with specified source nodes.
+     *
+     * @param G The graph.
+     * @param srcs The source nodes.
+     */
+    APSP(const Graph &G, std::vector<count>& srcs);
+
     virtual ~APSP() = default;
 
     /** Computes the shortest paths from each node to all other nodes. */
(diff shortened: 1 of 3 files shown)


fabratu:
Thanks for the PR!

While your additions are technically in a good shape, it would be better to introduce a new class `MPSP` and move (most of) the implementation of `APSP` there. For compatibility-reason the `APSP` constructor could then be a special case of `MPSP`. This would make the usage and interface clearer. Is it possible for you, to restructure the PR accordingly?

YanzheL:
Thanks for your response.
I'm glad to make further changes on this PR, but I'm quite busy right now. I'll come back later~

angriman:
Thank you for your contribution. We implemented the changes you suggested in #612, we did not merge this PR because we prefer this algorithm to be implemented into its own class rather than modifying the behavior of APSP.

PR #545, opened 2020-05-18
warning fix: clang-tidy/bugprone-fold-init-type

angriman:
Fix possible loss of precision of `std::accumulate` due to truncation or overflow.


avdgrinten:
LGTM.

PR #546, opened 2020-05-18
Fixes typo, leading to errors in python-tests

fabratu:


fabratu:
Merging this now, since it is a straightforward fix (and resolves problems with both python-tests and the PI-pipeline).

PR #548, opened 2020-05-19
warning fix: clang-tidy/bugprone-integer-division

angriman:
Fix integer divisions meant to be floating-point divisions.

### DIFF
--- a/networkit/cpp/centrality/Betweenness.cpp
+++ b/networkit/cpp/centrality/Betweenness.cpp
@@ -130,13 +130,11 @@ double Betweenness::maximum(){
     if (normalized) {
         return 1;
     }
-    double score;
+
     count n = G.numberOfNodes();
-    if (G.isDirected()) {
-        score = (n-1)*(n-2);
-    } else {
-        score = (n-1)*(n-2)/2;
-    }
+    double score = (n-1)*(n-2);
+    if (!G.isDirected())
+        score /= 2.;
     return score;
 }
 
--- a/networkit/cpp/community/AdjustedRandMeasure.cpp
+++ b/networkit/cpp/community/AdjustedRandMeasure.cpp
@@ -38,11 +38,11 @@ double NetworKit::AdjustedRandMeasure::getDissimilarity(const NetworKit::Graph &
         sumEta += s * (s - 1) / 2;
     }
 
-    count n = G.numberOfNodes();
+    const count denominator = (G.numberOfNodes() * (G.numberOfNodes() - 1)) / 2;
 
     double maxIndex = 0.5 * (sumZeta + sumEta);
 
-    double expectedIndex = sumZeta * sumEta / (n * (n-1) / 2);
+    double expectedIndex = static_cast<double>(sumZeta * sumEta) / static_cast<double>(denominator);
 
     if (maxIndex == 0) { // both clusterings are singleton clusterings
         return 0.0;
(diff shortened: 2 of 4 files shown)


networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -130,13 +130,11 @@ double Betweenness::maximum(){
     if (normalized) {
         return 1;
     }
-    double score;
-    count n = G.numberOfNodes();
-    if (G.isDirected()) {
-        score = (n-1)*(n-2);
-    } else {
-        score = (n-1)*(n-2)/2;
```
avdgrinten, line 138: The old code is fine here, since `(n-1)*(n-2)` is always even. In fact, the resulting `score` is an integer.

networkit/cpp/community/AdjustedRandMeasure.cpp
```diff
@@ -38,7 +38,7 @@ double NetworKit::AdjustedRandMeasure::getDissimilarity(const NetworKit::Graph &
         sumEta += s * (s - 1) / 2;
     }
 
-    count n = G.numberOfNodes();
+    const double n = G.numberOfNodes();
```
avdgrinten, line 41: Same here.

networkit/cpp/distance/HopPlotApproximation.cpp
```diff
@@ -66,7 +66,7 @@ void HopPlotApproximation::run() {
         }
     });
     // at zero distance, all nodes can only reach themselves
-    hopPlot[0] = 1/G->numberOfNodes();
+    hopPlot[0] = 1.0 / static_cast<double>(G->numberOfNodes());
```
avdgrinten, line 69: This looks somewhat strange. Is this what the algorithm intended to do?

networkit/cpp/centrality/Betweenness.cpp
angriman: But the return type is `double`.

networkit/cpp/community/AdjustedRandMeasure.cpp
angriman: Here `n` is used to compute `double expectedIndex = sumZeta * sumEta / (n * (n-1) / 2);` isn't the denominator computed as an integer?

avdgrinten: Yes, but the argument is the same as above: `n * (n-1)` is always even, so the divison never truncates.

networkit/cpp/distance/HopPlotApproximation.cpp
angriman: I don't know exactly but that line assigns `hopPlot[0]` to zero, which is probably not wanted.

networkit/cpp/community/AdjustedRandMeasure.cpp
angriman: Which is the same argument as betweenness, ok I will change this.

angriman:
@avdgrinten I addressed the issues in `Betweenness` and `AdjustedRandMeasure`.

networkit/cpp/distance/HopPlotApproximation.cpp
fabratu: The change should be right. In the code the hop plot at zero (given as fractional number) is `n` (since all nodes can only reach themselves) divided by all possible combinations of pairs `n^2`.

PR #555, opened 2020-05-21
warning fix: clang-tidy/performance-unnecessary-value-param

angriman:
Avoid unnecessary copies of function parameters.


avdgrinten:
This PR changes a lot of prototypes - that should be considered carefully. IMHO, it does not have a high priority (defer to after the next release).

PR #557, opened 2020-05-22
Fix use-after-move

angriman:
Fix clang-tidy use-after-move error in `ApproxSpanningEdge`.


avdgrinten:
IMHO, the fix makes the code less clean. Why not just `sequence.clear()` after moving it so that the container can be re-used?

angriman:
I did not know that was possible, I find it cleaner to do as you say so I'll do that.

fabratu:
Merging this now, since it is an easy fix and resolves an issue in the CI.

PR #558, opened 2020-05-22
Fix: Commute Time Distance

### LINKED ISSUE #447: CommuteTimeDistance returns different commute times on same node pair.
The CommuteTimeDistance class has, among others, a `run()` method and a `runSinglePair(u, v)` method to compute the commute time of given node pairs. `run` preprocesses the graph and computes the commute time for all node pairs; the commute time between a node pair u,v is then queried using `distance(u,v)`. `runSinglePair(u,v)` computes the the commute time between  u and v without preprocessing, and returns the commute time. 

The `distance(u,v)` and `runSinglePair(u,v)` should return the same commute time for the same node pairs, but do not. The returned values of both functions differ.

Here is my Python code to replicate this behaviour.

```
import networkit as nk
G = nk.Graph(6, False, False)
G.addEdge(0,2)
G.addEdge(1,2)
G.addEdge(2,3)
G.addEdge(2,4)
G.addEdge(3,5)
G.addEdge(4,5)
ctd = nk.distance.CommuteTimeDistance(G)
```
Using the `runSinglePair` method:
```
print(ctd.runSinglePair(0, 2))
2.449489742783178
 ```
Using the `distance()` method after preprocessing:

```
ctd.run()
print(ctd.distance(0, 2))
3.4641016151377544
```

### PR DESCRIPTION
**Pull request by fabratu**
Fixes #447 

The volume of the graph for `distance(u,v)` was computed differently then `runSinglePair(u,v)`, leading to a difference of factor `sqrt(2)` in the results. The update also includes a warning, when the algorithm is called on directed graph.

### DIFF
--- a/include/networkit/distance/CommuteTimeDistance.hpp
+++ b/include/networkit/distance/CommuteTimeDistance.hpp
@@ -57,11 +57,6 @@ public:
      */
     uint64_t getSetupTime() const;
 
-    /**
-     * Returns the commute time distance between node @a u and node @a v.
-     * @return commute time distance between the two nodes. Needs to call run() or runApproximation() first.
-     */
-    double distance(node u, node v);
 
     /**
      * Returns the commute time distance between node @a u and node @a v.
@@ -72,10 +67,11 @@ public:
     double runSinglePair(node u, node v);
 
     /**
-     * Returns the sum of the distances from node @a u.
-     * This method does not need the initial preprocessing.
-     * @return commute sum of the distances from the node.
+     * Returns the commute time distance between node @a u and node @a v.
+     * @return commute time distance between the two nodes. Needs to call run() or runApproximation() first.
      */
+    double distance(node u, node v);
+
     double runSingleSource(node u);
 
 protected:
--- a/networkit/cpp/distance/CommuteTimeDistance.cpp
+++ b/networkit/cpp/distance/CommuteTimeDistance.cpp
@@ -11,12 +11,15 @@
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/Timer.hpp>
 #include <networkit/distance/CommuteTimeDistance.hpp>
+#include <networkit/graph/GraphTools.hpp>
 
 namespace NetworKit {
 
 CommuteTimeDistance::CommuteTimeDistance(const Graph& G, double tol): Algorithm(), G(&G), tol(tol), lamg(1e-5) {
     // main purpose of method: preparing LAMG
-
+    if(G.isDirected()) 
+        throw std::runtime_error("Commute time distance is only supported for undirected graphs.");
+    
     // construct matrix from graph
     CSRMatrix matrix = CSRMatrix::laplacianMatrix(G);
 
@@ -150,14 +153,7 @@ void CommuteTimeDistance::runParallelApproximation() {
 double CommuteTimeDistance::distance(node u, node v) {
     assureFinished();
 
-    // compute volume
-    double volG = 0.0;
-    if (! G->isWeighted()) {
-        volG = 2.0 * G->numberOfEdges();
-    }
-    else {
-        volG = 2.0 * G->totalEdgeWeight();
-    }
+    double volG = GraphTools::volume(*G);
 
     if (exactly) {
         return sqrt(distances[u][v] * volG);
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by avdgrinten** [comment]
Is there no global helper function to compute the volume (e.g., in GraphTools)?

**Comment by fabratu** [comment]
Unfortunately not really, there is a deprecated volume method (for single node) pointing to `weightedDegree`(also single node). There is also a class for volumes, approx. the number of nodes in a given radii. Should this function be moved there instead?

**Comment by avdgrinten** [comment]
If there is no such method, it's OK to keep it as it is (but having it in GraphTools would be a bit nicer).

**Comment by hmeyerhenke** [comment]
I agree that having a volume method in GraphTools would be preferable (but I am not sure if you want to make such a change at this point).

**Comment by fabratu** [comment]
I agree that adding it to GraphTools would be better. I will do it then. Since this is not initially clear to me: In the directed case, is the volume the same just without the factor of 2 (for weighted/unweighted)?

**Comment by hmeyerhenke** [comment]
Hmmm, I am not 100% sure if there is a directed volume -- it would have to distinguish between in-volume and out-volume in any case. One could see the volume as the sum of the two, which would then coincide with the undirected volume. But these distinctions need to be clear. It would be okay at this point to do the volume only for undirected graphs and take care of the directed case after a more careful consideration later on.

File: networkit/cpp/distance/CommuteTimeDistance.cpp [hunk 1]
```diff
@@ -11,12 +11,14 @@
...
 #include <networkit/auxiliary/Timer.hpp>
 #include <networkit/distance/CommuteTimeDistance.hpp>
+#include <networkit/graph/GraphTools.hpp>
 
 namespace NetworKit {
 
 CommuteTimeDistance::CommuteTimeDistance(const Graph& G, double tol): Algorithm(), G(&G), tol(tol), lamg(1e-5) {
     // main purpose of method: preparing LAMG
-
+    if(G.isDirected()) WARN("Commute time distance is only supported for undirected graphs.");
```
angriman, line 20: Instead of printing a warning, it might be better to `throw` a `std::runtime_error`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,6 +200,19 @@ double density(const Graph &G) noexcept {
     return m / (n * (n - 1));
 }
 
+double volume(const Graph &G) {
+    if (G.isDirected())
+        WARN("Commute time distance is only supported for undirected graphs.");
```
angriman, line 205: Same here.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,6 +200,19 @@ double density(const Graph &G) noexcept {
...
+
+    double volG = 0.0;
+    if (!G.isWeighted()) {
+        volG = 2.0 * G.numberOfEdges();
+    } else {
+        volG = 2.0 * G.totalEdgeWeight();
+    }
+    return volG;
```
angriman, line 213: `volG` is not needed, we can simply replace the assignments with `return`.

include/networkit/distance/CommuteTimeDistance.hpp
```diff
@@ -75,7 +70,12 @@ class CommuteTimeDistance final : public Algorithm {
      * Returns the sum of the distances from node @a u.
      * This method does not need the initial preprocessing.
      * @return commute sum of the distances from the node.
+     */    /**
```
angriman, line 73: Fix comment begin.

networkit/cpp/graph/GraphTools.cpp
hmeyerhenke: Wouldn't it be nicer to have a more general method for a Graph G and a subset of the nodes V'? Volumes of subsets are of frequent interest, e.g. in clustering algorithms.

fabratu: Sounds like a good idea. Since we want the fix in the release and this also includes new unit-tests and notebook examples, the subset-volume will be handled in a different PR.

networkit/cpp/distance/CommuteTimeDistance.cpp
```diff
@@ -11,12 +11,14 @@
...
 #include <networkit/auxiliary/Timer.hpp>
 #include <networkit/distance/CommuteTimeDistance.hpp>
+#include <networkit/graph/GraphTools.hpp>
 
 namespace NetworKit {
 
 CommuteTimeDistance::CommuteTimeDistance(const Graph& G, double tol): Algorithm(), G(&G), tol(tol), lamg(1e-5) {
     // main purpose of method: preparing LAMG
-
+    if(G.isDirected()) throw std::runtime_error("Commute time distance is only supported for undirected graphs.");
```
avdgrinten, line 20: Coding style: put the throw onto the next line.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,6 +200,17 @@ double density(const Graph &G) noexcept {
     return m / (n * (n - 1));
 }
 
+double volume(const Graph &G) {
+    if (G.isDirected())
+        throw std::runtime_error("Commute time distance is only supported for undirected graphs.");
```
avdgrinten, line 205: Wrong error message.

PR #561, opened 2020-05-27
Reduce memory-footprint for BFS/Dijkstra

fabratu:
In the current state for `SSSP` (and inherited classes) there exists a `visited` vector, taking care of bookkeeping during runs. For `BFS` and `Dijkstra` this can be exchanged by comparing distances directly and removing the allocation of memory for `visited`. This saves roughly `#nodes * bytes * instances` amount of memory. 

The trade-off is comparing `double` instead of `uint_8`. Doing a quick-bench, this creates (depending on the compiler and flags) very low to none overhead in computational time. 
http://quick-bench.com/BKRXbz_iwoVyzk3W9Px_crliMLM

Thanks for @angriman to bring this up.

### DIFF
--- a/include/networkit/distance/SSSP.hpp
+++ b/include/networkit/distance/SSSP.hpp
@@ -167,8 +167,6 @@ protected:
     std::vector<edgeweight> distances;
     std::vector<std::vector<node>> previous; // predecessors on shortest path
     std::vector<bigfloat> npaths;
-    std::vector<uint8_t> visited;
-    uint8_t ts;
 
     std::vector<node> nodesSortedByDistance;
 
--- a/networkit/cpp/distance/BFS.cpp
+++ b/networkit/cpp/distance/BFS.cpp
@@ -23,15 +23,8 @@ void BFS::run() {
     const auto infDist = std::numeric_limits<edgeweight>::max();
     std::fill(distances.begin(), distances.end(), infDist);
 
-    if (distances.size() < z) {
+    if (distances.size() < z) 
         distances.resize(z, infDist);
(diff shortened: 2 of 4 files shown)


angriman:
Is the `visited` vector still used somewhere? If not we should remove it from `SSSP`.

fabratu:
Agreed and removed. After a second inspection (was mitguided before by the dynamic versions of BFS/Dijkstra) it is not used in other classes.

include/networkit/distance/SSSP.hpp
```diff
@@ -167,7 +167,7 @@ class SSSP : public Algorithm {
     std::vector<edgeweight> distances;
     std::vector<std::vector<node>> previous; // predecessors on shortest path
     std::vector<bigfloat> npaths;
-    std::vector<uint8_t> visited;
+    //std::vector<uint8_t> visited;
     uint8_t ts;
```
angriman, line 171: Remove both `visited` and `ts` if never used.

networkit/cpp/distance/BFS.cpp
```diff
@@ -23,15 +23,7 @@ void BFS::run() {
...
-    if (distances.size() < z) {
-        distances.resize(z, infDist);
-        visited.resize(z, ts);
-    }
-
-    if (ts++ == std::numeric_limits<uint8_t>::max()) {
-        ts = 1;
-        std::fill(visited.begin(), visited.end(), 0);
-    }
+    if (distances.size() < z) distances.resize(z, infDist);
```
angriman, line 26: Coding style, `resize` to a new line.

PR #564, opened 2020-05-28
Update CHANGES.md

fabratu:


avdgrinten:
Since we do not promise ABI stability over major releases, we only need to guarantee *API* stability. Which of these changes affect the API?

fabratu:
Removed all ABI-breaks.

Issue #535, opened 2020-04-27
networkit not installable on debian 11 (testing)


gabri94:
Hello, i have tried to install networkit on a fresh virtual environment on my debian testing system, but it failed. Howver i can instal it without any problem on Debian stable (10)
I am running Python 3.8.2.

It fails while:
```Building wheel for networkit```

The output is quite long, so you can find it attached.
[nk_err.log](https://github.com/networkit/networkit/files/4541371/nk_err.log)

**Comment by gabri94** [comment]
I managed to install python3.7 and and select it as runtime for my environment. It compiled.
I can assume that there's a problem with the version 3.8 of python

**Comment by Thijss** [comment]
Can confirm this. Installed Ubuntu 20.04, which ships with python 3.8
Having the same issue. I'll try python 3.7 too.

**Comment by avdgrinten** [comment]
Could this be a Cython issue? Could you install `cython` locally (e.g. via pip3) and try to rebuild it? Does the master branch show the same behavior as the released version of NetworKit?

**Comment by fabratu** [comment]
I can confirm that this is very likely a Cython-error. In Python 3.8 the signature of `PyCode_New()` has [changed](https://github.com/cython/cython/issues/2938). In the PyPI-version of NetworKit we used an older version of Cython, which is not aware of this changes. As suggested by @avdgrinten the problem is solved by rebuilding NetworKit from source using Cython 0.29 and newer:
```
git clone https://github.com/networkit/networkit networkit  
cd networkit  
python3 setup.py build_ext [-jX]  
pip3 install -e .
```

You can do actually build different versions of NetworKit:

`git checkout master` -> version 6.0
`git checkout 6.1` -> version 6.1
(use other tags for older versions)

fabratu:
In the next release, which we plan to release in the upcoming week, this error should be resolved.

fabratu:
Closing this issue - problem is fixed in version 7.0.

PR #565, opened 2020-05-28
Update README.md and requirements for 7.0

fabratu:
This brings the README.md and requirements up-to-date.

- ~~Raised min. compiler version for GCC/G++ to 5.1~~
- ~~Aligned compiler versions for testing + documentation (+ removed older versions)~~
- Raised min. Python version to 3.5. 3.4 is EOL since nearly one year and we only test against 3.5
- Raised min. Cython version to 0.29 in order to support Python 3.8 (see #535) 

~~Note: Can't say how important it is, but I could not find a feasible way to choose g++-5.1 instead of g++5 (which should be 5.4) for travis.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,11 @@ package:
 
 - A modern C++ compiler, e.g.: [g++] (&gt;= 4.8) or [clang++] (&gt;= 3.7)
 - OpenMP for parallelism (usually ships with the compiler)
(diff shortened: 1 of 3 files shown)


avdgrinten:
Changes to the GCC / Clang version should be done at the start of a new release cycle, not at the end of the current one. (But the Cython change is necessary.)

fabratu:
Ok. Updated the PR accordingly.

PR #566, opened 2020-05-31
Fix node sizes being passed as tuple-of-list

tinloaf:
The stray comma on that line causes computed node sizes to be passed as a one-tuple containing the desired list instead of the list itself. This makes networkx choke when it tries to access the list of node sizes.


fabratu:
LGTM.

PR #567, opened 2020-05-31
Fix wrong parameter in user guide

tinloaf:
The parameter controlling node sizes in the `drawGraph` method is named `node_size`, not `nodeSizes`.


fabratu:
LGTM.

PR #568, opened 2020-06-08
warning fix: clang-tidy/readability-avoid-const-params-in-decls

angriman:
Remove `const` from top level parameters in function declarations.


fabratu:
LGTM.

PR #569, opened 2020-06-10
clang-tidy: modernize-use-using

cndolo:
Replace `typedef` with `using`.

### DIFF
--- a/networkit/cpp/centrality/ApproxCloseness.cpp
+++ b/networkit/cpp/centrality/ApproxCloseness.cpp
@@ -14,10 +14,10 @@
 
 namespace NetworKit {
 
-typedef struct ListEntry_struct {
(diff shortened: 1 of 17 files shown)


networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -14,10 +14,10 @@
 
 namespace NetworKit {
 
-typedef struct ListEntry_struct {
+using ListEntry = struct ListEntry_struct {
     node node_val;
     edgeweight dist_val;
-} ListEntry;
+};
```
avdgrinten, line 20: Just get rid of `ListEntry_struct` here. (In C, struct names and type names are not necessarily identical but in C++ that is indeed the case.)

PR #571, opened 2020-06-12
Update .travis.yml and DevGuide to new development branch

cndolo:
- The documentation build is now deployed against the master branch.
- The DevGuide is updated according to the new development branch.

### DIFF
--- a/docs/DevGuide.rst
+++ b/docs/DevGuide.rst
@@ -24,20 +24,12 @@ Contribution Workflow
 
 The development of the NetworKit project takes place at our `official repository on GitHub <https://github.com/networkit/networkit>`__.
 
-There are two important branches you need to be aware of:
-
-* **master**: Release branch
-* **Dev**: Development (stage) branch used for active development
-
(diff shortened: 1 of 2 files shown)


docs/DevGuide.rst
```diff
@@ -68,10 +60,10 @@ Now you can update your fork on a regular basis with changes from the original r
    git fetch upstream
 
    # Merge changes into your local branch
-   git checkout --track origin/Dev
-   git merge upstream/Dev
+   git checkout --track origin/master
+   git merge upstream/master
 
-This will for example allow you to keep the development branch ``Dev`` up to date. Doing this will prevent an unpleasent surprise once you're ready to submit your pull request since there will be less or no merge conflicts.
+This will for example allow you to keep the development branch ``master`` up to date. Doing this will prevent an unpleasent surprise once you're ready to submit your pull request since there will be less or no merge conflicts.
```
angriman, line 66: Probably it is better to simply write "keep the master branch up to date", as we do not have a dedicated development branch anymore. This applies also in later parts of this file.

cndolo: I incorporated your suggestion, is it ok now?

PR #572, opened 2020-06-15
clang-tidy: readability-container-size-empty

cndolo:
Replace calls to `size()` with `empty()` where applicable.


angriman:
CI fails due to an homebrew issue in one of the osx builds, nothing related to the PR.

PR #576, opened 2020-06-24
Derive exception in CppClangFormat.py from a BaseException.

cndolo:
With Python 3.8 `CppClangFormat.py` throws an error stating the exception that is raised must be derived from a `BaseException`, so that this PR adds that. I also added a blank space to the allowed variants of `clang-format` so that versions such as `/usr/bin/clang-format` can be found.

### DIFF
--- a/extrafiles/tooling/CppClangFormat.py
+++ b/extrafiles/tooling/CppClangFormat.py
@@ -17,7 +17,7 @@ import os
 
 def findClangFormat():
 	"""Tries to find clang-format-XXX variants within the path"""
-	allowed = ["clang-format" + x for x in ["-8"]]
+	allowed = ["clang-format" + x for x in ["-8", ""]]
 	for candidate in allowed:
 		if not shutil.which(candidate) is None:
 			if nkt.isVerbose():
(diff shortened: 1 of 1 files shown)


fabratu:
Including the default name makes sense, since not all packages include numbering for binary-naming. However in a previous PR (#454, based on #421) all other (including the default name) names/versions of clang-format were deleted from `CppClangFormat.py`. But maybe 9 or newer is also supported. We should add a version test with a warning if a non-supported version is called.

fabratu:
Possible problem in line 36 https://github.com/networkit/networkit/pull/576/files#diff-566ff1356c1f592971a21575f02c01afL36 as `clang-format-8`is still referenced here?

cndolo:
> Possible problem in line 36 https://github.com/networkit/networkit/pull/576/files#diff-566ff1356c1f592971a21575f02c01afL36 as `clang-format-8`is still referenced here?
IMO, not really. That function is called in line 59 after the version of `clang-format` has been determined in L50.

fabratu:
Closing this PR in favor of #645, which covers this changes (and more).

PR #578, opened 2020-07-06
Refactor WeaklyConnectedComponents

angriman:
- `WeaklyConnectedComponents` uses the same algorithm as `ConnectedComponents`, but it is re-implemented in a different file. This PR merges the two implementations in a single one using pimpl.
- A new class `ConnectedComponentsInterface` that inherits from `Algorithm` implements all methods needed for all algorithms that compute components in a graph.
- New method to extract the largest strongly connected component from a directed graph.

### DIFF
--- a/networkit/cpp/components/ConnectedComponentsImpl.cpp
+++ b/networkit/cpp/components/ConnectedComponentsImpl.cpp
@@ -1,0 +1,107 @@
+// networkit-format
+
+#include "ConnectedComponentsImpl.hpp"
+
+#include <networkit/graph/GraphTools.hpp>
+
+namespace NetworKit {
+namespace ConnectedComponentsDetails {
+
+template <bool WeaklyCC>
+ConnectedComponentsImpl<WeaklyCC>::ConnectedComponentsImpl(const Graph &G, Partition &components)
+    : G(&G), componentPtr(&components) {
+
+    if (!WeaklyCC && G.isDirected())
+        throw std::runtime_error(
+            "Error, connected components of directed graphs cannot be "
+            "computed, use StronglyConnectedComponents or WeaklyConnectedComponents instead.");
+
+    if (WeaklyCC && !G.isDirected())
+        throw std::runtime_error("Error, weakly connected components of an undirected graph cannot "
+                                 "be computed, use ConnectedComponents instead.");
+}
+
+template <bool WeaklyCC>
+void ConnectedComponentsImpl<WeaklyCC>::run() {
+    index nComponents = 0;
+    count visitedNodes = 0;
+
+    std::queue<node> q;
+    auto &component = *componentPtr;
+    component.reset(G->upperNodeIdBound(), none);
+
+    // perform breadth-first searches
+    for (const node u : G->nodeRange()) {
+        if (component[u] != none)
+            continue;
+
+        component.setUpperBound(nComponents + 1);
+        const index c = nComponents;
+
+        q.push(u);
+        component[u] = c;
+
+        do {
+            const node v = q.front();
+            q.pop();
+            ++visitedNodes;
+
+            const auto visitNeighbor = [&](node w) -> void {
+                if (component[w] == none) {
+                    q.push(w);
+                    component[w] = c;
+                }
+            };
+
+            // enqueue neighbors, set component
+            G->forNeighborsOf(v, visitNeighbor);
+            if (WeaklyCC)
+                G->forInNeighborsOf(v, visitNeighbor);
+
+        } while (!q.empty());
+
+        ++nComponents;
+
+        if (visitedNodes == G->numberOfNodes())
+            break;
+    }
+
+    hasRun = true;
+}
+
+template <bool WeaklyCC>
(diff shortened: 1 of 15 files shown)


avdgrinten:
I don't think we should encourage the "template declaration in header, specialized implementation in source files" pattern. It's certainly a rather exotic approach. Can't we use Pimpl instead?

angriman:
@avdgrinten sounds good, I'll do that as soon as I have some time.

avdgrinten:
So far, the problem is only moved around, is it?

angriman:
Yes, still working on it

angriman:
Does it look better now? I also added an abstract class so that both ConnectedComponents and WeaklyCC provide the same APIs. In a future PR I'd also like to extend this to StronglyCC and Dyn(Weakly)CC.

fabratu:
Hm - I am not familiar with using pimpl, but it does look a bit strange.

What is the goal using this idiom here? For example what is gained in comparison to basic inheritance with a partial abstract component base class (since all component classes share some logic)?

angriman:
`ConnectedComponents` and `WeaklyConnectedComponents` use basically the same algorithm which is implemented twice. The idea here is to use pimpl to implement the algorithm only once for both classes.

include/networkit/components/ConnectedComponentsImpl.hpp
```diff
@@ -0,0 +1,37 @@
...
+#define NETWORKIT_COMPONENTS_CONNECTED_COMPONENTS_IMPL_HPP_
+
+#include <networkit/Globals.hpp>
+#include <networkit/graph/Graph.hpp>
+#include <networkit/structures/Partition.hpp>
+
+namespace NetworKit {
+namespace ConnectedComponentsDetails {
+
+class ConnectedComponentsImpl {
```
michitux, line 12: Why doesn't this class inherit from `Algorithm` but instead re-implements parts of its interface (`hasRun`, `run`, `assureFinished`)?

networkit/cpp/components/ConnectedComponentsImpl.cpp
```diff
@@ -0,0 +1,152 @@
...
+            [&](const node u) { return cc.componentOfNode(u) != largestCC->first; });
+
+    } else {
+        Graph S(*G);
+        const auto components = cc.getComponents();
+        for (size_t i = 0; i < components.size(); ++i) {
+            if (i != largestCC->first)
+                for (const node u : components[i])
+                    S.removeNode(u);
+        }
```
michitux, line 145: If we already have the connected components as sets of nodes, I wonder why `GraphTools::subgraphFromNodes` is not used instead. On the other hand, first calculating the sets of nodes for each component seems wasteful if could also instead just iterate over all nodes `u` and erase `u` if `cc.componentOfNode(u) != largestCC->first`.

michitux (dismissed):
The implementations of both connected component variants are still the same. However, this introduces quite some overhead due to the virtual function calls in methods like `componentOfNode` which are called in inner loops. Unless the interface shall be implemented in further classes, I wonder why the interface does not directly contain the implementations without declaring these common methods as virtual and then only the constructor is overridden in the subclasses.

Further, I wonder if the different classes should not be separated into different header files. So far I thought we had some kind of policy that every class should be in a separate header file unless it's some kind of helper.

I would also agree with @fabratu that having a base class that contains some common members like a partition to store the connected components and all those getter methods that are common to all kinds of connected components algorithms would make sense. The weakly connected components algorithm could also simply inherit from the connected components algorithm (that would already handle the directed case) and just set some flag in the constructor (or just check if the graph is directed). This would also solve the problem that `componentOfNode` is slow if this shall also be used for strongly connected components.

avdgrinten:
The original code did implement everything in one template that was defined in a public header. Hence, I suggested pimpl to reduce compile times. If we no longer use a templated implementation, that suggestion doesn't make sense anymore.

angriman:
Thanks for the review comments, now the interface inherits from `Algorithm` and implements all the methods that are common for every algorithm that computes components.
For (Weakly)ConnectedComponents I think that using pimpl is still reasonable.

avdgrinten:
Why? There is no private implementation here. The `ConnectedComponentsImpl` is a *public* class. pimpl would only make sense if this class was private, i.e., not exported in a public header.

In general, this would only make sense to hide a templated implementation (that uses templates to distinguish directed/undirected cases or similar). That is exactly what was done in the first iteration (where I recommended pimpl). Now that the templates are gone, pimpl doesn't make sense. You can either revert to a private template-based implementation (if that makes the code easier to maintain) or remove pimpl.

angriman:
The idea is to use one implementation for two algorithms and the template parameter seems the the easiest choice to me.
I restored it in the last commit, if it looks fine I will rebase the PR.

avdgrinten:
Just a minor detail: rename `ConnectedComponentsGeneral` to `ConnectedComponentsImpl` to express the intend more cleanly. Otherwise: LGTM.

angriman:
Right, I renamed the class and the files.

avdgrinten (approved):
LGTM.

PR #580, opened 2020-07-10
Update .travis.yml

fabratu:
This PR updates the travis-builds with the following addtions/removals:

- Add macOS 10.15 (AppleClang 12.0)
- Drops macOS 10.13
- Removes the forced update for AppleClang-based macOS images
- ~~Removes the non-monolithic case for macOS~~
- Updates compilers to most recent version: gcc 8 -> 10.1, llvm 8 -> 10
- ~~Adds basic caching for brew-packages. This can reduce the time for updating brew between runs.~~
- Splits the test-builds into two phases: "Core" and "Additional". Since the execution of "Additional" is depending on the success of "Core", we don't have to wait for the whole pipeline to finish if there is an error in one of the "Core"-builds. I know there is no clear rule on where to put each test. Did this on what the following reasoning. "Core" is everything like the basic minimum requirements (minimum compilers for Linux or an unchanged system for macOS) and contribution guidelines. The rest goes into "Additional".

By updating to the most recent compilers, I found some errors (mostly induced by clang-tidy). These are fixed in the second commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,109 +9,108 @@ env:
     - export CXX_STANDARD=11
     - export MONOLITH=ON
 
+stages:
+  - "Core tests"
+  - "Additional tests"
+
+cache:
+  timeout: 600
+  ccache: true
+  directories:
+    - $HOME/Library/Caches/Homebrew
+
 jobs:
   include:
+    - name: macOS 10.15, Apple Clang
+      env:
+        - CC=cc
+        - CXX=c++
+      os: osx
+      osx_image: xcode12
+      stage: "Core tests"
+      addons:
+        homebrew:
+          packages:
+            - libomp
+          update: false
+
+    - name: macOS 10.15, llvm 10
+      env:
+        - CC=/usr/local/opt/llvm@10/bin/clang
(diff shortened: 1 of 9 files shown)


angriman:
The tests look fine to me. Does the caching system work?

fabratu:
Unclear, tests on rebuilds show no improvement. I reverted the cache addition for this PR.

.travis.yml
```diff
@@ -9,109 +9,121 @@ env:
...
           packages:
-            - llvm@8
             - libomp
           update: true
+      script:
+        - $CXX --version
+        - mkdir build_non_monolith && cd "$_"
+        - cmake -DNETWORKIT_MONOLITH=OFF -DNETWORKIT_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug ..
+        - make -j2
+        - ctest -V
```
angriman, line 48: Actually, this block is not needed, there is already a block `script_cpp_only` already defined below that does the same thing. You could move that block here, and add `MONOLITH=OFF` in `env`.

PR #584, opened 2020-07-15
Refactor EdgeListReader

angriman:
Refactoring of `EdgeListReader`
- `getNodeMap` becomes `const` and returns a const ref to the map instead of a copy
- `read` does not crash anymore if there are more than 3 arguments per line, but reads the graph as unweighted, and warns that the additional arguments are ignored
- `commentPrefix` is passed as const ref, `const` is removed from value parameters
- documentation fixes


avdgrinten:
The entire EdgeListReader is somewhat antiquated and would benefit from an update. A better (and much faster) implementation strategy could make use of memory mapped files, like the SNAP and KONECT readers do.

avdgrinten:
Yes (but I didn't carefully review them.)

angriman:
Since these changes have already been included in #587, I am closing this PR.

PR #585, opened 2020-07-16
Remove bugprone-branch-clone check

angriman:
`bugprone-branch-clone` produces false positives if an if-else statement contains a omp parallel loop e.g.:
```
if (something)
#pragma omp parallel
{
    // some code
}
else
#pragma omp parallel
{
    // some other code         
}
```
triggers the warning, even if the the parallel blocks are different.


avdgrinten:
LGTM.

PR #587, opened 2020-07-26
Use Memory Mapped Files in EdgeListReader

cndolo:
This PR refactors the `EdgeListReader` by using memory mapped files as suggested in #584 . The fixes from that PR are included here.
The `EdgeListReader` now handles continuous and non-continuous  edgelists in the same function, no longer needs to read the file twice and follows the style of the `KONECTGraphReader` and `SNAPGraphReader`.

### DIFF
--- a/networkit/cpp/io/EdgeListReader.cpp
+++ b/networkit/cpp/io/EdgeListReader.cpp
@@ -34,6 +34,14 @@ const std::map<std::string,node> &EdgeListReader::getNodeMap() const {
     return this->mapNodeIds;
 }
 
+void EdgeListReader::malformedLineError(count lineNumber, const std::string &line) {
+    std::stringstream message;
+    message << "malformed line ";
+    message << lineNumber << ": ";
+    message << line;
+    throw std::runtime_error(message.str());
+}
+
 Graph EdgeListReader::readContinuous(const std::string& path) {
     std::ifstream file(path);
     Aux::enforceOpened(file);
@@ -45,6 +53,7 @@ Graph EdgeListReader::readContinuous(const std::string& path) {
     node maxNode = 0;
     bool weighted = false;
     bool checkedWeighted = false;
+    bool ignoredParameters = false;
 
     DEBUG("separator: " , this->separator);
     DEBUG("first node: " , this->firstNode);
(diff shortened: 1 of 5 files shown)


networkit/cpp/io/EdgeListReader.cpp
```diff
@@ -10,229 +10,142 @@
...
+            return 1;
+
+        return 0;
+    };
+
+    // Map nodes and increase graph size
+    auto mapNode = [&](node in) -> node {
+        if (!continuous) {
+            // mapNodeIds is <string, node>
+            auto it = mapNodeIds.find(std::to_string(in));
```
avdgrinten, line 87: This doesn't make sense. Why do we read the node IDs as numbers and then convert to strings? Was the old code able to handle non-numerical node IDs?

cndolo: AFAICT, it wasn't. It stored the node mapping as `<string, node>` (https://github.com/networkit/networkit/blob/master/networkit/cpp/io/EdgeListReader.cpp#L178) but converted the IDs to `node` when adding the edges, e.g: https://github.com/networkit/networkit/blob/master/networkit/cpp/io/EdgeListReader.cpp#L210. 

I went along with this so as not to change the interface of the `getNodeMap()` function.

michitux: Yes, the old code handled non-numerical node ids. For example the file
```
a b
a c
c d
foo bar
```

gives the node map `{'a': 0, 'b': 1, 'bar': 5, 'c': 2, 'd': 3, 'foo': 4}` and a graph with 6 nodes and 4 edges as expected.

avdgrinten: Let's continue to support this use case and add a test for this. (Note that this leaves some strange edge cases, e.g., nodes 01 is not equal to node 1 but hopefully does that anyway...)

cndolo: The refactored reader now supports non-numerical node IDs like before. I have added a unit test (and an edge list) for said use case.

networkit/cpp/io/EdgeListReader.cpp
```diff
@@ -4,235 +4,160 @@
...
-    while(std::getline(file,line)){
-        if (line.empty()) continue;
-        if(*line.rbegin() == '\r') line.pop_back();
-        ++i;
-        if (line.compare(0, this->commentPrefix.length(), this->commentPrefix) == 0) {
+    };
+
+    while (it < end) {
+        skipWhitespaceAndSeparator();
+        if (*it == '\n') {
```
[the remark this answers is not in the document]
cndolo: It is indeed better. Thanks for the suggestion.

fabratu (approved):

fabratu:
Merging this earlier to include this optimization in the release.

PR #589, opened 2020-08-07
Adds subgraph volume-computation, support for directed graphs

fabratu:
This PR would add the following:

- Adds support for volume of directed graphs. Implementation following initial thoughts from https://github.com/networkit/networkit/pull/558#issuecomment-632704544
- Adds support for volume of subgraphs. Implementation following initial thoughts from https://github.com/networkit/networkit/pull/558#discussion_r430247248
- Adds unit-tests and python-interface https://github.com/networkit/networkit/pull/558#pullrequestreview-421021520

### DIFF
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -129,12 +129,50 @@ std::pair<node, node> size(const Graph &G) noexcept;
 double density(const Graph &G) noexcept;
 
 /**
- * Returns the volume (sum of the out-degree of all nodes) of the graph.
+ * Returns the volume of the input graph.
  *
  * @param G The input graph.
  * @return Volume of the graph.
  */
 double volume(const Graph &G);
+
+/**
+ * Returns the volume (sum of the out-degree of all nodes) of a range
+ * of nodes.
+ *
+ * @param G The input graph.
+ * @param first,last The range of nodes to insert
+ * @return Volume of the graph.
+ */
+template <class InputIt>
+double volume(const Graph &G, InputIt first, InputIt last) {
+    double outVolume = 0.0;
+
+    // Compute volume of subgraph
+    for (; first != last; ++first)
+        outVolume += G.weightedDegree(*first, !G.isDirected());
+
+    return outVolume;
+}
+
+/**
+ * Returns the inVolume (sum of the in-degree of all nodes) of a range
+ * of nodes.
+ *
+ * @param G The input graph.
+ * @param first,last The range of nodes to insert
+ * @return inVolume of the subgraph.
+ */
+template <class InputIt>
+double inVolume(const Graph &G, InputIt first, InputIt last) {
+    double inVolume = 0.0;
+
+    // Compute volume of subgraph
+    for (; first != last; ++first)
+        inVolume += G.weightedDegreeIn(*first, !G.isDirected());
+
+    return inVolume;
+}
 
 /**
  * Copies all nodes of the input graph to a new graph (edges are not copied).
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -201,14 +201,12 @@ double density(const Graph &G) noexcept {
 }
 
 double volume(const Graph &G) {
-    if (G.isDirected())
-        throw std::runtime_error("Volume computation is only supported for undirected graphs.");
+    double volume = G.totalEdgeWeight();
 
-    if (!G.isWeighted()) {
-        return 2.0 * G.numberOfEdges();
-    } else {
-        return 2.0 * G.totalEdgeWeight();
+    if (!G.isDirected()) {
+        volume *= 2.0;
     }
+    return volume;
 }
 
 Graph copyNodes(const Graph &G) {
--- a/networkit/cpp/graph/test/GraphToolsGTest.cpp
+++ b/networkit/cpp/graph/test/GraphToolsGTest.cpp
@@ -101,6 +101,46 @@ TEST_P(GraphToolsGTest, testDensity) {
     }
 }
 
+TEST_P(GraphToolsGTest, testVolume) {
+    constexpr count n = 100;
+    constexpr double p = 0.1;
+
+    auto doTest = [&](const Graph &G) {
+        // Volume for graph G
+        const auto volume = GraphTools::volume(G);
+
+        // Volume for either directed/undirected
+        double mod = directed() ? 1.0 : 2.0;
+        double baseG = weighted() ? G.totalEdgeWeight() : static_cast<double>(G.numberOfEdges());
+        EXPECT_DOUBLE_EQ(volume, mod * baseG);
+
+        // Volume for subgraph G2
+        node seed = Aux::Random::integer(G.upperNodeIdBound());
+
+        const auto first = G.nodeRange().begin();
+        const auto lastOfSet = G.nodeRange().begin().operator++(seed);
+        const auto last = G.nodeRange().end();
+
+        const auto volumeSet = GraphTools::volume(G, first, lastOfSet);
+        const auto inVolumeSet = GraphTools::inVolume(G, first, lastOfSet);
+        const auto volumeComplementSet = GraphTools::volume(G, lastOfSet, last);
+        const auto inVolumeComplementSet = GraphTools::inVolume(G, lastOfSet, last);
+
+        EXPECT_NEAR(volumeSet + volumeComplementSet, volume, 1e-7);
+        EXPECT_NEAR(inVolumeSet + inVolumeComplementSet, volume, 1e-7);
+    };
+
+    for (int seed : {1, 2, 3}) {
+        Aux::Random::setSeed(seed, false);
+        auto G = ErdosRenyiGenerator(n, p, directed()).generate();
+        if (weighted()) {
+            G = generateRandomWeights(G);
+        }
+
+        doTest(G);
+    }
+}
+
 TEST_P(GraphToolsGTest, testMaxDegree) {
     constexpr count n = 100;
     constexpr double p = 0.1;
--- a/networkit/graphtools.pyx
+++ b/networkit/graphtools.pyx
@@ -27,6 +27,8 @@ cdef extern from "<networkit/graph/GraphTools.hpp>" namespace "NetworKit::GraphT
 	pair[count, count] size(_Graph G) nogil except +
 	double density(_Graph G) nogil except +
 	double volume(_Graph G) nogil except +
+	double volume[InputIt](_Graph G, InputIt first, InputIt last) nogil except +
+	double inVolume[InputIt](_Graph G, InputIt first, InputIt last) nogil except +
 	_Graph copyNodes(_Graph G) nogil except +
 	_Graph toUndirected(_Graph G) nogil except +
 	_Graph toUnweighted(_Graph G) nogil except +
@@ -321,21 +323,61 @@ cdef class GraphTools:
 		return density(graph._this)
 
 	@staticmethod
-	def volume(Graph graph):
+	def volume(Graph graph, nodes = None):
 		"""
-		Get the volume of the input graph.
+		Get the volume (for all outgoing edges) of a graph. If a list of nodes of the graph
+		is given, the volume for the corresponding subgraph is computed.
 
 		Parameters:
 		-----------
 		graph : networkit.Graph
 			The input graph.
+		nodes : list
+			(Optional) List of nodes from the graph.
 
 		Returns:
 		--------
 		double
-			The volume of the input graph.
+			The volume of the subgraph.
 		"""
-		return volume(graph._this)
+
+		cdef vector[node] cNodes
+
+		if nodes is not None:
+			try:
+				cNodes = <vector[node]?>nodes
+				return volume[vector[node].iterator](graph._this, cNodes.begin(), cNodes.end())
+			except TypeError:
+				raise RuntimeError("Error, nodes must be a list of nodes.")
+		else:
+			return volume(graph._this)
+
+	@staticmethod
+	def inVolume(Graph graph, nodes):
+		"""
+		Get the inVolume (for all incoming edges) of a subgraph, defined by the 
+		input graph and a corresponding subset of nodes.
+
+		Parameters
+		----------
+		graph : networkit.Graph
+			The input graph.
+		nodes : vector[node]
+			A vector of nodes from the graph. 
+
+		Returns
+		-------
+		double
+			The inVolume of the input graph.
+		"""
+
+		cdef vector[node] cNodes
+
+		try:
+			cNodes = <vector[node]?>nodes
+			return inVolume[vector[node].iterator](graph._this, cNodes.begin(), cNodes.end())
+		except TypeError:
+			raise RuntimeError("Error, nodes must be a list of nodes.")
 
 	@staticmethod
 	def copyNodes(Graph graph):
--- a/networkit/test/test_graphtools.py
+++ b/networkit/test/test_graphtools.py
@@ -131,6 +131,24 @@ class TestGraphTools(unittest.TestCase):
 						G.removeEdge(u, v)
 						doTest(G)
 
+	def testVolume(self):
+		for directed in [True, False]:
+			for weighted in [True, False]:
+				G = self.getSmallGraph(weighted, directed)
+
+				# Test total volume
+				v1 = nk.graphtools.volume(G)
+				mod = 1.0 if G.isDirected() else 2.0;
+				baseG = G.totalEdgeWeight() if G.isWeighted() else G.numberOfEdges();
+				self.assertEqual(v1, mod * baseG)
+
+				# Test partial volume
+				nodes = [nk.graphtools.randomNode(G)]
+				v2 = nk.graphtools.inVolume(G, nodes)
+				v3 = nk.graphtools.volume(G, nodes)
+				self.assertLessEqual(v2, v1)
+				self.assertLessEqual(v3, v1)
+
 	def testDensity(self):
 		def doTest(G):
 			d = nk.graphtools.density(G)


fabratu:
PR is ready for review (error in doc-build fixed in #590 )

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
-    if (!G.isWeighted()) {
-        return 2.0 * G.numberOfEdges();
+double outVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double outVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
+
+    // Compute volume of subgraph
+    if (!nodes.empty() > 0 && G.numberOfNodes() != nodes.size()) {
+        for (auto it = std::begin(nodes); it != std::end(nodes); ++it) {
+            outVolume += G.weightedDegree(*it, directed);
```
michitux, line 241: Counting self-loops twice for directed graphs seems wrong. The version for the whole graph counts them twice only for undirected graphs (as I would expect). I think this should be `!directed`. Or maybe not, as the variable is actually `true` if the graph is undirected?!

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
-    if (G.isDirected())
-        throw std::runtime_error("Volume computation is only supported for undirected graphs.");
+double inVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double inVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
+
+    // Compute volume of subgraph
+    if (!nodes.empty() && G.numberOfNodes() != nodes.size()) {
+        for (auto it = std::begin(nodes); it != std::end(nodes); ++it) {
+            inVolume += G.weightedDegreeIn(*it, directed);
```
michitux, line 210: Counting self-loops twice for directed graphs seems wrong. The version for the whole graph counts them twice only for undirected graphs (as I would expect). I think this should be `!directed`. Or maybe not, as the variable is actually `true` if the graph is undirected?!

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+
+            // This removes all incoming edges from the rest of the graph to the subgraph
+            G.forInNeighborsOf(*it, [&](node v) {
+                std::unordered_set<node>::const_iterator neighborIt = nodes.find(v);
+                if (neighborIt == nodes.end()) {
+                    inVolume -= static_cast<double>(G.weight(v, *it));
```
michitux, line 216: The edge weight should instead be taken as parameter in this lambda by defining it as `[&](node, node v, edgeweight w)`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
 
-double volume(const Graph &G) {
-    if (G.isDirected())
-        throw std::runtime_error("Volume computation is only supported for undirected graphs.");
+double inVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double inVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
+
+    // Compute volume of subgraph
+    if (!nodes.empty() && G.numberOfNodes() != nodes.size()) {
```
michitux, line 208: While the optimization for not counting the whole graph seems neat, I wonder if we should not check that the graph really contains all nodes that are in the set.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+        for (auto it = std::begin(nodes); it != std::end(nodes); ++it) {
```
michitux, line 209: I would suggest a quick check if the node actually exists here as the performance impact should be minimal (given that we iterate over all neighbors, anyway).

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+            inVolume += G.weightedDegreeIn(*it, directed);
+
+            // This removes all incoming edges from the rest of the graph to the subgraph
+            G.forInNeighborsOf(*it, [&](node v) {
+                std::unordered_set<node>::const_iterator neighborIt = nodes.find(v);
+                if (neighborIt == nodes.end()) {
+                    inVolume -= static_cast<double>(G.weight(v, *it));
+                }
```
michitux, line 217: For weighted graphs, this iterates over the neighbors twice: Once to collect the edge weights in `weightedDegree` and once here to remove some of them again. Why don't we directly only add those edge weights to neighbors that are in the set, i.e., remove the call to `G.weightedDegree` and add weights in the second loop when `neighborIt != nodes.end()`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+}
 
-    if (!G.isWeighted()) {
-        return 2.0 * G.numberOfEdges();
+double outVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double outVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
+
+    // Compute volume of subgraph
+    if (!nodes.empty() > 0 && G.numberOfNodes() != nodes.size()) {
```
michitux, line 239: Same here: While the optimization for not counting the whole graph seems neat, I wonder if we should not check that the graph really contains all nodes that are in the set.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+        for (auto it = std::begin(nodes); it != std::end(nodes); ++it) {
+            outVolume += G.weightedDegree(*it, directed);
+
+            // This removes all outgoing edges from the subgraph to the rest
+            G.forNeighborsOf(*it, [&](node v) {
+                std::unordered_set<node>::const_iterator neighborIt = nodes.find(v);
+                if (neighborIt == nodes.end()) {
+                    outVolume -= static_cast<double>(G.weight(*it, v));
+                }
```
michitux, line 248: Same here: For weighted graphs, this iterates over the neighbors twice: Once to collect the edge weights in `weightedDegree` and once here to remove some of them again. Why don't we directly only add those edge weights to neighbors that are in the set, i.e., remove the call to `G.weightedDegree` and add weights in the second loop when `neighborIt != nodes.end()`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
     return m / (n * (n - 1));
 }
 
-double volume(const Graph &G) {
-    if (G.isDirected())
-        throw std::runtime_error("Volume computation is only supported for undirected graphs.");
+double inVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double inVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
```
michitux, line 205: Do you mean `bool directed = !G.isDirected()`? If yes, this is extremely confusing. Please rename the variable to `undirected`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -200,15 +200,66 @@ double density(const Graph &G) noexcept {
...
+        }
+    }
+    return inVolume;
+}
 
-    if (!G.isWeighted()) {
-        return 2.0 * G.numberOfEdges();
+double outVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double outVolume = 0.0;
+    bool directed = G.isDirected() ? false : true;
```
michitux, line 236: Do you mean `bool directed = !G.isDirected()`? If yes, this is extremely confusing. Please rename the variable to `undirected`.

michitux (changes requested):
Why is the method `volume` removed without any deprecation? Further, the term _volume_ of a subset of nodes as I know it is usually the sum of the weights of the edges incident to the nodes, see, e.g., [Wikipedia on Conductance](https://en.wikipedia.org/wiki/Conductance_(graph)).

This is two times almost the same code. I would suggest to refactor this code to just one implementation.

fabratu:
Thanks for the thorough review, this helps a lot.

While most of it is clear, I would suggest leave two implementations. This split into out/in-functions is used in several places in nkt. However I changed the naming to align with other functions like `weightedDegree`. Also re-introduced the `volume`-function and split subgraph/graph-volume computation. Does it look better now?

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -201,14 +201,65 @@ double density(const Graph &G) noexcept {
...
+    return volume;
+}
+
+double subgraphInVolume(const Graph &G, const std::unordered_set<node> &nodes) {
+    double inVolume = 0.0;
+    bool countSelfLoopsTwice = !G.isDirected();
+    count nodesNotInGraph = 0;
+
+    // Check for nodes in subset, which are not part of the graph
+    for (auto it = std::begin(nodes); it != std::end(nodes); ++it) {
```
angriman, line 223: What about `for (node u : nodes)`?

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,36 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+/**
+ * Returns the volume (sum of the in-degree of all nodes) of a subgraph 
+ * defined by a set of nodes. If the set is empty, the volume of the complete
+ * graph is computed.
+ *
+ * @param G The input graph.
+ * @param nodes Subset of nodes.
+ * @return inVolume of the subgraph.
+ */
+double subgraphInVolume(const Graph &G, const std::unordered_set<node> &nodes);
```
angriman, line 149: For the input nodes, can we use ranges? So the user is not forced to create an `unordered_set`.

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,36 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+/**
+ * Returns the volume (sum of the out-degree of all nodes) of a subgraph 
+ * defined by a set of nodes. If the set is empty, the volume of the complete
+ * graph is computed.
+ *
+ * @param G The input graph.
+ * @param nodes Subset of nodes.
+ * @return Volume of the graph.
+ */
+double subgraphVolume(const Graph &G, const std::unordered_set<node> &nodes);
```
angriman, line 160: Same here.

networkit/cpp/graph/GraphTools.cpp
angriman: Of course my comment does not hold anymore if we are going to use ranges for the input nodes.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -201,14 +201,65 @@ double density(const Graph &G) noexcept {
...
+        if (!G.hasNode(*it)) {
+            nodesNotInGraph++;
+        } 
+    } 
+
+    // Compute volume of subgraph if node-subset is non-empty
+    if (!nodes.empty() > 0 && (G.numberOfNodes() - nodesNotInGraph) != nodes.size()) {
```
michitux, line 254: The `> 0` in the first condition should be removed. The `nodesNotInGraph` should be subtracted from `nodes.size()`, not `G.numberOfNodes()`.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+        const auto G2 = GraphTools::subgraphFromNodes(G, sNodes);
+
+        const auto outVolumeSub = GraphTools::subgraphVolume(G, sNodes);
+        const auto inVolumeSub = GraphTools::subgraphInVolume(G, sNodes);
+
+        // Volume for either directed/undirected
+        double baseG2 = weighted() ? static_cast<double>(G2.totalEdgeWeight())
+                                   : static_cast<double>(G2.numberOfEdges());
+        EXPECT_NEAR(outVolumeSub, mod * baseG2, 1e-7);
+        EXPECT_NEAR(inVolumeSub, mod * baseG2, 1e-7);
```
michitux, line 132: This does not match the new definition of volume and the test fails accordingly.

michitux (changes requested):
This looks much better. I agree with @angriman concerning the use of ranges, in particular as with the new implementation, no set membership needs to be tested anymore. I am not so sure about the names of the functions, though, as this is not the volume of a subgraph, but the volume of a set of nodes. Maybe just leaving them as `volume`/`inVolume` makes more sense?

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,86 @@ std::pair<node, node> size(const Graph &G) noexcept;
 double density(const Graph &G) noexcept;
 
 /**
- * Returns the volume (sum of the out-degree of all nodes) of the graph.
+ * Returns the volume of the input graph.
  *
  * @param G The input graph.
  * @return Volume of the graph.
+ *
```
angriman, line 136: This change is not needed

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,86 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+ * 
  */
 double volume(const Graph &G);
 
+/**
+ * Returns the volume (sum of the out-degree of all nodes) of a range
+ * of nodes.
+ *
+ * @param G The input graph.
+ * @param first Iterator for the starting node in the set.
```
angriman, line 145: Maybe it is clearer `@param first,last The range of nodes to insert` as they do in cppreference (e.g., https://en.cppreference.com/w/cpp/container/vector/insert).

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,86 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+ * @param last Iterator for the last node in the set.
+ * @return volume of the graph.
```
angriman, line 147:
```suggestion
 * @return volume of the graph.
```
```suggestion
 * @return Volume of the graph.
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,86 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+ */
+template <class InputIt>
+double volume(const Graph &G, InputIt first, InputIt last) {
+    double outVolume = 0.0;
+    bool countSelfLoopsTwice = !G.isDirected();
+    count nodesInGraph = 0;
+    InputIt actIt = first;
```
angriman, line 154: Probably it is cleaner to simply declare `actIt` in the for loops below. E.g., `for (auto actIt = first; actIt != last; ++actIt)`

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,86 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+
+/**
+ * Returns the inVolume (sum of the in-degree of all nodes) of a range
+ * of nodes.
+ *
+ * @param G The input graph.
+ * @param first Iterator for the starting node in the set.
+ * @param last Iterator for the last node in the set.
+ * @return inVolume of the subgraph.
+ */
```
angriman, line 183: Same comments as in `volume`, both code and documentation.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -201,14 +201,17 @@ double density(const Graph &G) noexcept {
 }
 
 double volume(const Graph &G) {
-    if (G.isDirected())
-        throw std::runtime_error("Volume computation is only supported for undirected graphs.");
+    auto volume = 0.0;
```
angriman, line 204: I don't think there is a need for `auto` here, simply use `double`.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+
+    auto doTest = [&](const Graph &G) {
+        // Volume for graph G
+        const auto volume = GraphTools::volume(G);
+
+        // Volume for either directed/undirected
+        double mod = directed() ? 1.0 : 2.0;
+        double baseG = weighted() ? static_cast<double>(G.totalEdgeWeight())
+                                  : static_cast<double>(G.numberOfEdges());
+        EXPECT_NEAR(volume, mod * baseG, 1e-7);
```
angriman, line 116: What about `EXPECT_DOUBLE_EQ`?

networkit/graphtools.pyx
```diff
@@ -337,6 +339,46 @@ cdef class GraphTools:
...
+		nodes : unordered_set
+			A set of nodes from the graph. 
+
+		Returns
+		-------
+		double
+			The inVolume of the input graph.
+		"""
+		return subgraphInVolume(graph._this, nodes)
+
```
angriman, line 381: Could you please add some simple Python tests for these APIs?

angriman (changes requested):

fabratu:
Thanks for the additional input, took care of everything.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+        Graph::NodeIterator lastOfSet = Graph::NodeIterator(&G, seed);
+        Graph::NodeIterator last = Graph::NodeIterator(&G, G.upperNodeIdBound());
+
+        const auto volumeSet = GraphTools::volume(G, first, lastOfSet);
+        const auto inVolumeSet= GraphTools::inVolume(G, first, lastOfSet);
+        const auto volumeComplementSet = GraphTools::volume(G, lastOfSet, last);
+        const auto inVolumeComplementSet= GraphTools::inVolume(G, lastOfSet, last);
+
+        EXPECT_NEAR(volumeSet + volumeComplementSet, volume, 1e-7);
+        EXPECT_NEAR(inVolumeSet + inVolumeComplementSet, volume, 1e-7);
```
angriman, line 132: I guess you can use `EXPECT_DOUBLE_EQ` here as well right?

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,79 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+
+    // Compute volume of subgraph if the subset is really a subset
+    if (nodesInGraph != G.numberOfNodes()) {
+        for (auto actIt = first; actIt != last; ++actIt) {
+            outVolume += G.weightedDegree(*actIt, countSelfLoopsTwice);
+        }
+    // Otherwise compute volume of the complete graph
+    } else {
+        outVolume = volume(G);
+    }
```
angriman, line 167: Does it really make sense to distinguish these two cases? I think that you can simply compute the volume as
```
for (auto actIt = first; actIt != last; ++actIt) {
    outVolume += G.weightedDegree(*actIt, countSelfLoopsTwice);
}
```
and we can assume that the input range is already a set, we should not check that.

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,79 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+
+    // Compute volume of subgraph if the subset is really a subset
+    if (nodesInGraph != G.numberOfNodes()) {
+        for (auto actIt = first; actIt != last; ++actIt) {
+            inVolume += G.weightedDegreeIn(*actIt, countSelfLoopsTwice);
+        }
+    // Otherwise compute volume of the complete graph
+    } else {
+        inVolume = volume(G);
+    }
```
angriman, line 199: Same thing here

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -201,14 +201,17 @@ double density(const Graph &G) noexcept {
...
+    double volume = 0.0;
 
-    if (!G.isWeighted()) {
-        return 2.0 * G.numberOfEdges();
+    if (G.isWeighted()) {
+        volume = G.totalEdgeWeight();
     } else {
-        return 2.0 * G.totalEdgeWeight();
+        volume = G.numberOfEdges();
+    }
```
angriman, line 210: `G.totalEdgeWeight()` already handles the weighted/unweighted case, you can simply use `double volume = G.totalEdgeWeight()` and drop the `if`.

angriman (changes requested):

networkit/cpp/graph/test/GraphToolsGTest.cpp
fabratu: No, there are differences in the range of approx. 1e-5

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+
+        // Volume for either directed/undirected
+        double mod = directed() ? 1.0 : 2.0;
+        double baseG = weighted() ? static_cast<double>(G.totalEdgeWeight())
+                                  : static_cast<double>(G.numberOfEdges());
+        EXPECT_DOUBLE_EQ(volume, mod * baseG);
+ 
+        // Volume for subgraph G2
+        node seed = Aux::Random::integer(G.upperNodeIdBound());
+        auto sNodes = std::unordered_set<node>();
```
angriman, line 120: This is never used

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+
+        Graph::NodeIterator first = Graph::NodeIterator(&G, 0);
+        Graph::NodeIterator lastOfSet = Graph::NodeIterator(&G, seed);
+        Graph::NodeIterator last = Graph::NodeIterator(&G, G.upperNodeIdBound());
+
+        const auto volumeSet = GraphTools::volume(G, first, lastOfSet);
```
angriman, line 126: This is not really needed, why don't you just use something like
```
const auto volumeSet = GraphTools::volume(G, G.nodeRange().begin(), G.nodeRange().begin() + seed);
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,55 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+ * @return inVolume of the subgraph.
+ */
+template <class InputIt>
+double inVolume(const Graph &G, InputIt first, InputIt last) {
+    double inVolume = 0.0;
+
+    // Compute volume of subgraph
+    for (auto actIt = first; actIt != last; ++actIt) {
+        inVolume += G.weightedDegreeIn(*actIt, !G.isDirected());
+    }
```
angriman, line 174: Even simpler:
```
for (; first != last; ++first)
    inVolume += G. weightedDegreeIn(*first, !G.isDirected());
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -129,13 +129,55 @@ std::pair<node, node> size(const Graph &G) noexcept;
...
+ * @return Volume of the graph.
+ */
+template <class InputIt>
+double volume(const Graph &G, InputIt first, InputIt last) {
+    double outVolume = 0.0;
+
+    // Compute volume of subgraph
+    for (auto actIt = first; actIt != last; ++actIt) {
+        outVolume += G.weightedDegree(*actIt, !G.isDirected());
+    }
```
angriman, line 154: I wrote below a simpler version of this loop.

angriman (changes requested):

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -101,6 +101,48 @@ TEST_P(GraphToolsGTest, testDensity) {
...
+    constexpr count n = 100;
+    constexpr double p = 0.1;
+
+    auto doTest = [&](const Graph &G) {
+        // Volume for graph G
+        const auto volume = GraphTools::volume(G);
+
+        // Volume for either directed/undirected
+        double mod = directed() ? 1.0 : 2.0;
+        double baseG = weighted() ? static_cast<double>(G.totalEdgeWeight())
```
angriman, line 114: The type of `totalEdgeWeight` is `edgeweight` (double), you don't need to cast it to double.

angriman (dismissed):

angriman:
These changes need to be approved also by @michitux

michitux (dismissed):
There are still some minor issues with the Python interface, see individual comments. Apart from that I approve this.

networkit/graphtools.pyx
```diff
@@ -321,21 +323,47 @@ cdef class GraphTools:
...
+			A vector of nodes from the graph.
 
 		Returns:
 		--------
 		double
-			The volume of the input graph.
+			The volume of the subgraph.
 		"""
-		return volume(graph._this)
+		if nodes.size() == 0:
```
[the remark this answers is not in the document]
fabratu: Thanks, yeah this would be cleaner. Adopted the changes (also the other comment).

angriman (approved):

networkit/graphtools.pyx
```diff
@@ -321,21 +323,61 @@ cdef class GraphTools:
...
+			try:
+				cNodes = <vector[node]?>nodes
+				return volume[vector[node].iterator](graph._this, cNodes.begin(), cNodes.end())
+			except TypeError:
+				raise RuntimeError("Error, nodes must be a list of nodes.")
+		else:
+			return volume(graph._this)
+
+	@staticmethod
+	def inVolume(Graph graph, nodes = None):
```
michitux, line 356: If for `inVolume`  the nodes parameter is not optional, why is it declared with a default value of `None`, which just raises an exception below?

fabratu: True. Thanks for all the input and being more careful than me.

michitux (approved):

michitux:
I see no further problems with this code. There is just one minor thing I would like to mention which can definitely wait for another round of improvements: Currently, the GIL is not released during the computation of the volume. This means that Python code in other threads cannot run in parallel. This would be possible though. What would be required is to first save the return value in a C++ variable, i.e., a variable declared, e.g.,  with `cdef double result`. Then the call to volume/inVolume could be wrapped in a `with nogil` block in which no Python objects are touched anymore and thus no GIL is required.

PR #601, opened 2020-10-01
Update Cython section of developer guide

cndolo:
Update the documentation on Cython development according to the new structure.


avdgrinten (approved):
LGTM. (But the entire DevGuide.rst file needs some revision).

PR #604, opened 2020-10-14
Fixes build order for documentation website

fabratu:
- This is neccessary for older cmake versions


avdgrinten:
LGTM, this makes more sense than the previous code anyway.

avdgrinten (approved):

fabratu:
Thanks and good to know concerning the code. Will merge this earlier, since it is a really small fix.

PR #605, opened 2020-10-16
Bugfix& Code Improvement: GlobalTradeSequence

manpen:
This PR addresses the following issues:
- `GlobalTradeSequence` produced one global trade too many. As a result, `GlobalCurveball` performed k+1 global trade if k were configured. For users this should have little implications (algorithms performs is slightly slower, but "adds more entropy" which I found when comparing the Markov Chain's mixing times)
- `FixedLinearCongruentialMap` requires ~2n bits to deal with n bit node ids, so it cannot process more than ~2 Billion nodes with 64bit arithmetic. This was previously ignored and is now explicitly checked.
- Small code improvements and deduplication.

**Question**: `randomization/GlobalTradeSequence.hpp` contains only implementation details  and only adds to the `networkit::CurveballDetails` namespace. So I suggest moving the header file into the source directory `networkit/cpp/randomization` (similarly to e.g. `CurveballImpl.hpp`). Should I do this in this PR, or do we need some deprecation period. The file is never referenced in a public interface (and I cannot think of a reason why it should be).

### DIFF
--- a/include/networkit/randomization/GlobalTradeSequence.hpp
+++ b/include/networkit/randomization/GlobalTradeSequence.hpp
@@ -22,6 +22,9 @@
 namespace NetworKit {
 namespace CurveballDetails {
 
+template <typename T>
+class FixedLinearCongruentialMap; // forward declaration (implementation below)
+
 /**
  * Computes a bijection f:[p]->[p], x -> (a*x+b) mod p where p is chosen
  * as the smallest p >= n and p prime. a and b are drawn unif at random.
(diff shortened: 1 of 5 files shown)


avdgrinten (dismissed):
LGTM.

Moving the header would require a deprecation period. We could add the private header, let it set some `NETWORKIT_PRIVATE_GLOBALTRADESEQUNCE_INCLUDE` macro, include the public header and warn in `GlobalTradeSequence.hpp` if that macro is not set. After removal of the header, we just move its contents.

Do you want to include that in the PR, or should this be merged independently?

avdgrinten (approved):
LGTM.

manpen:
Thank you for your suggestions and fast feedback!

Issue #532, opened 2020-04-16
Switch for loops in ApproxBetweeness for score aggregation?


rionda:
https://github.com/networkit/networkit/blob/ecb1af8c2c6ef3c484b45cdb9fdead584cbd94bc/networkit/cpp/centrality/ApproxBetweenness.cpp#L108

 https://github.com/networkit/networkit/blob/ecb1af8c2c6ef3c484b45cdb9fdead584cbd94bc/networkit/cpp/centrality/ApproxBetweenness.cpp#L109

(I admit this suggestion is not based on timing measurements).
 
There is a cost associated to the setup of an omp parallel loop. Rather than paying this cost as many times as there are processors/threads (i.e., local scores to be aggregated), switching the loops should allow to only pay the cost once.

(one also has to wonder whether there should be a special case that skips the for loops altogether if there is a single thread, and just assigns `local` to `scoreData`).

(thinking of the bigger picture, I suspect that this implementation can be reorganized to use an OMP reduction, avoiding this separate aggregation step, or better, incorporating it in the reduction)

Matteo

rionda:
Moreover isn't the current code potentially race-y? 
Two threads could be trying to increment `scoreData[v]` at the same time.
 It won't happen if the loops are switched.

angriman:
I think that switching the two loops makes sense. About the reduction, I am not sure how that would work since `scoreData` is a vector, so all threads need to work on their separate copy anyway. However, what about replacing
https://github.com/networkit/networkit/blob/ecb1af8c2c6ef3c484b45cdb9fdead584cbd94bc/networkit/cpp/centrality/ApproxBetweenness.cpp#L97
with an atomic operation? This would allow us to remove the `scorePerThread` vector and skip the aggregation step.

rionda:
I'm not a fan: the atomic puts a lock at the vector level, rather than at the single element level. It's too coarse. I find reductions much more elegant, but YMMV. =)

angriman:
Do you mean an OpenMP reduction? Because I am not sure how it could work here. Btw I compared the running time with 24 threads of swapping loops and atomic operations (here is the plot [atomic-agg.pdf](https://github.com/networkit/networkit/files/5222840/atomic-agg.pdf)) and also atomic operations vs the original implementation (here is the plot [atomic-orig.pdf](https://github.com/networkit/networkit/files/5222846/atomic-orig.pdf)).
[...]

PR #607, opened 2020-10-20
Refactor ApproxBetweenness

angriman:
This PR improves the performance of `ApproxBetweenness`:
- Atomic updates of vertex scores instead of keeping thread-local vectors (as discussed in #532)
- Preallocate memory for `SSSP` to avoid dynamic memory allocations inside the main loop

It also fixes a bug in `Dijkstra` (if `run` is called multiple times, the algorithm could start with a non-empty heap and yield wrong results).

Additional changes:
- Make `ApproxBetweenness` final
- Make member variables `epsilon` and `delta` const (they cannot be changed from outside)
- Make `numberOfSamples()` const
- Remove `const` from value parameters

### DIFF
--- a/networkit/cpp/centrality/ApproxBetweenness.cpp
+++ b/networkit/cpp/centrality/ApproxBetweenness.cpp
@@ -22,7 +22,7 @@
 
 namespace NetworKit {
 
-ApproxBetweenness::ApproxBetweenness(const Graph& G, const double epsilon, const double delta, const double universalConstant) : Centrality(G, true), epsilon(epsilon), delta(delta), universalConstant(universalConstant) {
+ApproxBetweenness::ApproxBetweenness(const Graph& G, double epsilon, double delta, double universalConstant) : Centrality(G, true), epsilon(epsilon), delta(delta), universalConstant(universalConstant) {
 
 }
 
--- a/networkit/cpp/centrality/ApproxBetweenness.cpp
+++ b/networkit/cpp/centrality/ApproxBetweenness.cpp
@@ -47,69 +47,54 @@ void ApproxBetweenness::run() {
     r = ceil((universalConstant / (epsilon * epsilon)) * (floor(log2(vd - 2)) + 1 - log(delta)));
 
     INFO("taking ", r, " path samples");
-    // parallelization:
-    count maxThreads = omp_get_max_threads();
-    DEBUG("max threads: ", maxThreads);
-    std::vector<std::vector<double> > scorePerThread(maxThreads, std::vector<double>(G.upperNodeIdBound()));
-    DEBUG("score per thread size: ", scorePerThread.size());
     handler.assureRunning();
-    #pragma omp parallel for
-    for (omp_index i = 1; i <= static_cast<omp_index>(r); i++) {
-        count thread = omp_get_thread_num();
-        DEBUG("sample ", i);
-        // if (i >= 1000) throw std::runtime_error("too many iterations");
-        // DEBUG
-        // sample random node pair
-        node u, v;
-        u = GraphTools::randomNode(G);
-        do {
-            v = GraphTools::randomNode(G);
-        } while (v == u);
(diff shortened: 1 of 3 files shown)


networkit/cpp/centrality/ApproxBetweenness.cpp
```diff
@@ -46,76 +45,72 @@ void ApproxBetweenness::run() {
...
         } while (v == u);
 
-        // runs faster for unweighted graphs
-        std::unique_ptr<SSSP> sssp;
-        if (G.isWeighted()) {
-            sssp.reset(new Dijkstra(G, u, true, false, v));
-        } else {
-            sssp.reset(new BFS(G, u, true, false, v));
-        }
+        auto &sssp = *(sssps[omp_get_thread_num()].get());
```
michitux, line 71: In my opinion, it would be more elegant to initialize the SSSP instance in an outer `omp parallel` section, in which then the `omp for` is nested. That way, the global vector could be avoided.

michitux (changes requested):
As already mentioned in #532, I would suggest to just keep a vector of doubles instead of using `std::atomic` and using `#pragma omp atomic`, as this also results in a CAS loop but without the overhead of having to copy the result.

networkit/cpp/centrality/ApproxBetweenness.cpp
angriman: That's a good idea, I've implemented it.

networkit/cpp/centrality/ApproxBetweenness.cpp
```diff
@@ -47,75 +47,61 @@ void ApproxBetweenness::run() {
...
+        for (omp_index i = 1; i <= static_cast<omp_index>(r); i++) {
+            DEBUG("sample ", i);
+            // sample random node pair
+            node u, v;
+            u = GraphTools::randomNode(G);
+            do {
+                v = GraphTools::randomNode(G);
+            } while (v == u);
+
+            auto &sssp = *(ssspPtr.get());
```
avdgrinten, line 66: Why not just `*ssspPtr`?

angriman: Ah, true

michitux:
Currently, each thread executes a static number of shortest path searches. I wonder if it could improve scheduling if instead all threads just independently executed shortest path searches until the required number of searches is reached. This condition could be checked by first atomically incrementing a global counter at the beginning of every iteration. When the incremented counter exceeds the wanted number of paths, the thread stops. That way, exactly the required number of shortest path searches is executed. As quite some work happens in each iteration, the atomic update should not impact the overall performance.

The drawback of the above-mentioned approach is that this makes the whole algorithm completely non-deterministic. At the moment, the algorithm should be deterministic if the number of threads is kept the same. The algorithm could be made completely deterministic by using a hash function for selecting the source/target node that uses the sample id combined with some random numbers that are drawn at the beginning of the algorithm as input.

michitux (approved):

avdgrinten:
It would be enough to re-initialize the PRNG in rach iteration to a seed derived from the atomic iteration count.

... but maybe we're diverging too far from the original scope of the PR here.

michitux:
> It would be enough to re-initialize the PRNG in each iteration to a seed derived from the atomic iteration count.

In principle yes, but initializing a Mersenne Twister with its 2.5 KiB state might be quite slow compared to a small graph / fast shortest path computation.

> ... but maybe we're diverging too far from the original scope of the PR here.

Yes, that's also why I approved the PR. If you think it would be worth considering the additional changes, feel free to open a separate issue for it.

rionda:
> Currently, each thread executes a static number of shortest path searches. I wonder if it could improve scheduling if instead all threads just independently executed shortest path searches until the required number of searches is reached.

Couldn't one use "#pragma omp for schedule(nonmonotonic : dynamic)` to at least alleviate the issue a bit?

Issue #592, opened 2020-09-01
Wrong normalization for undirected graphs in Betweenness?


rionda:
I may be wrong, but the normalization of the scores (when the `Betweenness` object is created with `normalized = true`) is not correct when the graph is undirected, due to the line:

https://github.com/networkit/networkit/blob/36a5fc57a2f6ba8d19d8fcee2c340f2434e763b6/networkit/cpp/centrality/Betweenness.cpp#L113

To see this issue, run the algorithm on the undirected graph `A - B - C` twice, once with `normalized = false` and once with `normalized = true`. The score of `B` is 2.0 both times. In the `normalized = true` case, it is obtained as 2.0 / ((2 * 1) / 2)), where the denominator is the value of the `pairs` variable (with the division by two coming from the line quoted above).

When the graph is undirected, Brandes' algorithm accounts *twice* the contribution of each *unordered* pair `(u,v)` to the non-normalized score of `w` (with `w \neq v` and `w \neq u`): one time the contribution is counted for the *ordered* pair `(u,v)`, and once for the *ordered* pair `(v,u)`. 

Thus, if one wants to get normalized scores, one should actually divide the non-normalized scores by the number of *ordered* pairs, not by the number of *unordered* pairs (which is what happens in the code right now, due to the line above).  Therefore, the factor by which one should divide is the same in both the directed and the undirected case. Thus, the line quoted above is wrong, and should be removed.

The same issue likely affects also the edge centrality (thus the whole `if (!G.isDirected())` block should be removed), but I leave that as an exercise to the reader =) .

As a side note, that `G.forNodes` can likely be a `G.parallelForNodes`. 

There is a similar issue also in the `Betweenness::maximum` function: the maximum non-normalized score in the undirected case is no different than in the directed case.

angriman:
Fixed in #608, thanks @rionda for bringing this up.

ramirogalvez:
I have had been obtaining different betweeness score across computers, and just realized that this is due to this comment (both of them have different networkit's versions). Now I see that at the end of @rionda comment it is stated that:

"There is a similar issue also in the Betweenness::maximum function: the maximum non-normalized score in the undirected case is no different than in the directed case."

However, when one sees the current code for Betweenness::maximum, this last bug does not seem to have been fixed (it still divides by 2). Maybe you checked this already and it was no neccesary to delete the division. But if not, I just wanted to report this issue.

PR #608, opened 2020-10-21
Refactor Betweenness

angriman:
This PR brings the following changes to `Betweenness`:
- Fix the normalization for unweighted graphs as suggested in #592 
- Faster parallel performances by aggregating the vertex and edge scores using `omp atomic` operations instead of using thread-local vectors as done in #607 
- Parallel score normalization
- Additional tests for the normalized betweenness

### DIFF
--- a/networkit/cpp/centrality/Betweenness.cpp
+++ b/networkit/cpp/centrality/Betweenness.cpp
@@ -32,9 +32,6 @@ void Betweenness::run() {
 
     // thread-local scores for efficient parallelism
     count maxThreads = omp_get_max_threads();
-    std::vector<std::vector<double> > scorePerThread(maxThreads, std::vector<double>(G.upperNodeIdBound()));
-    DEBUG("score per thread: ", scorePerThread.size());
-    DEBUG("G.upperEdgeIdBound(): ", G.upperEdgeIdBound());
     std::vector<std::vector<double> > edgeScorePerThread;
     if (computeEdgeCentrality) {
         edgeScorePerThread.resize(maxThreads, std::vector<double>(G.upperEdgeIdBound()));
@@ -82,21 +79,15 @@ void Betweenness::run() {
 
 
             }
-            if (t != s) {
-                scorePerThread[omp_get_thread_num()][t] += dependency[t];
-            }
+            if (t != s)
+#pragma omp atomic
+                scoreData[t] += dependency[t];
         }
     };
     handler.assureRunning();
     G.balancedParallelForNodes(computeDependencies);
     handler.assureRunning();
-    DEBUG("adding thread-local scores");
-    // add up all thread-local values
-    for (const auto &local : scorePerThread) {
-        G.parallelForNodes([&](node v){
-            scoreData[v] += local[v];
-        });
-    }
+
     if (computeEdgeCentrality) {
         for (const auto &local : edgeScorePerThread) {
             for (count i = 0; i < local.size(); i++) {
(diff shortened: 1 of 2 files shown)


fabratu (approved):
LGTM.

networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -30,20 +30,9 @@ void Betweenness::run() {
...
-    if (computeEdgeCentrality) {
-        edgeScorePerThread.resize(maxThreads, std::vector<double>(G.upperEdgeIdBound()));
-    }
-    DEBUG("edge score per thread: ", edgeScorePerThread.size());
-
-    std::vector<std::vector<double>> dependencies(maxThreads, std::vector<double>(z));
+    std::vector<std::vector<double>> dependencies(omp_get_max_threads(), std::vector<double>(z));
     std::vector<std::unique_ptr<SSSP>> sssps;
-    sssps.resize(maxThreads);
+    sssps.resize(omp_get_max_threads());
```
michitux, line 35: Similar to #607, the `SSSP` instance could instead be initialized here without the global `sssps` vector.

networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -59,68 +48,53 @@ void Betweenness::run() {
...
             scoreData[u] = scoreData[u] / pairs;
         });
-        if (computeEdgeCentrality) {
-            for (count edge = 0; edge < edgeScoreData.size(); edge++) {
-                edgeScoreData[edge] =  edgeScoreData[edge] / edges;
-            }
-        }
+
+        if (computeEdgeCentrality)
+            G.parallelForEdges([&](node, node, edgeweight, edgeid eid) {
```
michitux, line 95: I have not tested this, but using `G.parallelForEdges` here seems like a total waste of running time unless really a lot of edges have been deleted as this iterates over all nodes and their outgoing edges which involves a lot of random memory accesses compared to the simple sweep of the `edgeScoreData` vector. I would suggest to instead use the old loop with a simple `#pragma omp parallel for`.

michitux (changes requested):

networkit/cpp/centrality/Betweenness.cpp
angriman: Good point, thanks. Amended as suggested.

networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -59,68 +48,53 @@ void Betweenness::run() {
...
-        if (computeEdgeCentrality) {
-            for (count edge = 0; edge < edgeScoreData.size(); edge++) {
-                edgeScoreData[edge] =  edgeScoreData[edge] / edges;
-            }
-        }
+
+        if (computeEdgeCentrality)
+#pragma omp parallel for
+            for (omp_index i = 0; i < static_cast<omp_index>(edgeScoreData.size()); ++i)
+                edgeScoreData[i] =  edgeScoreData[i] / static_cast<double>(edges);
```
michitux, line 97: For the sake of readability I would suggest to add some braces around these code blocks, in particular with the `#pragma` that is not indented this looks weird.

michitux (approved):

networkit/cpp/centrality/Betweenness.cpp
angriman: This block might look suspicious indeed, I added braces.

michitux (approved):

Issue #579, opened 2020-07-09
Setting the number of OpenMP threads doesn't seem to affect performance


enourbakhsh:
I migrated from `igraph` to `networkit` to take advantage of its parallelism in finding connected components of huge graphs. I ran the following snippet first with `num_threads = 1` and then with `num_threads = 8` and `num_threads = 16`.  There was no noticeable difference in running time between the three. Am I missing something?

``` python
import networkit as nk
import numpy as np

def get_cc(edges, num_nodes, num_threads):
    nk.setNumberOfThreads(num_threads)
    G = nk.Graph(num_nodes, directed=False)    
    for edge in edges:
        G.addEdge(*edge)
    cc = nk.components.ConnectedComponents(G)
    cc.run()
    return cc.getComponents()

np.random.seed(0)
n = 10_000_000
a = np.random.randint(n//1.8, size=n)
b = np.random.randint(n//1.3, size=n)

edges = zip(a,b)
num_nodes = max(a)+max(b) # apparently it doesn't automatically detect the possible number of nodes?
num_threads = 16

%timeit res = get_cc(edges, num_nodes, num_threads)
```

**Timing results:**

`num_threads = 1`: 1.99 s ± 20.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
`num_threads = 8`: 2.08 s ± 34.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
`num_threads = 16`: 2.07 s ± 40.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

I later checked some values:
``` python
nk.setNumberOfThreads(5)
nk.getMaxNumberOfThreads()
```
> 5

``` python
nk.getCurrentNumberOfThreads()
```
> 1

Does this imply anything? I don't know why the current number of threads should be 1 after I ask for 5 for example.

angriman:
The implementation of `ConnectedComponents` does not use multithreading, you can try to use `ParallelConnectedComponents`.

enourbakhsh:
Thanks @angriman, I tried it but it throws the following error (maybe I don't need `.getComponents()`?):
>AttributeError: 'networkit._NetworKit.ParallelConnectedComponents' object has no attribute 'getComponents'

Is there a similar multithreading feauture in `addEdge(...)` that creates the graph?

[...]

PR #609, opened 2020-10-21
Added getComponents for ParallelComponents

fabratu:
From #579:
Adresses missing `getComponents()` for `ParallelComponents` (other component-variants do implement/expose this functionality)

### DIFF
--- a/include/networkit/components/ParallelConnectedComponents.hpp
+++ b/include/networkit/components/ParallelConnectedComponents.hpp
@@ -44,21 +44,25 @@ public:
     /**
      * This method returns the number of connected components.
      */
-    count numberOfComponents();
+    count numberOfComponents() const;
 
     /**
      * This method returns the the component in which node query is situated.
      *
      * @param[in]	query	the node whose component is asked for
      */
-    count componentOfNode(node u);
+    count componentOfNode(node u) const;
 
 
     /**
      * Return a Partition that represents the components
      */
-    Partition getPartition();
+    Partition getPartition() const;
 
+    /**
+     * @return Vector of components, each stored as (unordered) set of nodes.
+     */
+    std::vector<std::vector<node>> getComponents() const;
 
 private:
     const Graph* G;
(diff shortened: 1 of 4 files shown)


networkit/components.pyx
```diff
@@ -140,14 +141,43 @@ cdef class ParallelConnectedComponents(Algorithm):
...
 
+	def getComponents(self):
+		""" Get the connected components, each as a list of nodes.
+
+		Returns:
+		--------
+		list:
+			The connected components.
+		"""
+		return (<_ConnectedComponents*>(self._this)).getComponents()
```
angriman, line 179: Shouldn't you cast to `ParallelConnectedComponents`?

include/networkit/components/ParallelConnectedComponents.hpp
```diff
@@ -59,6 +59,10 @@ class ParallelConnectedComponents final : public Algorithm {
      */
     Partition getPartition();
 
+    /**
+     * @return Vector of components, each stored as (unordered) set of nodes.
+     */
+    std::vector<std::vector<node>> getComponents();
```
angriman, line 65: Can we make this function `const`?

networkit/cpp/components/ParallelConnectedComponents.cpp
```diff
@@ -164,20 +166,35 @@ void ParallelConnectedComponents::runSequential() {
...
 
 count ParallelConnectedComponents::componentOfNode(node u) {
+    assureFinished();
     assert (component[u] != none);
     return component[u];
 }
 
+std::vector<std::vector<node>> ParallelConnectedComponents::getComponents() {
+
+    // transform partition into vector of unordered_set
```
angriman, line 192: Hm not exactly

include/networkit/components/ParallelConnectedComponents.hpp
fabratu: Done - changed also the other functions

networkit/cpp/components/ParallelConnectedComponents.cpp
fabratu: Should be more careful with c&p

networkit/components.pyx
fabratu: Yes ;)

angriman (approved):
LGTM

PR #610, opened 2020-10-21
Travis CI: Updates llvm to most recent version 11.0.0

fabratu:


.travis.yml
```diff
@@ -48,10 +48,10 @@ jobs:
         - make -j2
         - ctest -V
 
-    - name: macOS 10.15, llvm 10
+    - name: macOS 10.15, llvm 11
```
angriman, line 51: Since we are installing the latest llvm version, what about changing `llvm 11` with something like "latest llvm"?

.travis.yml
```diff
@@ -90,10 +90,10 @@ jobs:
             - libomp
           update: false
 
-    - name: macOS 10.14, llvm 10
+    - name: macOS 10.14, llvm 11
```
angriman, line 93: Same here.

angriman (changes requested):
Since now we are installing the latest available llvm version, shouldn't we do the same for gcc?

.travis.yml
avdgrinten: Maybe `lastest llvm (version 11)`?

avdgrinten:
LGTM. It would be nice to also add the term "latest" to GCC builds and macOS versions but we can also merge as-is.

avdgrinten (approved):

angriman (dismissed):

.travis.yml
```diff
@@ -48,10 +48,10 @@ jobs:
         - make -j2
         - ctest -V
 
-    - name: macOS 10.15, llvm 10
+    - name: macOS 10.15, lastest llvm (version 11)
```
michitux, line 51: This should be `latest`, not `lastest`

.travis.yml
```diff
@@ -90,10 +90,10 @@ jobs:
             - libomp
           update: false
 
-    - name: macOS 10.14, llvm 10
+    - name: macOS 10.14, lastest llvm (version 11)
```
michitux, line 93: Same here, this should be `latest`, not `lastest`.

fabratu:
Modified also gcc (and fixed the typo ...)

michitux:
As far as I understand, for llvm we are now explicitly installing and using the latest version but for GCC we are explicitly installing and using version 10 instead of just using whatever the latest version is. Is there any reason for this?

.travis.yml
```diff
@@ -62,17 +62,17 @@ jobs:
             - libomp
           update: true
 
-    - name: macOS 10.15, GCC 10
+    - name: macOS 10.15, latest GCC (version 10)
       env:
-        - CC=/usr/local/opt/gcc@10/bin/gcc-10
-        - CXX=/usr/local/opt/gcc@10/bin/g++-10
+        - CC=/usr/local/opt/gcc/bin/gcc-10
+        - CXX=/usr/local/opt/gcc/bin/g++-10
```
michitux, line 68: Is it possible to remove the `-10` suffixes from the compiler names? Otherwise this will break after the release of GCC 11 and I would instead suggest to stay at the fixed compiler version.

.travis.yml
```diff
@@ -90,10 +90,10 @@ jobs:
             - libomp
           update: false
 
-    - name: macOS 10.14, llvm 10
+    - name: macOS 10.14, latest llvm (version 11)
```
angriman, line 93: Can we drop `(version 11)`? Otherwise this would be misleading once llvm 12 is released.

.travis.yml
fabratu: No it is not unfortunately. We could create links of course - but this does not solve the problem. And we could switch back of course, but a break caused by a new release was the reason for this PR in the first place :(

.travis.yml
fabratu: I think the one way or the other, we have to update the `.travis.yml` regularly to keep track of the compiler changes.

fabratu:
Hm, based on the last comments ... should we switch back to version specific packages and leave out the latest-reference for everything else then llvm (which is now installed without version numbering)?

avdgrinten:
I think the version number is valuable, (i) to reproduce the issue and (ii) to know what version we are tracking. As you said, *any* update will require manual changes anyway.

avdgrinten (approved):

angriman:
The builds on mac os 10.14 with latest GCC and llvm seem to take too long and they time out. Should we drop them? We already test mac os 10.14 with AppleClang.

fabratu:
I think a lot of it has to do with outdated images from travis (which by the time gets worse). They don't update the underlying packages in the images. So once we do brew update, it takes like 10 minutes to setup everything. Removing the update solves this (we had that in an earlier version), but then the new compiler are not available. So yes ... I would say we drop for the time being the tests for 10.14 gcc/llvm - or more in general the non-system compiler for older versions of macOS.

angriman (approved):

PR #612, opened 2020-10-23
Feature SPSP

angriman:
This PR implements a `SPSP` algorithm (Some-Pairs Shortest-Paths, but suggestions for better names are welcome) as suggested in #544. The algorithm computes shortest-path distances from a set of user-defined source vertices to all the other vertices in the graph.

As suggested by @fabratu, the implementation has been done in another class instead of within `APSP`.

### DIFF
--- a/networkit/cpp/distance/SPSP.cpp
+++ b/networkit/cpp/distance/SPSP.cpp
@@ -1,0 +1,42 @@
+/*
+ * SPSP.cpp
+ *
+ *  Created on: 23.10.2020
+ *      Author: Eugenio Angriman <angrimae@hu-berlin.de>
+ */
+
+// networkit-format
+
+#include <memory>
+#include <omp.h>
+
+#include <networkit/distance/BFS.hpp>
+#include <networkit/distance/Dijkstra.hpp>
+#include <networkit/distance/SPSP.hpp>
+#include <networkit/distance/SSSP.hpp>
+
+namespace NetworKit {
+
+void SPSP::run() {
+    distances.resize(sources.size());
+
+#pragma omp parallel
+    {
+        std::unique_ptr<SSSP> sssp;
+        if (G->isWeighted())
+            sssp = std::unique_ptr<SSSP>(new Dijkstra(*G, 0, false));
+        else
+            sssp = std::unique_ptr<SSSP>(new BFS(*G, 0, false));
+
+#pragma omp for
+        for (omp_index i = 0; i < sources.size(); ++i) {
+            sssp->setSource(sources[i]);
+            sssp->run();
+            distances[i] = sssp->getDistances();
+        }
+    }
+
+    hasRun = true;
+}
+
+} // namespace NetworKit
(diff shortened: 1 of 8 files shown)


networkit/cpp/distance/SPSP.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <networkit/distance/BFS.hpp>
+#include <networkit/distance/Dijkstra.hpp>
+#include <networkit/distance/SPSP.hpp>
+#include <networkit/distance/SSSP.hpp>
+
+namespace NetworKit {
+
+void SPSP::run() {
+    distances.clear();
+    distances.resize(sources.size(), std::vector<edgeweight>(G->upperNodeIdBound()));
```
michitux, line 22: Wouldn't it be more efficient to initialize the distances with empty vectors? Or is there any advantage in sequentially allocating the memory?

networkit/distance.pyx
```diff
@@ -924,6 +924,89 @@ cdef class APSP(Algorithm):
...
+    Parameters:
+	-----------
+	G : networkit.Graph
+		The graph.
+	sources : list.
+		Set of source nodes.
+    """
+	cdef Graph _G
+
+	def __cinit__(self, Graph G, sources):
```
michitux, line 952: This should be `Graph G not None` to avoid that `None` is accepted as graph. Further, I am wondering why the second parameter is not simply `vector[node] sources`, in which case Cython would automatically cast the input and raise a `TypeError` if the wrong type is passed.

networkit/distance.pyx
```diff
@@ -924,6 +924,89 @@ cdef class APSP(Algorithm):
...
+		self._G = G
+		cdef vector[node] sourcesVec
+		try:
+			sourcesVec = <vector[node]?>sources
+		except TypeError:
+			raise RuntimeError("Error, sources must be a list of nodes.")
```
michitux, line 958: Isn't a `TypeError` as raised by Cython more appropriate here?

networkit/distance.pyx
```diff
@@ -924,6 +924,89 @@ cdef class APSP(Algorithm):
...
+		Parameters
+		----------
+		sources : list
+			List of the new source nodes.
+		"""
+		cdef vector[node] sourcesVec
+		try:
+			sourcesVec = <vector[node]?>sources
+		except TypeError:
+			raise RuntimeError("Error, sources must be a list of nodes.")
```
michitux, line 1006: The same comments as above apply here, too.

michitux (changes requested):

networkit/cpp/distance/SPSP.cpp
angriman: True, the distance vectors are already allocated in `SSSP`, allocating memory here is useless.

networkit/distance.pyx
angriman: Good idea, this way Cython takes care checking the type of `sources`. I updated the PR accordingly.

michitux (approved):

Issue #241, opened 2018-10-08
Profiling module broken


angriman:
@hmeyerhenke noticed that after recent PRs were merged the profiling module is broken. The error seems to be related to some plots that cannot be drawn as they receive NoneType objects or they encounter divisions by zero.

I never worked with the profiling module before so I find it quite difficult to debug it as there is almost no documentation.

hmeyerhenke:
Maybe it's time for refactoring (and resulting documentation) then. How should it look like in your opinion?

mjspeck:
I'm having the same issue. Anymore progress on this issue?

angriman:
We recently fixed this issue by removing the feature causing the error, the profiling module should work now. Are you using the last version of NetworKit?

mjspeck:
Looks like it. I'm on version 8.0

mjspeck:
One thing I found: building from source fixes it. But Pypi version still seems to have the division by zero error.

angriman:
That's probably due to a wrong parameter initialization in `KatzCentrality` that has been fixed in #648. Unfortunately, that fix came after the 8.0 release. With the next release profiling should work also with the Pypi version.

mjspeck:
Ah, gotcha. Thanks for the quick responses.

PR #617, opened 2020-10-28
Profiling: remove broken features

angriman:
The profiling module was broken some time ago (see #241). This PR removes (temporarily) the features that cause the crashes so that this module can be used again.
Further, it corrects the default value for the `alpha` parameter in the `KatzCentrality` python interface (which did not match the one used in the C++ implementation).

### DIFF
--- a/include/networkit/centrality/DynKatzCentrality.hpp
+++ b/include/networkit/centrality/DynKatzCentrality.hpp
@@ -1,5 +1,5 @@
 /*
- * DynKatzCentrality.h
+ * DynKatzCentrality.hpp
  *
  *  Created on: April 2018
  *      Author: Alexander van der Grinten
--- a/include/networkit/centrality/KatzCentrality.hpp
+++ b/include/networkit/centrality/KatzCentrality.hpp
@@ -34,11 +34,12 @@ public:
      * Each iteration of the algorithm requires O(m) time. The number of iterations depends on how long it takes to reach the convergence.
      *
      * @param[in] G The graph.
-     * @param[in] alpha Damping of the matrix vector product result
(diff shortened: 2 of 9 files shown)


fabratu (dismissed):

avdgrinten (changes requested):
The Katz change is not guaranteed to work. Pick 1/(deg_max+1) instead to guarantee that the Katz centrality is well-defined. This should also be documented.

angriman:
Isn't that what I am already doing here?

https://github.com/angriman/networkit/blob/1de033991bfe3a9f6fed50e2b85c2953603a40e1/networkit/profiling/profiling.py#L273

avdgrinten:
Ah, yes. But I am referring to the `__cinit__` (the C++ code should probably be changed, too).

angriman:
Ah I see. So should we check in the constructor that alpha is `<= 1 / (max_deg + 1)`? Or should we always set alpha to `1 / (max_deg + 1)`?

avdgrinten:
I guess we should set the default value to 0 or some other special value and let the constructor compute it as 1/(deg_max + 1) if the user does not override the default.

angriman:
Alright now the code uses by default `1 / (max_deg + 1)` unless the user specifies another positive value.

avdgrinten (approved):
LGTM.

angriman:
Some `GTests` with Katz centrality failed because they used the previous default value for alpha, now they are fixed.

avdgrinten (dismissed):
Still LGTM.

avdgrinten:
This PR has conflicts. They need to be resolved before it is merged.

angriman:
Thanks for the feedback, I rebased the PR.

avdgrinten (approved):
LGTM.

PR #618, opened 2020-10-30
New Algorithm: Approximate Electrical Closeness

angriman:
This PR brings the new `ApproxElectricalCloseness` algorithm based on [Approximation of the Diagonal of a Laplacian’s Pseudoinverse for Complex Network Analysis](https://drops.dagstuhl.de/opus/volltexte/2020/12872/pdf/LIPIcs-ESA-2020-6.pdf), ESA 2020. The algorithm approximates the diagonal of the laplacian's pseudoinverse (Moore–Penrose inverse) of a graph which can be used to approximate the electrical closeness of every vertex of the graph.

The correctness of the algorithm is tested in two different ways:
- In the C++ tests using the LAMG solver to compute the nearly-exact values of the diagonal, this function is implemented in `ApproxElectricalCloseness`
- In the Python tests using the `pinv` function implemented in `numpy`.

Additional changes:
- New method `fill` for `Vector`
- New cmake flag `NETWORKIT_SANITY_CHECKS` to enable further sanity checks of an algorithm.


hmeyerhenke:
Side note: this algorithm provides approximation guarantees. Thus, please use "approximates" instead of "estimates".

angriman:
Yes good point, I updated the PR description and the docs.

fabratu (changes requested):

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
```diff
@@ -0,0 +1,669 @@
...
+        // Biconnected components with 3 nodes or less can be handled trivially.
+        if (curComponent.size() == 2) {
+            sequences.push_back({curComponent[0], curComponent[1]});
+            continue;
+        }
+
+        // We take the node with highest degree in the component as source
+        node source = curComponent[0];
+        for (node u : curComponent) {
+            source = G.degree(u) > G.degree(source) ? u : source;
```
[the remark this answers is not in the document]
avdgrinten: Hm? The ternary operator vs. `if` should not make a difference.

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
```diff
@@ -0,0 +1,669 @@
...
+    auto &bcc_ = *(bcc.get());
+    bcc_.run();
+    auto components = bcc_.getComponents();
+
+    std::queue<node> queue;
+    std::vector<node> curSequence;
+
+    for (const auto &curComponent : components) {
+        // Biconnected components with 3 nodes or less can be handled trivially.
+        if (curComponent.size() == 2) {
```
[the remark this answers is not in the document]
angriman: In a previous implementation I also had a specialization for 3 vertices, which was later removed because it complicates the code without bringing better performance. I'll update the comment.

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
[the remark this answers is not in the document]
angriman: On modern compilers `if-else` should not be faster than ternary operator. The bench might be slower for some other reason (also, note that `curComponent` has 100 zeros followed by 0, ... 99, probably you did not want to do that).

angriman:
Thanks for the comments, I've addressed them.

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
```diff
@@ -0,0 +1,669 @@
...
+                status[v] = NodeStatus::VISITED;
+                queue.push(v);
+                bfsParent[v] = currentNode;
+            }
+        });
+    } while (!queue.empty());
+
+#ifndef NDEBUG
+    checkBFSTree();
+#endif
```
avdgrinten, line 252: Do not guard expensive checks behind NDEBUG. Add a separate definition to CMakeLists.txt to enable sanity checking.

avdgrinten: Something like `NETWORKIT_SANITY_CHECKS`.

angriman: Sounds like a good idea, we might want set that flag to ON in some builds in our CI.

CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ option(NETWORKIT_WARNINGS_AS_ERRORS "Treat warnings as errors (except deprecated
 option(NETWORKIT_CLANG_TIDY "Check code with clang-tidy" OFF)
 option(NETWORKIT_FLATINSTALL "Install into a flat directory structure (useful when building a Python package)" OFF)
 option(NETWORKIT_COVERAGE "Build with support for coverage" OFF)
+option(NETWORKIT_SANITY_CHECKS "Algorithms will perform further (expensive) checks" OFF)
```
avdgrinten, line 20: The flag actually needs to be exported to the compiler by setting the `-DNETWORKIT_SANITY_CHECKS` option of the compiler's command line.

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
avdgrinten: Agreed.

CMakeLists.txt
angriman: So should we use a macro for additional tests? How would that work exactly?

avdgrinten (commented):
LGTM now.

fabratu (approved):

angriman:
I think we should merge this after #625 so I can enable `NETWORKIT_SANITY_CHECKS` in the debug builds.

angriman:
I enabled `NETWORKIT_SANITY_CHECKS` in the debug build.

avdgrinten (approved):

PR #619, opened 2020-10-30
Add node2vec algorithm in new embedding module

fidus58:
hopefully fixed CI errors


networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace NetworKit {
+namespace Embedding {
+
+std::mt19937_64 &getURNG() {
+    static bool seedPerThread = true;
+    static std::once_flag once;
+    std::call_once(once, [] { Aux::Random::setSeed(1, seedPerThread); });
+
+    return Aux::Random::getURNG();
```
michitux, line 31: Why do you set a custom seed here? And why the static seed? This makes reproducible experiments or running experiments with different seeds completely impossible as it is impossible to run the algorithm twice with the same seed or an explicitly set seed.

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,87 @@
+/*
+ * AliasSampler.cpp
+ *
+ *  Created on: 03.07.2020
+ *      Author: Klaus Ahrens  <ahrens@informatik.hu-berlin.de>
+ *
+ *  adapted and reimplemented in C++17
+ *  from node2vec [https://arxiv.org/pdf/1607.00653v1.pdf]
```
michitux, line 8: What do you mean by "adapted"? Is this based on existing code?

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+    for (auto i : Graph::NeighborRange<>(graph, t)) {
+        tNbrs.insert(i);
+    }
+    // for each t-neighbor v:
+    for (auto v : Graph::NeighborRange<>(graph, t)) {
+        double pSum = 0;
+        std::vector<float> pTable; // Probability distribution table
+        // for each v-neighbor x:
+        for (auto x : Graph::NeighborRange<>(graph, v)) {
+            auto weight = graph.weight(v, x);
```
michitux, line 55: This makes the running time quadratic in the number of neighbors, please use the `forNeighborsOf` method that provides the weights in constant time.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // graphData = std::make_unique<GraphData>(nn);
+    // back to c++11 ;-(
+    graphData = std::unique_ptr<GraphData>(new GraphData(nn));
+
+    index2node.resize(nn);
+
+    // pre-allocate unordered maps:
+    for (index v = 0; v < nn; ++v) {
+        auto degv = graph.degreeOut(v);
+        auto neighbors = Graph::NeighborRange<>(graph, v);
```
michitux, line 90: Node ids are not necessarily continuous in NetworKit. If you assume this, please add a check and throw an exception otherwise.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    if (walkLen == 1) {
+        return walk;
+    }
+    if (graph.degreeOut(start) == 0) {
+        walk.resize(1); // shorten walk to 1
+        return walk;
+    }
+    auto neighbors = Graph::NeighborRange<>(graph, start);
+    auto nn = std::distance(neighbors.begin(), neighbors.end());
```
michitux, line 126: Why not just `graph.degree(start)`?

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+/// Simulates walks from every node and writes it into Walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
+    auto nn = graph.numberOfNodes();
+    auto walksDone = 0;
+
+    WalkData walkData(nn, walkLen, walksPerNode);
+
+    std::vector<node> shuffled(nn);
+    //  std::iota(shuffled.begin(), shuffled.end(), 0);
+    int n = 0;
```
michitux, line 171: This should be `node`, not `int`.

networkit/embedding.pyx
```diff
@@ -0,0 +1,75 @@
...
+        The dimension of the calculated embedding. 
+    """
+
+    cdef Graph _G
+ 
+    def __cinit__(self, Graph G, P=1, Q=1, L=80, N=10, D=128):
+        self._G = G
+        self._this = new _Node2Vec(G._this, P, Q, L, N, D)
+
+    def run(self):
```
michitux, line 61: This is not necessary and actually makes things worse as the implementation in the parent class drops the GIL, allowing other Python code to run in parallel.

networkit/embedding.pyx
```diff
@@ -0,0 +1,75 @@
...
+from .base cimport _Algorithm, Algorithm
+from .graph cimport _Graph, Graph
+
+ctypedef uint64_t count
+
+cdef extern from "<networkit/embedding/Node2Vec.hpp>":
+
+    cdef cppclass _Node2Vec "NetworKit::Node2Vec"(_Algorithm):
+        _Node2Vec(_Graph G, double P, double Q, count L, count N, count D) except +
+        void run() nogil except +
```
michitux, line 17: The `run`-method is not necessary as it is already in the parent class.

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include "AliasSampler.hpp"
+
+namespace NetworKit {
+namespace Embedding {
+
+std::mt19937_64 &getURNG() {
+    static bool seedPerThread = true;
+    static std::once_flag once;
+    std::call_once(once, [] { Aux::Random::setSeed(1, seedPerThread); });
```
michitux, line 29: It is quite unexpected for a user that calling the embedding algorithm changes the global seed. If for some reason you require a custom seed, please do not use the global URNG of NetworKit but use a custom one.

michitux (changes requested):
I have left some comments on very obvious problems with this code. Some further comments:

* Why is most of the code not object oriented? Is there any reason why the different parts of the algorithm are not also implemented as subclasses of `Algorithm`?
* The dependence on continuous node ids seems quite bad and should be easy to fix.
* It would be great if the different parts of the algorithm that are parallelized and use randomness could be modified to be deterministic with respect to the parallel scheduling. Currently, the results - as far as I understand - depend on the dynamic scheduling.

I have not checked or tested the algorithmic correctness.

As you state this is adapted from node2vec, is this based on their actual implementation? If you have copied code from there, under which license is that code?

networkit/cpp/embedding/AliasSampler.cpp
avdgrinten: Klaus did re-implement the algorithm (without copying the code literally, as far as I understand). However, since the structure still matches the original code, it is probably a good idea to acknowledge the copyright. Luckily, this is not an issue since the original code is also under MIT.

networkit/cpp/embedding/AliasSampler.cpp
avdgrinten: Agreed, library functions should not change the global seed.

networkit/cpp/embedding/BiasedRandomWalk.cpp
angriman: To get the weights in constant time It is also possible to use `NeighborWeightRange`.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // back to c++11 ;-(
+    graphData = std::unique_ptr<GraphData>(new GraphData(nn));
+
+    index2node.resize(nn);
+
+    // pre-allocate unordered maps:
+    for (index v = 0; v < nn; ++v) {
+        auto degv = graph.degreeOut(v);
+        auto neighbors = Graph::NeighborRange<>(graph, v);
+        for (auto n : neighbors) {
```
angriman, line 92: This can be simplified with `for (node n : G->neighborRange(v))`
and the `neighbors` variable is not needed.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,200 @@
...
+            graphData->data[v][n] = AliasSampler(degv);
+        }
+    }
+
+    count nCnt = 0;
+
+#pragma omp parallel for schedule(dynamic)
+    // is forNodes parallelized for omp ?
+    //  for (node t = 0; t < nn; ++t) {
+    for (omp_index t = 0; t < static_cast<omp_index>(nn); ++t) {
```
angriman, line 104: The node ids of a graph are not guaranteed to be continue, use `G->balancedParallelForNodes([&](node u) { /* ... */ });` ?

networkit/cpp/embedding/LearnEmbedding.hpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+/// Learns embeddings using SGD, Skip-gram with negative sampling.
+Embeddings learnEmbeddings(AllWalks &walks, count dimensions, count winSize, count iterations);
+
+/*
+///print embeddings
+void printEmbeddings(std::ofstream& o, Embeddings& e);
+*/
+
```
angriman, line 60: Remove this code if unused.

angriman (changes requested):

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+    GraphData(size_t nn) : data{nn, NeighborMap()} {}
+};
+
+std::unique_ptr<GraphData> graphData;
+
+void preprocessNode(const Graph &graph, node t, double paramP, double paramQ) {
+
+    // for node t
+    NeighborSet tNbrs; // Neighbors of t
+    for (auto i : Graph::NeighborRange<>(graph, t)) {
```
angriman, line 49: I think that `for (auto/node i : graph->neighborRange(t))` is cleaner.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+        tNbrs.insert(i);
+    }
+    // for each t-neighbor v:
+    for (auto v : Graph::NeighborRange<>(graph, t)) {
```
angriman, line 53: As before.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+            graphData->data[v][n] = AliasSampler(degv);
+        }
+    }
+
+    graph.balancedParallelForNodes([&](node t) {
+        preprocessNode(graph, t, paramP, paramQ);
+    });
+}
+
+node nthNeighbor(Graph::NeighborRange<> range, index nth);
```
angriman, line 104: This method does not seem to be used anywhere. If so please remove it.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (walkLen == 1) {
+        return walk;
+    }
+    if (graph.degreeOut(start) == 0) {
+        walk.resize(1); // shorten walk to 1
+        return walk;
+    }
+    auto nn = graph.degree(start);
+
+    // node randomNeighbor = nthNeighbor(neighbors, nn * uniform_real());
```
angriman, line 122: Remove this line.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+    walk[nr++] = randomNeighbor;
+    node dst = randomNeighbor;
+
+    while (nr < walkLen) {
+        if (graph.degreeOut(dst) == 0) {
+            walk.resize(nr); // shorten walk to nr
+            return walk;
+        }
+        // auto neighbors = Graph::NeighborRange<>(graph, dst);
+        // auto nn = std::distance(neighbors.begin(), neighbors.end());
```
angriman, line 134: Remove these lines.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+        NeighborMap &map = graphData->data[dst];
+        AliasSampler &as = map[src];
+        // node next = nthNeighbor(neighbors, as.sample());
```
angriman, line 137: Remove this line.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+/// Simulates walks from every node and writes it into Walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
+    auto nn = graph.numberOfNodes();
+    auto walksDone = 0;
+
+    WalkData walkData(nn, walkLen, walksPerNode);
+
+    std::vector<node> shuffled(nn);
+    //  std::iota(shuffled.begin(), shuffled.end(), 0);
```
angriman, line 164: Remove this line.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+    count walksPerNode;
+
+    WalkData(count nn, count wl, count wpn)
+        : data{nn * wpn, Walk(wl)}, walkLength(wl), walksPerNode(wpn) {}
+};
+
+/// Simulates walks from every node and writes it into Walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
+    auto nn = graph.numberOfNodes();
+    auto walksDone = 0;
```
angriman, line 159: Instead of `auto`, would `count` be better?

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    std::vector<node> shuffled(nn);
+    //  std::iota(shuffled.begin(), shuffled.end(), 0);
+    node n = 0;
+    std::generate(shuffled.begin(), shuffled.end(), [&n]() { return n++; });
+
+    for (index c = 0; c < walksPerNode; ++c) {
+        std::shuffle(shuffled.begin(), shuffled.end(), Aux::Random::getURNG());
+#pragma omp parallel for schedule(dynamic)
+        // for (index i = 0; i < nn; ++i) { OMP needs signed index var
```
angriman, line 171: Remove this line.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    return walkData.data;
+}
+
+node nthNeighbor(Graph::NeighborRange<> range, index nth) {
+    auto it = range.begin();
+    std::advance(it, nth);
+    return *it;
+}
+
```
angriman, line 192: Remove this function if unused.

networkit/cpp/embedding/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_CXX_STANDARD 17)
```
angriman, line 1: I think we should not do that, NetworKit supports C++14 features.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+namespace Embedding {
+
+// Number of negative samples. Value taken from original word2vec code.
+const int negSamN = 5;
+
+const int maxExp = 6;
+
+// Learning rate for SGD. Value taken from original word2vec code.
+const double startAlpha = 0.025;
+const double minAlpha = startAlpha * 0.0001;
```
angriman, line 41: Use `constexpr`?

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+    Embeddings synPos = initPosEmb(vocab, dimensions);
+    AliasSampler as = vocabSampler(vocab);
+
+    double alpha = startAlpha; // learning rate
+    count wordCntAll = 0;
+
+    ModelData model(allWalks, dimensions, winSize, iterations, as, wordCntAll, alpha, synNeg,
+                    synPos);
+
+    //#pragma omp parallel for schedule(dynamic) collapse(2)
```
angriman, line 203: Remove this line.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+    for (index iterCnt = 0; iterCnt < iterations; ++iterCnt) {
+#pragma omp parallel for schedule(dynamic)
+        //      for (index run = 0; run < walks; ++run) {
```
angriman, line 206: Remove this line.

angriman (changes requested):

fidus58:
> Am 18.01.2021 um 10:25 schrieb Eugenio Angriman <notifications@github.com>:
> 
> 
> @angriman requested changes on this pull request.
> 
> In networkit/cpp/embedding/BiasedRandomWalk.cpp:
> 
> > +    //graphData = std::make_unique<GraphData>(nn);
> +    // back to c++11 ;-(
> +    graphData = std::unique_ptr<GraphData>(new GraphData(nn));
> +    // clang_tidy faults it but make_unique does not compile :-(
> 
> make_unique should work now, NetworKit now supports C++14 features.

i’m not sure if and how to update my branch to the current state of networkit - i thought i should NOT merge ?

...

> In networkit/cpp/embedding/AliasSampler.cpp:
> 
> > +#include <mutex>
> +#include <random>
> +#include <vector>
> +
> +#include <networkit/Globals.hpp>
> +#include <networkit/auxiliary/Random.hpp>
> +
> +#include "AliasSampler.hpp"
> +
> +namespace NetworKit {
> +namespace Embedding {
> +
> +double uniform_real() {
> +    static std::uniform_real_distribution<> dist;
> +
> +    return dist(Aux::Random::getURNG());
> 
> Instead of creating a new urng each time a random real is needed, would that be better to keep a urng as member variable?

but it’s a std::mt19937_64&, not a new one ?!

all the other remarks are clear and will be carried out …

thank you for your sharp view :-)

int main(){[](){[]{R":( 
)                     (
) mfg                 (
)      k ahrens       (
)                     (
):";};};}//{}()niam tni

angriman:
Ah sorry my bad, I thought that `getURNG` returned a copy of `std::mt19937_64`.

About updating your branch, you can do that by rebasing on NetworKit master branch:
```
git remote add networkit https://github.com/networkit/networkit
git pull --rebase networkit master
```

fidus58:
> Am 18.01.2021 um 11:03 schrieb Eugenio Angriman <notifications@github.com>:
> 
> 
> Ah sorry my bad, I thought that getURNG returned a copy of std::mt19937_64.
> 
> About updating your branch, you can do that by rebasing on NetworKit master branch:
> 
> git remote add networkit https://github.com/networkit/networkit
> git pull --rebase networkit master

after that i still get errors on all make_unique’s !?

[30/141] Building CXX object CMakeFile...t/cpp/embedding/BiasedRandomWalk.cpp.o
FAILED: CMakeFiles/networkit.dir/networkit/cpp/embedding/BiasedRandomWalk.cpp.o 
/usr/bin/g++  -Dnetworkit_EXPORTS -I../../include -I../../extlibs/ttmath -I../../extlibs/tlx -O3 -DNDEBUG -fPIC    -DNETWORKIT_RELEASE_LOGGING -fopenmp -std=gnu++11 -MD -MT CMakeFiles/networkit.dir/networkit/cpp/embedding/BiasedRandomWalk.cpp.o -MF CMakeFiles/networkit.dir/networkit/cpp/embedding/BiasedRandomWalk.cpp.o.d -o CMakeFiles/networkit.dir/networkit/cpp/embedding/BiasedRandomWalk.cpp.o -c ../../networkit/cpp/embedding/BiasedRandomWalk.cpp
../../networkit/cpp/embedding/BiasedRandomWalk.cpp: In function ‘void NetworKit::Embedding::preprocessTransitionProbs(const NetworKit::Graph&, double, double)’:
../../networkit/cpp/embedding/BiasedRandomWalk.cpp:82:22: error: ‘make_unique’ is not a member of ‘std’
     graphData = std::make_unique<GraphData>(nn);
                      ^~~~~~~~~~~

any idea?

 
int main(){[](){[]{R":( 
)                     (
) mfg                 (
)      k ahrens       (
)                     (
):";};};}//{}()niam tni

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,173 @@
...
+    WalkData walkData(nn, walkLen, walksPerNode);
+
+    std::vector<node> shuffled(nn);
+    node n = 0;
+    std::generate(shuffled.begin(), shuffled.end(), [&n]() { return n++; });
+
+    for (index c = 0; c < walksPerNode; ++c) {
+        std::shuffle(shuffled.begin(), shuffled.end(), Aux::Random::getURNG());
+#pragma omp parallel for schedule(dynamic)
+        for (omp_index i = 0; i < static_cast<omp_index>(nn); ++i) {
```
angriman, line 159: Use `graph->balancedParallelForNodes`, if node ids are not continuous this does not work.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,173 @@
...
+    count walkLength;
+    count walksPerNode;
+
+    WalkData(count nn, count wl, count wpn)
+        : data{nn * wpn, Walk(wl)}, walkLength(wl), walksPerNode(wpn) {}
+};
+
+/// Simulates walks from every node and writes it into Walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
+    auto nn = graph.numberOfNodes();
```
angriman, line 147: Use `upperNodeIdBound`, otherwise this function does not work with non-continuous node ids.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,219 @@
...
+    count dimensions = model.dimensions;
+    count winSize = model.winSize;
+    double alpha = model.alpha;
+    Embeddings &synPos = model.synPos;
+    Embeddings &synNeg = model.synNeg;
+
+    Walk &thisWalk = allWalks[walkNr];
+
+    for (index wordI = 0; wordI < thisWalk.size(); ++wordI) {
+        if (wordCntAll % 10000 == 0) {
```
angriman, line 104: I am not very familiar with the algorithm but this 1000 seems a bit strange. Why this number? Can we store it into a global variable and document it?

angriman (changes requested):

fidus58:
> Am 19.01.2021 um 09:58 schrieb Eugenio Angriman <notifications@github.com>:
> 
> 
> @angriman requested changes on this pull request.
> 
> In networkit/cpp/embedding/BiasedRandomWalk.cpp:
> 
> > +#pragma omp parallel for schedule(dynamic)
> +        for (omp_index i = 0; i < static_cast<omp_index>(nn); ++i) {
> 
> Use graph->balancedParallelForNodes, if node ids are not continuous this does not work.
> 
> In networkit/cpp/embedding/BiasedRandomWalk.cpp:
> 
> > +    return walk;
> +}
> +
> +struct WalkData {
> +    // assuming contiguous node numbers 0..n-1, they serve as indices
> +    AllWalks data;
> +    count walkLength;
> +    count walksPerNode;
> +
> +    WalkData(count nn, count wl, count wpn)
> +        : data{nn * wpn, Walk(wl)}, walkLength(wl), walksPerNode(wpn) {}
> +};
> +
> +/// Simulates walks from every node and writes it into Walks vector
> +AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
> +    auto nn = graph.numberOfNodes();
> 
> Use upperNodeIdBound, otherwise this function does not work with non-continuous node ids.

will be fixed …

> 
> In networkit/cpp/embedding/LearnEmbedding.cpp:
> 
> > +void trainModel(ModelData &model, count walkNr) {
> +    AllWalks &allWalks = model.allWalks;
> +    count &wordCntAll = model.wordCntAll;
> +    AliasSampler &as = model.as;
> +    count allWords = model.allWords;
> +    count iterations = model.iterations;
> +    count dimensions = model.dimensions;
> +    count winSize = model.winSize;
> +    double alpha = model.alpha;
> +    Embeddings &synPos = model.synPos;
> +    Embeddings &synNeg = model.synNeg;
> +
> +    Walk &thisWalk = allWalks[walkNr];
> +
> +    for (index wordI = 0; wordI < thisWalk.size(); ++wordI) {
> +        if (wordCntAll % 10000 == 0) {
> 
> I am not very familiar with the algorithm but this 1000 seems a bit strange. Why this number? Can we store it into a global variable and document it?
> 

this was indeed a magic number in the original snap code - both for progress report (dropped now) AND for parameter (alpha) influence reduction.
i did not do any experiments on it, either on the other ones (being constexpr now). 

int main(){[](){[]{R":( 
)                     (
) mfg                 (
)      k ahrens       (
)                     (
):";};};}//{}()niam tni

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,172 @@
...
+    count walksDone = 0;
+
+    WalkData walkData(nn, walkLen, walksPerNode);
+
+    std::vector<node> shuffled(nn);
+    node n = 0;
+    std::generate(shuffled.begin(), shuffled.end(), [&n]() { return n++; });
```
angriman, line 154: To generate a vector with all the nodes of a graph it's better to use `std::vector<node> shuffled(graph.nodeRange().begin(), graph.nodeRange().end());`.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return vocab;
+}
+
+// Precompute unigram table using alias sampling method
+AliasSampler vocabSampler(Vocab &vocab) {
+    double trainWordsPow = 0;
+    auto sz = vocab.size();
+    std::vector<float> probV(sz); // NOT ... {sz}
+    for (count i = 0; i < sz; ++i) {
+        double e = std::pow(vocab[i], 0.75);
```
angriman, line 62: Looks like `0.75` is another magic number.

networkit/cpp/embedding/Node2Vec.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+bool Node2Vec::isParallel() const {
+    return true;
+}
+
+std::vector<std::vector<float>> Node2Vec::getFeatures() {
+    assureFinished();
+    return std::move(features);
+}
```
angriman, line 76: What about returning a const ref?

angriman (changes requested):

fidus58:
…. soon done.

> 
> In networkit/cpp/embedding/Node2Vec.cpp:
> 
> > +std::vector<std::vector<float>> Node2Vec::getFeatures() {
> +    assureFinished();
> +    return std::move(features);
> +}
> 
> What about returning a const ref?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

that’s a reasonable alternative, i was not sure about the lifetime of the vector, with move it’s clearly on the python side
(even without move maybe?)

if you’d prefer const& i’ll change it.

int main(){[](){[]{R":( 
)                     (
) mfg                 (
)      k ahrens       (
)                     (
):";};};}//{}()niam tni

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// Initialize positive embeddings
+Embeddings initPosEmb(Vocab &vocab, int dimensions) {
+    Embeddings emb{vocab.size(), std::vector<float>(dimensions)};
+
+    for (index i = 0; i < emb.size(); ++i) {
+        std::generate(emb[i].begin(), emb[i].end(),
+                      [&]() { return (uniform_real() - 0.5) / dimensions; });
+    }
```
angriman, line 87: It looks like that each iteration of this loop modifies only the i-th element. Can this loop be parallelized? Of course, we need to make sure that `uniform_real()` is thread-safe, one idea to do this is to store a random generator per thread.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,169 @@
...
+};
+
+/// Simulates walks from every node and writes it into Walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count walksPerNode) {
+    auto nn = graph.numberOfNodes();
+    count walksDone = 0;
+
+    WalkData walkData(nn, walkLen, walksPerNode);
+    auto nodes = graph.nodeRange();
+    std::vector<node> shuffled(nodes.begin(), nodes.end());
```
angriman, line 151: If `nodes` is only used to generate the vector, wouldn't
```
std::vector<node> shuffled(graph.nodeRange().begin(), graph.nodeRange().end());
```
be better?

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,221 @@
...
+// Number of negative samples. Value taken from original word2vec code.
+constexpr int negSamN = 5;
+
+constexpr int maxExp = 6;
+
+// Learning parameters for SGD. Values taken from original word2vec code.
+constexpr double startAlpha = 0.025;
+constexpr int refreshAlphaCount = 10000;
+constexpr double minAlpha = startAlpha * 0.0001;
+constexpr double Pwr = 0.75;
```
angriman, line 43: Use camelCase for variables.

angriman: Also, maybe explain how this variable is used/affects the algorithm, the name does not say much about it.

angriman (changes requested):

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+struct GraphData {
+    // assuming contiguous node numbers 0..n-1, they serve as indices
+    using Data = std::vector<NeighborMap>;
+    Data data;
+
+    GraphData(size_t nn) : data{nn, NeighborMap()} {}
+};
+
+std::unique_ptr<GraphData> graphData;
```
angriman, line 43: Avoid to define global variables, store `graphData` as a member variable where needed.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,168 @@
...
+        });
+        // Normalizing table
+        float pfSum = (float)pSum;
+        std::for_each(pTable.begin(), pTable.end(), [pfSum](float &p) { p /= pfSum; });
+
+        graphData->data[v][t].unigram(pTable);
+    }
+}
+
+std::vector<std::vector<node>> index2node;
```
angriman, line 77: As before, avoid global variables.

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,82 @@
...
+ *  part of snap [https://github.com/snap-stanford/snap]
+ *  Copyright (c) 2007-2019, Jure Leskovec (under BSD license)
+ *
+ *  see [https://arxiv.org/pdf/1607.00653v1.pdf]
+ *
+ */
+
+// networkit-format
+
+#include <mutex>
```
angriman, line 17: Is this include needed?

networkit/cpp/embedding/BiasedRandomWalk.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *  see [https://arxiv.org/pdf/1607.00653v1.pdf]
+ *
+ */
+
+// networkit-format
+
+#ifndef BIASEDRANDOMWALK_HPP
+#define BIASEDRANDOMWALK_HPP
+
+#include <iostream>
```
angriman, line 20: Is this include needed?

networkit/cpp/embedding/BiasedRandomWalk.hpp
```diff
@@ -0,0 +1,41 @@
...
+using Walk = std::vector<node>;     // one walk
+using AllWalks = std::vector<Walk>; // n walks for all nodes: 2 dimensions in one big chunk
+
+/// preprocesses transition probabilities for random walks. Has to be called once before doWalks
+/// calls
+void preprocessTransitionProbs(const Graph &graph, double paramP, double paramQ);
+
+/// Simulates walks from every node and writes it into walks vector
+AllWalks doWalks(const Graph &graph, count walkLen, count numberOfWalks);
+
```
angriman, line 37: A way to avoid defining global variables is to create `BiasedRandomWalk` class here.

angriman (changes requested):

include/networkit/embedding/Node2Vec.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     This method computes all node embeddings.
+    */
+    void run() override;
+
+    /**
+     * Returns a string with the algorithm's name and its parameters, if there are any.
+     * @return The string representation of the algorithm.
+     */
+    std::string toString() const override;
```
angriman, line 57: `Algorithm::toString()` has been deprecated, this can be removed.

include/networkit/embedding/Node2Vec.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @return True if algorithm can run multi-threaded.
+     */
+    bool isParallel() const override;
```
angriman, line 62: `Algorithm::isParallel()` will be deprecated, just state in the documentation whether or not this algorithm is parallel.

angriman (changes requested):

networkit/cpp/embedding/BiasedRandomWalk.hpp
```diff
@@ -0,0 +1,67 @@
...
+    /// calls
+    void preprocessTransitionProbs(double paramP, double paramQ);
+
+    /// Simulates walks from every node and writes it into walks vector
+    AllWalks doWalks(count walkLen, count numberOfWalks);
+
+    BiasedRandomWalk(const Graph &graph);
+
+private:
+    const Graph &graph;
```
angriman, line 45: Probably it is better to keep a pointer to `graph` (as already done in `Node2Vec`).

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace NetworKit {
+namespace Embedding {
+
+double uniform_real() {
+    static std::uniform_real_distribution<> dist;
+
+    return dist(Aux::Random::getURNG());
+}
+
+void AliasSampler::unigram(std::vector<float> &probs) {
```
angriman, line 34: `probs` here is never modified, use `const`?

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,81 @@
...
+    while (!underV.empty()) {
+        auto curr = underV.back();
+        underV.pop_back();
+        U[curr] = 1;
+    }
+    while (!overV.empty()) {
+        auto curr = overV.back();
+        overV.pop_back();
+        U[curr] = 1;
+    }
```
angriman, line 69: These two loops can be simplified with
```
for (index i : underV/overV)
    U[i] = 1;
```
The two vectors will go out of scope anyway so there is no need to `pop_back` elements right?.

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+index AliasSampler::sample() {
+    double rx = uniform_real();
+    double ry = uniform_real();
+    index x = static_cast<index>(rx * K.size());
+    double y = ry;
+    return y < U[x] ? x : K[x];
```
angriman, line 77: Why do we need `rx` and `ry` here? Wouldn't something like
```
index x = static_cast<index>(uniform_real() * K.size());
return uniform_real() < U[x] ? x : K[x];
```
be simpler?

angriman (changes requested):

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "AliasSampler.hpp"
+
+namespace NetworKit {
+namespace Embedding {
+
+double uniform_real() {
+    static std::uniform_real_distribution<> dist;
+
+    return dist(Aux::Random::getURNG());
+}
```
michitux, line 32: What is the advantage of this function over `Aux::Random::real()`?

angriman: As reported in the docs [1], if that code is be used in "hot" sections, it might be better to keep a local `uniform_real_distribution` and avoid to use `Aux::Random::real()`.
[1] https://networkit.github.io/dev-docs/cpp_api/function_namespaceAux_1_1Random_1ab0532015f7ed7d0eddffeb10405720f9.html#exhale-function-namespaceaux-1-1random-1ab0532015f7ed7d0eddffeb10405720f9

michitux: Even though it is not completely clear, I assume that the slowdown mentioned in the documentation is compared to a) holding a local (!) reference to the URNG and b) a local distribution object. I assume that already calling `getURNG()` introduces most of the overhead. If this overhead shall be avoided, I would suggest adding a reference to the URNG and the distribution object in the `AliasSampler` class. Note that neither the URNG nor the `uniform_real_distribution` are thread-safe, but again this implementation is already not thread-safe, so I assume that this is only called from a single thread anyways?

avdgrinten: The implementation should be made thread-safe by removing global state.

avdgrinten: Taking a reference to getURNG() makes sense to avoid the overhead of the thread local access (but it should be stored in a member variable). The distribution object should also be re-used and not be recreated.

angriman (changes requested):
Another thing that is missing here is a new Jupyter Notebook that shows how to use this new feature.

avdgrinten:
Let's first merge this PR to avoid further stalling it, a notebook can be a added later.

avdgrinten:
What does still need to be done before merging the PR?

networkit/cpp/embedding/AliasSampler.hpp
```diff
@@ -0,0 +1,45 @@
...
+#define ALIASSAMPLER_HPP
+
+#include <vector>
+
+#include <networkit/Globals.hpp>
+
+namespace NetworKit {
+namespace Embedding {
+
+double uniform_real();
```
angriman, line 27: We probably don't need this anymore.

angriman:
I did not notice that `uniform_real` was discarded in favor of `Aux::Random::real`. Apart from the last one I left, I don't have further comments for this PR.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,154 @@
...
+            } else {
+                pTable.push_back(weight / paramQ);
+                pSum += weight / paramQ;
+            }
+        });
+        // Normalizing table
+        float pfSum = (float)pSum;
+        std::for_each(pTable.begin(), pTable.end(), [pfSum](float &p) { p /= pfSum; });
+
+        graphData->data[v][t].unigram(pTable);
```
michitux, line 75: This looks dangerous to me, as `graphData->data[v]` is an `std::unordered_map`, and this code is called in parallel loop over `t`. Or, in other words, why cannot two threads write to the unordered map (which is not thread-safe) of the same `v`?

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,154 @@
...
+    WalkData walkData(nn, walkLen, walksPerNode);
+    std::vector<node> shuffled(graph->nodeRange().begin(), graph->nodeRange().end());
+
+    for (index c = 0; c < walksPerNode; ++c) {
+        std::shuffle(shuffled.begin(), shuffled.end(), Aux::Random::getURNG());
+        graph->balancedParallelForNodes([&](node i) {
+            auto v = shuffled[i];
+            Walk thisWalk = oneWalk(v, walkLen);
+            walkData.data[c * nn + i] = std::move(thisWalk);
+            ++walksDone;
```
michitux, line 144: This must be an atomic increment or some increments to `walksDone` will be lost. However, I also cannot see where `walksDone` is used - can we simply remove this?

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,229 @@
...
+                }
+
+                for (index i = 0; i < dimensions; ++i) {
+                    eV[i] += grad * synNeg[target][i];
+                    synNeg[target][i] += grad * synPos[currWord][i];
+                }
+            }
+
+            for (index i = 0; i < dimensions; ++i) {
+                synPos[currWord][i] += eV[i];
```
michitux, line 172: Are we guaranteed that this walk is the only one writing at this position? If not this must be an atomic increment.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,229 @@
...
+                } else if (product < -maxExp) {
+                    grad = label * alpha;
+                } else {
+                    double exp = std::exp(product);
+                    grad = (label - 1 + 1 / (1 + exp)) * alpha;
+                }
+
+                for (index i = 0; i < dimensions; ++i) {
+                    eV[i] += grad * synNeg[target][i];
+                    synNeg[target][i] += grad * synPos[currWord][i];
```
michitux, line 167: Are we guaranteed that this walk is the only one writing at this position? If not this must be an atomic increment.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,229 @@
...
+        ++wordCntAll;
+    }
+}
+
+Embeddings learnEmbeddings(AllWalks &allWalks, count dimensions, count winSize, count iterations) {
+    count walks = allWalks.size();           // number of walks
+    count nodesPerWalk = allWalks[0].size(); // number of nodes per walk
+
+    std::unordered_map<node, node> renumbered;
+    std::unordered_map<node, node> original;
```
michitux, line 184: It seems to me that at least `original` could be a simple `std::vector` as it is indexed with continuous integers. Is that correct?

michitux (changes requested):
I quickly scanned this code and found several places where it seems to me that functions that are not thread-safe are called in parallel code. In some cases it seems a simple atomic update could solve this, in others I assume that even inserts to the same `std::unordered_map` could occur from several threads which would require locks for synchronization. I cannot guarantee that I have found all such places. Please carefully check again all parts that are parallelized. If these parallellizations did not exist in the original implementation I would also wonder if they affect the quality of the result in some places. If this is not clear, I think a flag to disable all parallelizations that could affect the quality would be good (I assume some parts of the parallelization do not affect the quality, so just executing the algorithm with a single thread might not be the solution).

networkit/cpp/embedding/LearnEmbedding.cpp
fidus58: Do you think, it's worth the asymmetry?

networkit/cpp/embedding/LearnEmbedding.cpp
fidus58: These problems where indeed part of the original code. My interpretation was, that it adds another level of indeterminism into an already indeterministic algorithm? Let me know, if you'd prefer atomic loads.

networkit/cpp/embedding/BiasedRandomWalk.cpp
fidus58: You are right, it's a relict of a former progress indicator. --> removed

networkit/cpp/embedding/BiasedRandomWalk.cpp
fidus58: Each `graphData->data[v] ` keeps a map of nextStep-probabilities depending on the previous step (coming from t). So while processing these in parallel over all nodes, every slot is filled exactly once. These maps are preallocated in the number of neighbors, so that the memory behind the maps does not change - only the values (AliasSampler) are changed from 0 to the calculated probs. That's why I think, there is no threading issue.

michitux: Ah, I see. Maybe this could be made clearer by explicitly using a method (like `find()`), that does not potentially trigger a modification of the `unordered_map`?

networkit/cpp/embedding/LearnEmbedding.cpp
michitux: I was not so much concerned about the load, more about the write. From what I understand updates to `synNeg[target][i]` could simply be lost if two threads write at the same time.

fidus58: sorry, i meant stores.

networkit/cpp/embedding/LearnEmbedding.cpp
michitux: Could we maybe replace both by simple `std::vector`s? If both are just indexed by node ids and have (almost) all nodes in the walks, this should be much faster. We could simply store `none` for nodes that have not been encountered yet.

networkit/embedding.pyx
```diff
@@ -0,0 +1,70 @@
...
+        P > max(Q,1) : less likely to sample an already-visited node in
+                       the following two steps
+        P < min(Q,1) : more likely to sample an already-visited node in
+                       the following two steps
+	Q : double
+        The ratio for the direction of the next step
+        Q > 1 : the random walk is biased towards nodes close to the
+                previous one.
+        Q < 1 : the random walk is biased towards nodes which are 
+                further away from the previous one.
```
angriman, line 45: The indentation here looks wrong

networkit/cpp/embedding/Node2Vec.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+Node2Vec::Node2Vec(const Graph &G, double P, double Q, count L, count N, count D)
+    : G(&G), P(P), Q(Q), L(L), N(N), D(D) {
+    if (G.numberOfNodes() != G.upperNodeIdBound()) {
+        throw std::runtime_error("The node ids of the graph must be continuous.");
+    }
+}
+
+void Node2Vec::run() {
+    using namespace NetworKit::Embedding;
```
angriman, line 42: This is already in the NetworKit namespace, would `using namespace Embedding;` be enough?

networkit/cpp/embedding/LearnEmbedding.cpp
michitux: I have not looked into this deeply enough to understand what consequences one or the way of doing this has. When I looked at it, it just seemed wrong to possibly loose some of these updates and atomic updates (implemented as compared-and-swap-loop, but you can simply use OpenMP atomic update) seems like a relatively easy solution. If it is okay to loose some of these updates (potentially also only to some dimensions), I think this should be clearly documented such that when any code analysis tool detects this, we know why it is implemented that way.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    Embeddings synNeg = initNegEmb(vocab, dimensions);
+    Embeddings synPos = initPosEmb(vocab, dimensions);
+    AliasSampler as = vocabSampler(vocab);
+
+    double alpha = startAlpha; // learning rate
+    count wordCntAll = 0;
+
+    ModelData model(allWalks, dimensions, winSize, iterations, as, wordCntAll, alpha, synNeg,
+                    synPos);
```
michitux, line 238: `wordCntAll` seems to be unused afterwards, why is this passed to the model as reference? Same for `alpha`.

networkit/cpp/embedding/BiasedRandomWalk.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (walkLen == 1) {
+        return walk;
+    }
+    if (graph->degreeOut(start) == 0) {
+        walk.resize(1); // shorten walk to 1
+        return walk;
+    }
+    auto nn = graph->degree(start);
+
+    node randomNeighbor = index2node[start][nn * Aux::Random::real()];
```
michitux, line 100: Why is this not using `Aux::Random::index(nn)`?

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,256 @@
...
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+                alpha = minAlpha;
+            }
+        }
+
+        node word = thisWalk[wordI];
+        std::vector<float> eV(dimensions);
+        index offset = Aux::Random::real() * winSize;
```
michitux, line 140: Why is this not using `Aux::Random::index(winSize)`?

networkit/cpp/embedding/AliasSampler.cpp
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+    for (index i : underV)
+        U[i] = 1;
+    for (index i : overV)
+        U[i] = 1;
+}
+
+index AliasSampler::sample() {
+    index x = static_cast<index>(Aux::Random::real() * K.size());
```
michitux, line 61: Why is this not using `Aux::Random::index(K.size())`?

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,256 @@
...
+        if (wordCntAll % refreshAlphaCount == 0) {
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+            alpha = startAlpha * (1 - wordCntAll / ((double)(iterations * allWords) + 1.0));
+            if (alpha < minAlpha) {
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+                alpha = minAlpha;
```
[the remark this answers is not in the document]
michitux: If you change this to a reference, additional care needs to be taken to avoid that other threads read a value of `alpha` below `minAlpha`. For this, the check for `minAlpha` needs to be before assigning the new value to `alpha` (i.e., first store the value in a local temporary, compare, and then assign either the temporary or `minAlpha`). I am not sure though if this needs to be a model parameter at all as it seems to be recomputed from scratch anyway and is not used afterwards.

michitux (changes requested):
I think I found some more (some new) issues with the parallelization.

networkit/cpp/embedding/BiasedRandomWalk.cpp
fidus58: I did not know this one but will adapt to it. Looking into its implementation I wonder why it uses a new `dist` for every call, whereas real has one `thread_local static` for all?

networkit/cpp/embedding/LearnEmbedding.cpp
fidus58: `alpha` is reduced along the progress of `wordCntAll`.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,256 @@
...
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+                synPos[currWord][i] = synPos[currWord][i] + eV[i];
+            }
+        }
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+        wordCntAll = wordCntAll + 1;
```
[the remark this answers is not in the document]
fidus58: compiler complains:

../../networkit/cpp/embedding/LearnEmbedding.cpp:201:20: error: expected ‘=’ before ‘+=’ token
         wordCntAll += 1;
                    ^~
same for          ++wordCntAll;

'#pragma amp atomic update' seems to work here ...

networkit/cpp/embedding/LearnEmbedding.cpp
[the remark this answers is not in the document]
fidus58: Yes, it should be a reference. The original code from snap does no protection against parallel accesses at all - so did I.
So I have the choice to take all these effects as additional randomness OR implement a version, which is not the 'same' as the original!? In discussion with Alexander van der Grinten we concluded, that it should be sufficient to prevent non-atomic writes.

networkit/cpp/embedding/LearnEmbedding.cpp
```diff
@@ -0,0 +1,256 @@
...
+#endif // NETWORKIT_OMP2
+                    synNeg[target][i] = synNeg[target][i] + grad * synPos[currWord][i];
+                }
+            }
+
+            for (index i = 0; i < dimensions; ++i) {
+#ifndef NETWORKIT_OMP2
+#pragma omp atomic write
+#endif // NETWORKIT_OMP2
+                synPos[currWord][i] = synPos[currWord][i] + eV[i];
```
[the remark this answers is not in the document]
avdgrinten: I think the algorithm makes no attempt to be deterministic; that's why an atomic write is enough. AFAICS the algorithm intentionally accepts the fact that values might be overwritten and iterates until a convergence criterion is met.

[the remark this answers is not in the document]
avdgrinten: Basically, this is a specialized version "asynchronous" SGD, see https://www.cs.toronto.edu/~ranzato/publications/DistBeliefNIPS2012_withAppendix.pdf

avdgrinten (approved):

avdgrinten:
I am merging this PR now. I think the code quality was improved substantially due to the reviews. There might be various small things that can be improved but I think this PR is ready for integration now.

PR #625, opened 2020-11-18
Migration from Travis CI to Github Actions

fabratu:
This migrates all tests from Travis CI to Github Actions. Motivation for this PR is the recently introduced price model for Travis CI and reoccuring test errors due to (likely) image problems.

- Windows tests are still done by appveyor
- Since Actions are only triggered / visible on the default branch, here is a preview from my fork: https://github.com/fabratu/networkit/actions (The documentation build fails, since the deployment token is not added to the fork)
- Added secrets for deployments to project settings

First impressions:
- There is more caching involved. Setup time therefore takes less time for consecutive runs.
- More concurrent jobs / cores per job.
- Reduced runtime: ~2h Travis CI vs. ~25min Github Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,279 @@
+name: build
+
+on: [push, pull_request]
+
+env:
+  OMP_NUM_THREADS: 2
+  CXX_STANDARD: 11
+  MONOLITH: ON
+
+jobs:
+  macos-build-apple-clang:
+    name: "macOS ${{ matrix.os }} (latest apple-clang, Python 3.9): ${{ matrix.variant }}"
+    runs-on: macos-${{ matrix.os }}
+    env:
+      CC: cc
+      CXX: c++
+    strategy:
+      matrix:
+        os: ['11.0', '10.15']
+        variant: ['full build', 'non-monolithic']
+    continue-on-error: ${{ matrix.os == '11.0' }}
+    steps:
+      - name: Install prerequisites 📦
+        run: |
+          brew install libomp
+          brew install ninja
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Checkout submodules
+        run: git submodule update --init --recursive
+      - name: Prepare environment and run checks (full build)
+        if: matrix.variant == 'full build'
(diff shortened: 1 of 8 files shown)


angriman:
Shouldn't you delete the `.travis.yml` file?

avdgrinten:
Why do we move scripts to their own files instead of keeping them inline (not necessarily a bad change, I am just asking for the rationale).

fabratu:
`.travis.yml`:
I should have mentioned it in the first comment: We haven't decided yet whether to have both systems for a small period of time and delete the Travis stuff once everything runs fine. So I wanted to leave this for discussion.

External scripts:
Github Actions doesn't support reusing scripts. Therefore to avoid having the same script several times, the core and full build are moved to an external script. I decided to do the same for the other builds (coverage, doc, clang-tidy) in order to have consistency.

fabratu:
Besides: I would vote for having both tools for a short time.

angriman:
Yes I think it makes sense to do so.

angriman (approved):

avdgrinten:
Please add proper shebangs to the scripts. Maybe also reconsider the directory structure: do we really need multiple subdirectories in .github/workflows? Isn't one subdirectory for scripts + utils enough? Otherwise, this looks good to me.

avdgrinten (dismissed):
LGTM now.

fabratu:
Thanks! Except that it doesn't work anymore (some folder access error). Looking into it.

avdgrinten:
By the way, why do we still use AppVeyor for Windows builds?

angriman:
I think we could migrate the Windows builds to GitHub actions in a separate PR.

avdgrinten:
Let's get this merged as soon as it works.

fabratu:
Technically it works, but for macOS 11 (Big Sur) some tests in Aux seems to fail. I have yet no idea, what this is about. The tests worked with the previous beta image, but not with the most recent one.

Should I first sort this out or do we want to merge the PR before that?

avdgrinten:
Travis does not test MacOS 11 either, so we can just drop these configurations / mark them as "allowed to fail" (if that is possible).

fabratu:
Ok, for now I added `continue-on-error` (which I think is what you suggest). You can see [here](https://github.com/fabratu/networkit/actions), that some actions fail. However if tagged with this option, the whole CI-pipeline will still be true. 

macOS 11.0 will get mandatory in a different PR.

angriman (approved):
Looks fine to me now

PR #626, opened 2020-12-07
Katz centrality: edge direction

angriman:
This PR allows the user to count in-edges or out-edges when computing Katz centrality.
It also includes additional minor fixes.

### DIFF
--- a/include/networkit/centrality/KatzCentrality.hpp
+++ b/include/networkit/centrality/KatzCentrality.hpp
@@ -12,6 +12,8 @@
 
 namespace NetworKit {
 
+enum EdgeDirection : char { InEdges = 0, OutEdges = 1 };
+
 /**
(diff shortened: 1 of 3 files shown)


include/networkit/centrality/KatzCentrality.hpp
```diff
@@ -12,6 +12,8 @@
 
 namespace NetworKit {
 
+enum KatzVariant { InEdges = 0, OutEdges = 1 };
```
avdgrinten, line 15: Do not call this KatzVariant (The same could be done in the case of PageRank, too) but something like EdgeDirection.

angriman: Agree, I changed the name

fabratu (approved):

PR #629, opened 2020-12-09
Update tools

angriman:
clang-tidy: updated to version 11
build with sanitizers: updated to gcc 10 (before we were using gcc 5 due to a Travis issue), and to the latest version of Python.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,7 @@ jobs:
           CXX_STANDARD: ${{ matrix.variant }}
 
   linux-build-clang-tidy:
-    name: "Linux (clang-9, Python 3.8): clang-tidy build"
+    name: "Linux (clang-11): clang-tidy build"
     runs-on: ubuntu-20.04
     env:
       CC: clang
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -198,23 +198,23 @@ jobs:
...
-    runs-on: ubuntu-18.04
+    name: "Linux (gcc-10, Python 3.9): core build, sanitizers, coverage"
+    runs-on: ubuntu-20.04
     env:
-      CC: gcc-5
-      CXX: g++-5
+      CC: gcc-10
+      CXX: g++-10
     steps:
       # gcc-5 has to be used with 18.04, because the integrated version from 16.04-image has a bug with linker flags:
```
fabratu, line 207: Only minor, but can you remove the comment? (Since we switch now to gcc-10 / 20.04)

fabratu:
Also good to know, that it works with clang-tidy 11. We stayed with 9, because 10 gave some problems.

.github/workflows/ci.yml
angriman: I removed it but in the C++14 PR, sorry.

fabratu: Ok :) Then we leave it as is. It would be good next time, to have it in one PR though.

fabratu (dismissed):

fabratu (approved):

PR #631, opened 2020-12-10
Refactor Algorithm and deprecate toString()

angriman:
I don't see a real use case for the `toString()` method in Algorithm (an algorithm's name is already known from the class name and the parameters can be checked in the docs) and very few subclasses implement it.
In this PR:
- deprecate `toString()`
- set `hasRun` to `false` by default and use the default constructor
- other minor changes

### DIFF
--- a/include/networkit/centrality/LocalPartitionCoverage.hpp
+++ b/include/networkit/centrality/LocalPartitionCoverage.hpp
@@ -42,7 +42,7 @@ public:
      * The name of this algorithm.
      * @return "Local partition coverage"
      */
-    std::string toString() const override;
+    std::string TLX_DEPRECATED(toString() const override);
 
 protected:
     const Partition& P;
--- a/networkit/community.pyx
+++ b/networkit/community.pyx
@@ -1080,7 +1080,7 @@ def detectCommunities(G, algo=None, inspect=True):
 	algo.run()
 	zeta = algo.getPartition()
 	t.stop()
-	print("{0} detected communities in {1} [s]".format(algo.toString(), t.elapsed))
+	print("Communities detected in {:.5f} [s]".format(t.elapsed))
 	if inspect:
 		print ("solution properties:")
 		inspectCommunities(zeta, G)
--- a/networkit/engineering.pyx
+++ b/networkit/engineering.pyx
@@ -98,18 +98,15 @@ def strongScaling(algorithmClass, threadSequence, inargs, inputTitle=None, repet
 		print("set number of threads to {0}".format(getMaxNumberOfThreads()))
 		for r in range(repetitions):
 			algorithm = algorithmClass(**inargs)
-			print("running {0}".format(algorithm.toString()))
+			print("running algorithm")
 			timer = stopwatch.Timer()
 			result = algorithm.run()
 			timer.stop()
 			print("elapsed time: {0}".format(timer.elapsed))
 			if inputTitle is None:
-				try:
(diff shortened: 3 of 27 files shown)


avdgrinten:
Is toString exposed to Python?

angriman:
Yes, same argument also there.

avdgrinten:
Why is it beneficial to deprecate the method? If its exposed to Python, people might use it in scripts / on the repl and its maintenance overhead is low.

angriman:
I just don't see the benefit of keeping it since almost all algorithms in NetworKit do not implement it. Also, because of how `toString` is implemented in the base class and not overridden in the derived classes, if a C++/Python script calls `toString` it just gets a runtime error.
So I think we should either deprecate it or make it pure virtual and implement it in the derived classes.

avdgrinten:
Okay, if it is not implement / errors out on current `master`, that's a good argument to remove it.

hmeyerhenke:
I think the original reason for toString was to print it into the output of experimental results. That way, one knew which algorithm produced which result. I guess more recently this has been handled with Jupyter notebooks in a different way. Bottom line: if you think this use case is not needed any more and toString() is not helpful in any other way, then deprecate it.

include/networkit/centrality/LocalPartitionCoverage.hpp
```diff
@@ -42,7 +42,7 @@ class LocalPartitionCoverage : public Centrality {
      * The name of this algorithm.
      * @return "Local partition coverage"
      */
-    std::string toString() const override;
+    TLX_DEPRECATED(std::string toString() const override);
```
fabratu, line 45: That will not make a difference in terms of functionality, but shouldn't it be like`std::string  TLX_DEPRECATED(toString() const override);`? This of course applies to all occurences.

networkit/community.pyx
```diff
@@ -1080,7 +1080,7 @@ def detectCommunities(G, algo=None, inspect=True):
 	algo.run()
 	zeta = algo.getPartition()
 	t.stop()
-	print("{0} detected communities in {1} [s]".format(algo.toString(), t.elapsed))
+	print("Communities detected in {:.5f} [s]".format(t.elapsed))
```
fabratu, line 1083: Even though `toString()` creates runtime-errors for many algorithms, deprecation and removal should be two steps.

networkit/engineering.pyx
```diff
@@ -98,18 +98,15 @@ def strongScaling(algorithmClass, threadSequence, inargs, inputTitle=None, repet
 		print("set number of threads to {0}".format(getMaxNumberOfThreads()))
 		for r in range(repetitions):
 			algorithm = algorithmClass(**inargs)
-			print("running {0}".format(algorithm.toString()))
```
fabratu, line 101: Same as above (deprecation/removal)

fabratu (changes requested):

networkit/community.pyx
angriman: I think that removing usages of a newly deprecated feature is fine.

avdgrinten: Yes, this change should be applied to prevent deprecation warnings when building NetworKit itself.

(But Fabian is right that the function definition should not be removed.)

fabratu: That makes sense of course.

fabratu (approved):

PR #632, opened 2020-12-10
Update Github Actions: Added Windows, llvm 11.0 for Ubuntu and let deployments only trigger on master

fabratu:
- Added llvm 11 for Ubuntu / Unix
- Added Windows tests + removed appveyor
- Removed `continue-on-error` for documentation build
- Add conditions, so that deployments for documentation and coveralls are only happening when a push is happening on master. This way it should be possible to pass secrets to the actions workflow.
- Changed minimum support for gcc from 4 to 5.5
- Changed Python version to 3.9 for Ubuntu / Unix


angriman (approved):
LGTM

PR #634, opened 2020-12-16
Fixes publishing of doc- and coverage-build. Temp. removes gcc-10 tests for macOS 11.0

fabratu:
(Hopefully) the last PR for ironing out the remaining problems with github actions. The following changes were made:

- Coverage build: Switched from ninja to make, since there are some problems with folder redirection and therefore detection of code-files for `cpp-coveralls`. This problems caused a report of 0% coverage in some cases.
- Documentation build: Moved from using `JamesIves/github-pages-deploy-action` to `peaceiris/actions-gh-pages` for publishing to github pages. Even though JamesIves-plugin seems to stable and well-known (and runs a set of common git-commands), it fails to publish in our case. The new plugin works very similar, is also well known and [tests](https://github.com/fabratu/networkit/runs/1558485371?check_suite_focus=true) publishing in our case is successful. 

Minor stuff:
- Documentation build: Removed the `continue-on-error` option for the build.
- macOS 11.0: For macOS 11.0 the gcc-10 build seems to fail at random points during the tests (with a process abort). Since the 11.0 image is preview anyway, the gcc-10 is commented out until the image is final. Reason: Github Actions as of now has no option to mark a job a really optional. The `continue-on-error` option let the build continue for other tests and also mark them as successful in the Actions-tab. However PRs and Pushes are marked as failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
     strategy:
       matrix:
         os: ['11.0', '10.15']
-        variant: ['llvm-11', 'gcc-10']
+        variant: ['llvm-11'] # 'gcc-10' is disabled for now, since the workflow exits randomly during test-phase
+        # Will be resumed once the 11.0 image is final. 
     continue-on-error: ${{ matrix.os == '11.0' }}
     steps:
       - name: Install prerequisites 📦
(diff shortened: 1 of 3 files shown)


angriman (approved):

avdgrinten:
Fast tracked because it fixes PR.

PR #635, opened 2020-12-16
Partition: warning fix and clang-format

angriman:
Fix `performance-unnecessary-value-param` in Partition, clang-format the class.


avdgrinten (approved):
LGTM, it's a trivial fix.

PR #638, opened 2020-12-17
New algorithm: group harmonic closeness

angriman:
This PR brings a new greedy algorithm for group-harmonic maximization. The algorithm is described in the paper by E. Angriman, R. Becker, G. D'Angelo, H. Gilbert, A. van der Grinten, H. Meyerhenke: "Group-Harmonic and Group-Closeness Maximization -- Approximation and Engineering", accepted at ALENEX 2021, arXiv version [here](https://arxiv.org/abs/2010.15435).

### DIFF
--- a/include/networkit/centrality/GroupHarmonicCloseness.hpp
+++ b/include/networkit/centrality/GroupHarmonicCloseness.hpp
@@ -1,0 +1,85 @@
+/*
+ * GroupHarmonicCloseness.hpp
+ *
+ * Created on: 15.12.2020
+ *     Author: Eugenio Angriman <angrimae@hu-berlin.de>
+ */
+
+// networkit-format
+
+#ifndef NETWORKIT_CENTRALITY_GROUP_HARMONIC_CLOSENESS_HPP_
+#define NETWORKIT_CENTRALITY_GROUP_HARMONIC_CLOSENESS_HPP_
+
+#include <memory>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+/**
+ * @ingroup centrality
+ */
+class GroupHarmonicCloseness final : public Algorithm {
+
+public:
+    /**
+     * Approximation algorithm for the group-harmonic maximization problem. The computed solutions
+     * have a guaranteed $\\lambda(1 - \\frac{1}{2e})$ (directed graphs) and
+     * $\\lambda(1 - \\frac{1}/{e})/2$ (undirected graphs) approximation ratio,
+     * where $\\lambda$ is the ratio
+     * between the minimal and the maximal edge weight. The algorithm is the one proposed in
+     * Angriman et al., ALENEX 2021. The worst-case running time of this approach is quadratic, but
+     * usually much faster in practice.
+     *
+     * @param G The input graph.
+     * @param k Size of the group of nodes.
+     */
+    GroupHarmonicCloseness(const Graph &G, count k = 1);
+
+    /**
+     * Runs the algorithm.
+     */
+    void run() override;
+
+    /**
+     * Returns the computed group.
+     *
+     * @return The computed group.
+     */
+    const std::vector<node> &groupMaxHarmonicCloseness() const;
+
+    /**
+     * Computes the group-harmonic score of the group of nodes in the given range.
+     *
+     * @param graph The input Graph.
+     * @param first,last The range that contains the vertices in the group.
+     *
+     * @returns The score of the group of nodes in the given range.
+     */
+    template <class InputIt>
+    static double scoreOfGroup(const Graph &graph, InputIt first, InputIt last);
+
+    class GroupHarmonicClosenessInterface : public Algorithm {
+    public:
+        std::vector<node> group;
+    };
+
+private:
+    const bool weighted;
+    // Always one between GroupHarmonicClosenessUnweighted and GroupHarmonicClosenessWeighted, see
+    // implementation.
+    std::unique_ptr<GroupHarmonicClosenessInterface> impl;
+
+    static double scoreOfGroup(const Graph &graph, const std::vector<node> &inputGroup);
+};
+
+template <class InputIt>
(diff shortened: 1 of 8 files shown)


angriman:
Before reviewing this we should merge #658 so I can simplify this PR with `ReachableNodes`.

angriman:
Now that #658 is merged this can be reviewed.

include/networkit/centrality/GroupHarmonicCloseness.hpp
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @returns The score of the group of nodes in the given range.
+     */
+    template <class InputIt>
+    static double scoreOfGroup(const Graph &graph, InputIt first, InputIt last);
+
+private:
+    const bool weighted;
+    std::unique_ptr<GroupHarmonicClosenessDetails::GroupHarmonicClosenessUnweighted> implUnweighted;
+    std::unique_ptr<GroupHarmonicClosenessDetails::GroupHarmonicClosenessWeighted> implWeighted;
```
avdgrinten, line 76: Two pointers here are kind of unnecessary. We could use a `std::unique_ptr<void *>` (maybe with a comment that it is always one of the two classes, based on the `weighted` variable) and change the code in `run()` to:

```cpp
if (weighted) {
    auto unweightedImpl = static_cast<GroupHarmonicClosenessUnweighted>(impl);
    unweightedImpl->run();
}else{
    auto weightedImpl = static_cast<GroupHarmonicClosenessWeighted>(impl);
    weightedImpl->run();
}
```

networkit/cpp/centrality/GroupHarmonicClosenessImpl.cpp
```diff
@@ -0,0 +1,596 @@
+/*
```
avdgrinten, line 1: This file implements a template specialization in a source file while the template is in a header file. This is an anti-pattern and has many pitfalls.

Move the implementation into the same file as the template declaration.

networkit/cpp/centrality/GroupHarmonicClosenessImpl.hpp
```diff
@@ -0,0 +1,106 @@
+/*
```
avdgrinten, line 1: After the .cpp file is merged into this one, I would suggest to merge both into `GroupHarmonicCloseness.cpp`. Then the entire template can be put into an anonymous namespace so that it is not visible outside of `GroupHarmonicCloseness.cpp`.

avdgrinten (changes requested):

angriman:
I addressed the comments.

networkit/cpp/centrality/GroupHarmonicCloseness.cpp
```diff
@@ -0,0 +1,695 @@
...
+
+private:
+    const Graph *G;
+    const count k;
+    WeightType minEdgeWeight = infDist;
+
+    std::unique_ptr<Partition> component;
+
+    std::vector<WeightType> distFromGroup, curBestDist;
+    std::vector<std::vector<WeightType>> curDistGlobal;
```
avdgrinten, line 58: It's not really clear what this stores, add a comment.

networkit/cpp/centrality/GroupHarmonicCloseness.cpp
```diff
@@ -0,0 +1,695 @@
...
+    };
+
+    template <class Type>
+    struct Less {
+        Less(const std::vector<Type> &dist) : dist(dist) {}
+        bool operator()(node x, node y) const noexcept { return dist[x] < dist[y]; }
+
+    private:
+        const std::vector<Type> &dist;
+    };
```
avdgrinten, line 81: Should these templates be in `Aux` such that multiple algos can use them? (See other PR.)

networkit/cpp/centrality/GroupHarmonicCloseness.cpp
```diff
@@ -0,0 +1,695 @@
...
+        *G, group.begin(), group.end(),
+        [&distFromGroup = distFromGroup](node x, count d) { assert(distFromGroup[x] == d); });
+}
+
+template <>
+void GroupHarmonicClosenessImpl<edgeweight>::checkDistFromGroup() const {
+    Traversal::DijkstraFrom(
+        *G, group.begin(), group.end(),
+        [&distFromGroup = distFromGroup](node x, edgeweight d) { assert(distFromGroup[x] == d); });
+}
```
avdgrinten, line 662: I wonder how much of these specializations could be avoided by specializing a `traversalFromGroup` function instead and then using that function in a non-specialized `checkDistFromGroup` (and other functions?).

networkit/cpp/centrality/GroupHarmonicCloseness.cpp
```diff
@@ -0,0 +1,684 @@
...
+
+    static double scoreOfGroup(const Graph &graph, const std::vector<node> &inputGroup);
+
+private:
+    const Graph *G;
+    const count k;
+    WeightType minEdgeWeight = infDist;
+
+    // Connected components partitioning
+    std::unique_ptr<Partition> component;
```
avdgrinten, line 57: Why is this a unique_ptr and not just a member?

networkit/cpp/centrality/GroupHarmonicCloseness.cpp
```diff
@@ -0,0 +1,684 @@
...
+    std::vector<WeightType> distFromGroup;
+    // Distance between the group to every node after the most promising swap is performed
+    std::vector<WeightType> curBestDist;
+    // Distance vectors used by SSSPs to avoid dynamic memory allocations
+    std::vector<std::vector<WeightType>> curDistGlobal;
+    std::vector<count> reachableNodesInComponent, reachableNodesUB;
+    std::vector<node> pred, nearerNodes;
+    std::vector<std::vector<node>> curNearerNodesGlobal;
+    std::vector<double> margGain;
+    std::vector<std::vector<bool>> visitedGlobal;
```
avdgrinten, line 69: Still a lot of undocumented members (that are also not self-documenting).

avdgrinten (approved):
LGTM.

PR #640, opened 2020-12-18
Warning fix: performance-noexcept-move-constructor

angriman:
As reported here [1], move constructors and assignment operators should be declared `noexcept` at least when used with STL containers, otherwise STL will choose copy constructors and thus there is a performance loss.

This PR adds the `noexcept` specifier to user-defined move constructors and assignment operators.

[1] https://clang.llvm.org/extra/clang-tidy/checks/performance-noexcept-move-constructor.html

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -983,13 +983,13 @@ public:
     Graph(const Graph &other) = default;
 
     /** Default move constructor */
-    Graph(Graph &&other) = default;
+    Graph(Graph &&other) noexcept = default;
 
     /** Default destructor */
     ~Graph() = default;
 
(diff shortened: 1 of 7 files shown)


fabratu (approved):

include/networkit/graph/Graph.hpp
```diff
@@ -983,13 +983,13 @@ class Graph final {
     Graph(const Graph &other) = default;
 
     /** Default move constructor */
-    Graph(Graph &&other) = default;
+    Graph(Graph &&other) noexcept(true) = default;
```
avdgrinten, line 986: Use `noexcept`, not `noexcept(true)`.

avdgrinten (approved):

PR #643, opened 2020-12-18
clang-format: algebraic/algorithms

angriman:
Right now only a small subset of C++ source files (123 over 697) comply with our coding style and this is checked in new PRs only if the authors add `// networkit-format` in new source files.
I think that this is a good temporary solution but it would be desirable to make the whole codebase adhere to the coding style. This would be helpful especially when refactoring, which is sometimes tedious to do if the file is not formatted.

I wrote a simple script to format and add `// networkit-format` to all the non-compliant files and my plan would be to format a few files per time (always excluding the ones involved in other PRs) until we cover the whole codebase. Afterwards I would remove `// networkit-format` and enforce our coding style to all source files. Here I started with `algebraic/algorithms`.

### DIFF
--- a/include/networkit/algebraic/algorithms/AlgebraicMatchingCoarsening.hpp
+++ b/include/networkit/algebraic/algorithms/AlgebraicMatchingCoarsening.hpp
@@ -5,6 +5,8 @@
  *      Author: Michael Wegner (michael.wegner@student.kit.edu)
  */
 
+// networkit-format
+
 #ifndef NETWORKIT_ALGEBRAIC_ALGORITHMS_ALGEBRAIC_MATCHING_COARSENING_HPP_
 #define NETWORKIT_ALGEBRAIC_ALGORITHMS_ALGEBRAIC_MATCHING_COARSENING_HPP_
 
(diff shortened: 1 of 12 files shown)


avdgrinten:
Reformating causes churn: for example, it requires all PRs to be rebased and that rebasing is quite tedious because reformatting also breaks all diffs. This applies to all branches of NetworKit that are not in the main repository.

We should not reformat for the sake of formatting. Our policy is to reformat individual pieces of code when the need arises. Here, I do not see how this yields a positive benefit to cost ratio.

angriman:
> Reformating causes churn: for example, it requires all PRs to be rebased and that rebasing is quite tedious because reformatting also breaks all diffs (requiring changes to be copied over manually). This additionally applies to all branches of NetworKit that are not in the main repository.
> 
That's why I suggested to only format files that are not touched by other PRs, on the other hand I see that this could affect other forks that did not post PRs yet.

> We should not reformat for the sake of formatting. Our policy is to reformat individual pieces of code when the need arises. Here, I do not see how this yields a positive benefit to cost ratio.

My point is that when it comes to refactoring non-formatted files one has the additional burden of formatting the code manually, which is something that can be avoided if all files comply to the same coding style.
The cost would be distributed over time and it is not that huge, formatting is done automatically for a few files per time and the changes are trivial to review.

I agree that this does not bring substantial benefit in the short term, but I think that it makes the codebase easier to maintain in the long term. Anyway I don't want to insist, if you don't agree I will close the PR.

include/networkit/algebraic/algorithms/AlgebraicMatchingCoarsening.hpp
```diff
@@ -69,9 +75,11 @@ AlgebraicMatchingCoarsening<Matrix>::AlgebraicMatchingCoarsening(const Graph& gr
     P = Matrix(graph.numberOfNodes(), numCoarse, triplets); // dimensions: fine x coarse
 }
 
-template<class Matrix>
+template <class Matrix>
 void AlgebraicMatchingCoarsening<Matrix>::run() {
-    Matrix coarseAdj = P.transpose() * A * P; // Matrix::mTmMultiply performs worse due to high sparsity of P (nnz = n)
+    Matrix coarseAdj =
+        P.transpose() * A
+        * P; // Matrix::mTmMultiply performs worse due to high sparsity of P (nnz = n)
```
fabratu, line 82: That is looking weird. How about putting the comment above line 80?

fabratu (changes requested):

angriman:
@fabratu I addressed your comment, thank you.

However, as far as we don't come to an agreement, I am not sure whether or not to proceed with this PR.

avdgrinten (changes requested):
I still suggest that this PR (or similar PRs that reformat large parts of the code base without improving them feature-wise) should not be merged. We should keep reformatting files as needed (e.g., when applying other changes) and not reformat for the sake of reformatting. I do not buy that either (i) the current code is hard to read due to formatting nor (ii) that reformatting as part of PRs that add features is a heavy burden.

PR #645, opened 2020-12-20
check_code.py: check version of clang-format

angriman:
Currently, `./check_code.py` checks the coding style by calling `clang-format-8` but this does not work with newer versions of clang-format and even with clang-format 8 if it can only be invoked using `clang-format`.

This PR fixes this issue and updates the coding style CI build to clang-format-11. Further, as suggested in #576, `CppClangFormat.py` will throw a `FileNotFoundError` upon failure.

### DIFF
--- a/extrafiles/tooling/CppClangFormat.py
+++ b/extrafiles/tooling/CppClangFormat.py
@@ -15,18 +15,39 @@ import shutil
 import subprocess
 import os
 
+def getEnv():
+	"""
+	Returns the current environment variables with 'LANG' set to 'C' to ensure
+	the output to be in English.
+	"""
+	env = dict(os.environ)
+	env['LANG'] = 'C'
+	return env
+
+def isSupported(cmd):
+	""" Checks if the given clang-format command is available and if its version is recent enough. """
+	if shutil.which(cmd) is None:
+		return False
+	# Read major revision number from "clang-format version XX.XX.XX ... "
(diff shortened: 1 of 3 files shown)


extrafiles/tooling/CppClangFormat.py
```diff
@@ -15,11 +15,22 @@
 import subprocess
 import os
 
+def isSupported(cmd):
+	""" Checks if the given clang-format command is available and if its version is recent enough. """
+	if shutil.which(cmd) is None:
+		return False
+	# Read major revision number from "clang-format version XX.XX.XX ... "
+	version = str(subprocess.check_output([cmd, "--version"],
+		universal_newlines=True)).strip().split()[2].split('.')[0]
```
avdgrinten, line 24: Make sure that this runs with `LANG=C` to force English output.

angriman: Good point, I fixed that.

fabratu (dismissed):

extrafiles/tooling/CppClangFormat.py
avdgrinten: Well, you added it below (which is good) but this check_output should also be changed ;)

angriman: Ah I missed that, now it should be ok.

avdgrinten (approved):

Issue #622, opened 2020-11-11
Potential data races


tamugang:
Hi, we found three potential races in this project, with a static analysis tool we are currently developing. These races do not seem severe issues, but we think it's still good to report them.

### Race on `totalCandidates`

/home/networkit/networkit/cpp/generators/HyperbolicGenerator.cpp:244: A variable defined here could be accessed by multiple threads simultaneously without correct synchronizations.

 244|    count totalCandidates = 0;

This is the OpenMP section causing the race (networkit/cpp/generators/HyperbolicGenerator.cpp:246):

```C++
count totalCandidates = 0;
    #pragma omp parallel for
    for (omp_index i = 0; i < static_cast<omp_index>(n); i++) {
        vector<index> near;
        totalCandidates += quad.getElementsProbabilistically(HyperbolicSpace::polarToCartesian(angles[i], radii[i]), edgeProb, anglesSorted, near);
        for (index j : near) {
            if (j >= n) ERROR("Node ", j, " prospective neighbor of ", i, " does not actually exist. Oops.");
            if (j > i) {
                result.addHalfEdge(i, j);
            }
        }

    }
```

This is not a reduction parallel loop, so the `+=` operator on `totalCandidates` causes races. It seems this variable is only used for debug purpose though.

### Race on `queued`

/home/networkit/networkit/cpp/distance/NeighborhoodFunctionApproximation.cpp:80: A variable defined here could be accessed by multiple threads simultaneously without correct synchronizations.

 80|    bool queued = true;

This is the OpenMP section causing the race (networkit/cpp/distance/NeighborhoodFunctionApproximation.cpp:88):

```C++
#pragma omp parallel for schedule(guided)
        for (omp_index v = 0; v < static_cast<omp_index>(activeNodes.size()); ++v) {
            if (!activeNodes[v]) continue;
            index tid = (index)omp_get_thread_num();

            for (count j = 0; j < k; j++) {
                // and to all previous neighbors of all its neighbors
                G->forNeighborsOf(v, [&](node u) {
                    mCurr[v][j] |= mPrev[u][j];
                });
            }
            // the least bit number in the bitmask of the current node/distance that has not been set
            double b = 0;
            for (count j = 0; j < k; j++) {
                for (count i = 0; i < lengthOfBitmask; i++) {
                    if (((mCurr[v][j] >> i) & 1) == 0) {
                        b += i;
                        break;
                    }
                }
            }
            // calculate the average least bit number that has not been set over all parallel approximations
[...]

PR #646, opened 2021-01-06
Fix potential data races

fabratu:
Fixes potential data races as reported in #622.

- `totalCandidates`: Added a reduction
- `queued`: ~~Make queued atomic~~ Since `msvc` does only support OpenMP 2.0, atomic writes are not available. Therefore a reduction for queued is used, which is equivalent to `queued = true`. Initial benchmarks on the artifical gtest graph instances also show better performance for reduction vs. atomic for this case. 
- `os_stack_trace_getter_`: This one is the most tricky one, since the error is not in NetworKit, but in googletest. From the report, the data race potentially occurs when executing `EXPECT_NE`. However from code-inspection (see https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L2670 which for every testcase calls https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L4859) it seems, that the problem may occur for every `EXPECT_*` or `ASSERT_*`, which is executed in parallel. One possibility would be to switch from `parallelForEdges` to `forEdges`. However other iterative methods like `parallelForNodes` are covered by their own test-cases. So instead of removing the parallel-tests, the PR refactors the test-cases in order to clearly distinquish between parallel and non-parallel functions. In addition all `EXPECT_*` and `ASSERT_*` are removed from parallel-sections.

### DIFF
--- a/networkit/cpp/distance/NeighborhoodFunctionApproximation.cpp
+++ b/networkit/cpp/distance/NeighborhoodFunctionApproximation.cpp
@@ -78,13 +78,13 @@ void NeighborhoodFunctionApproximation::run() {
     std::vector<count> localSumRemoved(omp_get_max_threads(), 0);
 
     bool queued = true;
-    while (queued) {
+    while(queued) {
         queued = false;
         count tmp = 0;
         for (index i = 0; i < (count)omp_get_max_threads(); ++i) {
             tmp += localSumRemoved[i];
         }
-        #pragma omp parallel for schedule(guided)
+        #pragma omp parallel for schedule(guided) reduction(||: queued)
         for (omp_index v = 0; v < static_cast<omp_index>(activeNodes.size()); ++v) {
             if (!activeNodes[v]) continue;
             index tid = (index)omp_get_thread_num();
@@ -125,7 +125,8 @@ void NeighborhoodFunctionApproximation::run() {
                 localSumRemoved[tid] += estimatedConnectedNodes;
                 activeNodes[v] = 0;
             } else {
-                queued = true;
+                // queued is not set to true directly because msvc does not support atomic write operations. 
+                queued = queued || true;
             }
         }
         for (const auto& elem : localEstimatesSum) {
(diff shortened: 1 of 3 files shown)


networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2007,19 +1992,79 @@ TEST_P(GraphGTest, testForWeightedEdgesWithIds) {
...
+        ASSERT_EQ(3u, m);
+    }
+}
+
+TEST_P(GraphGTest, testParallelForWeightedEdgesWithIds) {
+    std::vector<Graph> graphs;
+    graphs.push_back(Graph(10, false, false));
+    graphs.push_back(Graph(10, false, true));
+    graphs.push_back(Graph(10, true, false));
+    graphs.push_back(Graph(10, true, true));
```
angriman, line 2047: Same as above.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2007,19 +1992,79 @@ TEST_P(GraphGTest, testForWeightedEdgesWithIds) {
...
         }
+    }
+}
+
+TEST_P(GraphGTest, testParallelForEdgesWithIds) {
+    std::vector<Graph> graphs;
+    graphs.push_back(Graph(10, false, false));
+    graphs.push_back(Graph(10, false, true));
+    graphs.push_back(Graph(10, true, false));
+    graphs.push_back(Graph(10, true, true));
```
angriman, line 2003: I suggest to use [emplace_back](https://en.cppreference.com/w/cpp/container/vector/emplace_back)

networkit/cpp/distance/NeighborhoodFunctionApproximation.cpp
```diff
@@ -125,7 +125,8 @@ void NeighborhoodFunctionApproximation::run() {
                 localSumRemoved[tid] += estimatedConnectedNodes;
                 activeNodes[v] = 0;
             } else {
-                queued = true;
+                // queued is not set to true directly, because msvc does not support atomic write operations.
```
angriman, line 128: Remove the comma after "directly"

angriman (changes requested):

angriman (approved):
LGTM

PR #647, opened 2021-01-06
Fix isolated conductance algorithm

darabos:
I noticed that conductance came out as 1 for all clusters. Turns out #472 changed this comparison in the wrong direction by accident. (I'm surprised this even managed to run without crashing.) With this fix I'm getting reasonable-looking conductance values.


avdgrinten (approved):
LGTM. Looks like the previous change indeed slipped through code review unnoticed. It's good that this is fixed now. (And yes, I would also expect it to segfault, at least of u or v are large enough.)

PR #648, opened 2021-01-06
Fix alpha default initialization in Katz centrality

darabos:
Good news! A unit test in LynxKite caught this issue when I rebuilt after https://github.com/networkit/networkit/commit/e568f803a2862a43a514169225649bc1834e5bc7.

That commit looks super innocent. But instead of setting the `alpha` field of `KatzCentrality` it just changes the local `alpha` variable. I tried setting `this->alpha`, but it's `const`. We have to set it correctly right away in the initializer.

Without this fix, `alpha` is zero by default. This causes all the edges to be ignored and every node gets the same centrality.

### DIFF
--- a/include/networkit/centrality/KatzCentrality.hpp
+++ b/include/networkit/centrality/KatzCentrality.hpp
@@ -27,6 +27,7 @@ protected:
     const double alpha; // damping
     const double beta; // constant centrality amount
     const double tol; // error tolerance
+    static double defaultAlpha(const Graph& G);
 
 public:
     /**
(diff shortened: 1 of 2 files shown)


angriman (changes requested):
Thank you for this fix. I just have a minor comment, I'd suggest to declare `defaultAlpha` as a static method of `KatzCentrality`.

darabos:
> Thank you for this fix. I just have a minor comment, I'd suggest to declare `defaultAlpha` as a static method of `KatzCentrality`.

Good point. Should be done.

angriman (approved):
Thank you, LGTM.

PR #649, opened 2021-01-08
Disables CI tests for macos 11.0

fabratu:
Currently the macos 11.0 images (only available as preview) and pipelines from github actions produce incoherent results. Sometimes tests fail, doesn't even start at all and more. Therefore the tests are disabled until the image is final.


angriman (approved):

avdgrinten:
Can be fast tracked as it fixes CI issues.

Issue #653, opened 2021-01-12
Case configuration in clang-tidy


michitux:
While trying to fix the case of variable names, I noticed that the configuration provided in `.clang-tidy` is not consistent with the style guide. In particular, at least `MemberCase`, `ParameterCase` and `VariableCase` should be `camelBack` according to the style guide and not `CamelCase` as in the configuration. It would be nice if the configuration was consistent with the style guide such that new code could be checked and corrected using clang-tidy by manually enabling the checks on the command line and including only the headers that shall be changed. I suspect that globally adapting the style introduces too many changes but this could of course also be considered as an option, then the checks could be enabled by default.

avdgrinten:
Yes, these `.clang-tidy` settings should be fixed, agreed.

angriman:
Those checks do not even throw warnings, that's because we do not have the `readability-identifier-naming` enabled. I'll fix that.

angriman:
The case configuration for clang-tidy has been fixed, I am closing this.

PR #654, opened 2021-01-13
Fix case configuration in clang-tidy

angriman:
As described in #653


avdgrinten:
There are way too many violations to enable the check. We should change the configuration without enabling the check.

angriman:
True, I disabled them.

avdgrinten:
They are still enabled in the latest commit (and clang-tidy produces 80k lines of warnings).

avdgrinten (approved):
LGTM.

PR #657, opened 2021-01-15
Fix clang-analyzer warnings

angriman:
Fixes additional warnings spotted by clang-analyzer

### DIFF
--- a/networkit/cpp/io/MatrixMarketReader.cpp
+++ b/networkit/cpp/io/MatrixMarketReader.cpp
@@ -1,5 +1,5 @@
 /*
- * MatrixMarketReader.cpp
+ *  MatrixMarketReader.cpp
  *
  *  Created on: 25.07.2014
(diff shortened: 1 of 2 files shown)


fabratu (approved):
LGTM.

networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -26,7 +26,7 @@ namespace {
 
 CSRMatrix MatrixMarketReader::read(const std::string &path) {
     std::ifstream in(path);
-    if (!in) {
+    if (!in.is_open()) {
```
avdgrinten, line 29: What's the warning that the static analyzer throws here?

angriman: I don't remember the exact warning, the issue here is that we want to check if the file exists, but `if (!in)` does not do that.

PR #658, opened 2021-01-21
New module: reachability

angriman:
This PR implements algorithms to compute/estimate the number of reachable nodes from each node of a graph. These algorithms are already implemented in two NetworKit classes by just copy-pasting the code.

I did not find a suitable module for these algorithms so I created a new `reachability` module, and I moved the `AllSimplePaths` algorithm there.


include/networkit/reachability/ReachableNodes.hpp
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param u A node.
+     * @returns Upper bound of number of nodes reachable from @a u.
+     */
+    count numberOfReachableNodesUB(node u) const {
+        assureFinished();
+        if (G->isDirected())
+            return exact ? reachableLB[u] : reachableUB[u];
+        else
+            return reachableLB[u];
```
fabratu, line 67: For the undirected case: Since `reachableLB == reachableUB` (while not requires exact to be `true`), why does `numberOfReachableNodes` returns a runtime_error if exact is false?

include/networkit/reachability/ReachableNodes.hpp
```diff
@@ -0,0 +1,80 @@
...
+public:
+    /**
+     * Determines or estimates the number of reachable nodes from each node in the graph.
+     *
+     * @param G The graph.
+     * @param exact Whether or not to compute the number of reachable nodes exactly. Only used for
+     * directed graphs, on undirected graphs the number of reachable nodes from every node can be
+     * computed in linear time.
+     */
+    ReachableNodes(const Graph &G, bool exact = true);
```
fabratu, line 26: Since the `run()` method can be called repeatedly, how about adding a setter for `exact` (or exposing the variable)? In this way a user doesn't have to allocate 2*n memory a second time for having the same run with exact/non-exact.

notebooks/Reachability.ipynb
```diff
@@ -0,0 +1,137 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This algorithm computes all existing simple paths from a source node to a target node.\n",
+    "\n",
+    "The [AllSimplePaths(G, source, target, cutoff=none)](https://networkit.github.io/dev-docs/python_api/distance.html?highlight=all#networkit.distance.AllSimplePaths) constructor expects an unweighted graph, the source node and the target node as mandatory parameters. The maximum length of the paths can be fixed through `cutoff` parameter. This algorithm could take a lot of time on large networks (many edges), especially if the cutoff value is high or not specified.\n",
+    "\n",
+    "The algorithm is implemented only for unweighted graphs, so we shall convert the G to an unweighted graph."
```
fabratu, line 47: convert ~~the~~ G

fabratu (changes requested):

networkit/cpp/reachability/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+networkit_add_module(reachability
+    AllSimplePaths.cpp
+    ReachableNodes.cpp
+    )
+
+networkit_module_link_modules(reachability
+    base
```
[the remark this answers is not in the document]
angriman: I will remove the `auxiliary` include from `AllSimplePaths` since it is unused.

include/networkit/reachability/ReachableNodes.hpp
angriman: Only when `exact` is set to `true` then the number of reachable nodes is computed exactly. I think that `numberOfReachableNodes` should not be used if one does not compute the reachable nodes exactly.

networkit/cpp/centrality/TopCloseness.cpp
```diff
@@ -44,125 +44,25 @@ void TopCloseness::init() {
...
-    }
-}
+void TopCloseness::computeReachable() {
+    ReachableNodes rn(G, false);
+    rn.run();
 
-void TopCloseness::computeReachableNodesUndir() {
     reachLPtr = std::make_shared<std::vector<count>>(n);
+    reachUPtr = std::make_shared<std::vector<count>>(n);
+    auto &reachL = *reachLPtr, &reachU = *reachUPtr;
```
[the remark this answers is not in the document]
angriman: Agree, I fixed that.

include/networkit/reachability/ReachableNodes.hpp
angriman: Yes I think that's reasonable.

notebooks/Reachability.ipynb
angriman: Done

include/networkit/reachability/ReachableNodes.hpp
fabratu: Hmm ... it is just a bit strange, that this also applies to the undirected case. `exact` is not used in runUndirected() and the documentation also states that `exact` is`Only used for directed graphs, on undirected graphs the number of reachable nodes from every node can be computed in linear time.` I think this may cause confusion.

include/networkit/reachability/ReachableNodes.hpp
```diff
@@ -0,0 +1,81 @@
...
+     */
+    count numberOfReachableNodesUB(node u) const {
+        assureFinished();
+        if (G->isDirected())
+            return exact ? reachableLB[u] : reachableUB[u];
+        else
+            return reachableLB[u];
+    }
+
+    bool exact;
```
fabratu, line 70: Can you add a documentation for the member variable for doxygen/sphinx?

fabratu (approved):

PR #659, opened 2021-01-21
Update gitignore

angriman:
Ignore `include/ttmath` in `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ build/
 lib/
 lib64/
 include/tlx
+include/ttmath
 
 ### IDE
 *.project


fabratu:
Hm, since the source-files of `ttmath` are a part of the github-repo - shouldn't we keep them tracked in case we change them?

angriman:
`include/ttmath` is generated when NetworKit is installed, the actual library is stored in `extlibs/ttmath`. The same is already done for `tlx`, so it does not appear as an untracked file in `git status`.

fabratu:
Ah, you are right.

fabratu (approved):
LGTM.

PR #660, opened 2021-01-21
Fix minor warnings from GCC in C++17/20 mode

angriman:
Fixes new gcc C++17/20 warnings.

### DIFF
--- a/extlibs/ttmath/ttmath/ttmathobjects.hpp
+++ b/extlibs/ttmath/ttmath/ttmathobjects.hpp
@@ -484,7 +484,6 @@ public:
 
         if( i == table.end() )
         {
-            value.empty();
             *param = 0;
             return err_unknown_object;
         }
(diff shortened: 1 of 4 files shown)


avdgrinten (approved):
LGTM. What are the warnings that are fixed? Specifically, why does GCC complain about the `i ? y : x`? This seems to only be a stylistic issue to me (the old code expresses `i != 0` while the new code expresses `i > 0`).

The "unsigned cannot be negative" fixes make sense of course.

angriman:
Yes `i ? y : x` is just stylistic, I just find `i > 0 ? y : x` clearer to read.

avdgrinten:
What's the warning that GCC produced for the former?

angriman:
It produces `-Wunused-result` in `ttmathobjects.hpp` for:
```
value.empty();
```
And `-Wtype-limits` for the first condition in:
```
assert(levelIdx >= 0 && levelIdx < this->size());
```

avdgrinten:
So the ternary condition does not produce any warning and it was just refactored anyway?

angriman:
Yes it's a stylistic change (the only one in this PR), I just find it clearer to write it like that.

PR #661, opened 2021-01-21
ApproxElectricalCloseness: Fix minor issue with pivot lpinv column

mgoerg:
The vector that is computed as the column of the lpinv may differ by a constant from the actual column of the lpinv since L vanishes on constant vectors. This is not addressed explicitly in the part of the paper which explains how the pivot column is computed, but may affect the output centrality and diagonal entry values.

### DIFF
--- a/networkit/cpp/centrality/ApproxElectricalCloseness.cpp
+++ b/networkit/cpp/centrality/ApproxElectricalCloseness.cpp
@@ -507,6 +507,13 @@ void ApproxElectricalCloseness::run() {
             G.forNodes([&](node u) { rhs[u] = -1.0 / static_cast<double>(n); });
             rhs[root] += 1.;
             cg.solve(rhs, sol);
+
+            // ensure column sum of entries is 0.
+            double columnSum = 0.;
+            for (int i = 0; i < n; i++) {
+                columnSum += sol[i];
+            }
+            sol -= columnSum / n;
         }
 
         // All threads sample and aggregate USTs in parallel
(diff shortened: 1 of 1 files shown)


networkit/cpp/centrality/ApproxElectricalCloseness.cpp
```diff
@@ -507,6 +507,13 @@ void ApproxElectricalCloseness::run() {
             G.forNodes([&](node u) { rhs[u] = -1.0 / static_cast<double>(n); });
             rhs[root] += 1.;
             cg.solve(rhs, sol);
+
+            // ensure column sum of entries is 0.
+            double columnSum = 0.;
+            for (int i = 0; i < n; i++) {
+                columnSum += sol[i];
+            }
```
angriman, line 515: Instead of this for-loop, use
```
G.forNodes([&](node u) { columnSum += sol[u]; });
```

networkit/cpp/centrality/ApproxElectricalCloseness.cpp
```diff
@@ -507,6 +507,13 @@ void ApproxElectricalCloseness::run() {
...
+            sol -= columnSum / n;
```
angriman, line 516: There is an implicit `count` to `double` conversion here, use `columnSum / static_cast<double>(n)`

angriman (changes requested):
Thank you for the PR, I just have a couple of minor comments, see them below.

angriman (approved):
Thank you, LGTM.

PR #663, opened 2021-01-23
Parallelize ClusteredRandomGraphGenerator

angriman:
`ClusteredRandomGraphGenerator` is an O(m^2) algorithm and it is implemented sequentially.
A simple idea to parallelize it is to add all half-out edges in parallel (this is thread-safe), and then to use `GraphBuilder` to complete the work.

This already gives us some speedup, generating a graph with 15K nodes and 50 clusters took ~10s with the previous implementation and just ~130ms with the new one (on a 24 cores machine).

### DIFF
--- a/networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
+++ b/networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
@@ -3,55 +3,137 @@
  *
  *  Created on: 28.02.2014
  *      Author: cls
+ *              Eugenio Angriman <angrimae@hu-berlin.de>
  */
+
+// networkit-format
+
+#ifndef NDEBUG
+#include <algorithm>
+#endif
+
+#include <omp.h>
+#include <random>
+#include <stdexcept>
+#include <unordered_set>
 
 #include <networkit/auxiliary/Random.hpp>
 #include <networkit/generators/ClusteredRandomGraphGenerator.hpp>
-#include <networkit/structures/Partition.hpp>
+#include <networkit/graph/GraphBuilder.hpp>
 
 namespace NetworKit {
 
-ClusteredRandomGraphGenerator::ClusteredRandomGraphGenerator(count n, count k, double pin, double pout) : n(n), k(k), pin(pin), pout(pout) {
+ClusteredRandomGraphGenerator::ClusteredRandomGraphGenerator(count n, count k, double pIntra,
+                                                             double pInter)
+    : n(n), k(k), pIntra(pIntra), pInter(pInter), zeta(n) {
+
+    if (k == 0)
+        throw std::runtime_error("Error: k must be positive");
+
+    const auto isProbability = [](double p) noexcept -> bool { return (p >= 0) && (p <= 1); };
+
+    if (!isProbability(pIntra))
+        throw std::runtime_error("Error: pIntra must be in [0, 1]");
+
+    if (!isProbability(pInter))
+        throw std::runtime_error("Error: pInter must be in [0, 1]");
+
+    if (pIntra < pInter)
+        WARN("Intra-cluster probability is lower than inter-cluster probability.");
+
+    zeta.setUpperBound(k);
 }
 
 Graph ClusteredRandomGraphGenerator::generate() {
-    assert(pin >= pout);
+    GraphBuilder graphBuilder(n);
 
-    Graph G(n);
-    // assign nodes evenly to clusters
-    Partition zeta(n);
-    zeta.setUpperBound(k);
-    G.forNodes([&](node v){
-        index c = Aux::Random::integer(k-1);
-        zeta.addToSubset(c, v);
-    });
+    // Sequence of all nodes grouped by cluster. Here the vertices of each cluster are stored
+    // sequentially one after the other
+    std::vector<node> clustersSequence(n);
 
-    assert (zeta.numberOfSubsets() == k);
+    // Index in `clustersSequence` where each cluster ends. E.g., clusterEnd[i] is the index in
+    // `clustersSequence` of the first vertex in the (i + 1)-th cluster.
+    std::vector<size_t> clusterEnd(k);
 
-    G.forNodePairs([&](node u, node v){
-        if (zeta.subsetOf(u) == zeta.subsetOf(v)) {
-            if (Aux::Random::probability() <= pin) {
-                G.addEdge(u, v);
-            }
-        } else {
-            if (Aux::Random::probability() <= pout) {
-                G.addEdge(u, v);
-            }
+    {
+        // To assign vertices to clusters uniformly at random
+        std::uniform_int_distribution<index> clusterDistr{0, k - 1};
+
+        std::vector<count> clusterSizes(k);
+
+        // Assign nodes to clusters
+        graphBuilder.parallelForNodes([&zeta = zeta, &clusterDistr, &clusterSizes](node u) {
+            const index cluster = clusterDistr(Aux::Random::getURNG());
+            zeta.addToSubset(cluster, u);
+#pragma omp atomic
+            ++clusterSizes[cluster];
+        });
+
+        // Last cluster ends at n, each cluster i \in [0, ..., k - 2] ends at
+        // n - \sum_{j = i + 1}^{k - 1} cluster[j].size()
+        clusterEnd[k - 1] = n;
+        for (int64_t i = k - 2; i >= 0; --i)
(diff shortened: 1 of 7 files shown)


michitux:
There is no reason to use a quadratic algorithm, this can be done in linear time using the same skip distance sampling algorithm as the Erdős–Rényi generator. The algorithm just needs to group nodes by cluster and use different probabilities for sampling skip distances in parts of the adjacency matrix that is within clusters vs. between clusters.

angriman:
Using skip distances is a much better idea, if I understood correctly they are sampled using a geometric distribution. I updated the code accordingly.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,114 @@
...
+        } while (true);
+    };
 
-    G.shrinkToFit();
-    return G;
+#pragma omp parallel for schedule(guided)
+    for (omp_index i = 0; i < static_cast<omp_index>(n); ++i) {
+        const node u = clustersSequence[i];
+        const index clusterIdx = zeta.subsetOf(u);
+        const index start = clusterStart[clusterIdx];
```
michitux, line 99: Shouldn't start be `i`? From what I understood, otherwise, there will be twice as many intra-cluster edges added as desired to all nodes but the first node of the cluster and in particular also self-loops. Adding some documentation that documents the assumptions made here and in particular in `addHalfOutEdges` could help here (e.g., `addHalfOutEdges` seems to assume that the position of `u` is `start`, which seems wrong for inter-cluster edges).

angriman: That's true, thanks. I also added further documentation.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,122 @@
...
+        // Nodes grouped by cluster
+        std::vector<std::vector<node>> clusters(k);
+
+        // Assign nodes to clusters
+        graphBuilder.parallelForNodes([&zeta = zeta, &clusters, &clusterDistr](node u) {
+            const index cluster = clusterDistr(Aux::Random::getURNG());
+            zeta.addToSubset(cluster, u);
+#pragma omp critical
+            clusters[cluster].push_back(u);
+        });
```
avdgrinten, line 66: Entering a parallel region just to do an immediate `#pragma omp critical` is unlikely to be beneficial. You could use an array of per-cluster locks instead.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,128 @@
...
+            omp_init_lock(&locks[i]);
+
+        // Assign nodes to clusters
+        graphBuilder.parallelForNodes([&zeta = zeta, &clusters, &clusterDistr, &locks](node u) {
+            const index cluster = clusterDistr(Aux::Random::getURNG());
+            zeta.addToSubset(cluster, u);
+            omp_set_lock(&locks[cluster]);
+            clusters[cluster].push_back(u);
+            omp_unset_lock(&locks[cluster]);
+        });
```
michitux, line 72: This is basically a parallel counting sort. Two thoughts:

* What about applying a known and tested parallel counting sort algorithm?
* This most likely doesn't dominate the running time. The actual assignment (using `addToSubset` could still be parallelized as this might actually need some time to generate the random numbers. However, for simplicity I would suggest to instead implement the actual sorting sequentially without the additional `clusters` vector unless it turns out that this is actually a bottleneck.

Note that this implementation creates quite some overhead:
* We initialize and use locks, even if just a single thread is used
* We basically sort the nodes twice, once in the clusters vector and then in the final `clusterSequence` where actually just the sizes of the clusters would be needed to sort the nodes into `clusterSequence`.
* The `clusters` vector actually causes `1 + k * log(n/k)` memory allocations which could be completely avoided.

I wouldn't be surprised if sequentially this was a factor 2 to 4 slower than a simple sequential counting sort implementation and needed at least 8 cores to be as fast as a simple sequential counting sort implementation, if it becomes faster at all as randomly chosen cluster ids most likely cause heavy lock contention.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,128 @@
...
+            clusterEnd[i] = clusterEnd[i + 1] - clusters[i + 1].size();
+
+        // From a 2-D vector to a single 1-D vector.
+        for (auto &&cluster : clusters)
+            clustersSequence.insert(clustersSequence.end(), cluster.begin(), cluster.end());
+    }
+
+    // Add all the half-out edges from a node u to all the other nodes in the given sequence.
+    // To avoid adding multiple half-out edges, with respect to `clustersSequence` all the nodes
+    // in [start, end) should be subsequent to `u`.
```
michitux, line 87: This description contradicts the implementation, the implementation never adds an edge to the node at position `start`. In the usage below, this correctly avoids the self-loop to `u`, as below - in contradiction to the documentation -, `start` is not subsequent to `u`, but then it passes the end of the cluster (i.e., the next node after the cluster) as `start`, and thus the edge from `u` to the next node after the cluster is impossible as far as I can see.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,128 @@
...
+        const node u = clustersSequence[i];
+        // Index in `clustersSequence` where the cluster of node `u` ends.
+        const index end = clusterEnd[zeta.subsetOf(u)];
+
+        // Add intra-cluster half-out edges
+        addHalfOutEdges(u, i, end, pIntra);
 
-    G.shrinkToFit();
-    return G;
+        if (end < n) // Add inter-cluster half-out edges
```
michitux, line 119: This check is redundant, an equivalent check is in `addHalfOutEdges`.

networkit/cpp/generators/ClusteredRandomGraphGenerator.cpp
```diff
@@ -3,55 +3,128 @@
...
+
+#pragma omp parallel for schedule(guided)
+    for (omp_index i = 0; i < static_cast<omp_index>(n); ++i) {
+        // Current node
+        const node u = clustersSequence[i];
+        // Index in `clustersSequence` where the cluster of node `u` ends.
+        const index end = clusterEnd[zeta.subsetOf(u)];
+
+        // Add intra-cluster half-out edges
+        addHalfOutEdges(u, i, end, pIntra);
```
michitux, line 117: According to the documentation, `i` should be subsequent to `u`.

networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -75,7 +75,8 @@ TEST_F(GeneratorsGTest, testClusteredRandomGraphGenerator) {
     Partition part = gen.getCommunities();
     count nCommunities = part.getSubsetIds().size();
     EXPECT_EQ(n, G.upperNodeIdBound());
-    EXPECT_TRUE(nCommunities >= 1 && nCommunities <= c);
+    EXPECT_GE(nCommunities, 1);
+    EXPECT_LE(nCommunities, c);
```
michitux, line 79: I suggest to add a test with one or both of the probabilities set to 1 and the other to 0 and to test if the edges are generated as expected (i.e., no duplicate edges, no self-loops, test for all node pairs if they are connected iff they are intra/inter-cluster edges). Then I suggest another test, where one of the probabilities is set to 0 and the other to some value in `(0, 1)` and to test that there are no intra/inter-cluster edges.

michitux (changes requested):
According to my understanding of the code, there are still problems with nodes being incorrectly skipped.

michitux (approved):
I think this is good now. In parallel, the generated graph is no longer deterministic even for a fixed seed. I think this could be fixed by deterministically distributing the work among processors (this needs a different distribution of the edges to threads, though, for load balancing). I don't know if this is important enough to justify the effort, though. If we do not change this, it might be a good idea to note this in the documentation so users can switch to just a single thread if determinism is important.

hmeyerhenke:
Good point, @michitux ! 
@angriman , please document this non-deterministic behavior appropriately!

fabratu (approved):

michitux (approved):

PR #664, opened 2021-01-23
Deprecate Algorithm::isParallel()

angriman:
As for `toString()` in #631, `isParallel()` is implemented by only a small subset of NetworKit algorithms and by default it throws a `std::runtime_error`. Internally it is never used and I do not see a use case where someone wants to find out whether an algorithm is parallel at runtime.

### DIFF
--- a/networkit/base.pyx
+++ b/networkit/base.pyx
@@ -62,6 +62,8 @@ cdef class Algorithm:
 		bool
 			True if algorithm can run multi-threaded.
 		"""
+		from warning import warn
+		warn("isParallel() is deprecated and will be removed in upcoming releases.")
 		if self._this == NULL:
 			raise RuntimeError("Error, object not properly initialized")
 		return self._this.isParallel()
--- a/include/networkit/base/Algorithm.hpp
+++ b/include/networkit/base/Algorithm.hpp
@@ -4,6 +4,8 @@
 #include <string>
 #include <stdexcept>
 
+#include <tlx/define/deprecated.hpp>
+
 namespace NetworKit {
 
 class Algorithm {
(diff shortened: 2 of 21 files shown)


hmeyerhenke:
IIRC, this was once used to distinguish sequential from parallel runs, e.g. in experimental pipelines. Probably there are other ways to do this without losing information (such as checking the number of threads). Or maybe I am completely off here and it was used to expose if an algorithm _can_ be run in parallel or not (have not checked the source)? This could be helpful for users if it pays off at all to increase the number of threads. This may be harder to replace. Comments?

avdgrinten:
It returns the latter (whether an algorithm is parallelized or not, not necessarily if it runs in parallel). But as far as I see it, this is only useful when you are given an unknown `Algorithm` and want to know if it supports parallelism. However, in reality, you almost always know what algorithm you want to run anyway (and hence, you can look at the documentation to understand whether it is parallel or not).

hmeyerhenke:
I'd be fine with pushing this info to the documentation.

networkit/base.pyx
```diff
@@ -62,6 +62,8 @@ cdef class Algorithm:
 		bool
 			True if algorithm can run multi-threaded.
 		"""
+		from warning import warn
+		warn("isParallel() is deprecated and will no longer be supported.")
```
fabratu, line 66: Maybe rephrase this a bit? Like functionality will be removed in the upcoming release / Info about this can be found in the documentation.

fabratu (commented):
Since the agreement is to include the info explicitly/implicitly in the doc - is it already up to date (e.g. is every algorithm documented properly)?

fabratu (approved):

PR #665, opened 2021-01-26
clang tidy: check private headers

angriman:
The header filter of our `.clang-tidy` does not match private headers, so private implementations are not checked.
This PR updates our configuration to also match private implementations and fixes a warning.


avdgrinten:
`const` in the .cpp files is fine and can be kept. Other than that, this LGTM.

angriman:
Alright, I restored them.

avdgrinten (approved):

PR #668, opened 2021-01-27
Fixes missing Stemmer-module for doc-search

fabratu:
### What happens?

In the current version of the documentation build, the search-feature is broken due to a missing Javascript object (`Stemmer`).

### How to Fix

Likely an error in the bootstrap-theme. Until this is fixed, a workaround is to include `_static/language_data.js` as an additional js-file in the theme. In this way, the file is included before the js-searchlibrary. This produces a minimal overhead, since `Stemmer` is now defined on all pages, not just the search page.

### DIFF
--- a/extrafiles/sphinx-style/_templates/layout.html
+++ b/extrafiles/sphinx-style/_templates/layout.html
@@ -1,7 +1,7 @@
 {# Import the theme's layout. #}
 {% extends "!layout.html" %}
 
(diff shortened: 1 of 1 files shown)


angriman (approved):

fabratu:
Merging this early. It is only a small fix and every other PR-merge overwrites the manual workaround - leading to a non-functional search functionality on the documentation website.

PR #669, opened 2021-01-29
Remove addressed TODO in setup.py

angriman:
While reviewing @fabratu PR #667 I noticed that we already check whether the candidate compiler supports C++14 features by passing the `-std=c++14` option.


avdgrinten:
LGTM, this is a trivial PR and can be merged immediately.

avdgrinten (approved):

PR #672, opened 2021-02-03
New utility methods for (Bucket)PrioQueue

angriman:
I've been working recently with `BucketPQ` and I've noticed that it does not support basic (constant time) utility methods that can simplify the usage of this data structure. This PR adds the following methods to `PrioQueue` and `BucketPQ`:
- `empty`
- `contains`: whether or not the pq contains a given element.
- `getMin`: returns the top element without removing it (only for `BucketPQ`, `PrioQueue` already has `peekMin`).

Further, it adds tests for BucketPQ.

### DIFF
--- a/include/networkit/auxiliary/BucketPQ.hpp
+++ b/include/networkit/auxiliary/BucketPQ.hpp
@@ -8,11 +8,11 @@
 #ifndef NETWORKIT_AUXILIARY_BUCKET_PQ_HPP_
 #define NETWORKIT_AUXILIARY_BUCKET_PQ_HPP_
 
-#include <networkit/auxiliary/Log.hpp>
+#include <limits>
+#include <list>
+
 #include <networkit/Globals.hpp>
 #include <networkit/auxiliary/PrioQueue.hpp>
-#include <list>
-#include <limits>
 
 namespace Aux {
 
@@ -31,7 +31,22 @@ constexpr int64_t none = std::numeric_limits<int64_t>::max();
 class BucketPQ: public PrioQueue<int64_t, index> {
 private:
     std::vector<Bucket> buckets;			// the actual buckets
-    std::vector<Bucket::iterator> nodePtr;	// keeps track of node positions
+    static Bucket dummyBucket;
+    static const Bucket::iterator invalidPtr;
+
+    struct OptionalIterator {
+        bool valid;
+        Bucket::iterator iter;
+
+        void reset() {
+            valid = false;
+            iter = invalidPtr;
+        }
+        OptionalIterator() { reset(); }
+        OptionalIterator(bool valid, Bucket::iterator iter) : valid(valid), iter(iter) {}
+    };
+
+    std::vector<OptionalIterator> nodePtr; // keeps track of node positions
     std::vector<index> myBucket;			// keeps track of current bucket for each value
     int64_t currentMinKey;					// current min key
     int64_t currentMaxKey;					// current max key
@@ -40,7 +55,6 @@ private:
     count numElems;							// number of elements stored
     int64_t offset;							// offset from minAdmissibleKeys to 0
 
-private:
     /**
      * Constructor. Not to be used, only here for overriding.
      */
@@ -83,6 +97,11 @@ public:
     void insert(int64_t key, index value) override;
 
     /**
+     * Returns the element on top of the PrioQ.
+     */
+    std::pair<int64_t, index> getMin();
+
+    /**
      * Removes the element with minimum key and returns the key-value pair.
      */
     std::pair<int64_t, index> extractMin() override;
@@ -98,6 +117,16 @@ public:
      * @return Number of elements in PQ.
      */
     uint64_t size() const override;
+
+    /**
+     * @return Whether or not the PQ is empty.
+     */
+    bool empty() const noexcept override;
+
+    /**
+     * @return Whether or not the PQ contains the given element.
+     */
+    bool contains(const index &value) const override;
 
     /**
      * Removes key-value pair given by value @a val.
--- a/include/networkit/auxiliary/PrioQueue.hpp
+++ b/include/networkit/auxiliary/PrioQueue.hpp
@@ -95,6 +95,16 @@ public:
     virtual uint64_t size() const;
 
     /**
+     * @return Whether or not the PQ is empty.
+     */
+    virtual bool empty() const noexcept;
+
+    /**
+     * @return Whether or not the PQ contains the given value.
+     */
+    virtual bool contains(const Value &value) const;
+
+    /**
      * Removes key-value pair given by value @a val.
      */
     virtual void remove(const Value& val);
@@ -146,7 +156,7 @@ Aux::PrioQueue<Key, Value>::PrioQueue(uint64_t capacity) {
 }
 
 template<class Key, class Value>
-inline void Aux::PrioQueue<Key, Value>::insert(Key key, Value value) {
+void Aux::PrioQueue<Key, Value>::insert(Key key, Value value) {
     if (value >= mapValToKey.size()) {
         uint64_t doubledSize = 2 * mapValToKey.size();
         assert(value < doubledSize);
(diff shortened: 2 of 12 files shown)


include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -40,7 +40,9 @@ class BucketPQ: public PrioQueue<int64_t, index> {
     count numElems;							// number of elements stored
     int64_t offset;							// offset from minAdmissibleKeys to 0
 
-private:
+    Bucket dummyBucket = Bucket{};
+    const Bucket::iterator invalidPtr;
```
avdgrinten, line 44: These can probably be static constexpr (?).

include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -56,6 +58,8 @@ class BucketPQ: public PrioQueue<int64_t, index> {
      */
     void construct(uint64_t capacity);
 
+    std::pair<int64_t, index> extractMin(bool removeElement);
```
avdgrinten, line 61: Is this signature (in particular, returning a pair) in line with the other priority queues?

include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -82,6 +86,11 @@ class BucketPQ: public PrioQueue<int64_t, index> {
      */
     void insert(int64_t key, index value) override;
 
+    /**
+     * Returns the element on top of the PrioQ.
+     */
+    std::pair<int64_t, index> getMin();
```
avdgrinten, line 92: What about this signature?

include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -99,6 +108,16 @@ class BucketPQ: public PrioQueue<int64_t, index> {
...
 
+    /**
+     * @return Whether or not the PQ is empty.
+     */
+    bool empty() const noexcept override;
+
+    /**
+     * @return Whether or not the PQ contains the given element.
+     */
+    bool contains(const index &value) const override;
```
avdgrinten, line 119: Take `index` by value. `index` is a small integer type that is cheap to copy.

include/networkit/auxiliary/PrioQueue.hpp
```diff
@@ -198,6 +207,16 @@ inline uint64_t Aux::PrioQueue<Key, Value>::size() const {
     return pqset.size();
 }
 
+template<class Key, class Value>
+inline bool Aux::PrioQueue<Key, Value>::empty() const noexcept {
```
avdgrinten, line 211: `inline` is redundant for templates. Template functions are always inline.

include/networkit/auxiliary/PrioQueue.hpp
```diff
@@ -198,6 +207,16 @@ inline uint64_t Aux::PrioQueue<Key, Value>::size() const {
...
+    return pqset.empty();
+}
+
+template<class Key, class Value>
+inline bool Aux::PrioQueue<Key, Value>::contains(const Value &value) const {
```
avdgrinten, line 216: Same.

networkit/cpp/auxiliary/BucketPQ.cpp
```diff
@@ -92,8 +97,9 @@ void BucketPQ::remove(const index& value) {
     }
 }
 
-std::pair<int64_t, index> BucketPQ::extractMin() {
+std::pair<int64_t, index> BucketPQ::extractMin(bool removeElement) {
     if (size() == 0) {
+        WARN("The priority queue is empty!");
```
avdgrinten, line 102: Why does this warn?

It should either be allowed to call this on an empty queue (that's the current state), or it should be a precondition that the queue is non-empty (in which case this should be a hard error).

networkit/cpp/auxiliary/BucketPQ.cpp
```diff
@@ -92,8 +97,9 @@ void BucketPQ::remove(const index& value) {
     }
 }
 
-std::pair<int64_t, index> BucketPQ::extractMin() {
+std::pair<int64_t, index> BucketPQ::extractMin(bool removeElement) {
```
avdgrinten, line 100: This is an awkward signature. If removeElement = true, why doesn't the caller use getMin()?

avdgrinten (changes requested):

include/networkit/auxiliary/BucketPQ.hpp
angriman: `std::list` and `std::list::iterator` are not literals so I am not sure if that's possible (if I make them static constexpr I get errors).

include/networkit/auxiliary/BucketPQ.hpp
angriman: Yes, `PrioQueue::extractMin` returns a key-value pair.

include/networkit/auxiliary/BucketPQ.hpp
angriman: I added this to only get the first element without extracting it, this is consistent with `PrioQueue`.

include/networkit/auxiliary/BucketPQ.hpp
angriman: I agree but this is already defined in `PrioQueue` which takes a const ref. The same is done in other methods like `remove` because `PrioQueue` can also store object that are expensive to copy.

include/networkit/auxiliary/PrioQueue.hpp
angriman: I removed `inline` in the other methods as well.

include/networkit/auxiliary/BucketPQ.hpp
avdgrinten: What if they are static but not constexpr? Basically, we should avoid per-PQ dummy objects that waste a few bytes but do absolutely nothing.

angriman: Yes that works, now `dummyBucket` is static.

avdgrinten (approved):
LGTM now.

avdgrinten (dismissed):

fabratu (approved):
The failing CI-tests on Windows were caused by comparing iterators from different lists/vectors (msvc error: list iterators incompatible). LGTM now.

angriman:
Btw using a vector of `std::unique_ptr` to keep the pointers to the buckets makes BucketPQ not copyable.

avdgrinten (changes requested):
Wait, why is that done? Was that change already present when I reviewed?

Assuming that ::iterator is really an iterator, storing unique_ptrs to it is *very* strange!

angriman:
This is a new change, the issue here is that we cannot use the 'invalidPtr' iterator to indicate whether an element is in the BucketPQ or not. This does not work because comparing two iterators on two different lists is not allowed (at least by msvc).
The `unique_ptr` can be set to null so I used that as a workaround but I don't think that's a great solution.

An alternative solution to mark an element as not present in the BucketPQ could be to set its corresponding `nodePtr` iterator to the `end()` iterator of the element's bucket.

avdgrinten:
If anything, use an `std::optional` here, not a unique_ptr. Or use a pointer to the element instead of an iterator.

Iterators to vectors are invalidated on resize. Does our implementation take that into account?

angriman:
`std::optional` is only available from `C++17` right? To delete an element from a bucket in constant time we need an iterator to the element, I don't think that a pointer to the element would be enough.
Btw the buckets are implemented as `std::list` and are never resized.

I created a simple `OptionalIterator` struct that stores a list iterator and a boolean value to determine whether the element is in the prioq or not.

fabratu (approved):
Since `std::optional` is not available in C++ 14, this looks like a good solution with not much overhead.

avdgrinten (approved):
Agreed, that solution is fine (and avoids the unnecessary heap allocation that `unique_ptr` implied).

PR #674, opened 2021-02-04
New Grow-Shrink Algorithm for Group Closeness Maximization

angriman:
This PR brings the Grow-Shrink algorithm described in #473.
The code has been refactored and the other algorithm included in #473 (Local-Swaps) has been removed to simplify the review process.

I also added another CI build with `-DNETWORKIT_NATIVE=ON` (linux and clang), which was never tested before. Not sure if we should test native builds on more compilers and OSs.


fabratu (changes requested):

avdgrinten:
We should make an effort to keep the test matrix small (21 runs is already quite a lot). Can this be merged into one of the other runs?

fabratu:
Like with building against an external build TLX-lib, we could integrate a native-arch-build into the "full build". This would cover macOS with AppleClang and Linux with min. requirements and most recent (both gcc, clang).

Maybe we can further consolidate some of the tests in another PR.

avdgrinten:
If that means that we still test the "basic" configuration that everybody gets when installing over PyPI or other package managers, that'd be a good idea.

Also, a document to describe what each CI run does (and the rationale for having each run) would be great.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+#include <omp.h>
+#include <networkit/auxiliary/Random.hpp>
+
+#include "GroupClosenessGrowShrinkImpl.hpp"
+
+namespace NetworKit {
+namespace GroupClosenessGrowShrinkDetails {
+template <class Weight>
+GroupClosenessGrowShrinkImpl<Weight>::GroupClosenessGrowShrinkImpl(const Graph &G,
```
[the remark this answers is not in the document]
angriman: In this case most of the code is the same for both the weighted/unweighted implementations so I think that using templates is preferable here. Having two separate implementations would create too much duplicate code.
Anyway I agree that `WeightType` is a better name for the template parameter.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.cpp
```diff
@@ -0,0 +1,531 @@
...
+                                                                   std::vector<node> initGroup,
+                                                                   bool extended, count insertions,
+                                                                   count maxIterations)
+    : G(&G), group(std::move(initGroup)), extended(extended),
+      insertions(insertions == 0 ? computeConsecutiveInsertions(G, group.size()) : insertions),
+      maxIterations(maxIterations), heap(CompareDistance(distance)),
+      heap_(CompareDistance(distance_)) {
+
+    if (G.isDirected())
```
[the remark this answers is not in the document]
angriman: I think it's ok, the code starts at the 5th column. Anyway the file is checked with clang-format, wrong indentation would trigger an error.

angriman:
Looks better now?

include/networkit/centrality/GroupClosenessGrowShrink.hpp
```diff
@@ -0,0 +1,78 @@
...
+    std::vector<node> groupMaxCloseness() const;
+
+    /**
+     * Returns the total number of iterations performed by the algorithm.
+     */
+    count numberOfIterations() const;
+
+private:
+    const Graph *G;
+    std::unique_ptr<GroupClosenessGrowShrinkDetails::GroupClosenessGrowShrinkImpl<edgeweight>>
```
fabratu, line 71: Since the template variable is now called `WeightType`, I think the implementation pointers should also be renamed to avoid confusion.

fabratu (changes requested):

include/networkit/centrality/GroupClosenessGrowShrink.hpp
angriman: Do you have a better idea? I think `weighted/unweightedImpl` is already clear enough.

fabratu: Yeah, you are probably right. It is referring to unweighted/weighted graphs in the end.

fabratu (approved):
LGTM now. 

After this PR is done and merged, I will look into consolidation of CI-tests and post a PR for discussion.

PR #704, opened 2021-02-10
Adds a new Python module: csbridge

fabratu:
This PR adds a wrapper/bridge-module for getting a graph visualization widget based on ipycytoscape[1]. With this it is possible to draw colored graphs in Jupyter Notebooks without the need for other language-based tools like graph-viz or gephi. This can be useful in environments, where only access to the webbrowser is possible. Example image (colored karate-graph with harmonic closeness scores):

[1] https://ipycytoscape.readthedocs.io/en/latest/index.html

### DIFF
--- a/networkit/csbridge.py
+++ b/networkit/csbridge.py
@@ -17,7 +17,7 @@ except ImportError:
 else:
 	have_seaborn = True
 
-def widget_from_graph(G, node_scores = None, node_partition = None, node_colors = None, show_ids = True):
+def widget_from_graph(G, node_scores = None, node_partition = None, node_palette = None, show_ids = True):
 	""" 
 	Creates a ipycytoscape-widget from a given graph. The returned widget already contains
 	all nodes and edges from the graph. The graph is colored using an array of norm. rgb-values
@@ -36,7 +36,7 @@ def widget_from_graph(G, node_scores = None, node_partition = None, node_colors 
 	node_partition : networkit.structures.Partition
 		Partition object. This is used for color-calculation of the nodes (discrete distribution). 
 		Provide either node_scores or node_partition - not both. 	
-	node_colors : list of tuples
+	node_palette : list of tuples
 		Array consisting of normalized rgb-values. If none is given, seaborn.color_palette.colors is used
 	show_ids : 	boolean	
 		Set whether node ids should be visible in plot-widget. Is set to True by default. 
@@ -45,34 +45,41 @@ def widget_from_graph(G, node_scores = None, node_partition = None, node_colors 
 	if not have_cyto:
 		raise MissingDependencyError("ipycytoscape")
 
-	if node_scores is not None and node_partition is not None:
-		raise InputError("Provide either node_scores or node_partition - not both")
+	if node_scores is not None:
+		if node_partition is not None:
+			raise Exception("Provide either node_scores or node_partition - not both.")
+		if len(node_scores) != G.upperNodeIdBound():
+			raise Exception("node_scores should include scores for every node.")
+
+	# Set color palettes (maybe overwritten below)
+	if node_palette is not None:
+		palette = node_palette
+	else:
+		if not have_seaborn:
+			raise MissingDependencyError("seaborn")
+		palette = seaborn.color_palette("rocket_r", as_cmap=True).colors
 
 	# Color calculation: score = continuous distribution, partition = discrete distribution
 	hc_colors = []
 
 	# Partition
 	if node_partition is not None:
-		if node_colors is not None:
-			palette = node_colors
-		else:
-			if not have_seaborn:
-				raise MissingDependencyError("seaborn")
+		if node_palette is None:
 			palette = seaborn.color_palette("hls", node_partition.numberOfSubsets())
+		else:
+			if len(node_palette) < node_partition.numberOfSubsets():
+				raise Exception("Number of partitions higher than number of colors in provided palette. Provide node_palette with enough colors.")
 
 		partitions = node_partition.getVector()
+
+		if len(palette) < node_partition.numberOfSubsets():
+			raise Exception("Number of partitions to high for default coloring. Provide node_palette with enough colors.")
 
 		for i in G.iterNodes():
 			hc_colors.append((palette[partitions[i]][0] * 255, palette[partitions[i]][1] * 255, palette[partitions[i]][2] * 255))
 
 	# Score
-	elif node_scores is not None and len(node_scores) == G.numberOfNodes():
-		if node_colors is not None:
-			palette = node_colors
-		else:
-			if not have_seaborn:
-				raise MissingDependencyError("seaborn")
-			palette = seaborn.color_palette("rocket_r", as_cmap=True).colors
+	elif node_scores is not None:
 
 		minhc = min(node_scores)
 		maxhc = max(node_scores)
@@ -89,14 +96,16 @@ def widget_from_graph(G, node_scores = None, node_partition = None, node_colors 
 		if abs(maxhc - minhc) > 0:
 			for score in node_scores:
 				hc_colors.append(get_rgb(minhc, maxhc, score))
-		else: 
+		else:
+			color = palette[int((len(palette) -1) / 2)];
 			for i in G.iterNodes():
-				hc_colors.append((247, 137, 97))
+				hc_colors.append((color[0] * 255, color[1] * 255, color[2] * 255))
 
 	# No node values
 	else:
+		color = palette[0];
 		for i in G.iterNodes():
-			hc_colors.append((247, 137, 97))
+			hc_colors.append((color[0] * 255, color[1] * 255, color[2] * 255))
 	
 	# Set styling
 	if show_ids:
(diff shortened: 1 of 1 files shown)


networkit/__init__.py
```diff
@@ -89,6 +89,11 @@
 except ImportError:
 	raise MissingDependencyError("viztasks")
 
+try:
+	from . import csbridge
+except ImportError:
+	raise MissingDependencyError("csbridge")
```
avdgrinten, line 95: `csbridge` should not be available by default. (I consider it an anti-pattern that the existing modules are all available by default, but that cannot be changed without breaking lots of user code.)

fabratu: Ok, removed the import.

networkit/csbridge.py
```diff
@@ -0,0 +1,104 @@
...
+	have_seaborn = False
+else:
+	have_seaborn = True
+
+def widget_from_graph(G, node_scores = None, node_colors = None, show_ids = True):
+	""" Creates a ipycytoscape-widget from a given graph. The returned widget already contains
+		all nodes and edges from the graph. Optionally the graph can be colored using a seaborn
+		perceptually uniform color map (rocket) or a user given custom color map. See matplotlib
+		color maps for the correct formatting: 
+		https://matplotlib.org/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap
```
angriman, line 24: Indentation seems wrong here.

fabratu: Thanks!

As a side-node: I copied the indentation from other .py-files inside the project. Seems that there some different formats in place in respect to new-lines, indendation, etc. Now adapted the same styling as in the pyx-files. The doc string in these files were refactored when the documentation changed with the new sphinx-version.

fabratu:
Added some opt. for the insertion of Node and Edge-subwidgets into the ipycytoscape-object.

avdgrinten (commented):
The indentation is still broken.

For the interface, I think `node_scores` is nice for scalar scores but I'd like another option to display communities. Maybe a second argument `node_partition` that results in a "discrete" (and not continuous) color scheme by default?

fabratu:
Thanks. I will use another editor then.

Based on existing tutorial notebooks, `node_scores` is sufficient for visualization of communities via `getVector()`. Or do you recommend we should support partitions directly?

```
# 1. Create an instance of the Parallel Louvain Method
plm = nk.community.PLM(G_cluster)

# 2. Run the algorithm
plm.run()

# 3. Retrieve the resulting communities
communities = plm.getPartition().getVector()

# 4. Create widget
myPlot = widget_from_graph(G_cluster, communities, None, False)
```

fabratu:
I added a `node_partition` option for directly supporting discrete coloring (without interpolation).

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+		Array consisting of normalized rgb-values. If none is given, seaborn.color_palette.colors is used
+	show_ids : 	boolean	
+		Set whether node ids should be visible in plot-widget. Is set to True by default. 
+	"""
+	# Sanity checks
+	if not have_cyto:
+		raise MissingDependencyError("ipycytoscape")
+
+	if node_scores is not None and node_partition is not None:
+		raise InputError("Provide either node_scores or node_partition - not both")
```
avdgrinten, line 49: But providing neither is OK?

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+		Graph (nodes, edges), which should be visualized
+	node_scores : list of numbers
+		List of scores for each nodes, for example scores from a centrality measure. This is 
+		used for color-calculation of the nodes (continuous distribution). Provide either 
+		node_scores or node_partition - not both.
+	node_partition : networkit.structures.Partition
+		Partition object. This is used for color-calculation of the nodes (discrete distribution). 
+		Provide either node_scores or node_partition - not both. 	
+	node_colors : list of tuples
+		Array consisting of normalized rgb-values. If none is given, seaborn.color_palette.colors is used
```
avdgrinten, line 40: Rename to `node_palette` or so, since colors do not map to nodes 1:1.

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+			palette = node_colors
+		else:
+			if not have_seaborn:
+				raise MissingDependencyError("seaborn")
+			palette = seaborn.color_palette("hls", node_partition.numberOfSubsets())
+
+		partitions = node_partition.getVector()
+
+		for i in G.iterNodes():
+			hc_colors.append((palette[partitions[i]][0] * 255, palette[partitions[i]][1] * 255, palette[partitions[i]][2] * 255))
```
avdgrinten, line 66: What if node_colors does not have 256 colors?

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+
+	# Score
+	elif node_scores is not None and len(node_scores) == G.numberOfNodes():
```
avdgrinten, line 69: Why do we check len(node_scores) here? It should be a hard error if this does not match. Also: `numberOfNodes()` -> `upperNodeIdBound()` here.

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+		minhc = min(node_scores)
+		maxhc = max(node_scores)
+
+		# calculate coloring of nodes
+		def get_rgb(minimum, maximum, value):
+			minimum, maximum, value = float(minimum), float(maximum), float(value)
+			ratio = int((len(palette) - 1) * (value-minimum) / (maximum - minimum) * (value-minimum) / (maximum - minimum))
+			r = int(palette[ratio][0] * 255)
+			g = int(palette[ratio][1] * 255)
+			b = int(palette[ratio][2] * 255)
```
avdgrinten, line 86: See above.

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+			return r, g, b
+
+		if abs(maxhc - minhc) > 0:
+			for score in node_scores:
+				hc_colors.append(get_rgb(minhc, maxhc, score))
+		else: 
+			for i in G.iterNodes():
+				hc_colors.append((247, 137, 97))
```
avdgrinten, line 94: Maybe pick the central color instead?

networkit/csbridge.py
```diff
@@ -0,0 +1,141 @@
...
+
+	# No node values
+	else:
+		for i in G.iterNodes():
+			hc_colors.append((247, 137, 97))
```
avdgrinten, line 99: Maybe pick the first color instead?

networkit/csbridge.py
fabratu: At least that was my intuition. If none is provided (last else case), a default color for all nodes is choosen. So you get a widget with uniform coloring in the end.

networkit/csbridge.py
fabratu: In line 61 the palette is initialized with the number of partitions. This does scale the number of colors in the palette. I add a check in any case.

networkit/csbridge.py
fabratu: In this case ratio is scaled already to the number of colors in the palette (line 83). So ratio is always between 0 and len(palette) - 1.

networkit/csbridge.py
fabratu: Good idea. The initial idea was to pick a "neutral" light color from the perceptually uniform spectrum from seaborn. But letting the user decide is better.

networkit/csbridge.py
fabratu: I would use the same logic as above: take the central color of either a pre-defined or user-given palette.

networkit/csbridge.py
avdgrinten: Sounds reasonable.

networkit/csbridge.py
avdgrinten: Yes, but I can supply a `node_colors` array to bypass the palette generation, right? Shouldn't we dynamically adapt this to the number of elements in `node_colors`?

networkit/csbridge.py
avdgrinten: Well, if I specify a palette but no scores, I'd probably expect the function to pick the first color from the palette.

fabratu:
Took care of your comments (or added a note).

networkit/csbridge.py
fabratu: Hm, ok - makes sense

networkit/csbridge.py
fabratu: I would say we add another check, in case a user provides a color and it does not contain enough colors to match the number of partitions. Generating more colors (especially with seaborn) is easy - manipulating and extrapolating new colors in a given set maybe not that much :(

avdgrinten: Maybe it is enough to replace to scaling (0 - 255) to whatever the number of colors in `node_colors` is? Can we easily get the number of colors in a seaborn palette?

avdgrinten (approved):
LGTM now, thank you for the PR!

PR #705, opened 2021-02-16
Add binder-support

fabratu:
This PR adds supports for binder-integration.

In the past afaik, binder supported only direct repo integration. Additional packages would then be installed via configuration like environment.yml (conda) or requirements.txt (python). The final environment then included the complete repo, which is not minimal (since only `input` and `notebooks` are needed) and the repo2docker-build process also triggered errors due to missing Cython.

With direct Dockerfile-support we can supply a custom pre-build jupyter-notebook/lab instance with pre-defined content. This version build up on the latest release (8.0) and is a derivate of the image used for the upcoming cloud-instances.

A build (same Dockerfile, just from the fork) can be seen here: https://mybinder.org/v2/gh/fabratu/networkit/binder-support?urlpath=lab/tree/notebooks

Something to consider:
Binder uses its own mechanism to determine whether an Docker image should be rebuilt or not. Normally this is triggered by a new commit on the given branch. This commit does not have to include changes in the Dockerfile itself. The rebuilt could be partial (fast: just a diff on the image) or complete (approx. 20 minutes: new image). To avoid waiting time for the user, maybe it is a good idea to push the Dockerfile not directly to the `master`-branch, but to another branch like `binder`. The final link would then be: https://mybinder.org/v2/gh/networkit/networkit/binder?urlpath=lab/tree/notebooks (link includes the branch name). In this way one could also create new branches for special Dockerfile-versions, containing other documents or workflows. For example when doing a tutorial at a conference etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM jupyter/base-notebook:latest
+
+LABEL maintainer="Fabian Brandt-Tumescheit <brandtfa@hu-berlin.de>"
+
+# Install requirements
+USER root
+RUN mkdir /var/lib/apt/lists/partial && \
+    apt-get update && \
+    apt-get install -y  --no-install-recommends cmake git build-essential && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# UID 1000 (jovyan) is the default user for jupyter-images. The default group for correct permission-level is GID 100 (users)
+# Also see: https://jupyter.readthedocs.io/en/latest/community/content-community.html#what-is-a-jovyan
+USER 1000
+RUN pip install --upgrade pip 
+RUN pip install setuptools cython powerlaw sklearn seaborn pandas tabulate matplotlib networkx ipycytoscape
+
+# Create working environment
+# This has to be done as root in order to avoid access denied errors.
+USER root
+RUN mkdir -p ${HOME}/networkit
+COPY . ${HOME}/networkit/
+RUN ln -s ${HOME}/networkit/input ${HOME} && ln -s ${HOME}/networkit/notebooks ${HOME}
+RUN rm -rf ${HOME}/work
+RUN chown -R jovyan:users ${HOME}/*
+
+# Build and install networkit from current branch
+USER 1000
+RUN cd ${HOME}/networkit && python3 setup.py build_ext && pip install -e .
+
+# Configure Jupyter to enable ipycytoscape
+RUN conda install -y nodejs
+RUN jupyter labextension install @jupyter-widgets/jupyterlab-manager --no-build && jupyter labextension install jupyter-cytoscape
+RUN jupyter lab build
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
   <a href="https://github.com/networkit/networkit/actions"><img src="https://github.com/networkit/networkit/workflows/build/badge.svg"></a>
   <a href="https://badge.fury.io/py/networkit"><img src="https://badge.fury.io/py/networkit.svg"></a>
   <a href="https://coveralls.io/github/networkit/networkit?branch=master"><img src="https://coveralls.io/repos/github/networkit/networkit/badge.svg?branch=master"></a>
+  <a href="https://mybinder.org/v2/gh/networkit/networkit/master?urlpath=lab/tree/notebooks/User-Guide.ipynb"><img src="https://mybinder.org/badge_logo.svg"></a>
 </p>
 
 ## 
@@ -77,7 +78,9 @@ used for compilation.
 
 ## Usage example
 
-To get an overview and learn about NetworKit's different functions/classes, have a look at our interactive [notebooks-section][notebooks], especially the [Networkit UserGuide]. Note: To view and edit the computed output from the notebooks, it is recommended to use [Jupyter Notebook][jupyter-notebooks]. This requires the prior installation of NetworKit. You should really check that out before start working on your network analysis.
+To get an overview and learn about NetworKit's different functions/classes, have a look at our interactive [notebooks-section][notebooks], especially the [Networkit UserGuide]. Note: To view and edit the computed output from the notebooks, it is recommended to use [Jupyter Notebook][jupyter-notebooks]. This requires the prior installation of NetworKit. You should really check that out before start working on your network analysis. 
+
+We also provide a Binder-instance of our notebooks. To access this service, you can either click on the badge at the top or follow this [link][binder]. Disclaimer: Due to rebuilds of the underlying image, it can takes some time until your Binder instance is ready for usage.
 
 If you only want to see in short how NetworKit is used - the following example provides a climpse at that. Here we generate a random hyperbolic graph with 100k nodes and compute its communities with the PLM method:
 
@@ -237,3 +240,4 @@ in NetworKit, and simply using NetworKit. We ask you to cite the appropriate one
 [devguide]: https://networkit.github.io/dev-docs/DevGuide.html#devGuide
 [issues]: https://github.com/networkit/networkit/issues
 [jupyter-notebooks]: https://jupyter.org/install.html
+[binder]: https://mybinder.org/v2/gh/networkit/networkit/master?urlpath=lab/tree/notebooks


avdgrinten:
I guess we could link to a release branch instead of the master branch to avoid rebuilds.

Does Binder cache images indefinitely unless new commits appear?
Does Binder support prebuilt Docker images? We could build an image in a GitHub workflow and use GitHub's Docker registry to supply prebuilt images.

fabratu:
A release branch is also a good idea. Haven't found any info about the cache timout, but since they have their own registry (which they build the docker-images against), the cache should be rather longish.

Pre-built images: I think for Binder this is not possible, due to the scope of the project. How I understand it, you have to host [Binderhub](https://github.com/jupyterhub/binderhub) in order to provide such images.

avdgrinten:
According to https://github.com/jupyterhub/repo2docker-action it is possible to use prebuilt images on mybinder.org

avdgrinten:
Ah, the trick is that the action pushes to the official DockerHub registry. Then it uses a Dockerfile (for Binder) that solely consists of a `FROM ...` line to pull in that image.

avdgrinten:
Note that `FROM ...` can specify the source registry, so we could use GitHub's registry instead.

Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM jupyter/base-notebook:latest
```
avdgrinten, line 1: Why is the Dockerfile in the root of the repo?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN pip install seaborn
+RUN pip install networkit
+RUN pip install pandas
+RUN pip install tabulate
+RUN pip install matplotlib
+RUN pip install networkx
+RUN pip install ipycytoscape
+
+RUN git clone https://github.com/networkit/networkit.git
+WORKDIR networkit
```
avdgrinten, line 25: We should not clone the repository. If this Dockerfile is already checked out, we already have a copy of the repository. We should use that copy instead.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN git checkout 8.0
+RUN mv notebooks ${HOME}/
+RUN mv input ${HOME}/
+RUN rm ${HOME}/notebooks/test_notebooks.py
+WORKDIR ..
+RUN rm -rf networkit
+
+RUN conda install -y nodejs
```
avdgrinten, line 33: Why do we install nodejs from conda, not from apt?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN pip install matplotlib
+RUN pip install networkx
+RUN pip install ipycytoscape
+
+RUN git clone https://github.com/networkit/networkit.git
+WORKDIR networkit
+RUN git checkout 8.0
+RUN mv notebooks ${HOME}/
+RUN mv input ${HOME}/
+RUN rm ${HOME}/notebooks/test_notebooks.py
```
avdgrinten, line 29: Hm?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+WORKDIR ..
+RUN rm -rf networkit
+
+RUN conda install -y nodejs
+RUN jupyter labextension install @jupyter-widgets/jupyterlab-manager --no-build
+RUN jupyter labextension install jupyter-cytoscape
+RUN jupyter lab build
+
+USER root
+RUN fix-permissions /etc/jupyter/
```
avdgrinten, line 39: What does this do? Where is it documented? Please add a comment.

fabratu:
Ahh. That makes sense of course. In this way it be easier and faster to automate the process of dockerfile-generation via github workflow (with either github's or docker's repo).

So something like (branch to image-mapping):
master -> networkit:latest
release-X.Y -> networkit:vX.Y

Dockerfile
fabratu: Binder needs the file to be in the root-directory in order to detect it (especially if other configuraiton-files like requirements.txt are present)

Dockerfile
fabratu: Thats not how it works for Dockerfiles with Binder. If you provide configuration files in the end the repository-files are present in the jupyter-instance. With a Dockerfile it just builds the final image from this (without the remaining data).

Dockerfile
fabratu: It is based on recommendation: This is because the complete jupyter environment (the image is built upon it) is part of a conda environment. This also capsulates user permissions.

Dockerfile
fabratu: The idea is that the image-size is minimal. Like written above - the repository where the Dockerfile is hosted, is not part of the final image. The intuition is: check out our repo in the building process, but only include files you want to "use" in the end. So the user is only interested in input-graphs and notebooks. The rest is removed afterwards.

Dockerfile
avdgrinten: Well, it is expected that the repository is not automatically copied to the image. However, you can use `COPY ...` in the Dockerfile to achieve this.

fabratu: That makes sense! Thanks :)

Dockerfile
fabratu: Hard to find the source again, but it was recommended to avoid problems when building images based on jupyter-latest. In some cases you can get access denied from the underlying webserver if this is not included.

fabratu: It is a left-over from other images. Will delete it. Only used when modifying the configuration of jupyter.

fabratu:
Adressed your comments. With the `COPY` command the binder-image now includes the most recent notebooks, but a release version of networkit. Don't know if that is a problem though. Can change the link for the binder-badge to the release-8.0 branch (which we forgot to create at the last release).

avdgrinten:
I think we should copy the entire repository and build it inside the docker image.

fabratu:
Hm, this way maybe?

The image has now three different folders:
```
input
notebooks
networkit
```
The badge link opens the user guide inside the notebooks folder.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN pip install tabulate
+RUN pip install matplotlib
+RUN pip install networkx
+RUN pip install ipycytoscape
+
+USER root
+RUN mkdir -p ${HOME}/networkit
+COPY . ${HOME}/networkit/
+RUN mv ${HOME}/networkit/input ${HOME}
+RUN mv ${HOME}/networkit/notebooks ${HOME}
```
avdgrinten, line 27: Maybe symlink instead of moving? This would preserve the repository contents.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+USER 1000
+RUN pip install --upgrade pip setuptools cython
+RUN pip install powerlaw
+RUN pip install sklearn
+RUN pip install seaborn
+RUN pip install pandas
+RUN pip install tabulate
+RUN pip install matplotlib
+RUN pip install networkx
+RUN pip install ipycytoscape
```
avdgrinten, line 21: Docker generates an intermediate image after each COPY or RUN command. Hence, you can avoid a bit of overhead if you merge all of these into a single RUN. Use \ at the end of the line to do a multi-line RUN statement.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+USER root
+RUN mkdir -p ${HOME}/networkit
+COPY . ${HOME}/networkit/
+RUN mv ${HOME}/networkit/input ${HOME}
+RUN mv ${HOME}/networkit/notebooks ${HOME}
+RUN rm ${HOME}/notebooks/test_notebooks.py
```
avdgrinten, line 28: I still think that this is ugly. If we think that this file is distracting, we should move it to a different directory of the repo.

Dockerfile
avdgrinten: Apply this to the RUN statements below as well (e.g., by using && to merge several together).

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN chown -R 1000:100 ${HOME}/networkit
+RUN chown -R 1000:100 ${HOME}/input
+RUN chown -R 1000:100 ${HOME}/notebooks
```
avdgrinten, line 31: The gid looks off.

avdgrinten (commented):
Much better now; there are still some technical comments but nothing critical.

Dockerfile
fabratu: This was somewhat intended. It is a trade-off: More commands create smaller overlays, so if anything changes the updates in the image are faster. But since the image is rather static (apart from triggered rebuilds), this makes sense.

Dockerfile
fabratu: Moved it now instead of deleting. The reasoning for removing the file from the folder is to have focus on the first-time experience (which is likely the target-group for binder-notebooks).

Dockerfile
fabratu: The gid is correct (`jovyan:user` from Jupyter-base images). I exchanged for names instead of Ids to avoid confusion.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN pip install --upgrade pip 
+RUN pip install setuptools cython powerlaw sklearn seaborn pandas tabulate matplotlib networkx ipycytoscape
+
+# Create working environment
+# This has to be done as root in order to avoid access denied errors.
+USER root
+RUN mkdir -p ${HOME}/networkit
+COPY . ${HOME}/networkit/
+RUN ln -s ${HOME}/networkit/input ${HOME} && ln -s ${HOME}/networkit/notebooks ${HOME}
+RUN mv ${HOME}/notebooks/test_notebooks.py ${HOME}/networkit && rm -rf ${HOME}/work
```
avdgrinten, line 23: I am not happy with the move either. If we think that the file is distracting, this argument also applies to users who download the repo manually. Hence, in that case, the file should be moved to a different directory altogether (`scripts/`), in a different PR.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN chown -R jovyan:users ${HOME}/*
```
avdgrinten, line 24: Add a comment that this is the default user (UID 1000) within the Jupyter image; maybe link to https://jupyter.readthedocs.io/en/latest/community/content-community.html#what-is-a-jovyan

README.md
```diff
@@ -3,6 +3,7 @@
   <a href="https://github.com/networkit/networkit/actions"><img src="https://github.com/networkit/networkit/workflows/build/badge.svg"></a>
   <a href="https://badge.fury.io/py/networkit"><img src="https://badge.fury.io/py/networkit.svg"></a>
   <a href="https://coveralls.io/github/networkit/networkit?branch=master"><img src="https://coveralls.io/repos/github/networkit/networkit/badge.svg?branch=master"></a>
+  <a href="https://mybinder.org/v2/gh/networkit/networkit/master?urlpath=lab/tree/notebooks"><img src="https://mybinder.org/badge_logo.svg"></a>
```
avdgrinten, line 6: I think that this needs a disclaimer that the image might take a while to build.

Dockerfile
fabratu: Agreed - removed the removal.

avdgrinten (dismissed):
LGTM now.

fabratu:
Thanks for the review (again). 

Sry for the dismiss. Missed the changes to the README in the last push.

avdgrinten (approved):

fabratu:
Merging this earlier, since there is a low interaction with the rest of the code-base and it is useful for the upcoming release.

PR #710, opened 2021-02-17
Follow-up fixes for neighbors being swapped with the end

michitux:
As neighbors are swapped with the last neighbor when they are removed, removing edges from within `forNeighborsOf()` is no longer safe but we also do not need to check for `none` there anymore. As `removeNode()` removes incident edges, this is also not necessary anymore in `AdamicAdarDistance`. I have briefly looked for other uses of `removeEdge()`, but on first glance they seem to be safe.

### DIFF
--- a/include/networkit/distance/AdamicAdarDistance.hpp
+++ b/include/networkit/distance/AdamicAdarDistance.hpp
@@ -22,8 +22,6 @@ class AdamicAdarDistance final : public NodeDistance {
 private:
     std::vector<double> aaDistance; //result vector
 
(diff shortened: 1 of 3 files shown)


angriman (approved):

fabratu:
> As neighbors are swapped with the last neighbor when they are removed, [...]

Can you point to the file/line where to find this?

angriman:
https://github.com/networkit/networkit/blob/master/networkit/cpp/graph/Graph.cpp#L647

PR #711, opened 2021-02-19
Github actions: raise minimum supported clang version to 3.9

fabratu:
Since min. version of clang is raised from 3.8 to 3.9 - CI tests have to be adjusted. This is a copy from work done in #516 by @angriman


angriman (approved):

fabratu:
Merging this early, since it is only a small change and a depending PR is needed for an upcoming release.

PR #720, opened 2021-03-03
New Algorithm: Normalized PageRank

Relux-the-Relux:
Implements the normalized PageRank algorithm as request in #685 .


angriman:
Since the implementation is almost exactly the same as the original `PageRank`, I think that we should implement the normalization directly into `PageRank`. For example, we could add a public `normalized` boolean variable into `PageRank`, and normalize the scores only if it is set to `true`.

hmeyerhenke:
I agree with Eugenio, the default option should be not to normalize.

avdgrinten:
Please fix up the commit structure using `git rebase -i` and condense fixes into previous commits (but do not squash everything into a single commit: e.g., having the tests in an own commit can make sense). In the future, you may want to use `git commit --fixup` + `git rebase --autosquash` to fix earlier commits.

networkit/cpp/centrality/PageRank.cpp
```diff
@@ -71,17 +71,39 @@ void PageRank::run() {
...
 double PageRank::maximum() {
-    return 1.0; // upper bound, could be tighter by assuming e.g. a star graph with n nodes
+    if (!normalized)
+        return 1.0; // upper bound, could be tighter by assuming e.g. a star graph with n nodes
+
+    auto max = scoreData[0]; // Unlike regular Page Rank there isn't a universal upper bound for
+                             // normalized page rank. So we need to iterate.
+    G.balancedParallelForNodes([&](const node u) {
+        if (max < scoreData[u])
+            max = scoreData[u];
```
michitux, line 104: This is not a true maximum calculation due to race conditions, i.e., if there are two threads that find a larger value at the same time, there is no guarantee that the larger of the two will be stored. There is [Aux::Parallel::atomic_max()](https://github.com/networkit/networkit/blob/master/include/networkit/auxiliary/Parallel.hpp) in NetworKit that allows computing the maximum in a thread-safe way. I also think there is no need to use the `balanced` iteration as the computation for all nodes should be identical. Further, I would suggest to move this calculation into the run method and store the maximum in a member variable.

hmeyerhenke:
Please follow @michitux's recommendations.

networkit/cpp/centrality/PageRank.cpp
```diff
@@ -71,17 +72,42 @@ void PageRank::run() {
...
+                max, scoreData[u]); // Unlike regular Page Rank there isn't a universal upper bound
+                                    // for normalized page rank. So we need to iterate.
+        });
+    }
     hasRun = true;
 }
 
 double PageRank::maximum() {
-    return 1.0; // upper bound, could be tighter by assuming e.g. a star graph with n nodes
+    return max;
```
angriman, line 110: Since no synchronization is needed here, I recommend to replace this with `max.load(std::memory_order_relaxed)`.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -584,6 +627,48 @@ TEST_F(CentralityGTest, testPageRankCentrality) {
...
+    G.addEdge(5, 6, 3);
+    G.addEdge(5, 7, 2);
+
+    double damp = 0.85;
+    PageRank centrality(G, damp, 1e-8, true);
+    centrality.run();
+    std::vector<double> cen = centrality.scores();
+
+    // compare to Matlab results
+    const double tol = 1;
```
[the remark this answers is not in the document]
Relux-the-Relux: It is because PageRank uses absolute error tolerance instead of relative. Since the test uses an undirected Graph the values for Normalized are incredibly high, as there are no dangling nodes in a connected undirected graph. So in terms of relative error tolerance, the error here is about equal to the one in the non-normalized PageRank test.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -501,6 +501,49 @@ TEST_P(CentralityGTest, testPageRank) {
...
+        pr.norm = norm;
+        pr.run();
+
+        auto pr_ranking = pr.ranking();
+        constexpr double eps = 1;
+        if(norm == PageRank::Norm::L1Norm) {
+            if (G.isDirected()) {
+                constexpr double eps = 1e-2;
+            EXPECT_EQ(pr_ranking[0].first, 326);
+            EXPECT_NEAR(pr_ranking[0].second, 133.1727, eps);
```
angriman, line 519: Indentation here is wrong. Also, I don't think that using `constexpr` here is helpful, just use `const` (the same applies in other parts of this function).

angriman: Also, avoid to re-declare `eps`. Can't we test normalized and non-normalized results with the same precision?

networkit/cpp/centrality/test/CentralityGTest.cpp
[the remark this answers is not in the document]
Relux-the-Relux: That would be a much higher accuracy than the PageRank Test.

Because GTest's EXPECT_NEAR uses absolute tolerance. While regular PageRank goes from 0 to 1 normalized page rank goes from 1 to an upper limit that is solely graph dependant. In this case, the results are all over 40 million so if the result is 40 000 000.5 or 40 000 000.6 doesn't make that much difference. while the same test in Pagerank are 0.0565 or higher and use 1e-4(not 1e-6 or lower) is a much bigger fraction of the result than 1 is for the results of normalized Pagerank, so currently the tests for normalized PageRank are actually more precise than the ones for regular Page Rank

P.S. keep in mind that does not refer to the tolerance within the PageRank algorithm that still uses the default 1e-8 when calculating the PageRank since the normalization is the last step.

networkit/cpp/centrality/test/CentralityGTest.cpp
Relux-the-Relux: We can't test both tests with the same precision because of the difference in the magnitude of results and EXPECT_NEAR uses an absolute difference between the results.

networkit/cpp/centrality/test/CentralityGTest.cpp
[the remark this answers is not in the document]
angriman: My point is: if we have Matlab results that are precise up to `1e-4` why do we test with `eps = 1`? If our implementation is correct I would expect it to return (nearly) the same results as Matlab. Do we get different results compared to Matlab if we test with `eps = 1e-4`?

networkit/cpp/centrality/PageRank.cpp
```diff
@@ -71,17 +72,39 @@ void PageRank::run() {
...
+        G.parallelForNodes([&](const node u) { scoreData[u] /= sum; });
+    } else {
+        // calculate sum of the Pagerank of dangling Nodes for normalization
+        const auto sum = G.parallelSumForNodes([&](const node u) {
+            if (G.degree(u) == 0.0)
+                return scoreData[u];
+            return 0.0;
+        });
+
+        const auto normFactor = (1.0 / n) * (tol + ((1.0 - tol) * sum));
```
fabratu, line 89: I think the usage of `tol` is wrong here. According to the paper, `r_low`is computed with the teleport probabilty (or `1-damp`). This also could explain the high values for normalized page rank.

fabratu (changes requested):

networkit/cpp/centrality/PageRank.cpp
Relux-the-Relux: You're right. Eps there is not the tolerance but 1-damp.

fabratu: Better, but still not correct. Admitted - terms are bit confusing. If you look at the rest of the implementation of PageRank and compare this with the paper, you see that `damp` (from our code) = `1 - eps` (even though it is called damping factor in the paper). You can see this for example when looking at the computation of the final teleport prob.

fabratu:
With the computation of norm. PageRank now fixed - the tests and tolerances look much better now. 

Can you squash the PR into 2-3 commits? One for the implementation, one for the tests and opt. one for changes to the previous implementation. Also please use proper capitilization in commit messages. It makes them more readable.

fabratu (dismissed):

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -501,6 +501,34 @@ TEST_P(CentralityGTest, testPageRank) {
...
+        auto pr_ranking = pr.ranking();
+        const double tol = 1e-4;
+        if (G.isDirected()) {
+            EXPECT_EQ(pr_ranking[0].first, 326);
+            EXPECT_NEAR(pr_ranking[0].second, 57.6657, tol);
+        } else {
+            EXPECT_EQ(pr_ranking[0].first, 699);
+            EXPECT_NEAR(pr_ranking[0].second, 205.7324, tol);
+        }
+        constexpr count maxIterations = 2;
```
angriman, line 522: Use `const`

angriman (approved):

networkit/centrality.pyx
```diff
@@ -1735,7 +1735,7 @@ class Norm(object):
 cdef extern from "<networkit/centrality/PageRank.hpp>":
 
 	cdef cppclass _PageRank "NetworKit::PageRank" (_Centrality):
-		_PageRank(_Graph, double damp, double tol) except +
+		_PageRank(_Graph, double damp, double tol, normalized) except +
```
michitux, line 1738: Why is there no type specified for `normalized`? Is this not necessary?

angriman (approved):

Issue #594, opened 2020-09-08
Make Partition.forEntries take a reference as argument


rionda:
https://github.com/networkit/networkit/blob/02de60d064527410c0acf8e1c8cb0d403c5c1409/include/networkit/structures/Partition.hpp#L295

This function, and the similar `parallelForEntries` should likely take a reference `Callback &` to give more flexibility to the caller.

fabratu:
Thanks for the suggestion @rionda. Can you maybe elaborate more (or give examples) why this would be beneficial?

avdgrinten:
In particular, this might introduce another indirection in a critical code path if the compile decides not to inline this function.

rionda:
Having a reference would allow the caller to save and retrieve state/results from the `Callback` object.

avdgrinten:
Sure but is that really important? The callback can just store a reference / pointer to some external data structure. I don't think the cost of the reference outweights the benefits.

rionda:
Ok, whatever, I thought it would be more useful that way =)

fabratu:
Closing this for now. If there is more / new input, we may open this again in the future.

PR #722, opened 2021-03-10
making forEntries and parallelForEntries take the callback function by reference

Relux-the-Relux:
Addressing #594.


avdgrinten:
It's not clear to me that this is beneficial.

fabratu:
@Relux-the-Relux Thanks for your work! Based on the discussion in #594 however, we should drop the idea.

Issue #721, opened 2021-03-10
EdmondsKarp is missing .run() method


AndrewHannigan:
EdmondsKarp is missing the `.run()` method, but it also complains that the .run() method needs to be called before `getMaxFlow()` can be called.

```
>>> g = nk.generators.ErdosRenyiGenerator(nNodes=1000, prob=0.2, directed=True).generate()
>>> gw = nk.graphtools.toWeighted(g)
>>> ek = nk.flow.EdmondsKarp(graph=gw, source=100_000_000, sink=200_000_000)
>>> ek.run()

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/IPython/core/interactiveshell.py", line 3427, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-20-9ec6715874fc>", line 1, in <module>
    ek.run()
AttributeError: 'networkit.flow.EdmondsKarp' object has no attribute 'run'
```

Calling getMaxFlow without run first yields SIGSEGV
```
>>> ek.getMaxFlow()

terminate called after throwing an instance of 'std::runtime_error'
  what():  Error, run must be called first
Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

angriman:
@AndrewHannigan we merged a fix for this issue, thanks for pointing this out.

PR #723, opened 2021-03-10
Make CI fail on failed python test, fix failing python tests

angriman:
The python interface of `EdmondsKarp` does not properly inherit from `Algorithm`, and thus the `run` method is not visible (#721).
The CI has system has been fixed so that it fails when a python test fails.
Further fixes in `test_graphio.py`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,16 +29,16 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.9
+      - name: Checkout networkit
+        uses: actions/checkout@v2
+        with:
+          submodules: true
       - name: Checkout tlx
         if: matrix.variant == 'full build'
         uses: actions/checkout@v2
         with: 
           repository: tlx/tlx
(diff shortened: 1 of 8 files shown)


fabratu:
Why this is not triggered is currently investigated. One problem is, that "fast failing" is not active by default (even though the documentation of Github Actions) states it. Will post a PR soon.

michitux:
@angriman I think you need to remove the `__dealloc__`-method, too, or at least it is superfluous, depending on the order in which they are executed it might lead to a double free.

angriman:
@michitux thanks, that's probably what causes the segfault.

avdgrinten:
This PR contains a bunch of unrelated CI changes; these need to be extracted to a separate PR.

angriman:
The CI changes have been included so that the fixes that this PR applies are actually tested. Our CI system had a bug that made the CI succeed even it some python tests failed, I updated the description of the PR to clarify this.

fabratu (approved):

Issue #633, opened 2020-12-10
Infinite loop when samples are too big


darabos:
If you run PivotMDS or ApproxCloseness with a sample size that is greater than the size of the graph, the sampling code will run in an infinite loop.

- https://github.com/networkit/networkit/blob/7.1/networkit/cpp/viz/PivotMDS.cpp#L105
- https://github.com/networkit/networkit/blob/7.1/networkit/cpp/centrality/ApproxCloseness.cpp#L68

I don't think that's great, but I'm not sure what would be the right fix.

- EstimateBetweenness will just sample the same node multiple times. (https://github.com/networkit/networkit/blob/7.1/networkit/cpp/centrality/EstimateBetweenness.cpp#L33)
- Another option is to throw an exception.
- We could just use a smaller sample.

I think the smaller sample is the most user-friendly option. Maybe I could add a utility function for picking a sample and use that everywhere? It could switch to randomly _discarding_ nodes when the requested sample size is more than half the node count. (So when the requested sample size is greater than the number of nodes this would just finish immediately and return all nodes.)

I know that a sample should generally be way smaller than the whole graph. But I think a reasonable use case is to use a fixed sample size that you expect gives good general performance, like 100. Your code processes a bunch of graphs. One day it hits a graph that has less than 100 nodes and 💥  fails.

avdgrinten:
Thank you for the report. Yes, running into an infinite loop is certainly undesirable. A pragmatic fix would clamp the sample size to the graph size. In the case of closeness, that should yield an exact result. I am not sure what happens in the case of PivotMDS. As you mentioned, other algorithms do not care about drawing multiple identical samples, so they should be fine.

Relux-the-Relux:
Solved by the merge of #724

PR #724, opened 2021-03-12
Fix infinite loop in PivotMDS and approxCloseness when sample size is too large.

Relux-the-Relux:
addressing #633.

### DIFF
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -53,6 +53,15 @@ edgeweight maxWeightedInDegree(const Graph &G);
 node randomNode(const Graph &G);
 
 /**
+ * Returns n distinct random nodes of the input graph.
+ *
+ * @param G The input graph.
+ * @param n The number of desired nodes.
+ * @return A vector of n disntinct random nodes.
+ */
+std::vector<node> randomNodes(const Graph &G, count n);
+
+/**
  * Returns a random neighbor of node @a u. Returns none if degree is zero.
  *
  * @param G The input graph.
--- a/networkit/cpp/centrality/ApproxCloseness.cpp
+++ b/networkit/cpp/centrality/ApproxCloseness.cpp
@@ -68,43 +68,7 @@ void ApproxCloseness::run() {
 }
 
 void ApproxCloseness::estimateClosenessForUndirectedGraph() {
-    std::vector<node> sampledNodes;
-    // sample nodes
-    std::vector<bool> alreadySampled(G.upperNodeIdBound(), false);
-    
-    if (nSamples >= G.numberOfNodes()){
-        for (const auto sample: G.nodeRange()) {
-            sampledNodes.push_back(sample);
-            alreadySampled[sample] = true;
-        }
-    } else if (nSamples >  G.numberOfNodes()/2) {//in order to minimize the calls to randomNode
-                                                 //we randomize the ones that aren't pivot
-                                                 //if the are more samples than non-samples
-        std::fill(alreadySampled.begin(), alreadySampled.end(), true);
-        
-        for (count i = 0; i < G.numberOfNodes() - nSamples; ++i) { // we have to sample distinct nodes
-            node v = GraphTools::randomNode(G);
-            while (!alreadySampled[v]) {
-                v = GraphTools::randomNode(G);
-            }
-            alreadySampled[v] = false;
-        }
-        
-        for (const auto sample: G.nodeRange()) {
-            if (alreadySampled[sample]) {
-                sampledNodes.push_back(sample);
-            }
-        }
-    } else {
-        for (count i = 0; i < nSamples; ++i) { // we have to sample distinct nodes
-            node v = GraphTools::randomNode(G);
-            while (alreadySampled[v]) {
-                v = GraphTools::randomNode(G);
-            }
-            sampledNodes.push_back(v);
-            alreadySampled[v] = true;
-        }
-    }
+    std::vector<node> sampledNodes = GraphTools::randomNodes(G, nSamples);
 
     LCSum = std::vector<double>(G.upperNodeIdBound());
     LCNum = std::vector<count>(G.upperNodeIdBound());
--- a/networkit/cpp/centrality/ApproxCloseness.cpp
+++ b/networkit/cpp/centrality/ApproxCloseness.cpp
@@ -20,10 +20,11 @@ struct ListEntry {
 };
 
 ApproxCloseness::ApproxCloseness(const Graph& G, count nSamples, double epsilon, bool normalized, CLOSENESS_TYPE type) : Centrality(G, normalized), nSamples(nSamples), epsilon(epsilon), type(type) {
-    assert(nSamples > 0 && nSamples <= G.numberOfNodes() && epsilon >= 0);
+    assert(nSamples > 0 && epsilon >= 0);
 }
 
 void ApproxCloseness::run() {
+    if (nSamples > G.numberOfNodes()) nSamples = G.numberOfNodes();
     if (G.isDirected()) {
         switch (type) {
             case OUTBOUND:
--- a/networkit/cpp/centrality/ApproxCloseness.cpp
+++ b/networkit/cpp/centrality/ApproxCloseness.cpp
@@ -24,7 +24,10 @@ ApproxCloseness::ApproxCloseness(const Graph& G, count nSamples, double epsilon,
 }
 
 void ApproxCloseness::run() {
-    if (nSamples > G.numberOfNodes()) nSamples = G.numberOfNodes();
+    if (nSamples > G.numberOfNodes()) {
+        WARN("Number of samples higher than the number of nodes. Setting number of samples to number of nodes");
+        nSamples = G.numberOfNodes();
+    }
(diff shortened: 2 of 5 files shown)


avdgrinten:
Is this change correct? Does it preserve the approximation guarantee?

Relux-the-Relux:
> Is this change correct? Does it preserve the approximation guarantee?

Yes, since we're changing at the very start of the algorithm. It means that PivotMDS will simply act exactly as the Classical MDS (which is what Pivot tries to approximate) and approxCentrality will simply calculate the exact centrality.

angriman:
Note that both algorithms try to sample a fixed number of distinct pivots/samples among `n` nodes, setting this number to `n` makes the algorithm very inefficient and for no good reason.
I recommend to refactor the code so that it minimizes the number of calls to `GraphTools::randomNode` (which should be zero if the number of pivots/sample is `n`).

networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -67,13 +68,37 @@ void ApproxCloseness::estimateClosenessForUndirectedGraph() {
...
-        node v = GraphTools::randomNode(G);
-        while (alreadySampled[v]) {
-            v = GraphTools::randomNode(G);
+    
+    if (nSamples >= G.upperNodeIdBound()){
+        for (const auto sample: G.nodeRange()) {
+            sampledNodes.push_back(sample);
+            alreadySampled[sample] = true;
+        }
+    } else if (nSamples >  G.upperNodeIdBound()/2) {//in order to minimize the calls to randomNode we randomize the ones that aren't pivot if the are more samples than non-samples
```
angriman, line 77: Avoid to write beyond the 100th column, put the comment below instead.

networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -93,7 +118,7 @@ void ApproxCloseness::estimateClosenessForUndirectedGraph() {
     for (count i = 0; i < nSamples; ++i) {
         runOnPivot(i, pivot, delta, sampledNodes);
     }
-
+
```
angriman, line 121: This change should be removed.

networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -20,10 +20,11 @@ struct ListEntry {
 };
 
 ApproxCloseness::ApproxCloseness(const Graph& G, count nSamples, double epsilon, bool normalized, CLOSENESS_TYPE type) : Centrality(G, normalized), nSamples(nSamples), epsilon(epsilon), type(type) {
-    assert(nSamples > 0 && nSamples <= G.numberOfNodes() && epsilon >= 0);
+    assert(nSamples > 0 && epsilon >= 0);
 }
 
 void ApproxCloseness::run() {
+    if (nSamples > G.numberOfNodes()) nSamples = G.numberOfNodes();
```
angriman, line 27: The assignment should go in the next line.

networkit/cpp/viz/PivotMDS.cpp
```diff
@@ -17,6 +17,7 @@ PivotMDS::PivotMDS(const Graph &graph, count dim, count numPivots)
 
 void PivotMDS::run() {
     count n = G->numberOfNodes();
+    if (n < numPivots) numPivots = n;
```
angriman, line 20: The assignment should go in the next line.

angriman: Also, here we should add a warning that we are changing `numPivots` to `n`.

networkit/cpp/viz/PivotMDS.cpp
```diff
@@ -108,14 +109,38 @@ std::vector<node> PivotMDS::computePivots() {
...
     index pivotIdx = 0;
-    while (pivotIdx < numPivots) {
-        node pivotCandidate = GraphTools::randomNode(*G);
-        if (!pivot[pivotCandidate]) {
+    if (numPivots == n){
+        for (const auto pivotCandidate: G->nodeRange()) {
             pivots[pivotIdx++] = pivotCandidate;
             pivot[pivotCandidate] = true;
         }
+    } else if (numPivots > n/2) { //in order to minimize the calls to randomNode we randomize the ones that aren't pivot if the are more pivots than non-pivots
```
angriman, line 117: As above, avoid long lines.

angriman:
Looks better now. Another thing, instead of implementing the same functionality twice, I think it would be cleaner to implement it in a `randomNodes` function in `GraphTools`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +74,45 @@ node randomNode(const Graph &G) {
     return v;
 }
 
+std::vector<node> randomNodes(const Graph &G, count n) {
```
angriman, line 77: Add an assertion that checks that `n <= G.numberOfNodes()`

include/networkit/graph/GraphTools.hpp
```diff
@@ -52,6 +52,15 @@ edgeweight maxWeightedInDegree(const Graph &G);
...
 node randomNode(const Graph &G);
 
+/**
+ * Returns n distinct random nodes of the input graph.
+ *
+ * @param G The input graph.
+ * @param n The number of desired nodes.
+ * @return A vector of n disntinct random nodes.
+ */
+std::vector<node> randomNodes(const Graph &G, count n);
```
angriman, line 62: Add some tests for this method.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +74,45 @@ node randomNode(const Graph &G) {
...
+    std::vector<node> selectedNodes;
+    std::vector<bool> alreadySelected(G.numberOfNodes(), false);
+
+    if (n >= G.numberOfNodes()) {
+        for (const auto sample : G.nodeRange()) {
+            selectedNodes.push_back(sample);
+        }
```
angriman, line 84: This can be simplified with `std::insert(selectedNodes.begin(), G.nodeRange().begin(), G.nodeRange().end())`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -74,6 +74,45 @@ node randomNode(const Graph &G) {
...
+            node v = GraphTools::randomNode(G);
+            while (!alreadySelected[v]) {
+                v = GraphTools::randomNode(G);
+            }
+            alreadySelected[v] = false;
+        }
+
+        for (const auto sample : G.nodeRange()) {
+            if (alreadySelected[sample]) {
+                selectedNodes.push_back(sample);
```
angriman, line 100: Break if `selectedNodes.size() == n`

Relux-the-Relux: Does this make sense? The for loop ends already when you inset the n-th node since that would be all nodes the node range.

Relux-the-Relux: Does this make sense? if selectedNodes were to reach `n` would mean that they would be caught by the first case, since this the for the case that there are more `n/2` samples but less than `n`, so this break condition would never be reached.

Also for loop ends already when you check the n-th node since that would be all nodes in the node range.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -222,6 +222,40 @@ TEST_P(GraphToolsGTest, testRandomNode) {
     }
 }
 
+TEST_P(GraphToolsGTest, testRandomNodes) {
+    constexpr count n = 20;
+    constexpr double p = 0.01;
+
```
angriman, line 228: Here `constexpr` is not needed, we do not need to evaluate `n` and `p` at compile time. In these cases I think it's fine to just use `const`, for more check here https://en.cppreference.com/w/cpp/language/constexpr

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -222,6 +222,40 @@ TEST_P(GraphToolsGTest, testRandomNode) {
...
+    auto G = ErdosRenyiGenerator(n, p, directed()).generate();
```
angriman, line 229: Set a random seed before generating the graph

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -222,6 +222,40 @@ TEST_P(GraphToolsGTest, testRandomNode) {
...
+        G = Graph(G, true, G.isDirected());
+    }
+
+    count sampleAll = G.numberOfNodes();
+    std::vector<node> allSample = GraphTools::randomNodes(G, sampleAll);
+    EXPECT_EQ(allSample.size(), sampleAll);
+    for (count i = 0; i < sampleAll; i++) {
+        for (count j = i + 1; j < sampleAll; j++)
+            EXPECT_NE(allSample[i], allSample[j]);
+    }
```
angriman, line 240: Just use `EXPECT_EQ(std::unordered_set<node>{allSample.begin(), allSample.end()}.size(), sampleAll);`, same for the cases below.

angriman (approved):

Issue #651, opened 2021-01-11
Definition of neighbor inclusion in GraphTools::subgraphFromNodes unclear


michitux:
When I read the documentation of the method `GraphTools::subgraphFromNodes`, I understand that enabling the neighbor inclusion options will simply include outgoing or incoming neighbors into the induced subgraphs. However, this is not the case. Instead, only edges between the originally given nodes and the neighbors are included but not edges between the neighbors. What surprised me even more is that if just outgoing neighbors shall be included, incoming edges from these neighbors to the original nodes are included, too (and vice-versa if just incoming neighbors shall be considered). Is this whole behavior intentional? If yes, the documentation should be changed to describe it. Currently it says that the method "Returns an induced subgraph of the input graph", which is certainly wrong when any of the neighbor inclusion options is true as edges between the neighbors are missing.

michitux:
I just found #373 which contained a much longer documentation of the methods that describes the current behavior which is apparently intentional. I suggest to restore the longer description from #373.

avdgrinten:
Restoring a longer description is of course a good idea.

To be honest, that functionality is so specific that it is probably better to move it to its own function.

PR #726, opened 2021-03-22
Restoring the documentation of GraphTools::subgraphFromNodes.

Relux-the-Relux:
Addressing the unclarity in the function bought up in #651 . By bringing the documentation more in line with what was originally in #374 .

### DIFF
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -195,6 +195,19 @@ Graph copyNodes(const Graph &G);
 /**
  * Returns an induced subgraph of this graph (including potential edge weights/directions)
  *
+ * The subgraph contains all nodes in Nodes and all edges which have one end point
+ * in Nodes and the other in Nodes.
+ *
+ * @param G The input graph.
+ * @param nodes Nodes of the induced subgraph.
+ *
+ * @return Induced subgraph.
+ */
+Graph subgraphFromNodes(const Graph &G, const std::unordered_set<node> &nodes);
+
+/**
+ * Returns an induced subgraph of this graph (including potential edge weights/directions)
+ *
  * There a two relevant sets of nodes:
  *  - Nodes are such passed as arguments.
  *  - Neighbors are empty by default.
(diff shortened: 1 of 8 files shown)


Relux-the-Relux:
Also should we maybe have a function subgraphFromNodesAndNeighbours for what michitux had expectedd?

hmeyerhenke:
I would expect this method to return the induced subgraph of the nodes specified as sequence in the parameter list.

michitux:
@Relux-the-Relux I have no use case for such a method. If this is necessary, I think the user should simply include the neighbors in the set of nodes.

Relux-the-Relux:
@manpen do you still need the neighbor functionality of this function? we're planning to move it to its own separate function.

Relux-the-Relux:
@manpen do you still need the neighbor functionality? we're going to move it to its own function.

manpen:
Dear @Relux-the-Relux, sorry for my delayed response --- I'm currently without a computer of my own and hence cannot look into the details; if I recall correctly, the functionality was used by internal code that was never merged. Thus, feel free to change it (if it does not break NetworKit internals). If necessary I can always fix my code later on.

fabratu (changes requested):
- No need to re-implement the function, just call `subgraphFromNodesAndEdgesToTheirNeighbors` with everything set to false.
- Also the name of is not optimal. It may imply, that the input-set should already contain the neighbors. Maybe also not ideal, but how about `subgraphAndNeighborsFromNodes`?
- Add a second implementation of `subgraphFromNodes` with all parameters but without default values and deprecate it. The same should be done for Python.

michitux:
@Relux-the-Relux Sorry for breaking the formatting and there seems to be an include for `<array>` missing. Further, clang seems to be unhappy with my array-loops, probably one should just initialize the array as a proper variable. Do you want to fix that or shall I provide a fix? I don't mind fixing it, I just want to avoid that we both work on the same things.

Relux-the-Relux:
@michitux If you could apply the fix that'd be great

michitux:
I have (hopefully) fixed the issues, rebased on current master, renamed the new method as suggested and added a new, deprecated compatibility wrapper. Note that in Python overloading is not possible, therefore there is just a single method with the old parameters first and the new `compact` parameter last. This is a bit unfortunate, but named parameters make it easily possible to just use the new compact parameter.

fabratu (approved):
LGTM now.

Issue #725, opened 2021-03-15
BUG: EdgeListReader Error Segmentation fault (core dumped)


XikunHuang:
**Describe the bug**
Use EdgeListReader to read tab-separated graph file and get Segmentation fault (core dumped) error. 
I encoutered this error in networkit **v7.1, v8.0 and v8.1**. And v7.0 is fine.   

**Environment**
 - OS: Ubuntu18.0.4
 - Networkit Version: 7.1, 8.0, 8.1
 - Python Version: 3.9.2

**To Reproduce**
1. Toy graph file
```
1    2
3    4
5    6
5    8
6    10
```

2. Source code 
```
import networkit as nk

filepath = "./input/toy_graph"
net_reader = nk.graphio.EdgeListReader(separator='\t', firstNode=1, continuous=False, directed=True)
net = net_reader.read(filepath)
print(net.numberOfNodes(), net.numberOfEdges())
```
3. Full error output
Segmentation fault (core dumped)

I also tried the method in [How to read an edge list file?](https://lists.ira.uni-karlsruhe.de/pipermail/networkit/2015-April/000359.html), and it does not work. Any suggestions? 
Thanks a lot.
Xikun

angriman:
Hi Xikun, if you set `continuous=False` then `EdgeListReader` will count the number of unique node ids by creating a map from the node ids in the edge list file to `[0, 1, ..., n - 1]`. Thus, in your case, you just have to remove `firstNode=1`.
Setting `firstNode=1` decreases by 1 the node id of every node in the graph, and this is not meant to be used with `continuous=False`.

XikunHuang:
Hi, angriman, thanks for your quick reply and it really helps.
But simply remove ```firstNode=1``` will raise another error under networkit version 8.0. Based on your comment, I tried 3 ways to read an edge list file with ```continuous=False```, and two of them work. Below are the results.

**Case 1: remove firstNode=1**
```
net_reader = nk.graphio.EdgeListReader(separator='\t', continuous=False, directed=True)
net = net_reader.read(filepath)
print(net.numberOfNodes(), net.numberOfEdges())
for item in net.iterNodes():
   print(item)
```
Error:
```
Traceback (most recent call last):
  File "analysis.py", line 41, in <module>
    net_reader = nk.graphio.EdgeListReader(separator='\t', continuous=False, directed=True)
  File "networkit/graphio.pyx", line 279, in networkit.graphio.EdgeListReader.__cinit__
[...]

PR #727, opened 2021-03-22
Segmentation fault fix on EdgeListReader in the case of firstNode > 0 and continuous = false

Relux-the-Relux:
to especially tell that firstNode should be 0 if continuous = false to address the confusion in #725 .

### DIFF
--- a/networkit/cpp/io/EdgeListReader.cpp
+++ b/networkit/cpp/io/EdgeListReader.cpp
@@ -20,7 +20,13 @@ namespace NetworKit {
 EdgeListReader::EdgeListReader(char separator, node firstNode, const std::string &commentPrefix,
                                bool continuous, bool directed)
     : separator(separator), commentPrefix(commentPrefix), firstNode(firstNode),
-      continuous(continuous), mapNodeIds(), directed(directed) {}
+      continuous(continuous), mapNodeIds(), directed(directed) {
+    if (!continuous && firstNode != 0) {
+        // firstNode not being 0 in the continuous = false case leads to a segmentation fault
+        WARN("firstNode set to 0 since continuous is false");
+        firstNode = 0;
+    }
+}
 
 const std::map<std::string, node> &EdgeListReader::getNodeMap() const {
     if (this->continuous)
--- a/networkit/graphio.pyx
+++ b/networkit/graphio.pyx
@@ -259,7 +259,7 @@ cdef class EdgeListReader(GraphReader):
 	and can be accessed using getNodeMap().
 
 	To shift continuous integer node labels which are not zero-indexed, set firstNode to
-	the smallest id used in the file.
+	the smallest id used in the file. firstNode will be ignored in the non-continuous case.
 
 	The file may also include line comments which start with the commentPrefix.
 


michitux:
I think the goal in NetworKit should be to make it impossible to trigger a segmentation fault from Python. Instead of specifying this implementation detail in the documentation such that every possibly user needs to read and understand the problem and what it means, I would instead suggest to implement a simple check in the code that sets `firstNode` to 0 if `continuous == false`.

networkit/cpp/io/EdgeListReader.cpp
```diff
@@ -20,7 +20,12 @@ namespace NetworKit {
 EdgeListReader::EdgeListReader(char separator, node firstNode, const std::string &commentPrefix,
                                bool continuous, bool directed)
     : separator(separator), commentPrefix(commentPrefix), firstNode(firstNode),
-      continuous(continuous), mapNodeIds(), directed(directed) {}
+      continuous(continuous), mapNodeIds(), directed(directed) {
+    if (!continuous && firstNode != 0) {
+        WARN("firstNode set to 0 since continuous is true");
```
michitux, line 25: This should be `false`, not `true`.
```suggestion
        WARN("firstNode set to 0 since continuous is false");
```

networkit/graphio.pyx
```diff
@@ -259,7 +259,9 @@ cdef class EdgeListReader(GraphReader):
 	and can be accessed using getNodeMap().
 
 	To shift continuous integer node labels which are not zero-indexed, set firstNode to
-	the smallest id used in the file.
+	the smallest id used in the file. As a value for firstNode different from 0 is not 
+    allowed in the non-continuous case firstNode will be set to 0 regardless of input if
+    continuous is false.
```
michitux, line 264: I would instead write that the value of `firstNode` will be ignored in the non-continuous case, as, again, it is an implementation detail that we set it to 0 instead of not using it.

michitux (changes requested):
Apart from the two comments this looks good to me.

networkit/cpp/io/EdgeListReader.cpp
```diff
@@ -20,7 +20,13 @@ namespace NetworKit {
 EdgeListReader::EdgeListReader(char separator, node firstNode, const std::string &commentPrefix,
                                bool continuous, bool directed)
     : separator(separator), commentPrefix(commentPrefix), firstNode(firstNode),
-      continuous(continuous), mapNodeIds(), directed(directed) {}
+      continuous(continuous), mapNodeIds(), directed(directed) {
+    if (!continuous && firstNode != 0) {
+        WARN("firstNode set to 0 since continuous is false");
+        firstNode =
+            0; // firstNode not being 0 in the continuous = false case leads to a segmentation fault
```
angriman, line 27: Just a minor thing, this line looks odd, please write the comment in another line

networkit/graphio.pyx
```diff
@@ -259,8 +259,8 @@ cdef class EdgeListReader(GraphReader):
 	and can be accessed using getNodeMap().
 
 	To shift continuous integer node labels which are not zero-indexed, set firstNode to
-	the smallest id used in the file.
-
+	the smallest id used in the file. firstNode will be ignored in the non-continuous case.
+
```
angriman, line 263: Avoid trailing white spaces / tabs.

angriman:
I think that now it looks good, what do you think @michitux?

michitux (approved):

PR #729, opened 2021-03-28
New Algorithm: Overlapping Normalized Mutual Information

jonasspinner:
This implements the Overlapping Normalized Mutual Information from [NMI13]. I'm happy to receive some feedback.

There are some points which might need to be changed.

1. Documentation. Most of the algorithm is implemented in separate functions. Some of the doc comments might be unnecessary. The inclusion of LaTeX formulas could be too much and the formatting could also be wrong.
2. Naming conventions. The public methods conform to camelCase, but the hidden functions and variable names are in snake_case. The only attribute is called `m_normalization`. The string parameters for normalization are also in snake_case.
3. Normalization (`normalize`). The current implementation allows the specification of the method used to normalize the mutual information. The authors use "max". My impression is that it is the most useful choice. It could be better to remove the ability to choose the normalization method.
4. Numerical error handling (`clamp_below`, `clamp_above`). To make sure that the value actually is in [0, 1], it is clamped. Discrepancies less than `Aux::NumericTools::acceptableError` are silently corrected, otherwise they are corrected and an error is logged with `ERRORF`. I have only seen errors within the acceptable error range, but this might not be true for all inputs.
5. Only looking at intersections (`conditional_entropy`). I(X;Y) is defined as a sum of minima of H\*(X_i|Y_j) over all cluster pairs (X_i, Y_j) between the covers. The [reference implementation](https://github.com/aaronmcdaid/Overlapping-NMI) from [NMI13] and my code only calculate H\*(X_i|Y_j) if X_i and Y_j intersect. In the case that there exists a X_i that does not intersect with any Y_j, the entropy of X_i is used. This can lead to different results. A fix for that case should be possible as described in the comments of `conditional_entropy`, but I'm not sure if it is worth the additional complexity. A simple fix would be to iterate over all pairs of clusters, but this would make the algorithm very inefficient for covers with a lot of clusters, O(|X||Y|) instead of O(number of intersecting pairs of clusters).

[NMI13]
McDaid, Aaron F., Derek Greene, and Neil Hurley. "Normalized Mutual Information to Evaluate Overlapping
Community Finding Algorithms." ArXiv:1110.2515 [Physics], August 2, 2013. http://arxiv.org/abs/1110.2515.

### DIFF
--- a/include/networkit/auxiliary/HashUtils.hpp
+++ b/include/networkit/auxiliary/HashUtils.hpp
@@ -1,0 +1,29 @@
+#ifndef NETWORKIT_AUXILIARY_HASH_UTILS_HPP_
+#define NETWORKIT_AUXILIARY_HASH_UTILS_HPP_
+// networkit-format
+
+#include <utility>
+
+namespace Aux {
+
+template <typename T>
+void hashCombine(std::size_t &seed, T const &v) {
+    // The Boost Software License, Version 1.0 applies to this function.
+    // See https://www.boost.org/LICENSE_1_0.txt
+    // https://www.boost.org/doc/libs/1_75_0/doc/html/hash/reference.html#boost.hash_combine
+    seed ^= std::hash<T>()(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
+
+struct PairHash {
+    template <typename A, typename B>
+    std::size_t operator()(const std::pair<A, B> &pair) const {
+        std::size_t seed = 0;
+        hashCombine(seed, pair.first);
+        hashCombine(seed, pair.second);
+        return seed;
+    }
+};
+
+} // namespace Aux
+
+#endif // NETWORKIT_AUXILIARY_HASH_UTILS_HPP_
--- a/networkit/cpp/community/OverlappingNMIDistance.cpp
+++ b/networkit/cpp/community/OverlappingNMIDistance.cpp
@@ -1,0 +1,270 @@
+// networkit-format
+
+#include <cmath>
+#include <iomanip>
+#include <numeric>
+
+#include <networkit/auxiliary/NumericTools.hpp>
+#include <networkit/auxiliary/SignalHandling.hpp>
+#include <networkit/community/OverlappingNMIDistance.hpp>
+
+namespace NetworKit {
+
+OverlappingNMIDistance::SizesAndIntersections
+OverlappingNMIDistance::calculateClusterAndIntersectionSizes(const Graph &graph, const Cover &X,
+                                                             const Cover &Y) {
+    SizesAndIntersections result;
+    result.sizesX.resize(X.upperBound());
+    result.sizesY.resize(Y.upperBound());
+
+    // nodeRange() ignores deleted nodes.
+    for (node u : graph.nodeRange()) {
+        for (auto i : X[u]) {
+            for (auto j : Y[u]) {
+                ++result.intersectionSizes[{i, j}];
+            }
+
+            ++result.sizesX[i];
+        }
+        for (auto j : Y[u]) {
+            ++result.sizesY[j];
+        }
+    }
+
+    return result;
+}
+
+double OverlappingNMIDistance::h(count w, count n) {
+    auto v = static_cast<double>(w);
+    auto value = v > 0.0 ? -v * std::log2(v / n) : 0.0;
+    assert(!std::isnan(value));
+    return value;
+}
+
+double OverlappingNMIDistance::entropy(count size, count n) {
+    assert(0 <= size && size <= n);
+    auto value = h(size, n) + h(n - size, n);
+    assert(value >= 0.0);
+    return value;
+}
+
+double OverlappingNMIDistance::entropy(const std::vector<count> &sizesX, count n) {
+    double entropyX = 0.0;
+    for (auto sizeXi : sizesX) {
+        if (sizeXi == 0)
+            continue;
+        entropyX += entropy(sizeXi, n);
+    }
+    assert(!std::isnan(entropyX));
+    assert(entropyX >= 0.0);
+    return entropyX;
+}
+
+double OverlappingNMIDistance::adjustedConditionalEntropy(count sizeXi, count sizeYj,
+                                                          count intersectionSize, count n) {
+    count a = n + intersectionSize - sizeXi - sizeYj; // X_{i,u} = 0 and Y_{j,u} = 0
+    count b = sizeYj - intersectionSize;              // X_{i,u} = 0 and Y_{j,u} = 1
+    count c = sizeXi - intersectionSize;              // X_{i,u} = 1 and Y_{j,u} = 0
+    count d = intersectionSize;                       // X_{i,u} = 1 and Y_{j,u} = 1
+
+    auto entropyXiYj = h(a, n) + h(b, n) + h(c, n) + h(d, n);
+    auto entropyYj = entropy(sizeYj, n);
+    auto entropyXi = entropy(sizeXi, n);
+
+    if (h(a, n) + h(d, n) >= h(b, n) + h(c, n)) {
+        return entropyXiYj - entropyYj;
+    } else {
+        return entropyXi;
+    }
+}
+
+double OverlappingNMIDistance::conditionalEntropy(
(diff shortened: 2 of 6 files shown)


jonasspinner:
Regarding only looking at intersecting pairs of clusters. An example is (X_i, Y_j) where X_i and X_j do not intersect and have 
0 to 100 nodes. The error occurs at around (|X_i|, |Y_j|) = (80, 3) and (3, 80).

entropies
selection

The maximum error is 9.3076 at (3, 87).
errors

jonasspinner:
The non-succesful CI builds seem to fail at installing prequisites. The commits pass the CI builds with my fork. Not sure how to handle that.

fabratu:
The CI-errors happen occasionally due to updates in the underlying Linux-images (syncing of package-repositories). After a restart the tests now are successful.

avdgrinten:
This addition fits well into NetworKit. Thank you for the PR.

The code looks good in general but let's wait for a proper code review to see if any changes need to be done.

networkit/cpp/community/OverlappingNMIDistance.cpp
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+public:
+    std::size_t operator()(const std::pair<index, index> &pair) const {
+        std::size_t seed = 0;
+        hash_combine(seed, pair.first);
+        hash_combine(seed, pair.second);
+        return seed;
+    }
+};
```
angriman, line 30: This might be useful also in other parts of NetworKit, it's probably a good idea to move it to the `auxiliary` module.

networkit/cpp/community/OverlappingNMIDistance.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Calculating cluster sizes of covers and intersection sizes between two covers.
+ *
+ * @param X
+ * @param Y
+ * @return The cluster sizes and intersection sizes of X, Y.
+ */
+std::tuple<std::vector<count>, std::vector<count>,
+           std::unordered_map<std::pair<index, index>, count, pair_hash>>
```
angriman, line 40: I think that wrapping the result into a `struct` is cleaner than using a tuple.

networkit/cpp/community/OverlappingNMIDistance.cpp
```diff
@@ -0,0 +1,393 @@
...
+    if (std::isnan(nmi)) {
+        ERROR("Set nmi ", nmi, " to 0.0");
+        nmi = 0.0;
+    }
+    clamp_below(nmi, 0.0, "Set nmi lower than 0.0 to 0.0: %s");
+    clamp_above(nmi, 1.0, "Set nmi larger than 1.0 to 1.0: %s");
+
+    return nmi;
+}
+} // namespace
```
angriman, line 312: I think it's a better idea to implement these utility methods as private static members of `OverlappingNMIDistance`.
Also, please use `camelBack`.

networkit/cpp/community/OverlappingNMIDistance.cpp
jonasspinner: That seems like a good idea. I added `Aux::hashCombine` and `Aux::PairHash` in a `HashUtils.hpp` (e9590cd).

networkit/cpp/community/OverlappingNMIDistance.cpp
jonasspinner: Done (b073403)

networkit/cpp/community/OverlappingNMIDistance.cpp
jonasspinner: Done (4685253 and 18f2d4d)

angriman (approved):
Thank you, the PR looks good to me. However, I am not very familiar with this topic so I encourage other NetworKit developers/maintainers to review it.

angriman (approved):

include/networkit/auxiliary/HashUtils.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <utility>
+
+namespace Aux {
+
+template <typename T>
+void hashCombine(std::size_t &seed, T const &v) {
+    // https://www.boost.org/doc/libs/1_75_0/doc/html/hash/reference.html#boost.hash_combine
+    seed ^= std::hash<T>()(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
```
jonasspinner, line 13: This might need to be changed or include the [Boost Software License](https://www.boost.org/users/license.html).

hmeyerhenke: I'd rather avoid to include another license. It simply complicates matters and I wonder if there is no simpler other way.

jonasspinner: A simple combination of two hash values which is better than just an xor would be:
```
constexpr auto halfWidth = std::numeric_limits<size_t>::digits / 2;
auto a = std::hash<A>()(pair.first);
auto b = std::hash<B>()(pair.second);
return a ^ (b << halfWidth);
```

avdgrinten: I think it is enough to add something like `The Boost license applies to this function.` Boost's license is non-viral and compatible with the MIT license.

hmeyerhenke: If that's the case, then fine.

jonasspinner: Thank you for the input. I added a license notice to the function.

michitux (approved):
I think this looks good. I have only one minor note which is how to represent the enum in Python. Currently, it is not exported and instead a string is used. It is possible to export the enum to Python, but then again NetworKit uses string parameters in a lot of other places, too, so I think this can stay as it is. Concerning the normalization I wouldn't mind keeping the choice, in particular as `NMIDistance` currently does not use the `max` normalization. It would make sense to also implement at least the `max` normalization for `NMIDistance`.

Just as a side a note, @jonasspinner implemented this as part of the practical part of my lecture on graph partitioning and graph clustering.

jonasspinner:
That's good to hear.

By coincidence, I was just looking into how to replace it with an enum. There seem to be different variants used in Networkit like `MaxentStress` with  [cdef enum](https://github.com/networkit/networkit/blob/e6a9ffe92b926d984d5c0961c396f87463e64dde/networkit/viz.pyx#L98) and [integer member constants](https://github.com/networkit/networkit/blob/e6a9ffe92b926d984d5c0961c396f87463e64dde/networkit/viz.pyx#L160), `MultipleEdgesHandling` with a [class on module level](https://github.com/networkit/networkit/blob/a9bcf06d32ffe40fcb58878af0c62b4cf92ae966/networkit/graphio.pyx#L73), or as you have mentioned with just using strings.

I also encountered problems with using a `enum class` in the C++ class in combination with the `cdef enum` in cython, but it should be [possible](https://stackoverflow.com/questions/31001918/wrap-enum-class-for-cython).

So I would say that either an python enum `OverlappingNMIDistanceNormalization` in `community` or integer constants like `OverlappingNMIDistance.MAX` seem to be possible for me. But I am not really familiar with cython, so there might be a better way.

michitux:
I do not think integer constants are a good solution as inconsistencies between C++ and Python could easily be introduced. I implemented something similar to the solution you mentioned, see [here](https://github.com/michitux/networkit/blob/f2610311f7d948ba1889e35af5cbe570b2a62468/networkit/community.pyx#L1109-L1137) for an example (that is not yet part of NetworKit). Using strings seems to be a common pattern in Python as far as I am aware, so I do not want to say that this needs to be changed. A consistent approach in NetworKit would make sense, though.

jonasspinner:
I rebased and squashed the commits, decided for using the enum in cython and removed `nogil` annotations.

michitux:
Why did you remove the `nogil` annotations? Using `nogil` allows running code in parallel from threads in Python and is in general a good idea for methods whose running time can be long.

jonasspinner:
I misunderstood the effects/the need for `nogil`. I thought that it is only needed for parallelism in C++. I will fix that shortly. Thanks for catching that.

michitux (approved):
Thanks, I think this is good now. I still think there should be a standard how to expose enums/choices to Python but for now I would suggest to merge this.

PR #730, opened 2021-03-31
PathGrowingMatcher: throw if input graph is directed

angriman:
`PathGrowingMatcher` does not seem to support directed graphs, when launched on a directed graph one gets an error like this ```IndexError: vector<bool>::_M_range_check: __n (which is 140354493912512) >= this->size() (which is 100)```.
I think that we should check in the constructor if the graph is undirected, and this should also be documented.

### DIFF
--- a/include/networkit/matching/LocalMaxMatcher.hpp
+++ b/include/networkit/matching/LocalMaxMatcher.hpp
@@ -5,12 +5,12 @@
  *      Author: Christian Staudt
  */
 
+// networkit-format
+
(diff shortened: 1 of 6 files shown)


hmeyerhenke:
Indeed a good and necessary idea, which should be done for all matching algorithms, I think. Matching in NetworKit is likely older than the support for directed graphs.

fabratu:
@angriman Can you add also documentation to `LocalMaxMatcher` (it does not work with directed graphs)?

fabratu (dismissed):
LGTM.

fabratu (dismissed):

PR #732, opened 2021-03-31
Updating the documentation for matching.

Relux-the-Relux:
Updating the python documentation. Also includes a small fix for the c++ documentation.

### DIFF
--- a/networkit/matching.pyx
+++ b/networkit/matching.pyx
@@ -20,39 +20,144 @@ cdef class Matching:
 		return self
 
 	def match(self, node u, node v):
+		"""
+		Set two nodes u and v as each others matching partners.
+		
+		Parameters:
+		-----------
+		u : node
+		v : node
+		"""
 		self._this.match(u,v)
 
 	def unmatch(self, node u,  node v):
+		"""
+		Reset the two nodes u and v to unmatched.
+		
+		Parameters:
+		-----------
+		u : node
+		v : node
+		"""
 		self._this.unmatch(u, v)
 
 	def isMatched(self, node u):
+		"""
+		Check if node u is matched.
+		
+		Parameters:
+		-----------
+		u : node
+		
+		Returns:
+		-----------
+		bool
+			true if u has a matching partner.
+		"""
 		return self._this.isMatched(u)
 
 	def areMatched(self, node u, node v):
+		"""
+		Check if the nodes u and v are matched to each other.
+		
+		Parameters:
+		-----------
+		u : node
+		v : node
+		
+		Returns:
+		-----------
+		bool
+			true if node u and v are matched to each other.
+		"""
 		return self._this.areMatched(u,v)
 
 	def isProper(self, Graph G):
+		"""
+		Check whether this is a proper matching in the graph, i.e. there is an edge between
+		each pair and if the matching partner of v is u then the matching partner of u is v.
+		
+		Parameters:
+		-----------
+		G : networkit.Graph
+			Graph to be checked.
+		Returns:
+		-----------
+		bool
+			true if this is a proper matching.
+		"""
 		return self._this.isProper(G._this)
 
 	def size(self, Graph G):
+		"""
+		Get the number of edges in this matching.
+		
+		Parameters:
+		-----------
+		G : networkit.Graph
+		
+		Returns:
+		-----------
+		int
+			total number of edges in the matching.
+		"""
 		return self._this.size(G._this)
 
 	def mate(self, node v):
+		"""
+		Get the matched neighbor of v if it exists, otherwise +inf.
+		
+		Parameters:
+		-----------
+		v : node
+		
+		Returns:
+		-----------
+		int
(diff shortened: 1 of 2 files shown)


networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		Returns:
+		-----------
+		bool
+			true if u has a matching partner.
+		"""
 		return self._this.isMatched(u)
 
 	def areMatched(self, node u, node v):
+		"""
+		Check if node u and v are matched to each other.
```
angriman, line 61: the nodes

networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		Returns:
+		-----------
+		bool
+			true if node u and v are matched to each other.
+		"""
 		return self._this.areMatched(u,v)
 
 	def isProper(self, Graph G):
+		"""
+		Check whether this is a proper matching in the graph, i.e. there is a edge between
```
angriman, line 77: an edge

networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		each pair and if the matching partners of v is u then the matching partners of u is v.
```
angriman, line 78: matching partner (not partners)

angriman (commented):
Thanks for improving the docs, maybe check again for further grammar mistakes.

networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		Returns:
+		-----------
+		count
+			total number of edges in the matching.
+		"""
 		return self._this.size(G._this)
 
 	def mate(self, node v):
+		"""
+		Get the matched neighbor of v if it exists, otherwise none.
```
angriman, line 108: None is used in the C++ code, not sure that this holds also in Python.

Relux-the-Relux: None also comes up in the python code. (see below with edge scores).

Relux-the-Relux: What would the return value for a non-existent matching if not None?

angriman: In `Globals.hpp` none is defined as `std::numeric_limits<index>::max()`, which is not the Python `None`. I think that, if there is no matching partner, this method just returns that value, and we should point this out in the docs. However, this behavior is not ideal, I think we should change the return value to `None` instead of +inf.

Relux-the-Relux: Ok, I will change it to +inf for now. Maybe also open a new issue to change c++ to none to also be None in python?

angriman: yes let's do that

networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		-----------
+		bool
+			true if node u and v are matched to each other.
+		"""
 		return self._this.areMatched(u,v)
 
 	def isProper(self, Graph G):
+		"""
+		Check whether this is a proper matching in the graph, i.e. there is an edge between
+		each pair and if the matching partner of v is u then the matching partners of u is v.
```
angriman, line 78:
```suggestion
		each pair and if the matching partner of v is u then the matching partner of u is v.
```

networkit/matching.pyx
```diff
@@ -20,39 +20,144 @@ cdef class Matching:
...
+		"""
+		Get the number of edges in this matching.
+		
+		Parameters:
+		-----------
+		G : Graph
+		
+		Returns:
+		-----------
+		count
```
angriman, line 101: `int`? There is no `count` in python.

angriman (changes requested):
A general comment: for those functions that return `Graph` or `Partition` (or that take those types as input parameters), in the docs you should write `networkit.Graph` or `networkit.Partition`. In this way the documentation creates the links to the corresponding objects. Take a look at how this is done in other modules, like [centrality](https://github.com/networkit/networkit/blob/master/networkit/centrality.pyx).

networkit/matching.pyx
Relux-the-Relux: If you check graph.pyx for example count is used.

angriman: the same applies there, I don't think it makes sense to mention a type that does not exist.

networkit/matching.pyx
```diff
@@ -92,10 +198,17 @@ cdef class PathGrowingMatcher(Matcher):
 	"""
 	Path growing matching algorithm as described by  Hougardy and Drake.
 	Computes an approximate maximum weight matching with guarantee 1/2.
+	
+	Parameters:
+	-----------
+	G : networkit.Graph
+		Graph to apply matching.
+	edgeScores : list, optional
+		list with edgeScores if not None.
```
angriman, line 207: Remove `if not None` from here.

angriman (approved):

PR #733, opened 2021-04-03
Faster Matching::isProper

angriman:
`Matching::isProper` uses `forNodes` to check the matching of every vertex. However, we can use `nodeRange()` and return `false` as soon as we find a vertex that does not satisfy the matching property.


angriman:
The failures are due to `networkx`, @fabratu do you know how can we solve this?

fabratu (approved):

PR #734, opened 2021-04-07
Suitor matcher

angriman:
This PR brings the 1/2-approximation algorithm for maximum (weighted) matching by Manne and Halappanavar (New Effective Multithreaded Matching Algorithms, IPDPS 2014).
There are two versions of the algorithm, one of which only supports graphs where the adjacency lists of every vertex are sorted by edge weight in decreasing order. To this end I also added a new `sortEdgesByWeight` method in `Graph`, which allows to sort the edges of a graph by edge weight in increasing or decreasing order. This method uses a new generic `sortEdges` method in `Graph`, which allows to sort the edges of a graph according to a custom criterion.

I am a bit hesitant to add the new sorting methods to `Graph`, they are probably too specific. An alternative idea is to implement them in `GraphTools`.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -14,6 +14,8 @@
 
 #include <algorithm>
 #include <functional>
+#include <numeric>
+#include <omp.h>
 #include <queue>
 #include <stack>
 #include <stdexcept>
@@ -22,6 +24,7 @@
 #include <vector>
 
 #include <networkit/Globals.hpp>
+#include <networkit/auxiliary/ArrayTools.hpp>
 #include <networkit/auxiliary/FunctionTraits.hpp>
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/Random.hpp>
@@ -52,6 +55,13 @@ struct WeightedEdge : Edge {
     WeightedEdge() : Edge(), weight(std::numeric_limits<edgeweight>::max()) {}
 
     WeightedEdge(node u, node v, edgeweight w) : Edge(u, v), weight(w) {}
+};
+
+struct WeightedEdgeWithId : WeightedEdge {
+    edgeid eid;
+
+    WeightedEdgeWithId(node u, node v, edgeweight w, edgeid eid)
+        : WeightedEdge(u, v, w), eid(eid) {}
 };
 
 inline bool operator==(const Edge &e1, const Edge &e2) {
@@ -1083,6 +1093,15 @@ public:
     void sortEdges();
 
     /**
+     * Sorts the adjacency arrays by a custom criterion.
+     *
+     * @param lambda Lambda function used to sort the edges. It takes two WeightedEdge
+     * e1 and e2 as input parameters, returns true if e1 < e2, false otherwise.
+     */
+    template <class Lambda>
+    void sortEdges(Lambda lambda);
+
+    /**
      * Set edge count of the graph to edges.
      * @param edges the edge count of a graph
      */
@@ -1549,6 +1568,20 @@ public:
         if (!hasNode(u) || i >= outEdges[u].size())
             return none;
         return outEdges[u][i];
+    }
+
+    /**
+     * Return the weight to the i-th (outgoing) neighbor of @a u.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return @a weight of the i-th (outgoing) neighbor of @a u, or @c +inf if no such
+     * neighbor exists.
+     */
+    edgeweight getIthNeighborWeight(node u, index i) const {
+        if (!hasNode(u) || i >= outEdges[u].size())
+            return nullWeight;
+        return isWeighted() ? outEdgeWeights[u][i] : defaultEdgeWeight;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


avdgrinten:
Does the increase in runtime performance really justify the templated implementation? Can't you just branch on a sortSuitor variable at runtime?

Also, the sortEdgesByWeight method is not appropriate for `Graph`, as it can be implemented based on simpler primitives (i.e., the sortEdges method).

angriman:
> Does the increase in runtime performance really justify the templated implementation? Can't you just branch on a sortSuitor variable at runtime?

I removed the templated implementation, it that was not really necessary.

> Also, the sortEdgesByWeight method is not appropriate for `Graph`, as it can be implemented based on simpler primitives (i.e., the sortEdges method).

I moved it in `GraphTools` and it uses the `sortEdges` primitive.

include/networkit/graph/Graph.hpp
```diff
@@ -2162,6 +2186,57 @@ std::pair<count, count> Graph::removeAdjacentEdges(node u, Condition condition,
...
+
+        std::vector<node> sortedAdjList(deg);
+        std::vector<edgeweight> sortedWeights(deg);
+        for (index i = 0; i < deg; ++i) {
+            sortedAdjList[i] = adjList[indices[i]];
+            sortedWeights[i] = weights[indices[i]];
+        }
+
+        adjList = std::move(sortedAdjList);
+        weights = std::move(sortedWeights);
```
michitux, line 2216: Wouldn't it be possible to apply this permutation in-place using just a single temporary value? This could also be done in a utility method that is then applied to the vectors as needed.

networkit/graph.pyx
```diff
@@ -394,6 +394,15 @@ cdef class Graph:
...
 
+	def sortEdgesBy(self, object callback):
+		cdef SortEdgesLambdaWrapper* wrapper
+		try:
+			wrapper = new SortEdgesLambdaWrapper(callback)
+			with nogil:
+				self._this.sortEdges[SortEdgesLambdaWrapper](dereference(wrapper))
+		finally:
+			del wrapper
+
```
michitux, line 405: Is there any test for this method? As this may be a bit fragile I would suggest to add a test case for it.

michitux (changes requested):

include/networkit/graph/Graph.hpp
angriman: That's a good idea, I implemented this feature in `ArrayTools::applyPermutation`. Unfortunately I a have to use a bool vector to keep track of which elements have already been swapped, not sure if we can avoid that.

networkit/graph.pyx
angriman: This was a first attempt to provide `sortEdgesByWeight` to Python, which is now implemented in `nk.graphtools`. Since it is not needed anymore I think it is a better idea to keep it for C++ only.

include/networkit/graph/Graph.hpp
michitux: It is not easily possible to avoid this, the best solution seems to use a bit of the permutation to mark the swapped elements (e.g., used signed integers and change each value `x` to `- 1 - x`). Afterwards, one would need to reverse this.

angriman: This would work if the permutation does not use all the bits right? If it does, we still need to use an additional vector. I could add a condition that checks if the length of the permutation requires all the bits, and use the additional vector only if necessary.

michitux: If the permutation needs all 64 bits, this means that the graph has more than 2^63 nodes. I do not think that NetworKit intends to support such graphs in the foreseeable future.

angriman: So do we always assume that the permutation is a 64-bit integer? Since this is a general utility method it might make sense to support also shorter integers (e.g., (u)int8_t).

networkit/cpp/matching/SuitorMatcher.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    if (G.isDirected())
+        throw std::runtime_error("This algorithm does not support directed graphs.");
+
+    if (sortSuitor && checkSortedEdges && !hasEdgesSortedByWeight(G))
+        throw std::runtime_error("Edges are not sorted by weight");
+}
+
+bool SuitorMatcher::hasEdgesSortedByWeight(const Graph &G) {
+    for (node u : G.nodeRange()) {
```
michitux, line 32: This could be parallelized easily but on the other hand it might not be worth the effort.

networkit/cpp/matching/SuitorMatcher.cpp
```diff
@@ -0,0 +1,148 @@
...
+        G.forNeighborsOf(u, [&](node v, edgeweight ew) {
+            if (ew == prevWeight) {
+                if (prevNode > v)
+                    sorted = false;
+
+            } else if (ew > prevWeight)
+                sorted = false;
+
+            prevNode = v;
+            prevWeight = ew;
```
michitux, line 45: This seems like a re-implementation of [`std::is_sorted`](https://en.cppreference.com/w/cpp/algorithm/is_sorted) on `G.weightNeighborRange` with a custom comparison function.

networkit/cpp/matching/SuitorMatcher.cpp
```diff
@@ -0,0 +1,148 @@
...
+void SuitorMatcher::findSortSuitor(node current) {
+    bool done = false;
+
+    do {
+        node partner = suitor[current];
+        edgeweight heaviest = ws[current];
+
+        for (index &iter = neighborIterators[current]; iter < G->degree(current); ++iter) {
+            const node v = G->getIthNeighbor(current, iter);
+            const edgeweight weight = G->getIthNeighborWeight(current, iter);
```
michitux, line 97: Why does `neighborIterators` not actually store an iterator (e.g., a `NeighborWeightRange`)?

include/networkit/graph/Graph.hpp
```diff
@@ -1535,6 +1556,20 @@ class Graph final {
...
 
+    /**
+     * Return the weight to the i-th (outgoing) neighbor of @a u.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return @a edge weight to the i-th (outgoing) neighbor of @a u, or @c +inf if no such
+     * neighbor exists.
+     */
+    edgeweight getIthNeighborWeight(node u, index i) const {
```
michitux, line 1567: Is it intentional that this method is not added to the Python interface?

include/networkit/graph/Graph.hpp
```diff
@@ -54,6 +56,16 @@ struct WeightedEdge : Edge {
     WeightedEdge(node u, node v, edgeweight w) : Edge(u, v), weight(w) {}
 };
 
+struct WeightedEdgeWithId : WeightedEdge {
+    edgeid eid;
+
+    // Needed by cython
```
michitux, line 62: There seems to be no Cython interface for this currently, so this comment seems misleading.

michitux (approved):
In general this looks good now, I have just a few small comments that you may implement or not.

Do you have any idea how fast sorting neighbors by id with your new function is compared to the existing `sortEdges` functionality? My suspicion would be that due to increased locality it may be faster in practice, in particular when multiple cores are used. If it should be faster, I would suggest to replace the existing `sortEdges` method by your new method.

networkit/cpp/matching/SuitorMatcher.cpp
angriman: If the input graph is unweighted then `NeighborWeightRange` cannot be used because it requires an unweighted graph. To make this work we need to use two vectors with two different iterators and add some checks to pick the correct one at runtime, but I find it easier to just store the index to the current neighbor.
However, I think it would be a better idea to make `NeighborWeightRange` also work with unweighted graphs, that would make things easier.

hmeyerhenke: You mean "...because it requires a weighted graph", right?

include/networkit/graph/Graph.hpp
angriman: `getIthNeighbor` does not have a Python interface, so we either provide a Python interface for both or for none.

networkit/cpp/matching/SuitorMatcher.cpp
angriman: Yes, thanks.

include/networkit/graph/Graph.hpp
```diff
@@ -2162,6 +2197,61 @@ std::pair<count, count> Graph::removeAdjacentEdges(node u, Condition condition,
     return {removedEdges, removedSelfLoops};
 }
 
+template <class Lambda>
+void Graph::sortEdges(Lambda lambda) {
+
+    const auto sortAdjacencyArrays = [&](node u, std::vector<node> &adjList,
+                                         std::vector<edgeweight> &weights,
+                                         std::vector<edgeid> &edgeIds) -> void {
+        std::vector<index> indices(adjList.size());
```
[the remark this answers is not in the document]
fabratu: Although the code slightly changed in the most recent version, I think time can still be safed if resize is only done if `adjList.size() > indices.size()`. Reasoning: resize calls erase on indices if `adjList.size() < indices.size()`, while iota is called anyways afterwards.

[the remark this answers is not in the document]
angriman: Now resize is called only when necessary.

fabratu (approved):
LGTM.

fabratu (approved):

PR #738, opened 2021-04-16
Github Actions documentation/opt. and AVX2-detection changes

fabratu:
As discussed in #674 (and in preperation for the distribution of a wheel-package) this PR reduces the total number of tests slightly. This also changes:

- Remaining tests are renamed according to a certain scheme, which is now documented at the beginning of the `ci.yml` file. This also gives a rationale, why each test is done.
- ~~Shortening the name of each test, so that the full name is visible in the Actions-overview for a run. Hopefully the emojis are not to fancy~~
- Remove duplicating lines of code in `ci.myl` by using a "[fake ternary operator](https://github.community/t/do-expressions-support-ternary-operators-to-change-their-returned-value/18114)"
- Removing custom tests in main `CMakeLists.txt` for AVX2. `-march=native` already should enable all hardware-specific features like AVX2.
- Added custom-test for Windows/MSVC in order to detect AVX2-support. `-march=native` or similar is not available in MSVC.

Any input is appreciated :) Choosing the "right" test-cases is always depending on ones point of view.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@
 # - Latest version (gcc/Python, clang/Python): core
 #
 # 3. Windows
-# - Latest version (msvc): core, core (release)
+# - Latest version (msvc): core, core (release), core (native)
 #
 #
 # Job naming
@@ -328,7 +328,7 @@ jobs:
       VS_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\
     strategy:
       matrix:
-        build-configuration: ['core', 'core (release)']
+        build-configuration: ['core', 'core (release)', 'core (native)']
     steps:
       - name: Checkout networkit
         uses: actions/checkout@v2
@@ -339,7 +339,7 @@ jobs:
           cd "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\"
           ./vcvarsall.bat x64
           md ${{ github.workspace }}\build && cd ${{ github.workspace }}\build
-          cmake -G "Visual Studio 16 2019" -A x64 -DNETWORKIT_STATIC=ON -DNETWORKIT_BUILD_TESTS=ON ..
+          cmake -G "Visual Studio 16 2019" -A x64 -DNETWORKIT_STATIC=ON -DNETWORKIT_BUILD_TESTS=ON -DNETWORKIT_NATIVE=${{ matrix.build-configuration == 'core (native)' }} ..
           cd "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\"
           ./devenv.com $Env:GITHUB_WORKSPACE\build\networkit.sln /build ${{ matrix.build-configuration == 'core' && 'Debug' || 'Release' }}
           cd ${{ github.workspace }}\build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,94 @@
+########################################################################################################
+# Build configurations
+# ====================
+#
+# full
+# ----
+# Description: Basic build with all components (TLX, C++, Cython). Has to be tested on every fully 
+#              supported platform. Close to user-experience, except that warnings are also treated 
+#              as errors during build and TLX is provided as externally build library (normal 
+#              installation builds both NetworKit and TLX in one step).
+# Script:      scripts/full.sh
+# Build-type:  Release
+# Variants:    native = Build for native architecture
+# 
+#         
+# core
+# ----
+# Description: Builds the C++ code into one shared library. Has also sanity checks and debugging. 
+#              Since this only modifies the C++ building-process, Cython + Python-tests are removed in 
+#              order to increase speed.
+# Script:      scripts/core.sh
+# Build-type:  Debug
+# Variants:    release = Release-build (only used on Windows, since "full" is not supported)
+#              native = Build for native architecture (only used on Windows, since "full" is not supported)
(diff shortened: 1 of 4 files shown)


avdgrinten:
The Emoji are too confusing to people who don't know NetworKit's CI infrastructure.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+#              C++17 = Check for C++17 conformance
+#              C++20 = Check for C++20 conformance
+# 
+#
+# clang-tidy
+# ----------
+# Description: Uses clang-tidy for several static analysis checks. These involve bug prone programming
+#              errors, possible modernizations and performance optiomizations.
+# Script:      scripts/clang_tidy.sh
+# Config:      /.clang-tidy
```
angriman, line 44: Remove the `/`?

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+# Build-type:  Debug
+#
+# 
+# coverage
+# --------
+# Description: Enables code-coverage, leak-sanitizers and tests in order to create cpp-coveralls data.
```
angriman, line 50: Maybe specify address and leak sanitizers (address sanitizers are enabled by default if one sets `-DNETWORKIT_WITH_SANITIZERS=leak`).

angriman:
Some build names are called something like `Linux (gcc-10)` and others `Linux (llvm-11)`. However, llvm is not just a compiler, the name should be changed to something like `Linux (llvm clang-11)` or just `Linux (clang-11)`.

fabratu:
Adressed all comments.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+# Description: Builds the C++ code into one shared library. Has also sanity checks enabled.
+# Script:      scripts/core.sh
+# Build-type:  Debug
+#
+# 
+# core (variant)
+# ---------------------
+# Description: These builds are variants of core with different options. 
+# Script:      scripts/core.sh
+# Variants:    release = Release-build
```
avdgrinten, line 33: Why does `core` need a `release` variant?

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+#              non-monolithic = Each C++ module is a library
+#              C++17 = Check for C++17 conformance
+#              C++20 = Check for C++20 conformance
```
avdgrinten, line 36: I think the C++17 build can probably be dropped. I know it has existed for quite some time, but if C++20 works, there is little reason that C++17 wouldn't.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+#              native = Build for native architecture
```
avdgrinten, line 37: Why do we need a native core build if we also have a native full build?

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+# full (variant)
+# ---------------------
+# Description: These builds are variants of full with different options.
+# Script:      scripts/full.sh
+# Variants:    native = Build for native architecture
+# 
+#         
+# core
+# ----
+# Description: Builds the C++ code into one shared library. Has also sanity checks enabled.
```
avdgrinten, line 24: This description misses the rationale for `core`: the reason why we do these core builds is that they are faster than `full` builds, right?

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+# ----
+# Description: Basic build with all components (TLX, C++, Cython). Has to be tested on every fully 
+#              supported platform. Close to user-experience, except that warnings are also treated 
+#              as errors during build and TLX is provided as externally build library (normal 
+#              installation builds both NetworKit and TLX in one step).
+# Script:      scripts/full.sh
+# Build-type:  Release
+#
+# 
+# full (variant)
```
avdgrinten, line 15: Merge the variants into the description of `full`.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+# 
+#         
+# core
+# ----
+# Description: Builds the C++ code into one shared library. Has also sanity checks enabled.
+# Script:      scripts/core.sh
+# Build-type:  Debug
+#
+# 
+# core (variant)
```
avdgrinten, line 29: Merge the variants into the description of `core`.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,104 @@
...
+#              when a pull request is merged.
+# Script:      scripts/documentation.sh 
+# Build-type:  Debug
+#
+#
+# Jobs:
+# =====
+# 0. Rationale:
+# - Main platform is Linux, which is therefore used for most of the tests. Basic user experience (full) 
+#   should be tested on all platforms  with the default compiler.
```
avdgrinten, line 68: What does "on all platforms" mean here if we previously only refer to Linux?

.github/workflows/ci.yml
avdgrinten: If there is a reason, state it in the description. (I guess it is only used for Windows?)

.github/workflows/ci.yml
fabratu: Yes, main argument is heaving more speed for Debug-builds. `full` is by default a Release-build, while `core` is a Debug-build. Added this information to the description.

.github/workflows/ci.yml
fabratu: Should refer to all supported platforms. Restructured this point to make it clearer.

avdgrinten (approved):
LGTM now.

PR #739, opened 2021-04-18
New Approximation Algorithm for Group Closeness

angriman:
This PR brings the local search approximation algorithm for group closeness presented in [Group-Harmonic and Group-Closeness Maximization – Approximation and Engineering](https://epubs.siam.org/doi/10.1137/1.9781611976472.12) (ALENEX 2021).

The main benefit of the template implementation is that distances are stored as integers or floats depending on whether the graph is weighted or not (and thus we can do integer operations instead of floating point operations on unweighted graphs).

### DIFF
--- a/include/networkit/centrality/GroupClosenessLocalSearch.hpp
+++ b/include/networkit/centrality/GroupClosenessLocalSearch.hpp
@@ -32,7 +32,6 @@ public:
      * @param maxIterations Maximum number of swaps allowed. Prevents the algorithm from performing
      * too many swaps by giving up the approximation guarantee.
      */
-
     template <class InputIt>
     GroupClosenessLocalSearch(const Graph &G, InputIt first, InputIt last,
                               bool runGrowShrink = true,
--- a/include/networkit/centrality/GroupClosenessLocalSearch.hpp
+++ b/include/networkit/centrality/GroupClosenessLocalSearch.hpp
@@ -1,0 +1,80 @@
+#ifndef NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SEARCH_HPP_
+#define NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SEARCH_HPP_
+
+// networkit-format
+
+#include <limits>
+#include <memory>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class GroupClosenessLocalSearch final : public Algorithm {
+public:
+    /**
+     * Local search approximation algorithm for Group Closeness Maximization presented in
+     * "Group-Harmonic and Group-Closeness Maximization – Approximation and Engineering", Angriman
+     * et al., ALENEX 2021. The algorithm evaluates all possible swaps between a vertex in the group
+     * and the vertices outside, and performs a swap iff the decrement in farness is at least $$(1 -
+     * 1 / (k \\cdot (n - k)))$$, where $$k$$ is the number of vertices in the group. Thus,
+     * even in a best-case scenario the time complexity of this algorithm is $$O(n \\cdot k)$$. To
+     * keep the number of swaps low, it is recommended to use this algorithm as a refinement step of
+     * an already good solution computed by a faster algorithm e.g., greedy (GroupCloseness), or
+     * GrowShrink (GroupClosenessGrowShrink). In undirected graphs the approximation ratio is 1/5,
+     * on directed graphs it has not been demonstrated.
+     *
+     * @param G A graph.
+     * @param first, last A range that contains the initial group of nodes.
+     * @param runGrowShrink Whether or not to run the Grow-Shrink algorithm on the initial group.
+     * @param maxIterations Maximum number of swaps allowed. Prevents the algorithm from performing
+     * too many swaps by giving up the approximation guarantee.
+     */
+
+    template <class InputIt>
+    GroupClosenessLocalSearch(const Graph &G, InputIt first, InputIt last,
+                              bool runGrowShrink = true,
+                              count maxIterations = std::numeric_limits<count>::max())
+        : GroupClosenessLocalSearch(G, std::vector<node>(first, last), runGrowShrink,
+                                    maxIterations) {}
+
+    GroupClosenessLocalSearch(const Graph &G, const std::vector<node> &group,
+                              bool runGrowShrink = true,
+                              count maxIterations = std::numeric_limits<count>::max());
+
+    /**
+     * Runs the algorithm.
+     */
+    void run() override;
+
+    /**
+     * Returns the computed group.
+     */
+    std::vector<node> groupMaxCloseness() const;
+
+    /**
+     * Returns the number of iterations performed by the algorithm.
+     */
+    count numberOfIterations() const;
+
+    class GroupClosenessLocalSearchInterface : public Algorithm {
+    public:
+        template <class InputIt>
+        GroupClosenessLocalSearchInterface(InputIt first, InputIt last) : group(first, last) {}
+
+        std::unordered_set<node> group;
+        count nIterations;
+    };
+
+private:
+    const bool weighted;
+    // Is always one between GroupClosenessLocalSearchImpl<count/edgeweight>,
+    // see implementation.
+    std::unique_ptr<GroupClosenessLocalSearchInterface> impl;
+};
+
+} // namespace NetworKit
+
+#endif // NETWORKIT_CENTRALITY_GROUP_CLOSENESS_LOCAL_SEARCH_HPP_
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -18,6 +18,9 @@ from .graph cimport _Graph, Graph
 from .structures cimport _Cover, Cover, _Partition, Partition
 from networkit.algebraic import adjacencyEigenvector, PageRankMatrix, symmetricEigenvectors
 
+cdef extern from "limits.h":
+	cdef uint64_t ULONG_MAX
+
 cdef extern from "<networkit/centrality/Centrality.hpp>":
 
 	cdef cppclass _Centrality "NetworKit::Centrality"(_Algorithm):
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -1559,6 +1559,76 @@ cdef class GroupHarmonicCloseness(Algorithm):
 		return _GroupHarmonicCloseness.scoreOfGroup[vector[node].iterator](
 				graph._this, inputGroup.begin(), inputGroup.end())
 
+cdef extern from "<networkit/centrality/GroupClosenessLocalSearch.hpp>":
+	cdef cppclass _GroupClosenessLocalSearch "NetworKit::GroupClosenessLocalSearch"(_Algorithm):
+		_GroupClosenessLocalSearch(_Graph, const vector[node] &group, bool_t runGrowShrink, count maxIterations) except +
+		_GroupClosenessLocalSearch(_Graph, const vector[node] &group, bool_t runGrowShrink) except +
+		_GroupClosenessLocalSearch(_Graph, const vector[node] &group) except +
+		vector[node] groupMaxCloseness() except +
+		count numberOfIterations() except +
+		count maxIterations
+
+cdef class GroupClosenessLocalSearch(Algorithm):
+	cdef Graph _G
+
+	def __cinit__(self, Graph G not None, group, runGrowShrink, maxIterations):
+		"""
+		Local search approximation algorithm for Group Closeness Maximization presented in
+		"Group-Harmonic and Group-Closeness Maximization – Approximation and Engineering", Angriman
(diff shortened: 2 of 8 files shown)


angriman:
I restructured the private implementation as suggested in #638.

networkit/centrality.pyx
```diff
@@ -865,7 +868,7 @@ cdef extern from "<networkit/centrality/TopHarmonicCloseness.hpp>":
 
 
 cdef class TopHarmonicCloseness(Algorithm):
-	""" 
+	"""
```
michitux, line 871: This change seems unrelated.

networkit/centrality.pyx
```diff
@@ -1367,7 +1370,7 @@ cdef class GroupCloseness(Algorithm):
 		return (<_GroupCloseness*>(self._this)).groupMaxCloseness()
 
 	def computeFarness(self, S, H=0):
-		""" 
+		"""
```
michitux, line 1373: This change seems unrelated, too.

networkit/centrality.pyx
```diff
@@ -1498,6 +1501,65 @@ cdef class GroupClosenessLocalSwaps(Algorithm):
...
+		1 / (k \\cdot (n - k)))$$, where $$k$$ is the number of vertices in the group. Thus,
+		even in a best-case scenario the time complexity of this algorithm is $$O(n \\cdot k)$$. To
+		keep the number of swaps low, it is recommended to use this algorithm as a refinement step of
+		an already good solution computed by a faster algorithm e.g., greedy (GroupCloseness), or
+		GrowShrink (GroupClosenessGrowShrink). In undirected graphs the approximation ratio is 1/5,
+		on directed graphs it has not been demonstrated.
+
+		Parameters
+		----------
+			A connected, undirected, unweighted graph.
```
michitux, line 1529: The parameter name seems to be missing here.

include/networkit/centrality/GroupClosenessLocalSearch.hpp
```diff
@@ -32,7 +32,6 @@ class GroupClosenessLocalSearch final : public Algorithm {
      * @param maxIterations Maximum number of swaps allowed. Prevents the algorithm from performing
      * too many swaps by giving up the approximation guarantee.
      */
-
```
michitux: This change seems unrelated.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1168 @@
...
+        impl = std::make_unique<GroupClosenessLocalSearchUnweighted>(G, group, runGrowShrink,
+                                                                     maxIterations);
+}
+
+// We need to call the correct destructor
+GroupClosenessLocalSearch::~GroupClosenessLocalSearch() {
+    if (weighted)
+        delete static_cast<GroupClosenessLocalSearchWeighted *>(impl.release());
+    else
+        delete static_cast<GroupClosenessLocalSearchUnweighted *>(impl.release());
```
michitux, line 1140: This shouldn't be necessary. The unique pointer has type `Algorithm`, which has a virtual destructor and thus the destructor of the inherited classes that also have virtual descructor should be called.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1168 @@
...
+}
+
+void GroupClosenessLocalSearch::run() {
+    if (weighted)
+        static_cast<GroupClosenessLocalSearchWeighted *>(impl.get())->run();
+    else
+        static_cast<GroupClosenessLocalSearchUnweighted *>(impl.get())->run();
```
michitux, line 1147: These casts are not necessary, the `run` method is inherited from the `Algorithm` class which is the base class of both implementations.

michitux (commented):
Apart from the minor comments I wonder why the templated implementation doesn't inherit from a new base class that has the two results as members and implements the two getter methods (`groupMaxCloseness` and `numberOfIterations`). It is not even necessary to make them virtual, the base class can simply have the `group` and `nIterations` members including the two methods to access them and a constructor to set the group initially.

avdgrinten:
If we go for the base class, we can just make the result members public and don't need to bother with getters/setters (since no concrete implementation will be accessible outside of the `.cpp` file anyway).

avdgrinten:
I don't have a strong opinion on a base class vs. casting around `Algorithm` pointers; both options are better than the initial version (which had an incomplete class in the public header and the definition in a private header).

angriman:
I added a new base class, which simplifies a bit the code as we do not need to cast `Algorithm` to the implementation class anymore.

include/networkit/centrality/GroupClosenessLocalSearch.hpp
```diff
@@ -0,0 +1,81 @@
...
+                              bool runGrowShrink = true,
+                              count maxIterations = std::numeric_limits<count>::max())
+        : GroupClosenessLocalSearch(G, std::vector<node>(first, last), runGrowShrink,
+                                    maxIterations) {}
+
+    GroupClosenessLocalSearch(const Graph &G, const std::vector<node> &group,
+                              bool runGrowShrink = true,
+                              count maxIterations = std::numeric_limits<count>::max());
+
+    ~GroupClosenessLocalSearch() override = default;
```
avdgrinten, line 46: Why do we add this `= default` destructor?

include/networkit/centrality/GroupClosenessLocalSearch.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Returns the number of iterations performed by the algorithm.
+     */
+    count numberOfIterations() const;
+
+    class GroupClosenessLocalSearchInterface : public Algorithm {
+    public:
+        GroupClosenessLocalSearchInterface(const std::vector<node> &initGroup)
+            : group(initGroup.begin(), initGroup.end()) {}
```
avdgrinten, line 66: Taking iterators would be more efficient here (we're constructing a temp vector just to put it into a set?).

networkit/centrality.pyx
```diff
@@ -1498,6 +1501,66 @@ cdef class GroupClosenessLocalSwaps(Algorithm):
...
+		group : list
+			The initial group of nodes.
+		useGrowShrink : bool
+			Whether or not to run the GrowShrink algorithm on the initial group.
+		maxIterations : int
+			Maximum number of swaps allowed. Prevents the algorithm from performing
+			too many swaps by giving up the approximation guarantee.
+		"""
+		self._G = G
+		self._this = new _GroupClosenessLocalSearch(G._this, group, runGrowShrink, maxIterations if maxIterations is not None else ULONG_MAX)
```
avdgrinten, line 1540: Can we call an overload of this function without the last parameter instead of duplicating the default parameter here?

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1143 @@
...
+    void run() override;
+
+    ~GroupClosenessLocalSearchImpl() override = default;
+
+private:
+    const Graph *G;
+    const bool runGrowShrink;
+    const count maxIterations;
+
+    std::vector<WeightType> distance, distance2;
```
avdgrinten, line 57: Why do we need two of these vectors? Add a comment.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1143 @@
...
+    std::vector<WeightType> oldDistance, oldDistance2;
+    std::vector<std::vector<WeightType>> distanceGlobal, distance2Global;
```
avdgrinten, line 59: For several of these vectors, it's not clear what we store here (especially the `std::vector<std::vector<...>>` variables). Add comments.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1143 @@
...
+    std::vector<float> approxFarnessDecrease;
+    std::vector<count> sumOfMins;
+    std::vector<edgeweight> sumOfDistancesGreaterThan;
+    count diam;
+
+    count stackSize;
+    WeightType groupFarness;
+
+    template <class Type>
+    struct Less {
```
avdgrinten, line 82: Rename to a more descriptive name, for example `LessInVector` or similar. Same for `Greater`.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1143 @@
...
+        // Swaps with u did not improve the closeness enough, restore u into the group
+        group.insert(u);
+        const WeightType decreaseU = computeFarnessDecrease(u);
+        tlx::unused(decreaseU);
+#ifndef NDEBUG
+        if (G->isWeighted())
+            assert(std::abs(static_cast<double>(decreaseU - increaseU)) <= 1e-9);
+        else
+            assert(decreaseU == increaseU);
+#endif // NDEBUG
```
avdgrinten, line 1019: Wrapping these statements in `#ifndef NDEBUG` is unnecessary, isn't it? Surely the compiler will be able to optimize them away if `NDEBUG` is set? (Unless you're doing more complicated stuff than just `assert`ing something.)

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1150 @@
...
+    std::vector<std::vector<bool>> visitedGlobal;
+
+    std::vector<node> nodesAtDistance, nodesSortedByDistance;
+    std::vector<std::vector<node>> nodesLeftToExploreAtDistanceGlobal;
+    std::vector<std::pair<node, WeightType>> decreasePerThread;
+
+    std::vector<node> stack, candidatesToAddVec;
+    std::vector<float> approxFarnessDecrease;
+    std::vector<count> sumOfMins;
+    std::vector<edgeweight> sumOfDistancesGreaterThan;
```
avdgrinten, line 77: There are still lots of vectors with no description. If the name is descriptive (`nodesSortedByDistance`), that's fine, but for others (`nodesAtDistance`, `nodesLeftToExploreAtDistanceGlobal`, `oldNearest`, `oldNearest2`, etc.) it would make the code much easier to read if there were comments. (The comments that you added to some of the vectors already help a lot.)

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1150 @@
...
+            sumOfDistancesGreaterThan[i - 1] =
+                sumOfDistancesGreaterThan[i] + distance[nodesSortedByDistance[i - 1]];
+    }
+
+#ifndef NDEBUG
+    G->forNodes([&](node x) {
+        assert(nearest[x] != u);
+        assert(nearest2[x] != u);
+    });
+#endif // NDEBUG
```
avdgrinten, line 907: This is still wrapped in `NDEBUG`.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1150 @@
...
+
+    return decr;
+}
+
+// Unweighted graphs
+template <>
+void GroupClosenessLocalSearchImpl<count>::updateAfterRemoval(node u, bool updateNodesAtDistance) {
+#ifndef NDEBUG
+    G->forNodes([&](node x) { assert(nearest[x] != nearest2[x]); });
+#endif
```
avdgrinten, line 767: This is still wrapped in `NDEBUG`.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1150 @@
...
+            }
+        });
+    }
+
+#ifndef NDEBUG
+    G->forNodes([&](node x) {
+        assert(nearest[x] != u);
+        assert(nearest2[x] != u);
+    });
+#endif // NDEBUG
```
avdgrinten, line 830: This is still wrapped in NDEBUG.

networkit/cpp/centrality/GroupClosenessLocalSearch.cpp
```diff
@@ -0,0 +1,1150 @@
...
+}
+
+// Weighted graphs
+template <>
+void GroupClosenessLocalSearchImpl<edgeweight>::updateAfterRemoval(node u,
+                                                                   bool updateNodesAtDistance) {
+#ifndef NDEBUG
+    G->forNodes([&](node x) { assert(nearest[x] != nearest2[x]); });
+#endif // NDEBUG
```
avdgrinten, line 839: This is still wrapped in NDEBUG.

angriman:
Before merging this I suggest to merge #638 first so that I can integrate the `Less/GreaterInvector` comparators.

avdgrinten:
#638 is merged now.

angriman:
I updated the implementation so that it uses the comparators in `Aux`.

avdgrinten (approved):
LGTM now.

angriman:
Merging this for the new release.

PR #743, opened 2021-04-21
Undeprecate timeStep methods and treat deprecation warnings as errors

### LINKED ISSUE #742: Building NetworKit causes warnings about deprecated functionality
Building NetworKit involves tons of warnings. This is not caught by the CI due to `-Wno-error=deprecated-declarations`. Remove all use of the deprecated functionality, then remove `-Wno-error=deprecated-declarations`.

NetworKit should never make use of its own deprecated functionality. If it has to use deprecated functions because no other alternative exists, why is that functionality deprecated?

avdgrinten:
Undeprecate the `Graph::time()`, `Graph::timeStep()` and `GraphEventProxy::timeStep()` functions because:

* There is are no alternatives to these features within NetworKit.
* There are internal uses of this functionality (in `DGSReader`) and they are not deprecated. That means that we *have to* introduce a replacement for these features (or remove `DGSReader`).
* Builds cause tons of warnings due to the deprecation.

The functions were deprecated in #421. I consider it a mistake on my side that I did not spot this issue during review (in #458, I explicitly asked for a separate PR for timeStep changes). These functions can be deprecated again when a suitable replacements exists. We still warn about their use.

Furthermore, we remove `-Wno-error=deprecated-declarations` from `CMakeLists.txt` to ensure that similar situations do not occur again.

Fix #742.


angriman:
If `TLX_DEPRECATED` is not needed anymore in `Graph.hpp`, we can remove `#include <tlx/define/deprecated.hpp>`.

avdgrinten:
Good point.

In case anybody wonders how we can fix this situation: once we have general graph attributes, we can re-route this functionality to a graph attribute and re-deprecate the old ways of accessing the timestamp information.

angriman:
> If `TLX_DEPRECATED` is not needed anymore in `Graph.hpp`, we can remove `#include <tlx/define/deprecated.hpp>`.

The same also holds for `GraphEventProxy.hpp`, I think we can merge this as soon as you remove the two includes.

angriman (approved):

angriman (changes requested):
I have been too optimistic, apparently `TLX_DEPRECATED` is used in other classes where `deprecated.hpp` is not included.

avdgrinten:
CI succeeded; this can be merged now.

angriman (approved):

PR #748, opened 2021-04-29
Bug Fix: ignoreSelfLoops property of DegreeCentrality now actually ignores self loops

### LINKED ISSUE #745: Degree centrality does not actually ignore self-loops
The `DegreeCentrality`-class has a parameter `ignoreSelfLoops`. If `true`, "self-loops will not be taken into account". However, there is no code to actually not consider self-loops for node degrees. Just the maximum is set to `G.numberOfNodes() - 1`. The documentation claims that the code runs in `O(m)` time where `m` is the number of edges, which would be enough to exclude self-loops, but the current code actually runs in `O(n)` time. If it is intentional that `ignoreSelfLoops` only influences the maximum value and must only be used on graphs without self-loops, then this should be clearly documented. Further, the graph knows if self-loops exist so the normalization could be chosen automatically.

Relux-the-Relux:
The documentation now has also been updated to reflect its actual running time.

Closes #745

### DIFF
--- a/include/networkit/centrality/DegreeCentrality.hpp
+++ b/include/networkit/centrality/DegreeCentrality.hpp
@@ -21,9 +21,8 @@ class DegreeCentrality: public Centrality {
 public:
     /**
      * Constructs the DegreeCentrality class for the given Graph @a G. If the centrality scores should be normalized,
-     * then set @a normalized to <code>true</code>. The run() method runs in O(n) time, where n is the number of
-     * nodes in the graph if there are no selfloops or selfloops are not being ignored. If they are being ignored
-     * and there are self loops in the graph then it runs in O(m) where m is the number of edges.
+     * then set @a normalized to <code>true</code>. run() runs in O(n) time if ignoreSelfLoops is false or the graph 
+     * has no self-loops; otherwise it runs in O(m).
      *
      * @param G The graph.
      * @param normalized Set this parameter to <code>true</code> if scores should be normalized in the interval [0,1].
--- a/include/networkit/centrality/DegreeCentrality.hpp
+++ b/include/networkit/centrality/DegreeCentrality.hpp
@@ -21,8 +21,9 @@ class DegreeCentrality: public Centrality {
 public:
     /**
      * Constructs the DegreeCentrality class for the given Graph @a G. If the centrality scores should be normalized,
-     * then set @a normalized to <code>true</code>. The run() method runs in O(m) time, where m is the number of
-     * edges in the graph.
+     * then set @a normalized to <code>true</code>. The run() method runs in O(n) time, where n is the number of
+     * nodes in the graph if there are no selfloops or selfloops are not being ignored. If they are being ignored
+     * and there are self loops in the graph then it runs in O(m) where m is the number of edges.
      *
      * @param G The graph.
      * @param normalized Set this parameter to <code>true</code> if scores should be normalized in the interval [0,1].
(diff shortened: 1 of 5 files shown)


networkit/centrality.pyx
```diff
@@ -733,8 +733,9 @@ cdef extern from "<networkit/centrality/DegreeCentrality.hpp>":
 
 cdef class DegreeCentrality(Centrality):
 	""" Node centrality index which ranks nodes by their degree.
-	Optional normalization by maximum degree. The run() method runs in O(m) time, where m is the number of
-	edges in the graph.
+	Optional normalization by maximum degree.  The run() method runs in O(n) time, where n is the number of
+    nodes in the graph if there are no selfloops or selfloops are not being ignored. If they are being ignored
+    and there are self loops in the graph then it runs in O(m) where m is the number of edges.
```
angriman, line 738: Mixed spaces and tabs

include/networkit/centrality/DegreeCentrality.hpp
```diff
@@ -21,8 +21,9 @@ class DegreeCentrality: public Centrality {
 public:
     /**
      * Constructs the DegreeCentrality class for the given Graph @a G. If the centrality scores should be normalized,
-     * then set @a normalized to <code>true</code>. The run() method runs in O(m) time, where m is the number of
-     * edges in the graph.
+     * then set @a normalized to <code>true</code>. The run() method runs in O(n) time, where n is the number of
+     * nodes in the graph if there are no selfloops or selfloops are not being ignored. If they are being ignored
+     * and there are self loops in the graph then it runs in O(m) where m is the number of edges.
```
angriman, line 26: Maybe simpler: run() runs in O(n) time if ignoreSelfLoops is false or the graph has no self-loops; otherwise it runs in O(m).

angriman (changes requested):
This fix does not work if the graph has multiple self-loops right? Should we take this case into account?

Also, please add tests cases for degree centrality on graphs with (multiple) self-loops.

Relux-the-Relux:
> This fix does not work if the graph has multiple self-loops right? Should we take this case into account?
> 
> Also, please add tests cases for degree centrality on graphs with (multiple) self-loops.

It should work with multiple self-loops since it checks for every node if the node has a self-loop.

But will add the test of course.

angriman:
I meant if a vertex (not a graph) has multiple self-loops. In that case removing 1 is not enough.

Relux-the-Relux:
Yes, but if the graph has multi edges then the maximum degree is no longer n (or n-1 in case of no self-loops). So I don't think multi edges are to be supported by the algorithm. From my understanding is that NetworKit algorithms are not to support multi edges due to all problems they bring.

avdgrinten:
Yes, we should not support multi-edges at all. Multi-edges break the equivalence between graphs and matrices. We should probably sanitize them on input and ignore them otherwise.

avdgrinten:
Or rather, assume that they don't exist after input is done.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1953,4 +1954,125 @@ TEST_P(CentralityGTest, testGroupClosenessGrowShrink) {
...
+        EXPECT_EQ(2, dc.score(0));
+        EXPECT_EQ(1, dc.score(1));
+        EXPECT_EQ(4, dc.score(2));
+        EXPECT_EQ(2, dc.score(3));
+        EXPECT_EQ(2, dc.score(4));
+        EXPECT_EQ(5, dc.score(5));
+        EXPECT_EQ(1, dc.score(6));
+        EXPECT_EQ(1, dc.score(7));
+    }
+}
```
angriman, line 2076: Instead of writing a lot of `EXPECT_EQ` I recommend to create a `std::array` with the expected result and do the comparison in a loop (this applies to the previous tests as well).

angriman (changes requested):
In addition to the C++ tests it would be good to also have a simple Python test for `DegreeCentrality`.

angriman (approved):

PR #749, opened 2021-05-04
Small fixes: breath-update, workaround for clang-12 errors & adding includes for gcc-11

fabratu:
This PR introduces the following fixes:

- Introducing workarounds for clang-12 errors in `GraphGTest`, `GraphBuilderAutoCompleteGTest`and `GraphBuilderDirectSwapGTest` by rewriting the loops and removing `critical`-sections.
- Introducing workarounds for clang-12 error in `LocalSimilarityScore` by exchanging function call of `balancedParallelForNodes` by it's logic. 
- The pinning for breathe is removed, since the new release doesn't break anymore. 
- Slight changes to documentation code (see suggestions in https://github.com/michaeljones/breathe/releases/tag/v4.29.2). 
- Add explicit includes for `<limits>` due to changes in gcc-11 (errors already visible on CI for most recent Ubuntu-runners) (see #755)

Background to clang-12 errors:

The most recent version (12.0.0 & 12.0.1) crashes on IR generation for four occasions: [GraphGTest.cpp (line 1346)](https://github.com/networkit/networkit/blob/master/networkit/cpp/graph/test/GraphGTest.cpp#L1346), [GraphBuilderAutoCompleteGTest.cpp (line 486)](https://github.com/networkit/networkit/blob/master/networkit/cpp/graph/test/GraphBuilderAutoCompleteGTest.cpp#486), [GraphBuilderDirectSwapGTest.cpp (line 531)](https://github.com/networkit/networkit/blob/master/networkit/cpp/graph/test/GraphBuilderDirectSwapGTest.cpp#L531), [LocalSimilarityScore.cpp (line 51)](https://github.com/networkit/networkit/blob/master/networkit/cpp/sparsification/LocalSimilarityScore.cpp#L51). In all cases this happens during `parallelForNodes`, after `#pragma omp critical` is set. All other usages of `critical` don't involve `parrallelForNodes` (and don't cause crashes). If `critical` is removed (which we don't want of course), building the code completes without error. The errors are very likely bugs, the code itself seems ok.

### DIFF
--- a/networkit/cpp/graph/test/GraphBuilderAutoCompleteGTest.cpp
+++ b/networkit/cpp/graph/test/GraphBuilderAutoCompleteGTest.cpp
@@ -482,10 +482,9 @@ TEST_P(GraphBuilderAutoCompleteGTest, testForNodes) {
 }
 
 TEST_P(GraphBuilderAutoCompleteGTest, testParallelForNodes) {
-    std::vector<node> visited;
+    std::vector<node> visited(bHouse.upperNodeIdBound());
     this->bHouse.parallelForNodes([&](node u) {
-        #pragma omp critical
-        visited.push_back(u);
+        visited[u] = u;
     });
 
     Aux::Parallel::sort(visited.begin(), visited.end());
--- a/networkit/cpp/graph/test/GraphBuilderDirectSwapGTest.cpp
+++ b/networkit/cpp/graph/test/GraphBuilderDirectSwapGTest.cpp
@@ -527,10 +527,9 @@ TEST_P(GraphBuilderDirectSwapGTest, testForNodes) {
 }
 
 TEST_P(GraphBuilderDirectSwapGTest, testParallelForNodes) {
-    std::vector<node> visited;
+    std::vector<node> visited(bHouse.upperNodeIdBound());
     this->bHouse.parallelForNodes([&](node u) {
-        #pragma omp critical
-        visited.push_back(u);
+        visited[u] = u;
     });
 
     Aux::Parallel::sort(visited.begin(), visited.end());
--- a/networkit/cpp/graph/test/GraphGTest.cpp
+++ b/networkit/cpp/graph/test/GraphGTest.cpp
@@ -1342,10 +1342,9 @@ TEST_P(GraphGTest, testForNodes) {
 }
 
 TEST_P(GraphGTest, testParallelForNodes) {
-    std::vector<node> visited;
+    std::vector<node> visited(Ghouse.upperNodeIdBound());
     this->Ghouse.parallelForNodes([&](node u) {
-#pragma omp critical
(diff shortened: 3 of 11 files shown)


angriman:
Not being able to use a critical section within a `parallelForNodes` loop sounds restrictive. Do you know why those crashes happen? Do we have any other alternative solution?

networkit/cpp/graph/test/GraphBuilderAutoCompleteGTest.cpp
```diff
@@ -483,9 +483,9 @@ TEST_P(GraphBuilderAutoCompleteGTest, testForNodes) {
 
 TEST_P(GraphBuilderAutoCompleteGTest, testParallelForNodes) {
     std::vector<node> visited;
+    visited.resize(bHouse.upperNodeIdBound());
```
angriman, line 486: You can drop the `resize` by passing the size to the constructor of the vector.

networkit/cpp/graph/test/GraphBuilderDirectSwapGTest.cpp
```diff
@@ -528,9 +528,9 @@ TEST_P(GraphBuilderDirectSwapGTest, testForNodes) {
 
 TEST_P(GraphBuilderDirectSwapGTest, testParallelForNodes) {
     std::vector<node> visited;
+    visited.resize(bHouse.upperNodeIdBound());
```
angriman, line 531: Same as above.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -1343,9 +1343,9 @@ TEST_P(GraphGTest, testForNodes) {
 
 TEST_P(GraphGTest, testParallelForNodes) {
     std::vector<node> visited;
+    visited.resize(Ghouse.upperNodeIdBound());
```
angriman, line 1346: Also here.

fabratu:
> Do you know why those crashes happen? Do we have any other alternative solution?

I think all occasions are bugs, since clang crashes on IR-generation. The only alternative solution I came up with, is to have workarounds. Thats means either rewriting the critical-section (while still `parallelForNodes`) or exchange the call to `parallelForNodes` with it's logic. Both approaches are used in the PR. As a result, clang-12 compiles correctly.

avdgrinten:
Are the bugs reported to Clang's upstream?

fabratu:
> Are the bugs reported to Clang's upstream?

Not yet, currently working on it.

angriman (approved):
I think we can merge this, I don't like the workaround for critical sections in `parallelForNodes`, but on the other hand it affects very few parts of the code base and I don't see alternative solutions.

fabratu:
Will merge this early (small fixes, needed for correct CI-tests)

Issue #495, opened 2020-02-06
Cannot pickle graph


saibalmars:
Hi, I got some problem when I tried to pickle an nk graph:
``` python
import netwokit as nk
import pickle
g =nk.graph.Graph(2)
g.addEdge(0,1)
pickle.dumps(g)
```

This code gives me errors as follows:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3325, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-11-79ae2e0d8bd4>", line 1, in <module>
    pickle.dumps(g)
  File "stringsource", line 2, in networkit._NetworKit.Graph.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
```

Is there any way to fix this? 
Or is there any way to let it be readable for multiprocessing.pool?

avdgrinten:
That is expected. You can serialize the graph to a file using the `NetworkitBinaryWriter` and deserialize it using the `NetworkitBinaryReader` (or the `readGraph` / `writeGraph`) helper functions. Other format also work but the NetworKit binary format generally has lowest performance overhead.

michitux:
Note also that all `run()`-methods as well as some other long-running methods in NetworKit release the GIL, therefore if you mainly want to run several algorithms implemented in NetworKit in parallel, a pool of threads (like `multiprocessing.pool.ThreadPool`) should be enough to achieve decent parallelism.

gijzelaerr:
I just ran into the same issue. isn't it possible to integrate the NetworkitBinaryReader & Writer by implementing the pickling interfaces? Methods like:

https://docs.python.org/3/library/pickle.html#object.__getstate__

Or are there low-level reasons why this might nog work?

AndrewHannigan:
@gijzelaerr I think currently things like edge indexes don't get written to the Networkit binary format.  So a round trip from in-memory nk graph to binary format back to in-memory nk graph does not necessarily recreate an identical graph, and you would have to rebuild the edge indexes aagain.  So that complicates a pickle interface, where you'd expect the round trip to return something identical.

avdgrinten:
It would be possible (and desirable) to extend the NetworKit binary format to also support edge indices though. The binary format stores a feature bitmask; it would be possible to add this without sacrificing backwards compatibility.

AndrewHannigan:
[...]

PR #751, opened 2021-05-07
NetworkitBinaryGraph is now able to save Edge Indices

Relux-the-Relux:
Addressing #495 .

### DIFF
--- a/include/networkit/io/NetworkitBinaryGraph.hpp
+++ b/include/networkit/io/NetworkitBinaryGraph.hpp
@@ -29,6 +29,8 @@ struct Header {
     uint64_t offsetAdjTranspose;
     uint64_t offsetWeightLists;
     uint64_t offsetWeightTranspose;
+    uint64_t offsetAdjIdLists;
+    uint64_t offsetAdjIdTranspose;
 };
 
 enum WEIGHT_FORMAT {
@@ -43,6 +45,8 @@ static constexpr uint8_t DELETED_BIT = 0x1; // bit 0
 static constexpr uint64_t DIR_MASK = 0x1; // bit 0
 static constexpr uint64_t WGHT_MASK = 0xE; //bit 1-3
 static constexpr uint64_t WGHT_SHIFT = 0x1;
+static constexpr uint64_t INDEX_MASK = 0x10; //bit 4
+static constexpr uint64_t INDEX_SHIFT = 0x4;
 
 /**
  * Serializes value into a buffer and returns the number of bytes written.
--- a/include/networkit/io/NetworkitBinaryReader.hpp
+++ b/include/networkit/io/NetworkitBinaryReader.hpp
@@ -7,8 +7,11 @@
 #ifndef NETWORKIT_IO_NETWORKIT_BINARY_READER_HPP_
 #define NETWORKIT_IO_NETWORKIT_BINARY_READER_HPP_
 
+#include <cstring>
+#include <string>
 #include <networkit/graph/Graph.hpp>
 #include <networkit/io/GraphReader.hpp>
+#include <networkit/io/MemoryMappedFile.hpp>
 
 namespace NetworKit {
 
@@ -24,12 +27,18 @@ public:
     NetworkitBinaryReader(){};
 
     Graph read(const std::string &path) override;
+    Graph readFromBuffer(const std::vector<uint8_t> &data);
 
 private:
     count nodes;
     count chunks;
     bool directed;
     bool weighted;
+    bool indexed;
+    count version;
+
+    template <class T>
+    Graph readData(const T &source);
 };
 } // namespace NetworKit
 
--- a/include/networkit/io/NetworkitBinaryWriter.hpp
+++ b/include/networkit/io/NetworkitBinaryWriter.hpp
@@ -21,6 +21,8 @@ enum class NetworkitBinaryWeights {
     autoDetect
 };
 
+enum class NetworkitBinaryEdgeIDs { noEdgeIDs, writeEdgeIDs, autoDetect };
+
 /**
  * @ingroup io
  *
@@ -30,13 +32,20 @@ class NetworkitBinaryWriter final : public GraphWriter {
 
 public:
     NetworkitBinaryWriter(uint64_t chunks = 32,
-                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
+                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect,
+                          NetworkitBinaryEdgeIDs edgeIndex = NetworkitBinaryEdgeIDs::autoDetect);
 
     void write(const Graph &G, const std::string &path) override;
+    std::vector<uint8_t> writeToBuffer(const Graph &G);
 
 private:
     count chunks;
     NetworkitBinaryWeights weightsType;
+    NetworkitBinaryEdgeIDs edgeIndex;
+    bool preserveEdgeIndex;
+
+    template <class T>
+    void writeData(T &outStream, const Graph &G);
 };
 
 } // namespace NetworKit
--- a/networkit/cpp/io/NetworkitBinaryGraph.md
+++ b/networkit/cpp/io/NetworkitBinaryGraph.md
@@ -23,15 +23,17 @@ struct Header {
 };
 ```
 - magic: A constant value used to identify the file format version.
-    - The current version is '*nkbg002*' which supports weighted, undirected and directed graphs.
+    - The current version is '*nkbg003*' which supports weighted, undirected and directed graphs.
 - checksum: Currently not used
 - features: Contains the graph information bitwise
     - Bit 0 : directed or undirected
-    - Bit 1-2 : the weight format of the graph. Below are the possible formats:
-        - 0 = Weights are unsigned integers
-        - 1 = Weights are signed integers
-        - 2 = Weights are doubles
-        - 3 = Weights are floats
+    - Bit 1-3 : the weight format of the graph. Below are the possible formats:
+		- 0 = Graph is unweighted
+        - 1 = Weights are unsigned integers
+        - 2 = Weights are signed integers
+        - 3 = Weights are doubles
+        - 4 = Weights are floats
+	- Bit 4 = proper edge indexing
 - nodes : The number of nodes the graph has
 - chunks: The number of chunks the nodes have been divided in
 - offsetBaseData: Offset of base data in the file 
@@ -47,6 +49,7 @@ Base data
 ```
 uint64_t nodeFlags[nodes]: Flags storing information about a node
 uint64_t firstVertex[chunks-1]: The index of the first vertex of each chunk excluding the first chunk
+uint64_t edgeIndex[edges]:
 ```
 Adjacency lists
 -----------------
--- a/networkit/cpp/io/NetworkitBinaryGraph.md
+++ b/networkit/cpp/io/NetworkitBinaryGraph.md
@@ -20,20 +20,22 @@ struct Header {
     uint64_t offsetTranspose;
     uint64_t offsetWeightLists;
     uint64_t offsetWeightTranspose;
+    uint64_t offsetAdjIdLists;
+    uint64_t offsetAdjIdTranspose;
 };
 ```
 - magic: A constant value used to identify the file format version.
-    - The current version is '*nkbg003*' which supports weighted, undirected and directed graphs.
+    - The current version is '*nkbg002*' which supports weighted, undirected and directed graphs.
 - checksum: Currently not used
 - features: Contains the graph information bitwise
     - Bit 0 : directed or undirected
     - Bit 1-3 : the weight format of the graph. Below are the possible formats:
-		- 0 = Graph is unweighted
+        - 0 = Graph is unweighted
         - 1 = Weights are unsigned integers
         - 2 = Weights are signed integers
         - 3 = Weights are doubles
         - 4 = Weights are floats
-	- Bit 4 = proper edge indexing
+    - Bit 4 = edge indices are present
 - nodes : The number of nodes the graph has
 - chunks: The number of chunks the nodes have been divided in
 - offsetBaseData: Offset of base data in the file 
@@ -41,6 +43,10 @@ struct Header {
 - offsetTranspose: Offset of the transposed adjaceny lists in the file
 - offsetWeightLists: Offset of the adjacency weights in the file
 - offsetWeightTranspose: Offset of the transposed adjacency weights in the file
+- offsetWeightLists: Offset of the adjacency weights in the file
+- offsetWeightTranspose: Offset of the transposed adjacency weights in the file
+- offsetAdjIdLists: Offset of the adjacency edge ids in the file
+- offsetAdjIdTranspose: Offset of the transposed adjacency edge ids in the file
 
 All offsets are relative to the beginning of the section.
 
@@ -49,7 +55,6 @@ Base data
 ```
 uint64_t nodeFlags[nodes]: Flags storing information about a node
 uint64_t firstVertex[chunks-1]: The index of the first vertex of each chunk excluding the first chunk
-uint64_t edgeIndex[edges]:
 ```
 Adjacency lists
 -----------------
@@ -85,4 +90,16 @@ Depending on the type of weights:
  - unsigned/signed weights: varint data [...]: Varint encoded weight lists
  - double weights: double data [...]: Weight lists as doubles
  - float weights: float data [...]: Weight lists as floats
+```
+Index lists
+--------------------
+```
+uint64_t offset[chunks-1]: Offset of the file where the indices are
+varint data [...]: Varint encoded weight lists
+```
+Index transpose
+--------------------
+```
+uint64_t offset[chunks-1]: Offset of the file where the transposed indices are:
+varint data [...]: Varint encoded index lists
 ```
--- a/networkit/cpp/io/NetworkitBinaryReader.cpp
+++ b/networkit/cpp/io/NetworkitBinaryReader.cpp
@@ -7,7 +7,6 @@
  */
 
 #include <atomic>
-#include <cstring>
 #include <fstream>
 
 #include <networkit/auxiliary/Log.hpp>
@@ -15,15 +14,34 @@
 #include <networkit/io/NetworkitBinaryGraph.hpp>
 #include <networkit/io/MemoryMappedFile.hpp>
 
+
+const char *accessData(const NetworKit::MemoryMappedFile &source) {
+    return source.cbegin();
+}
+
+const char *accessData(const std::vector<uint8_t> &source) {
+    return (char*)source.data();
+}
+
 namespace NetworKit {
 
 Graph NetworkitBinaryReader::read(const std::string& path) {
+    MemoryMappedFile mmfile(path);
+    Graph G = readData(mmfile);
+    return G;
+}
+
+Graph NetworkitBinaryReader::readFromBuffer(const std::vector<uint8_t>& data) {
+    return readData(data);
+}
+
+template<class T>
+Graph NetworkitBinaryReader::readData(const T &source) {
     nkbg::Header header;
     nkbg::WEIGHT_FORMAT weightFormat;
-
-    MemoryMappedFile mmfile(path);
-    auto it = mmfile.cbegin();
-
+    
+    const char* startIt = accessData(source);
+    const char* it = startIt;
     auto readHeader = [&] () {
         memcpy(&header.magic, it, sizeof(uint64_t));
         it += sizeof(uint64_t);
@@ -31,6 +49,19 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
         it += sizeof(uint64_t);
         memcpy(&header.features, it, sizeof(uint64_t));
         it += sizeof(uint64_t);
+        if (!memcmp("nkbg002", header.magic, 8)) {
+            version = 2;
+        }else if (!memcmp("nkbg003", header.magic, 8)) {
+            version = 3;
+        }else{
+            throw std::runtime_error("Reader expected another magic value");
+        }
+        directed = (header.features & nkbg::DIR_MASK);
+        weightFormat = static_cast<nkbg::WEIGHT_FORMAT>((header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT);
+        indexed = false;
+        if (version >= 3) {
+            indexed = (header.features & nkbg::INDEX_MASK) >> nkbg::INDEX_SHIFT;
+        }
         memcpy(&header.nodes, it, sizeof(uint64_t));
         it += sizeof(uint64_t);
         memcpy(&header.chunks, it, sizeof(uint64_t));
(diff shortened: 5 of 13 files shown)


avdgrinten:
This is a nice feature.

We also need to update the documentation of the file format to describe how the edge IDs are stored, in:
https://github.com/networkit/networkit/blob/master/networkit/cpp/io/NetworkitBinaryGraph.md

Do we need to bump the version? Can't the version 2 reader read a graph with edge IDs, too? As far as I can see, the old reader will just ignore the edge IDs but it will continue to work, right?

Relux-the-Relux:
> Do we need to bump the version? Can't the version 2 reader read a graph with edge IDs, too? As far as I can see, the old reader will just ignore the edge IDs but it will continue to work, right?

No, I didn't bump the version it was just  error in the documentation(from my side) I fixed it it is still nkbg02

Relux-the-Relux:
Also going to update the documentation

avdgrinten:
This does not solve the pickle issue yet -- we still need to implement logic to pickle and unpickle graphs in Python.

avdgrinten:
Your implementation stores the graph in a file and then saves the file name to implement the pickle operation. That is not really what users expect -- the pickle operation should store the graph itself. One option would be reading the file back into a Python 3 `bytes` object.

Relux-the-Relux:
Ok made the change that it now reads from the file into memory. However, this causes 4 hard disk accesses to get and then set the states (2 for getstate and 2 for setstate).

maybe we should implement a version of BinaryReader/Writer that take/output binary strings? Unsure how often get/setstate will get called so that might be a non-issue.

include/networkit/io/NetworkitBinaryReader.hpp
```diff
@@ -24,12 +27,17 @@ class NetworkitBinaryReader final : public GraphReader {
     NetworkitBinaryReader(){};
 
     Graph read(const std::string &path) override;
+    Graph readState(const std::string &data);
```
avdgrinten, line 30: Rename to `readFromBuffer`. Let this take a `std::vector<uint8_t>` instead of a `std::string`.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -30,13 +30,19 @@ class NetworkitBinaryWriter final : public GraphWriter {
 
 public:
     NetworkitBinaryWriter(uint64_t chunks = 32,
-                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
+                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect,
+                          bool preserveEdgeIndex = false);
 
     void write(const Graph &G, const std::string &path) override;
+    std::string writeState(const Graph &G);
```
avdgrinten, line 37: Rename to `writeToBuffer` and let this return a `std::vector<uint8_t>`.

networkit/cpp/io/NetworkitBinaryGraph.md
```diff
@@ -20,25 +20,33 @@ struct Header {
...
-        - 1 = Weights are signed integers
-        - 2 = Weights are doubles
-        - 3 = Weights are floats
+    - Bit 1-3 : the weight format of the graph. Below are the possible formats:
+		- 0 = Graph is unweighted
+        - 1 = Weights are unsigned integers
+        - 2 = Weights are signed integers
+        - 3 = Weights are doubles
+        - 4 = Weights are floats
+	- Bit 4 = proper edge indexing
```
avdgrinten, line 38: "proper edge indexing" -> "edge indices are present". Remove the unrelated changes above this one.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -7,23 +7,44 @@
...
+
+template <>
+const char *getIterator(const NetworKit::MemoryMappedFile &source) {
+    return source.cbegin();
+}
+
+template <>
+const char *getIterator(const std::string &source) {
+    return source.data();
+}
```
avdgrinten, line 28: Do not use specialization here. Use multiple function overloads instead (basically, just drop the `template <>`).

avdgrinten: These functions are local to the file and should be in an anonymous namespace.

avdgrinten: Since these functions return a `const char *`, `accessData()` would be a better name.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -89,12 +116,15 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
-    const char *adjIt = mmfile.cbegin() + header.offsetAdjLists;
-    const char *transpIt = mmfile.cbegin() + header.offsetAdjTranspose;
-    const char *adjWghtIt = mmfile.cbegin() + header.offsetWeightLists;
-    const char *transpWghtIt = mmfile.cbegin() + header.offsetWeightTranspose;
+    const char *adjIt = getIterator(source) + header.offsetAdjLists;
+    const char *transpIt = getIterator(source) + header.offsetAdjTranspose;
+    const char *adjWghtIt = getIterator(source) + header.offsetWeightLists;
+    const char *transpWghtIt = getIterator(source) + header.offsetWeightTranspose;
+    const char *adjIdIt = getIterator(source) + header.offsetAdjIdLists;
+    const char *transpIdIt = getIterator(source) + header.offsetAdjIdTranspose;
```
avdgrinten, line 125: Save the result of `getIterator(source)` to a local variable to avoid duplication here.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -249,74 +282,97 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
...
     }
+
+    if(preserveEdgeIndex && weightFormat != nkbg::WEIGHT_FORMAT::NONE) {
+        header.offsetAdjIdLists = header.offsetWeightTranspose
+            + (chunks - 1) * sizeof(uint64_t)//transpWeightOffsets
+            + transpWghtOffsets.back();
+        header.offsetAdjIdTranspose = header.offsetAdjIdLists
+            + (chunks - 1) * sizeof(uint64_t)
+            + adjIndexOffsets.back();
+    } else if(preserveEdgeIndex) { //Even if there is no weight data, offset for weights are still written
```
avdgrinten, line 293: `//Even if there is no weight data, offset for weights are still written` -> Why is this done?

avdgrinten (changes requested):

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -17,12 +18,23 @@
...
+}
+
+std::string NetworkitBinaryWriter::writeState(const Graph &G) {
+    std::stringstream outdata;
+    writeData(outdata, G);
+    return outdata.str();
+}
+
+template <class T>
+void NetworkitBinaryWriter::writeData(T &outStream, const Graph &G){
```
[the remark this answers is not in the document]
Relux-the-Relux: In this case, the MemoryMappedFile functionality of Networkit would need to be expanded as it currently can only put existing files into memory (and will fail when opening empty files) and has currently no write functionality.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -249,74 +282,97 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
         header.offsetWeightLists = 0;
         header.offsetWeightTranspose = 0;
     }
+
+    if(preserveEdgeIndex && weightFormat != nkbg::WEIGHT_FORMAT::NONE) {
```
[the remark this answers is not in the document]
avdgrinten: Maybe we should just have a `nextOffset` variable that starts at 0 (or at the size of the headers) and that we increment whenever we "allocate" space for more data within the file? That approach would be less complicated than calculating later offsets based on earlier offsets.

networkit/cpp/io/NetworkitBinaryWriter.cpp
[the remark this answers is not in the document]
Relux-the-Relux: yes, it works since string to std::vector<uint8_t> is a easy conversion. So no problem on the c++ side. I currently just need to test with cython since the documentation recommends using std::string when interfacing bytes data with python.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -17,12 +18,24 @@
...
+    writeData(outfile, G);
+    INFO("Written graph to ", path);
+}
+
+std::vector<uint8_t> NetworkitBinaryWriter::writeToBuffer(const Graph &G) {
+    std::stringstream outdata;
+    writeData(outdata, G);
+    std::string str = outdata.str();
+    return std::vector<uint8_t>(str.begin(), str.end());
+}
```
avdgrinten, line 35: This incurs a further copy. If we use a writable memory mapped file, the copy could be avoided. It would be fine to keep this code for now (while changing it later to a writable memory mapped file). On the other hand, using a writable file and replacing `outStream.write` by `memcpy` does not seem to be that difficult either. Do you want to change that now or keep it as it is for this PR?

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -28,7 +28,9 @@ struct Header {
     uint64_t offsetAdjLists;
     uint64_t offsetAdjTranspose;
     uint64_t offsetWeightLists;
-    uint64_t offsetWeightTranspose;
+    uint64_t offsetWeightTranspose;
```
avdgrinten, line 31: There is extra whitespace at the end of this file.

networkit/cpp/io/NetworkitBinaryGraph.md
avdgrinten: This still needs to be fixed.

networkit/test/test_graph.py
```diff
@@ -253,5 +254,10 @@ def testAddEdgeWithMissingNodes(self):
 		self.assertEqual(G.numberOfNodes(), 3)
 		self.assertEqual(G.numberOfEdges(), 2)
 
+	def testGraphPickling(self):
+		G = nk.Graph(2)
+		G.addEdge(0,1)
+		pickle.dumps(G)
```
avdgrinten, line 260: The test should also verify that unpickling works and that the deseralized graph is identical to the serialized one.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -30,13 +30,19 @@ class NetworkitBinaryWriter final : public GraphWriter {
 
 public:
     NetworkitBinaryWriter(uint64_t chunks = 32,
-                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
+                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect,
+                          bool preserveEdgeIndex = false);
```
avdgrinten, line 34: Why is this `false` by default? I think we should default to true if the graph has edge indices.

What happens if the users sets `true` here but the graph has no edge indices?

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -106,26 +133,34 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
+        uint64_t transIndexOff = 0;
         if(vertex) {
             memcpy(&off, adjIt + (c-1)* sizeof(uint64_t), sizeof(uint64_t));
             memcpy(&transpOff, transpIt + (c-1)* sizeof(uint64_t), sizeof(uint64_t));
             memcpy(&wghtOff, adjWghtIt + (c-1) * sizeof(uint64_t), sizeof(uint64_t));
             memcpy(&transWghtOff, transpWghtIt + (c-1) * sizeof(uint64_t), sizeof(uint64_t));
+            if(indexed){
+            memcpy(&indexOff, adjIdIt + (c-1) * sizeof(uint64_t), sizeof(uint64_t));
+            memcpy(&transIndexOff, transpIdIt + (c-1) * sizeof(uint64_t), sizeof(uint64_t));
+            }
```
avdgrinten, line 153: Doesn't match NetworKit's coding style.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -106,26 +133,34 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
         transpOff += (chunks - 1) * sizeof(uint64_t);
         off += sizeof(uint64_t);
         transpOff += sizeof(uint64_t);
         wghtOff += (chunks - 1) * sizeof(uint64_t);
         transWghtOff += (chunks - 1) * sizeof(uint64_t);
+        indexOff += (chunks - 1) * sizeof(uint64_t);
+        transIndexOff += (chunks - 1) * sizeof(uint64_t);
         uint64_t n = firstVert[c+1] - firstVert[c];
-
         for (uint64_t i = 0; i < n; i++) {
```
avdgrinten, line 164: Unrelated change.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -216,12 +258,17 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
                         break;
                 }
+                if(indexed){
+                    transIndexOff += nkbg::varIntDecode(reinterpret_cast<const uint8_t*>(transpIdIt + transIndexOff), id);
+                }
                 if(!directed) {
                     if(curr != add) {
-                        G.addPartialEdge(unsafe, curr, add, weight);
+                        if(indexed) G.addPartialEdge(unsafe, curr, add, weight, id);
+                        else G.addPartialEdge(unsafe, curr, add, weight);
```
avdgrinten, line 267: This calls the same function, right? We can get rid of the `if`.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -216,12 +258,17 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
                     }
                 } else {
-                    G.addPartialInEdge(unsafe, curr, add, weight);
+                    if(indexed) G.addPartialInEdge(unsafe, curr, add, weight, id);
+                    else G.addPartialInEdge(unsafe, curr, add, weight);
```
avdgrinten, line 271: Likewise.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -232,6 +279,7 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
     for(omp_index c = 0; c < static_cast<omp_index>(chunks); c++) {
         constructGraph(c);
     }
+
     G.setNumberOfSelfLoops(unsafe, selfLoops);
```
avdgrinten, line 283: Unrelated change.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -172,10 +208,15 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
                     case nkbg::WEIGHT_FORMAT::NONE:
                         break;
                 }
+                if(indexed){
+                    indexOff += nkbg::varIntDecode(reinterpret_cast<const uint8_t*>(adjIdIt + indexOff), id);
+                }
                 if(!directed) {
-                    G.addPartialEdge(unsafe, curr, add, weight);
+                    if(indexed) G.addPartialEdge(unsafe, curr, add, weight, id);
+                    else G.addPartialEdge(unsafe, curr, add, weight);
```
avdgrinten, line 216: See below.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -172,10 +208,15 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
...
                 } else {
-                    G.addPartialOutEdge(unsafe, curr, add, weight);
+                    if(indexed) G.addPartialOutEdge(unsafe, curr, add, weight, id);
+                    else G.addPartialOutEdge(unsafe, curr, add, weight);
```
avdgrinten, line 219: See below.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -139,6 +174,7 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
             }
             //Read adjacency lists.
             for (uint64_t j = 0; j < outNbrs; j++) {
+                edgeid id;
```
avdgrinten, line 177: Don't we have to set the `Graph`'s maximal edge ID somewhere? Otherwise the `Graph` will allocate the same ID again when adding new edges.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -180,6 +200,13 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
         return size;
     };
 
+    auto computeIndexOffsets = [&] (edgeid id) {
+        uint64_t size = 0;
+        uint8_t tmp [10];
+        if(preserveEdgeIndex) size = nkbg::varIntEncode(id, tmp);
```
avdgrinten, line 206: Doesn't match NetworKit's coding style.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -191,23 +218,27 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
                         outNbrs++;
                         adjSize += nkbg::varIntEncode(v, tmp);
                         adjWeightSize += computeWeightsOffsets(w);
+                        if(preserveEdgeIndex) adjIndexSize += computeIndexOffsets(G.edgeId(n, v));
```
avdgrinten, line 221: Coding style, also below.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -222,32 +253,50 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
         transpOffsets.push_back(transpSize);
         adjWghtOffsets.push_back(adjWeightSize);
         transpWghtOffsets.push_back(transpWeightSize);
+        adjIndexOffsets.push_back(adjIndexSize);
+        transpIndexOffsets.push_back(transpIndexSize);
     }
     // Write header.
-    strncpy(header.magic,"nkbg002",8);
+    strncpy(header.magic,"nkbg002", 8);
```
avdgrinten, line 260: Unrelated change.

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -347,7 +397,51 @@ void NetworkitBinaryWriter::write(const Graph &G, const std::string &path) {
...
     });
-
-    INFO("Written graph to ", path);
+    
+    // Write adj index.
+    for (uint64_t c = 1; c < chunks; c++) {
+        outStream.write(reinterpret_cast<char*>(&adjIndexOffsets[c-1]), sizeof(uint64_t));
+    }
+    if(preserveEdgeIndex){
+    G.forNodes([&](node u) {
```
avdgrinten, line 406: Coding style.

avdgrinten:
The commit structure is a bit messy. Squash all the changes to the C++ reader and writer to a single commit. The commits that deal with tests and/or the Python interface can stay separate.

networkit/cpp/io/NetworkitBinaryReader.cpp
Relux-the-Relux: No, because we are reading the ids from file, not generating ourselves. See line 213.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -30,13 +30,20 @@ class NetworkitBinaryWriter final : public GraphWriter {
 
 public:
     NetworkitBinaryWriter(uint64_t chunks = 32,
-                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect);
+                          NetworkitBinaryWeights weightsType = NetworkitBinaryWeights::autoDetect,
+                          bool ignoreEdgeIndex = false);
```
avdgrinten, line 34: I don't like this `bool` flag here. Do we need it?

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -45,6 +63,10 @@ Graph NetworkitBinaryReader::read(const std::string& path) {
         it += sizeof(uint64_t);
         memcpy(&header.offsetWeightTranspose, it, sizeof(uint64_t));
         it += sizeof(uint64_t);
+        memcpy(&header.offsetAdjIdLists, it, sizeof(uint64_t));
+        it += sizeof(uint64_t);
+        memcpy(&header.offsetAdjIdTranspose, it, sizeof(uint64_t));
+        it += sizeof(uint64_t);
```
avdgrinten, line 69: We need to make sure that these fields actually exist, do we? Old files will not have these entries.

avdgrinten: In fact, the header of old files will simply have the wrong size (since these fields are present whether edge IDs are in use or not). This probably requires a version bump.

Note that we still need to be able to read files of version 2 (and we should probably only write version 3 files if edge IDs are in use).

avdgrinten (commented):
It would be nice to have this merged soon but there are a few remaining issue that I commented on below.

include/networkit/io/NetworkitBinaryWriter.hpp
Relux-the-Relux: Not really I could remove it and make always and only save indices of previously indexed graphs.

networkit/cpp/io/NetworkitBinaryReader.cpp
Relux-the-Relux: I don't think it is necessary, since the size of the header shouldn't matter since we get a pointer(explicitly written) to where the data begins. It is the same case with unweighted graphs, where no weight data is written but we reader the header for it anyway, since it will be later handled with the header checker depending on the features flags.

yeah, so the size of the header doesn't matter because of the feature flags. since we will never read the pointers for IDs if that feature is not turned on, otherwise we just seek the data at the pointers specified in the header (not necessarily where the header immediately ends).

It is a similar situation to unweighted graphs where no weight data is written.

But I should test it.

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -49,6 +49,21 @@ Graph NetworkitBinaryReader::readData(const T &source) {
...
+                weightFormat = static_cast<nkbg::WEIGHT_FORMAT>((header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT);
+                indexed = false;
+                version = 2;
+            }
+        } else {
+            directed = (header.features & nkbg::DIR_MASK);
+            weightFormat = static_cast<nkbg::WEIGHT_FORMAT>((header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT);
+            indexed = (header.features & nkbg::INDEX_MASK) >> nkbg::INDEX_SHIFT;
+            version = 3;
+        }
```
avdgrinten, line 66: There is some redundancy here. I think this part of the code should be restructured, for example like this:
```suggestion
if (!memcmp("nkbg002", header.magic, 8)) {
    version = 2;
}else if (!memcmp("nkbg003", header.magic, 8)) {
    version = 3;
}else{
    throw std::runtime_error("Reader expected another magic value");
}
directed = (header.features & nkbg::DIR_MASK);
weightFormat = static_cast<nkbg::WEIGHT_FORMAT>((header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT);
indexed = false;
if (version >= 3)
    indexed = (header.features & nkbg::INDEX_MASK) >> nkbg::INDEX_SHIFT;
```

networkit/cpp/io/NetworkitBinaryReader.cpp
```diff
@@ -118,10 +125,8 @@ Graph NetworkitBinaryReader::readData(const T &source) {
     const char *transpIt = startIt + header.offsetAdjTranspose;
     const char *adjWghtIt = startIt + header.offsetWeightLists;
     const char *transpWghtIt = startIt + header.offsetWeightTranspose;
-    const char *adjIdIt = startIt + header.offsetAdjIdLists;
-    const char *transpIdIt = startIt + header.offsetAdjIdTranspose;
-
-
+    const char *adjIdIt = (version > 2) ? startIt + header.offsetAdjIdLists : 0;
+    const char *transpIdIt = (version > 2) ? startIt + header.offsetAdjIdTranspose : 0;
```
avdgrinten, line 129: Use `nullptr` here.

avdgrinten (commented):
Can be merged after some minor fixes. When rebasing, please also change the commit messages to be in imperative: `expands Graph constructor ...` -> `Expand Graph constructor ...`

avdgrinten (approved):
LGTM.

PR #752, opened 2021-05-18
improved (parallel) coarsening implementation for clusterings

larsgottesbueren:
The algorithm generates all incident edges of a coarse vertex at the same time (on the same thread) and thus avoids linear searches in the neighbor lists.
This is the first in a potential series of pull-requests that contain the features from the 'huge' ego-splitting pull-request.

### DIFF
--- a/networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
+++ b/networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
@@ -15,8 +15,6 @@
 #include <networkit/coarsening/ParallelPartitionCoarsening.hpp>
 #include <networkit/graph/GraphBuilder.hpp>
 
-#include <iostream>
-
 namespace NetworKit {
 
 ParallelPartitionCoarsening::ParallelPartitionCoarsening(const Graph &G, const Partition &zeta,
--- a/networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
+++ b/networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
@@ -5,6 +5,8 @@
  *      Author: cls
  */
 
+// networkit-format
+
 #include <numeric>
 #include <omp.h>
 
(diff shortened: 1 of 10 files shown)


larsgottesbueren:
The CI failures don't seem to be related to the changes in this PR. I hope that's fine.

angriman:
That's true, we are working on fixing those CI issues in #749

networkit/cpp/structures/Partition.cpp
```diff
@@ -134,6 +134,10 @@ std::vector<index> Partition::getVector() const {
     return data;
 }
 
+std::vector<index> Partition::moveVector() {
+    return std::move(this->data);
+}
```
angriman, line 139: I don't think that this is a good idea, it makes the instance of `Partition` useless after it is called. What about returning a const ref?

larsgottesbueren: I guess the second copy during coarsening doesn't hurt too bad. I've made the getVector() function return a const ref, and deleted the moveVector() function.

michitux:
The coarsening currently accepts (both in the old and the new implementation as far as I can see) directed graphs but always returns undirected graphs which might be broken in the case that edges exist just in one direction. I'm wondering if the implementation could be fixed to work for directed graphs. This could be done by aggregating incoming edges the same way outgoing edges are aggregated for each cluster.

networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
```diff
@@ -13,124 +15,97 @@
...
+    // sort fine vertices by coarse vertices
+    std::vector<index> partBegin(numParts + 2, 0);
+    std::vector<node> nodesSortedByPart(G->numberOfNodes());
+    G->forNodes([&](const node u) { partBegin[nodeToSuperNode[u] + 2]++; });
+    std::partial_sum(partBegin.begin(), partBegin.end(), partBegin.begin());
+    G->forNodes([&](const node u) { nodesSortedByPart[partBegin[nodeToSuperNode[u] + 1]++] = u; });
+
+    Gcoarsened = Graph(numParts, true, false);
+
+    auto aggregate_edge_weights = [&](node su, count &numEdges, count &numSelfLoops,
```
angriman, line 40: Use camelCase

angriman (commented):
Looks ok to me, just a minor comment. Please rebase on master.

networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
```diff
@@ -13,124 +15,97 @@
 #include <networkit/coarsening/ParallelPartitionCoarsening.hpp>
 #include <networkit/graph/GraphBuilder.hpp>
 
+#include <iostream>
```
angriman, line 18: This is not needed anymore right?

angriman (commented):
A general comment, if you format existing files please do that in a separate commit, otherwise it is difficult to see the changes that are made in the PR.

networkit/cpp/coarsening/ParallelPartitionCoarsening.cpp
larsgottesbueren: Correct. Is there a way to remove it without re-triggering the CI?

angriman: No, CI is triggered after every change. It's quite annoying that it needs to be done manually, we should probably change this setting.

angriman (approved):

PR #753, opened 2021-05-18
Community Detection Algorithm based on Map Equation

larsgottesbueren:
The main algorithm is basically Louvain.
There are two parallelization strategies: RelaxMap which uses locks to update cluster cuts, and Synchronous, which performs some moves on stale data and frequently synchronizes the cuts and volumes.
Second part of the ego-splitting PR series.

Collaborators @michitux  @ArminWiebigke

### DIFF
--- a/include/networkit/auxiliary/SparseVector.hpp
+++ b/include/networkit/auxiliary/SparseVector.hpp
@@ -71,11 +71,6 @@ public:
     const T &operator[](index i) const;
 
     /**
-     * Get all indexes into which elements were inserted.
-     */
-    const std::vector<index> &insertedIndexes() const;
-
-    /**
      * Returns true iff an element was previously inserted at the given index.
      * @param idx
      */
@@ -115,6 +110,18 @@ public:
      * @param emptyValue new emptyValue
      */
     void resize(size_t size, T emptyValue);
+
+    /**
+     * Applies the given lambda to each inserted index and associated value
+     * @tparam ElementHandler
+     * @param lambda
+     */
+    template <typename ElementHandler>
+    void forElements(ElementHandler &&lambda) {
+        for (index i : usedIndexes) {
+            lambda(i, data[i]);
+        }
+    }
 
 private:
     std::vector<T> data;
@@ -171,11 +178,6 @@ index SparseVector<T>::upperBound() const {
 template <typename T>
 count SparseVector<T>::size() const {
     return usedIndexes.size();
-}
-
-template <typename T>
-const std::vector<index> &NetworKit::SparseVector<T>::insertedIndexes() const {
-    return usedIndexes;
 }
 
 template <typename T>
--- a/include/networkit/auxiliary/SparseVector.hpp
+++ b/include/networkit/auxiliary/SparseVector.hpp
@@ -1,0 +1,229 @@
+/*
+ * SparseVector.hpp
+ *
+ * Created: 2019-10-15
+ * Author: Armin Wiebigke
+ */
+// networkit-format
+#ifndef NETWORKIT_AUXILIARY_SPARSE_VECTOR_HPP_
+#define NETWORKIT_AUXILIARY_SPARSE_VECTOR_HPP_
+
+#include <algorithm>
+#include <cassert>
+#include <vector>
+
+#include <networkit/Globals.hpp>
+
+namespace NetworKit {
+
+/**
+ * A vector that imitates a map with unsigned integer keys.
+ * This class has faster access than a map, but needs space linear in the maximum key value.
+ */
+template <typename T>
+class SparseVector {
+public:
+    SparseVector();
+
+    /**
+     * Construct an empty vector. Empty values are created using the default constructor.
+     * @param size upper bound for the maximum usable index
+     */
+    explicit SparseVector(index size);
+
+    /**
+     * Construct an empty vector.
+     * @param size upper bound for the maximum usable index
+     * @param emptyValue value used for empty entries
+     */
+    SparseVector(index size, T emptyValue);
+
+    /**
+     * Resize the vector so that indexes up to size-1 can be used.
+     */
+    void setUpperBound(index size);
+
+    /**	 *
+     * @return the upper bound of the indexes that can be used
+     */
+    index upperBound() const;
+
+    /**
+     * @return the number of inserted elements.
+     */
+    count size() const;
+
+    /**
+     * Insert an value at position a given position.
+     * @param i index where the value is inserted
+     * @param value
+     */
+    void insert(index i, T value);
+
+    /**
+     * Access operator. Before accessing an element, insert it by using the insert() method.
+     */
+    T &operator[](index i);
+
+    /**
+     * Const access operator. Before accessing an element, insert it by using the insert() method.
+     */
+    const T &operator[](index i) const;
+
+    /**
+     * Get all indexes into which elements were inserted.
+     */
+    const std::vector<index> &insertedIndexes() const;
+
+    /**
+     * Returns true iff an element was previously inserted at the given index.
+     * @param idx
+     */
+    bool indexIsUsed(index idx);
+
+    void insertOrSet(index i, T value);
+
+    /**
+     * Remove all indexes for which the value is set to the emptyValue.
+     */
+    void removeUnusedIndexes() {
+        auto new_end = std::remove_if(usedIndexes.begin(), usedIndexes.end(),
+                                      [&](index i) { return data[i] == emptyValue; });
+        usedIndexes.erase(new_end, usedIndexes.end());
+    }
+
+    /**
+     * Reset all values to the default value, so it is "empty". The upper bound is not changed.
+     */
+    void reset();
+
+    /**
+     * Clear the vector, setting the upper bound of usable indexes to 0.
+     */
+    void clear();
+
+    /**
+     * Reallocate the datastructure if size exceeds current upper bound
+     * This is different from setUpperBound() since we want to make sure both usedIndexes and data
+     * are allocated on the socket of the calling thread
+     * @param size
+     * @param emptyValue new emptyValue
+     */
+    void resize(size_t size, T emptyValue);
+
+    /**
+     * Expose internal vector
+     */
+    std::vector<T> &getVector();
+
+    void clearIndexes() { usedIndexes.clear(); }
+
+    void resetEntry(index i) { data[i] = emptyValue; }
+
+    bool isClean() {
+        return usedIndexes.empty() && std::all_of(data.begin(), data.end(), [&](const T &x) {
+                   return x == emptyValue;
+               });
+    }
+
+private:
+    std::vector<T> data;
+    std::vector<index> usedIndexes;
+    T emptyValue;
+};
+
+template <typename T>
+SparseVector<T>::SparseVector() : emptyValue(T{}) {}
+
+template <typename T>
+SparseVector<T>::SparseVector(count size) : SparseVector(size, T{}) {}
+
+template <typename T>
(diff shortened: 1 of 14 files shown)


angriman (commented):
I recommend to integrate the `SparseVector` data structure into [tlx](https://github.com/tlx/tlx), which already implements several advanced data structures used by NetworKit.

larsgottesbueren:
> I recommend to integrate the `SparseVector` data structure into [tlx](https://github.com/tlx/tlx), which already implements several advanced data structures used by NetworKit.

I prefer not to. It doesn't really fit tlx since it's pretty basic and more of a programming pattern than a data structure. Also, I don't know how actively Timo maintains tlx, so that would be a whole other step that can take a lot of time.
Is there a reason not to integrate it here directly? If so, I would just remove the SparseVector class and employ the pattern manually where needed.

angriman:
The only reason is to avoid to implement in NetworKit general-purpose data structures because the main focus of the project are graph algorithms. However, I see that `SparseVector` is quite simple and might not be a good fit for tlx either, so I guess it's fine if we make an exception (as we already did in the past) and we integrate it into NetworKit.

Two first general comments:
- Try to clean up the commit history
- Make sure you follow our [style guide](https://networkit.github.io/dev-docs/DevGuide.html#style-guide) and format your code accordingly

larsgottesbueren:
I added an include in PrioQueue.hpp which was necessary for the code to compile (after I upgraded my compilers this morning).

include/networkit/auxiliary/SparseVector.hpp
```diff
@@ -0,0 +1,229 @@
...
+#include <networkit/Globals.hpp>
+
+namespace NetworKit {
+
+/**
+ * A vector that imitates a map with unsigned integer keys.
+ * This class has faster access than a map, but needs space linear in the maximum key value.
+ */
+template <typename T>
+class SparseVector {
```
angriman, line 24: Does this data structure bring any advantage? As far as I see it does not save space or time.

michitux: Save time or space compared to what other data structure? This data structure is a pattern we already use in NetworKit, e.g., in `PLM`, which could be refactored to use this new class. Compared to a hash map, it gives you guaranteed constant time operations and is also faster in practice. Note that it cannot always replace a hash map, only in cases where you can spend the time to initialize it once in time linear in the maximum key value, after which you get constant time insertion and querying of elements and reset linear in the number of inserted elements.

angriman: Compared to a std vector. I was a bit confused by the name of this class, with `Sparse` I expected some sort of compression strategy like the one implemented a CSR matrix. This is not the case here but I see that `usedIndexes` is used to iterate over the "used" slots only, which is of course faster than iterating over all the elements if the elements stored are sparse.

hmeyerhenke: @angriman : What you describe is the compression strategy. You don't save space in this particular case, but time.

networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -0,0 +1,583 @@
...
+    calculateInitialClusterCutAndVolume();
+    timer.stop();
+    DEBUG("init ", timer.elapsedMilliseconds(), " ms");
+
+#ifndef NDEBUG
+    updatePLogPSums();
+#endif
+
+    bool clusteringChanged = false;
+    std::vector<node> nodes = GraphTools::nodeSet(*G);
```
angriman, line 74: Use `std::vector<node> nodes{G->nodeRange().begin(), G->nodeRange().end};`

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -259,6 +259,17 @@ Graph copyNodes(const Graph &G) {
...
+std::vector<node> nodeSet(const Graph &G) {
+    std::vector<node> nodes;
+    nodes.reserve(G.numberOfNodes());
+    for (node u = 0; u < G.upperNodeIdBound(); ++u) {
+        if (G.hasNode(u)) {
+            nodes.push_back(u);
+        }
+    }
+    return nodes;
+}
```
angriman, line 271: Once you use the graph nodes iterator, this can be removed.

networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+    assert(vol == G->weightedDegree(u, true));
+
+    if (neighborClusterWeights.size() > 0) {
+        // Calculate best cluster
+        index targetCluster = currentCluster;
+        double weightToTargetCluster = weightToCurrent;
+        double bestChange = fitnessChange(u, vol, loop, currentCluster, currentCluster,
+                                          weightToCurrent, weightToCurrent, totalCut);
+        for (index neighborCluster : neighborClusterWeights.insertedIndexes()) {
```
angriman, line 307: I think it would be better to add a method in `SparseVector` that iterates over all the "used" slots (e.g., `forNonZeroElements` or something like that).

include/networkit/auxiliary/SparseVector.hpp
```diff
@@ -0,0 +1,229 @@
...
+     */
+    const std::vector<index> &insertedIndexes() const;
+
+    /**
+     * Returns true iff an element was previously inserted at the given index.
+     * @param idx
+     */
+    bool indexIsUsed(index idx);
+
+    void insertOrSet(index i, T value);
```
angriman, line 84: Documentation is missing for this method.

include/networkit/auxiliary/SparseVector.hpp
```diff
@@ -0,0 +1,229 @@
...
+     * are allocated on the socket of the calling thread
+     * @param size
+     * @param emptyValue new emptyValue
+     */
+    void resize(size_t size, T emptyValue);
+
+    /**
+     * Expose internal vector
+     */
+    std::vector<T> &getVector();
```
angriman, line 117: This is never used, either remove it or return a const ref.

include/networkit/auxiliary/SparseVector.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+    void clearIndexes() { usedIndexes.clear(); }
+
+    void resetEntry(index i) { data[i] = emptyValue; }
+
+    bool isClean() {
+        return usedIndexes.empty() && std::all_of(data.begin(), data.end(), [&](const T &x) {
+                   return x == emptyValue;
+               });
+    }
```
angriman, line 127: Documentation is missing.

include/networkit/community/LouvainMapEquation.hpp
```diff
@@ -0,0 +1,147 @@
...
+     * Calculate the change in the map equation if the node is moved from its current cluster to the
+     * target cluster. To simplify the calculation, we remove terms that are constant for all target
+     * clusters. As a result, "moving" the node to its current cluster gives a value != 0, although
+     * the complete map equation would not change.
+     */
+    double fitnessChange(node, double degree, double loopWeight, node currentCluster,
+                         node targetCluster, double weightToTarget, double weightToCurrent,
+                         double totalCutCurrently);
+
+    count idiv_ceil(count a, count b) const { return (a + b - 1) / b; }
```
angriman, line 115: Coding style

include/networkit/community/LouvainMapEquation.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    std::pair<count, count> chunk_bounds(count i, count n, count chunk_size) const {
```
angriman, line 117: Coding style

include/networkit/community/LouvainMapEquation.hpp
```diff
@@ -0,0 +1,147 @@
...
+        return std::make_pair(i * chunk_size, std::min(n, (i + 1) * chunk_size));
+    }
+
+    ParallelizationType
+    convert_string_to_parallelization_type(const std::string &para_strat) const {
```
angriman, line 122: Coding style

networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -0,0 +1,583 @@
...
+}
+
+void LouvainMapEquation::run() {
+    if (hasRun)
+        throw std::runtime_error("Algorithm was already run!");
+
+    Aux::SignalHandler handler;
+
+    Aux::Timer timer;
+    timer.start();
```
angriman, line 47: Consider to remove time measurements.

networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -0,0 +1,583 @@
...
+
+#pragma omp for schedule(guided) nowait
+            for (omp_index i = 0; i < static_cast<omp_index>(nodes.size()); ++i) {
+                if (tryLocalMove(nodes[i], neighborClusterWeights, dummyMoves, false)) {
+                    nm += 1;
+                }
+            }
+
+#pragma omp atomic
+            nodesMoved += nm;
```
angriman, line 133: Here you could use an OpenMP reduction.

larsgottesbueren: Yes, but the loop before that is nowait, so there will be little to no contention.

networkit/cpp/community/LouvainMapEquation.cpp
larsgottesbueren: Do you mean the `forNonZeroElements` function should take a lambda and do the iteration, or should I rather expose a range of iterators?

angriman: I think that for this case taking a lambda is fine since you don't need to interrupt the loop.

angriman: Iterators could be added later if needed.

networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -0,0 +1,565 @@
...
+// networkit-format
+
+#include <algorithm>
+#include <cassert>
+
+#include <omp.h>
+
+#include <networkit/auxiliary/Parallelism.hpp>
+#include <networkit/auxiliary/SignalHandling.hpp>
+#include <networkit/auxiliary/Timer.hpp>
```
angriman, line 18: `Timer.hpp` is not needed anymore.

include/networkit/community/LouvainMapEquation.hpp
```diff
@@ -0,0 +1,147 @@
...
+     * @param[in] maxIterations maximum number of iterations for move phase
+     * @param[in] parallelization strategy (default synchronous):
+     *            none
+     *            relaxmap    -> one lock per community to update cuts
+     *            synchronous -> work on stale cuts and volumes, update in second step
+     *
+     */
+    explicit LouvainMapEquation(const Graph &graph, bool hierarchical = false,
+                                count maxIterations = 32,
+                                const std::string parallelizationStrategy = "relaxmap")
```
fabratu, line 50: CI (clang-tidy checks) complains about this: `the const qualified parameter 'parallelizationStrategy' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param,-warnings-as-errors]`

It would be good to also fix the narrowing-conversion warnings in the CI-log (at least in the files you added). However they only trigger warnings as of now - so it is not neccessary.

larsgottesbueren: I'm already on it. Is this check new and how does it not fail on the main branch? I rebased on the most recent version.

angriman (approved):

PR #755, opened 2021-05-18
Refactor edgelistwriter

### LINKED ISSUE #754: Abysmal write performance
I only tested this for the EdgeListWriter specifically, but it's throughput is extremely bad (at most 100-200 kbit/s)
This can easily be fixed by replacing all std::endl in the respective write handles with newlines (and calling flush once at the end if needed. I dont know, I did, just to be safe.) which yields me about 40mb/s

I.e. replace this

`auto writeUnweightedEdge = [&](node u, node v){
        file << (u + firstNode) << separator << (v + firstNode) << std::endl;
    };`

with this

`auto writeUnweightedEdge = [&](node u, node v){
        file << (u + firstNode) << separator << (v + firstNode) << "\n";
    };`

CoverWriter seems to do the same, and there's probably more.
I would've sent a PR but I literally don't have a clue how they work and I don't wanna mess anything up for you.
Greets.

angriman:
As suggested in #754, `std::endl` may degrade the performance of graph writers. This PR replaces `std::endl` with `\n` in the `io` module,
Closes #754

### DIFF
--- a/networkit/cpp/io/CoverWriter.cpp
+++ b/networkit/cpp/io/CoverWriter.cpp
@@ -20,7 +20,7 @@ void CoverWriter::write(Cover &zeta, const std::string &path) const {
         for (auto &v : nodes) {
             file << v << " ";
         }
-        file << std::endl;
+        file << '\n';
     }
 }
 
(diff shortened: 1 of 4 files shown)


angriman:
Without that I cannot compile NetworKit with new compilers like gcc 11. Since it is a trivial fix I think it's fine to include it (we have to do that anyway).

fabratu:
We have also seen this recently with clang-11 builds on CI. Background is likely, that gcc-11 introduced changes in implicit header dependencies https://gcc.gnu.org/gcc-11/porting_to.html Maybe clang-11 still uses some header/libs from gnu-gcc. 

For #749 I also had to add some includes, otherwise CI doesn't succeed anymore for Ubuntu-runners. Maybe it is a better idea to gather this in a different PR to add all includes?

angriman:
Alright, I removed the additional include.

fabratu (approved):
LGTM.

avdgrinten (approved):
LGTM.

PR #756, opened 2021-05-19
Python: Add support for Windows 7-10

fabratu:
This PR adds support for building/installing NetworKit for modern Windows-OS. Since only msvc-builds are supported by now, the port will be labeled as "experimental" as of version 9.0.

So far the following works:

- Building C++ code via Visual Studio project files, NMake, Ninja (msvc)
- Creating Python extensions with default installation and external tlx
- Installing via pip and using Python-module with CPython 3.9
- Running Python unit tests + notebooks
- CI-pipeline for native and non-native builds

Currently not supported / tested:
- mingw64/msys2
- cygwin

Errors:
~~- CI-builds for Windows should fail on errors by default. However this is not alway the case (currently for `core`-build)~~
~~- Non-static builds produce errors during compilation. I think this could be ignored, if there are no other errors left.~~
~~- Reproducibility of random numbers. This is quite puzzling. Currently the test for randomEdges is disabled on Windows, cause it produces different results for same random seeds. The following happens for both CI and non-virtualized OS if the test is enabled:~~

Caveats:
- A shared build on Windows is currently not supported - this is due to the fact, that (apart from global shared variables) no other symbols are exported.
- This also disables building against an external core-lib, since dlls (such as Cython-extensions) and the static core-lib must reside in the same directory.
- CI uses git-bash (mingw) for test-scripts. This is the only way to reliably fail on errors. For Powershell Core/Desktop and cmd several errors are hidden during both cmake and Python script phases. Using git-bash removes the possibility to test for Windows-styled paths.


avdgrinten:
I think it is not a problem that only static msvc builds are supported for now.

However, the reproducibility issue of the RNG is an issue that should be fixed before merging. Does this also happen for the underlying `std::mt19937` RNG that we use? My suspicion is that this is somehow caused by NetworKit's code and not by `std::mt19937`.

We should also ensure that errors always fail the CI build. Why happens on error right now?

fabratu:
> We should also ensure that errors always fail the CI build. Why happens on error right now?

Hard to tell. With the previous way of building the C++ core (creating and building Visual Code project files) errors in the pipelines caused also an immediate failure of GA run. For cmake-configuration errors this seems not to be the case. The pipeline fails, but GA marks the run as success. This is likely fixable though.

fabratu:
While the commit history needs some proper cleaning, all CI-builds look fine now.

I think it is a good idea to get some first feedback on the PR. The building system takes leverage of some workarounds concerning placement of libs/support for build-variants.

include/networkit/auxiliary/GlobalState.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <mutex>
+
+#include <networkit/auxiliary/Log.hpp>
+#include <networkit/auxiliary/SignalHandling.hpp>
+
+#include <networkit/networkit_export.h>
+
+namespace Aux {
+
+namespace GlobalState {
```
avdgrinten, line 19: Instead of using `Aux::GlobalState`, I'd prefer `NetworKit::GlobalState`. Further namespaces within `GlobalState` should not be necessary.

include/networkit/auxiliary/GlobalState.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace Random {
+
+NETWORKIT_EXPORT uint64_t getSeed();
+NETWORKIT_EXPORT void setSeed(uint64_t seed);
+
+NETWORKIT_EXPORT uint64_t getGlobalSeed();
+NETWORKIT_EXPORT void incGlobalSeed();
+
+NETWORKIT_EXPORT void setSeedUseThreadId(bool useThreadId);
+NETWORKIT_EXPORT bool getSeedUseThreadId();
```
avdgrinten, line 30: I wonder if this is a good time to get rid of non-global seeds. Are there any users of `setSeed()` without also specifying `useThreadId` as `true`?

include/networkit/auxiliary/Log.hpp
```diff
@@ -73,28 +75,12 @@ void setLogLevel(const std::string &logLevel);
...
-void setLogLevel(LogLevel p = LogLevel::info);
-
-void setPrintTime(bool b);
-bool getPrintTime();
-
-void setPrintLocation(bool b);
-bool getPrintLocation();
-
-void setLogfile(const std::string &filename);
-}
```
avdgrinten, line 88: We need to implement these functions by delegating to the `GlobalState` ones (to avoid API breaks).

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifndef NETWORKIT_DEPRECATED_NO_EXPORT
+#  define NETWORKIT_DEPRECATED_NO_EXPORT NETWORKIT_NO_EXPORT NETWORKIT_DEPRECATED
+#endif
+
+#if 0 /* DEFINE_NO_DEPRECATED */
+#  ifndef NETWORKIT_NO_DEPRECATED
+#    define NETWORKIT_NO_DEPRECATED
+#  endif
+#endif
```
avdgrinten, line 46: What's this?

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
...
+#  define NETWORKIT_DEPRECATED __declspec(deprecated)
+#endif
+
+#ifndef NETWORKIT_DEPRECATED_EXPORT
+#  define NETWORKIT_DEPRECATED_EXPORT NETWORKIT_EXPORT NETWORKIT_DEPRECATED
+#endif
+
+#ifndef NETWORKIT_DEPRECATED_NO_EXPORT
+#  define NETWORKIT_DEPRECATED_NO_EXPORT NETWORKIT_NO_EXPORT NETWORKIT_DEPRECATED
+#endif
```
avdgrinten, line 40: I don't understand the purpose of this code.

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
+
```
avdgrinten, line 1: The naming convention of this file does not match NetworKit's. Maybe `Macros.hpp` would be a better name?

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
...
+#        define NETWORKIT_EXPORT
+#      endif
+#    else
+#      if defined(WIN32)
+#        define NETWORKIT_EXPORT __declspec(dllimport)
+#      else
+#        define NETWORKIT_EXPORT 
+#      endif
+#    endif
+#  endif
```
avdgrinten, line 23: Why is this guarded by `#ifndef NETWORKIT_EXPORT`?

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
...
+
+#  ifndef NETWORKIT_NO_EXPORT
+#    define NETWORKIT_NO_EXPORT 
+#  endif
```
avdgrinten, line 27: What does this do?

include/networkit/networkit_export.h
```diff
@@ -0,0 +1,48 @@
+
+#ifndef NETWORKIT_EXPORT_H
+#define NETWORKIT_EXPORT_H
+
+#ifdef NETWORKIT_STATIC_DEFINE
+#  define NETWORKIT_EXPORT
+#  define NETWORKIT_NO_EXPORT
+#else
```
avdgrinten, line 8: Why is this case needed? I think the entire file needs some comments to explain the purpose of each macro.

networkit/cpp/auxiliary/GlobalState.cpp
```diff
@@ -0,0 +1,146 @@
+// networkit-format
```
avdgrinten, line 1: Maybe this file should not be in `Aux` but at the top level in `networkit/cpp/` since it (potentially) affects all modules.

networkit/cpp/auxiliary/Random.cpp
```diff
@@ -10,13 +10,14 @@
...
+#include <networkit/auxiliary/GlobalState.hpp>
 #include <networkit/auxiliary/Random.hpp>
 
 namespace Aux {
 namespace Random {
 
-static std::atomic<uint64_t> seedValue{0};
-static std::atomic<bool> seedUseThreadId{false};
+//static std::atomic<uint64_t> seedValue{0};
+//static std::atomic<bool> seedUseThreadId{false};
```
avdgrinten, line 20: This needs to be fixed.

networkit/cpp/auxiliary/Random.cpp
```diff
@@ -26,12 +27,15 @@ static std::atomic<bool> seedUseThreadId{false};
 // relaxed.
 // As long as globalSeedGeneration is zero, the getURNG()
 // ignores these two variables.
-static std::atomic<uint64_t> globalSeedGeneration{0}; //< global seed generation, updated on every setSeed-call
+//static std::atomic<uint64_t> globalSeedGeneration{0}; //< global seed generation, updated on every setSeed-call
```
avdgrinten, line 30: Likewise.

networkit/cpp/auxiliary/Random.cpp
```diff
@@ -26,12 +27,15 @@ static std::atomic<bool> seedUseThreadId{false};
...
 void setSeed(uint64_t seed, bool useThreadId) {
-    seedValue.store(seed, std::memory_order_relaxed);
-    seedUseThreadId.store(useThreadId, std::memory_order_relaxed);
-    globalSeedGeneration.fetch_add(1, std::memory_order_release) ;
+    Aux::GlobalState::Random::setSeed(seed);
+    //seedValue.store(seed, std::memory_order_relaxed);
+    Aux::GlobalState::Random::setSeedUseThreadId(useThreadId);
+    //seedUseThreadId.store(useThreadId, std::memory_order_relaxed);
+    Aux::GlobalState::Random::incGlobalSeed();
+    //globalSeedGeneration.fetch_add(1, std::memory_order_release);
```
avdgrinten, line 38: Likewise.

avdgrinten: And everywhere else (do not keep commented out code unless there is a *very* good reason).

networkit/cpp/numerics/LAMG/MultiLevelSetup.cpp
```diff
@@ -1,247 +0,0 @@
-/*
```
avdgrinten, line 1: @angriman has more knowledge of the LAMG than I do. What do we lose if we remove these specializations? (vs. the generic implementation in the header?)

include/networkit/auxiliary/GlobalState.hpp
fabratu: Yeah, we should do it. However it is better to move it to another PR.

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
-			INTERFACE_INCLUDE_DIRECTORIES "${NETWORKIT_EXT_TLX}/include/")
+	if(WIN32)
+		set_target_properties(tlx PROPERTIES
+				IMPORTED_LOCATION "${NETWORKIT_EXT_TLX}/lib/tlx.lib"
+				INTERFACE_INCLUDE_DIRECTORIES "${NETWORKIT_EXT_TLX}/include/")
+	else()
+		set_target_properties(tlx PROPERTIES
+				IMPORTED_LOCATION "${NETWORKIT_EXT_TLX}/lib/libtlx.a"
+				INTERFACE_INCLUDE_DIRECTORIES "${NETWORKIT_EXT_TLX}/include/")
+	endif()
```
avdgrinten, line 240: Instead of duplicating these lines, maybe it's better to define a `${NETWORKIT_LIB_SUFFX}`? Maybe CMake even defines such a variable already.

CMakeLists.txt
```diff
@@ -656,7 +712,12 @@ install(DIRECTORY extlibs/ttmath/ttmath
 # Prepare pkg-config file
 
 set(NETWORKIT_LIBNAME networkit)
-execute_process(COMMAND python3 -c "from version import version; print(version, end='')"
+if(WIN32)
+	set(NETWORKIT_CMAKE_PYTHON_INTERPRETER python)
+else()
+	set(NETWORKIT_CMAKE_PYTHON_INTERPRETER python3)
+endif()
+execute_process(COMMAND ${NETWORKIT_CMAKE_PYTHON_INTERPRETER} -c "from version import version; print(version, end='')"
```
avdgrinten, line 720: What about using [the FindPython3 module](https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3) that you are already using above?

CMakeLists.txt
```diff
@@ -428,13 +473,24 @@ if (NETWORKIT_BUILD_TESTS)
...
+			)
+		else()
+			target_link_libraries(networkit_tests
+				PRIVATE
+					gtest
+					networkit
+					tlx
+					OpenMP::OpenMP_CXX
+			)
+		endif()
```
avdgrinten, line 493: Can't we just call `target_link_libraries` twice to add another library?

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
+
+		set_target_properties(networkit_export PROPERTIES
+				CXX_STANDARD ${NETWORKIT_CXX_STANDARD}
+				COMPILE_FLAGS "${NETWORKIT_CXX_FLAGS}"
+				LINK_FLAGS "${NETWORKIT_LINK_FLAGS}")
+
+		install(TARGETS networkit
+				LIBRARY DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				ARCHIVE DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				RUNTIME DESTINATION "${NETWORKIT_LIB_DEST}/networkit")
```
avdgrinten, line 261: Why is this in `if(WIN32)`? It seems to affect the `networkit` library, not the new library containing the state?

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
 # NETWORKIT MODULES
 if(NETWORKIT_BUILD_CORE AND NETWORKIT_MONOLITH)
-	if(NETWORKIT_STATIC)
-		add_library(networkit networkit/cpp/networkit.cpp)
+
+	if(WIN32)
+		# Static and non-static is ignored for Windows-builds. Shared-build only is not supported and getting linked against multiple dll's 
+		# needs a dll-based globalstate to enable consistent global variable share.
+		add_library(networkit_export SHARED networkit/cpp/GlobalState.cpp)
+		add_library(networkit networkit/cpp/networkit.cpp)
```
avdgrinten, line 251: Why is this done?

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
 ################################################################################
 # NETWORKIT MODULES
 if(NETWORKIT_BUILD_CORE AND NETWORKIT_MONOLITH)
-	if(NETWORKIT_STATIC)
-		add_library(networkit networkit/cpp/networkit.cpp)
+
+	if(WIN32)
+		# Static and non-static is ignored for Windows-builds. Shared-build only is not supported and getting linked against multiple dll's 
+		# needs a dll-based globalstate to enable consistent global variable share.
+		add_library(networkit_export SHARED networkit/cpp/GlobalState.cpp)
```
avdgrinten, line 250: Instead of calling the library `networkit_export`, I would prefer `networkit_state`.

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
+				LIBRARY DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				ARCHIVE DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				RUNTIME DESTINATION "${NETWORKIT_LIB_DEST}/networkit")
+
+		install(TARGETS networkit_export
+				LIBRARY DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				ARCHIVE DESTINATION "${NETWORKIT_LIB_DEST}/networkit"
+				RUNTIME DESTINATION "${NETWORKIT_LIB_DEST}/networkit")
+
+		target_link_libraries(networkit PRIVATE OpenMP::OpenMP_CXX PUBLIC tlx networkit_export)
```
avdgrinten, line 268: OpenMP and TLX are also linked in the in the general case. Move these out of the `if(WIN32)` and only add `networkit_export` for WIN32.

CMakeLists.txt
```diff
@@ -311,13 +356,13 @@ if(NETWORKIT_PYTHON)
...
-			install(TARGETS ${TARGET_LIB}
-				LIBRARY DESTINATION "${NETWORKIT_LIB_DEST}/networkit")
+		target_link_libraries(${TARGET_LIB} PRIVATE OpenMP::OpenMP_CXX PUBLIC tlx)
+		install(TARGETS ${TARGET_LIB}
+			LIBRARY DESTINATION "${NETWORKIT_LIB_DEST}/networkit")
+
+		if(WIN32)
+			FIND_PACKAGE(Python3 COMPONENTS Interpreter Development)
+			target_link_libraries(${TARGET_LIB} PRIVATE ${Python3_LIBRARIES})
 		endif()
```
avdgrinten, line 366: Why is this in `if(WIN32)`? Don't we need to link against Python libs on Linux/macOS?

include/networkit/GlobalState.hpp
```diff
@@ -0,0 +1,94 @@
...
+#endif
+#else
+#if defined(WIN32)
+#define NETWORKIT_EXPORT __declspec(dllimport)
+#else
+#define NETWORKIT_EXPORT
+#endif
+#endif // networkit_EXPORTS
+#endif // NETWORKIT_EXPORT
+#endif // NETWORKIT_STATIC_DEFINE
```
avdgrinten, line 23: This is quite hard to read without indentation. Is the indentation removed by CMake? Is it also removed if we place it after the `#`?

include/networkit/GlobalState.hpp
```diff
@@ -0,0 +1,94 @@
+#ifndef NETWORKIT_GLOBAL_STATE_HPP_
+#define NETWORKIT_GLOBAL_STATE_HPP_
+
+#ifdef NETWORKIT_STATIC_DEFINE
+#define NETWORKIT_EXPORT
+#define NETWORKIT_NO_EXPORT
+#else
```
avdgrinten, line 7: Why is this case for? Can it be deleted?

include/networkit/GlobalState.hpp
```diff
@@ -0,0 +1,94 @@
...
+#ifndef NETWORKIT_EXPORT
+#ifdef networkit_EXPORTS
```
avdgrinten, line 9: Seems that we're using this definition from CMake to control the export behavior here. Is that hard coded by CMake? It does not follow our usual naming convention. I would prefer a name such as: `NETWORKIT_BUILDING_STATELIB` or something similar.

include/networkit/auxiliary/Log.hpp
```diff
@@ -44,7 +46,9 @@
 #	define TRACEPOINT LOG_AT(::Aux::Log::LogLevel::trace, "tracepoint")
 #endif // defined(NETWORKIT_RELEASE_LOGGING)
 
-namespace Aux { namespace Log {
+namespace Aux { 
+
+    namespace Log {
 
 struct Location {
```
avdgrinten, line 53: Indentation looks strange here.

networkit/test/test_graph.py
```diff
@@ -155,9 +156,10 @@ def testRandomEdgesReproducibility(self):
 				self.assertNotEqual(results[0][0], results[seed][0])
 
 			# assert results are consistent for same seeds
-			for repeats in results:
-				for x in repeats[1:]:
-					self.assertEqual(repeats[0], x)
+			if sys.platform != "win32":
+				for repeats in results:
+					for x in repeats[1:]:
+						self.assertEqual(repeats[0], x)
```
avdgrinten, line 162: Why is this change done?

notebooks/test_notebooks.py
```diff
@@ -17,6 +18,9 @@ def run_notebook(path):
     proc = ExecutePreprocessor(timeout=600, kernel_name='python3')
     proc.allow_errors = True
 
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith('win'):
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
avdgrinten, line 22: Why does this `if` check for?

setup.py
```diff
@@ -82,14 +82,15 @@ def determineCompiler(candidates, std, flags):
...
 	for compiler in candidates:
-		cmd = [compiler,"-o","test_build","-std={}".format(std)]
+		cmd = [compiler,"-o","test_build","-std{}".format(std)]
 		cmd.extend(flags)
 		cmd.append("sample.cpp")
 		try:
 			if subprocess.call(cmd,stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL) == 0:
 				os.remove("sample.cpp")
-				os.remove("test_build")
+				os.remove("test_build*")
```
avdgrinten, line 93: Why is that change done? Does it even work? I don't think `os.remove` expands shell wildcards.

CMakeLists.txt
```diff
@@ -222,32 +229,69 @@ if(NOT NETWORKIT_EXT_TLX)
...
 
 ################################################################################
 # NETWORKIT MODULES
 if(NETWORKIT_BUILD_CORE AND NETWORKIT_MONOLITH)
-	if(NETWORKIT_STATIC)
-		add_library(networkit networkit/cpp/networkit.cpp)
+
+	if(WIN32)
+		# Static and non-static is ignored for Windows-builds. Shared-build only is not supported and getting linked against multiple dll's 
+		# needs a dll-based globalstate to enable consistent global variable share.
```
avdgrinten, line 249: Doesn't it make more sense to error out if `NETWORKIT_STATIC` is not set on Windows?

setup.py
```diff
@@ -128,8 +134,12 @@ def buildNetworKit(install_prefix, externalCore=False, externalTlx=None, withTes
...
 	from sysconfig import get_paths, get_config_var
 	comp_cmd.append("-DNETWORKIT_PYTHON="+get_paths()['include']) #provide python.h files
-	comp_cmd.append("-DNETWORKIT_PYTHON_SOABI="+get_config_var('SOABI')) #provide lib env specification
-	if externalCore:
+	os_soabi = get_config_var('SOABI')
+	if os_soabi is None:
+		os_soabi =  get_config_var('EXT_SUFFIX').split(".")[1] # get_config_var('SOABI') is None on win32-systems
+	comp_cmd.append("-DNETWORKIT_PYTHON_SOABI="+os_soabi) #provide lib env specification
+	if externalCore and sys.platform != "win32":
+		# Since Windows build includes a core dll-lib, builds from different folders are not supported. Hence external core builds are disabled.
```
avdgrinten, line 142: Doesn't make more sense to error out on `externalCore`?

avdgrinten, line 142: Is that the only reason that externalCore is not supported? AFAICS the real reason is that we cannot build the core as a DLL, right?

CMakeLists.txt
fabratu: Good point. Had something like that in a previous version (introducing new if/else-cases), but this did not include a cmake variable.

CMakeLists.txt
fabratu: Right, maybe that is also a good point to directly switch to use the module on all platforms.

CMakeLists.txt
fabratu: Removed - mistake from a merge.

notebooks/test_notebooks.py
fabratu: https://github.com/jupyter/nbconvert/issues/1372

CMakeLists.txt
fabratu: Normally they are found by default. But maybe it is a good idea to use it on all systems.

notebooks/test_notebooks.py
avdgrinten: Add a comment to explain the problem and link to the issue.

include/networkit/GlobalState.hpp
fabratu: Yes-  and moving the intendation behind `#` does not work.

networkit/test/test_graph.py
fabratu: Left-over from merge.

setup.py
fabratu: It does work, but this is also a left-over from previous merge.

CMakeLists.txt
fabratu: Is changed now (also the other occassions). Reasoning was not to interrupt the user with build-errors, which are not expected.

CMakeLists.txt
```diff
@@ -656,7 +698,10 @@ install(DIRECTORY extlibs/ttmath/ttmath
 # Prepare pkg-config file
 
 set(NETWORKIT_LIBNAME networkit)
-execute_process(COMMAND python3 -c "from version import version; print(version, end='')"
+if(${Python3_Interpreter_FOUND})
+	set(NETWORKIT_CMAKE_PYTHON_INTERPRETER ${Python3_EXECUTABLE})
+endif()
```
avdgrinten, line 703: What if the interpreter is not found? Can we add something like a `REQUIRED` clause to `find_package`?

fabratu: Sure, should have used `REQUIRED` in the first place.

avdgrinten: But then we can also remove this `if`.

README.md
```diff
@@ -26,14 +26,15 @@ Furthermore, NetworKit's core can be built and used as a native library if neede
...
+- A modern C++ compiler, e.g.: [g++] (&gt;= 5.3), [clang++] (&gt;= 3.9) or MSVC (&gt;= 14.13)
 - OpenMP for parallelism (usually ships with the compiler)
 - Python3 (3.5 or higher is supported)
   - Development libraries for Python3. The package name depends on your distribution. Examples: 
     - Debian/Ubuntu: `apt-get install python3-dev`
     - RHEL/CentOS: `dnf install python3-devel`
+    - Windows: Use the official release installer from [www.python.org][https://www.python.org/downloads/windows/]
 - [Pip]
-- [CMake] version 3.5 or higher (e.g., `pip3 install cmake`)
+- [CMake] version 3.16 or higher (e.g., `pip3 install cmake`)
```
avdgrinten, line 37: Are there any distributions that do not have 3.16 available? 3.16 is only 2 years old, is it?.

include/networkit/GlobalState.hpp
```diff
@@ -0,0 +1,81 @@
+#ifndef NETWORKIT_GLOBAL_STATE_HPP_
+#define NETWORKIT_GLOBAL_STATE_HPP_
+
+#ifndef NETWORKIT_EXPORT
+#if defined(WIN32)
+#define NETWORKIT_EXPORT __declspec(dllexport)
+#else
+#define NETWORKIT_EXPORT
+#endif
+#endif // NETWORKIT_EXPORT
```
avdgrinten, line 10: This looks broken. Surely we also need `dllimport`? We should set an appropriate variable in the CMake script to distinguish between export and import (= the `NETWORKIT_BUILDING_STATELIB` that I mentioned earlier).

include/networkit/auxiliary/Log.hpp
```diff
@@ -6,49 +6,59 @@
...
 
+#include <tlx/define/deprecated.hpp>
+
 #ifdef _MSC_VER
-    #define NETWORKT_PRETTY_FUNCTION __FUNCSIG__
+#define NETWORKT_PRETTY_FUNCTION __FUNCSIG__
 #else
-    #define NETWORKT_PRETTY_FUNCTION __PRETTY_FUNCTION__
+#define NETWORKT_PRETTY_FUNCTION __PRETTY_FUNCTION__
 #endif // _MSC_VER
```
avdgrinten, line 15: This file has a lot of unrelated reformatting. Please remove these changes or factor them out into a commit of their own (that *only* does reformatting and nothing else).

include/networkit/GlobalState.hpp
avdgrinten: Also, there is still no indentation.

networkit/cpp/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+networkit_add_module(globals
+    GlobalState.cpp
+    )
```
avdgrinten, line 3: There is already a `global` module. Let's name this `globalstate`, not `globals`.

README.md
fabratu: Thats certainly possible. Thats why we suggest to install cmake via `pip3`. Since they deploy via wheel-package, it is supported up to CentOS 5.

Another way would be to document, not to mix conda and pip environments.

include/networkit/GlobalState.hpp
fabratu: For intendation we need to change `.clang_format` config. This will be done in a seperate PR, since it likely changes more than only a few files.

README.md
fabratu: Based on external discussion: Setting the minimum cmake version to 3.6 (youngest release in 5 year support span). Also adding an conditional for `FindPython` (supported since 3.12). For older versions we use `FindPythonInterp` and `FindPythonLibs`.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(NETWORKIT_CLANG_TIDY "Check code with clang-tidy" OFF)
 option(NETWORKIT_FLATINSTALL "Install into a flat directory structure (useful when building a Python package)" OFF)
 option(NETWORKIT_COVERAGE "Build with support for coverage" OFF)
 option(NETWORKIT_SANITY_CHECKS "Algorithms will perform further (expensive) checks" OFF)
+option(NETWORKIT_BUILDING_STATELIB "Builds networkit/GlobalState.hpp as seperate library." OFF)
```
avdgrinten, line 21: Why is this a configuration option?

CMakeLists.txt
```diff
@@ -75,6 +76,8 @@ elseif(NOT NETWORKIT_RELEASE_LOGGING STREQUAL "OFF")
 	message(FATAL_ERROR "Unsupported setting ${NETWORKIT_RELEASE_LOGGING} for NETWORKIT_RELEASE_LOGGING")
 endif()
 
+message(STATUS "COMPILER ID: ${CMAKE_CXX_COMPILER_ID}")
```
avdgrinten, line 79: Do we want to keep these messages?

CMakeLists.txt
```diff
@@ -223,31 +231,59 @@ if(NOT NETWORKIT_EXT_TLX)
...
+			message(FATAL_ERROR
+					"Windows does only support static builds. Call cmake with '-DNETWORKIT_STATIC=ON'.")
+		endif()
 		add_library(networkit SHARED networkit/cpp/networkit.cpp)
 	endif()
 
+	if(WIN32 AND NETWORKIT_BUILDING_STATELIB)
+		# For Windows all global shared variables have to be exported additionally as dll
+		add_library(networkit_state SHARED networkit/cpp/GlobalState.cpp)
+		set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /DNETWORKIT_BUILDING_STATELIB")
```
avdgrinten, line 255: Instead of changing `NETWORKIT_CXX_FLAGS`, we should pass this directly to `set_target_properties` (since `NETWORKIT_CXX_FLAGS` applies to all build targets).

CMakeLists.txt
```diff
@@ -223,31 +231,59 @@ if(NOT NETWORKIT_EXT_TLX)
...
 		add_library(networkit networkit/cpp/networkit.cpp)
 	else()
+		if(WIN32)
+			message(FATAL_ERROR
+					"Windows does only support static builds. Call cmake with '-DNETWORKIT_STATIC=ON'.")
+		endif()
 		add_library(networkit SHARED networkit/cpp/networkit.cpp)
 	endif()
 
+	if(WIN32 AND NETWORKIT_BUILDING_STATELIB)
```
avdgrinten, line 252: What happens if `NETWORKIT_BUILDING_STATELIB` is disabled? Does the build still work? AFAICS it won't because GlobalState.cpp is not compiled in.

CMakeLists.txt
avdgrinten: (Also, this should be `.cpp`, not `.hpp`?)

CMakeLists.txt
avdgrinten: I would be fine with discarding the configuration option to avoid this problem for now (but of course, we need to keep the `/DNETWORKIT_BUILDING_STATELIB` flag).

CMakeLists.txt
```diff
@@ -656,7 +705,9 @@ install(DIRECTORY extlibs/ttmath/ttmath
 # Prepare pkg-config file
 
 set(NETWORKIT_LIBNAME networkit)
-execute_process(COMMAND python3 -c "from version import version; print(version, end='')"
+set(NETWORKIT_CMAKE_PYTHON_INTERPRETER ${Python_EXECUTABLE})
```
avdgrinten, line 708: Why do we use this `set()` instead of replacing all occurrences of `NETWORKIT_CMAKE_PYTHON_INTERPRETER` by `Python_EXECUTABLE`?

avdgrinten (approved):
LGTM now.

PR #757, opened 2021-05-25
Fix EdgeListReader

angriman:
In #727 we did not really fix the issue of `EdgeListReader` because https://github.com/networkit/networkit/blob/9128af222f2ddac8bfe079aa9697ed87026c39a9/networkit/cpp/io/EdgeListReader.cpp#L27 simply changes the input parameter of the constructor, not the actual class member. This PR:
- Performs the actual fix: `this->firstNode = 0;`
- Adds a test that triggers a segfault with the previous implementation


angriman:
While fixing this I noticed that several other clang-tidy checks were not caught due to an issue in our `.clang-tidy` file. I included this fix in #758.

PR #758, opened 2021-05-25
Fix clang-analyzer and clang-tidy checks

angriman:
This PR:
- Fixes the `.clang-tidy` file: it was malformed and yaml failed to parse it correctly
- Fixes some warnings that went unnoticed due to the previous issue (not all of them, `cppcoreguidelines-narrowing-conversions` are too many to be fixed at once, so they are not treated as errors anymore) 
- Adds several `clang-analyzer` checks and fixes the warnings they trigger. However, I am not sure we want to include _all_ those clang-analyzer checks, so I am open to suggestions.
- Adds a new check that makes CI fail if `.clang-format` or `.clang-tidy` are malformed.


avdgrinten:
How did a malformed YAML file go unnoticed? Shouldn't it fail the CI run?

.clang-tidy
```diff
@@ -66,4 +111,4 @@ CheckOptions:
   - key:   readability-identifier-naming.TemplateParameterCase
     value: CamelCase
   - key:   performance-unnecessary-value-param.AllowedTypes
-    value: [Ff]unction$
+    value: '[Ff]unction$'
```
avdgrinten, line 114: I assume that was the reason why it was malformed?

.clang-tidy
```diff
@@ -44,10 +48,51 @@ Checks: >
 # To turn all warnings into errors, simply write '*'
 # in the first line after WarningsAsErrors
 WarningsAsErrors:  >
-    *
+    bugprone-assert-side-effect,
```
avdgrinten, line 51: Isn't there an easier way to turn all but a single check into an error? Something like `*,-foobar`?

angriman:
Probably a failure in reading the configuration file is not treated as an error and the build continues. In the original file for example if I run `clang-tidy -dump-config` I get
```
YAML:69:13: error: unexpected scalar
    value: [Ff]unction$
            ^
Error parsing .../.clang-tidy: Invalid argument
---
Checks:          'clang-diagnostic-*,clang-analyzer-*'
WarningsAsErrors: ''
HeaderFilterRegex: ''
AnalyzeTemporaryDtors: false
FormatStyle:     none
User:            eugenio
CheckOptions:
... list of check options ...
```
and the configurations written in `.clang-tidy` are just ignored. They are not ignored after the fix.

.clang-tidy
angriman: That's indeed much better and it works.

.clang-tidy
angriman: Yes exactly (see my comment below).

fabratu:
Based on the minor version of clang-tidy, the handling of this malforming seems to have changed. We have recent runs, where the clang-tidy complains about the `unexpected scalar` and fails immediately. There are also runs, where additional warnings are triggered, but the run completes. Not so recent runs didn't complain and didn't catch any additional warnings.

Bottom line: good idea to check for malformed configs.

fabratu:
Since it better fits in this PR, maybe move/copy the clang-tidy fix from #758 here?

angriman:
That's a good idea, after we merge this CI should work correctly again and then I can rebase the other PRs.

fabratu:
While I have no practical experience using these checks, from the description all `-core.*`, `-cplusplus.*`, `-deadcode.*` and `-nullability.*`seems reasonable. Some of the unsafe patterns will likely get attention in PR-reviews, but for new developers this may help in avoiding mistakes with memory leaks/pointer.

networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -317,7 +317,7 @@ void ApproxCloseness::runOnPivot(index i, const std::vector<node> &pivot, const
             }
             list[j].clear();
         } else { // u is another node, estimate its score
-            if (epsilon == 0 || (d <= delta[u] * (1.0 / epsilon - 1.0)) || ((last[pivot[u]] == i) && (dist[pivot[u]] <= delta[u] / epsilon))) {
+            if (epsilon == 0 || (d <= delta[u] * (1.0 / epsilon - 1.0)) || ((last[pivot[u]] == i) && ((double) dist[pivot[u]] <= delta[u] / epsilon))) {
```
fabratu, line 320: Why using no `static_cast` here ?

networkit/cpp/centrality/ApproxGroupBetweenness.cpp
```diff
@@ -39,9 +39,9 @@ void ApproxGroupBetweenness::run() {
     std::vector<int64_t> bucketInitializer(n);
     std::vector<std::vector<node>> incidencyList(n);
     std::vector<count> hyperEdges;
-    count samples = groupSize * log(n) / (epsilon * epsilon);
+    const auto samples = static_cast<count>(std::ceil((double)groupSize * std::log(n) / (epsilon * epsilon)));
```
fabratu, line 42: Why using no `static_cast` here ?

networkit/cpp/centrality/Betweenness.cpp
```diff
@@ -84,11 +84,11 @@ void Betweenness::run() {
 
     if (normalized) {
         // divide by the number of possible pairs
-        count n = G.numberOfNodes();
-        count pairs = (n-2) * (n-1);
-        count edges =  n    * (n-1);
+        const double n = static_cast<double>(G.numberOfNodes());
+        const double pairs = (n-2.) * (n-1.);
+        const double edges =  n    * (n-1.);
```
fabratu, line 89: You can remove the static_cast for `edges` in line 97.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.cpp
```diff
@@ -16,12 +16,10 @@
...
 namespace GroupClosenessGrowShrinkDetails {
 template <class WeightType>
-GroupClosenessGrowShrinkImpl<WeightType>::GroupClosenessGrowShrinkImpl(const Graph &G,
-                                                                       std::vector<node> initGroup,
-                                                                       bool extended,
-                                                                       count insertions,
-                                                                       count maxIterations)
-    : G(&G), group(std::move(initGroup)), extended(extended),
+GroupClosenessGrowShrinkImpl<WeightType>::GroupClosenessGrowShrinkImpl(
+    const Graph &G, std::vector<node> &&initGroup, bool extended, count insertions,
```
fabratu, line 20: Changing parameter `initGroup` to r-value reference does look odd. If a l-value is given, does this still work? In the unittests only [iterator-based constructor](https://github.com/networkit/networkit/blob/master/networkit/cpp/centrality/test/CentralityGTest.cpp#L1937) is used.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.hpp
```diff
@@ -40,7 +40,7 @@ class GroupClosenessGrowShrinkImpl final {
     static constexpr float maxInt16 = static_cast<float>(std::numeric_limits<uint16_t>::max());
 
 public:
-    GroupClosenessGrowShrinkImpl(const Graph &G, std::vector<node> group, bool extended = false,
+    GroupClosenessGrowShrinkImpl(const Graph &G, std::vector<node> &&group, bool extended = false,
```
fabratu, line 43: Same here.

fabratu (changes requested):

include/networkit/distance/SSSP.hpp
```diff
@@ -151,7 +151,7 @@ class SSSP : public Algorithm {
      */
     count getSumOfDistances() const {
         assureFinished();
-        return sumDist;
+        return static_cast<count>(sumDist);
```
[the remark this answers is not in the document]
angriman: This would work only on unweighted graphs, for weighted graph we need `sumDist` to be a double. Actually, the method should return a `double`, not a `count`.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.cpp
angriman: I think this is ok because it's the private implementation, not the public interface. The algorithms needs to change `group` anyway, so instead of copying a const ref it just takes an r-value ref.

fabratu: Ah, it works because the public interface does make a copy, which is then handed over to the private interface via r-value ref. Makes sense.

fabratu (approved):
LGTM.

networkit/cpp/algebraic/CSRMatrix.cpp
```diff
@@ -136,8 +136,8 @@ void CSRMatrix::setValue(const index i, const index j, const double value) {
 
     if (colIdx != none && colIdx >= rowIdx[i] && columnIdx[colIdx] == j) { // the matrix already has an entry at (i,j) => replace it
         if (value == getZero()) { // remove the nonZero value
-            columnIdx.erase(columnIdx.begin() + colIdx);
-            nonZeros.erase(nonZeros.begin() + colIdx);
+            columnIdx.erase(std::next(columnIdx.begin(), static_cast<int64_t>(colIdx)));
+            nonZeros.erase(std::next(nonZeros.begin(), static_cast<int64_t>(colIdx)));
```
avdgrinten, line 140: Why is this change done?

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.cpp
avdgrinten: Let the constructor take the group by value in this case (and move the value to the member variable).

networkit/cpp/auxiliary/BucketPQ.cpp
```diff
@@ -133,7 +133,7 @@ bool BucketPQ::empty() const noexcept {
 }
 
 int64_t BucketPQ::getKey(const index& val) {
-    int64_t key =myBucket[val]-offset;
+    int64_t key = static_cast<int64_t>(myBucket[val]) - offset;
     return key;
 }
```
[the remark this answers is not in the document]
angriman: Yes they are, it is often used with negative numbers to get a max-prioq

networkit/cpp/centrality/ApproxCloseness.cpp
```diff
@@ -304,7 +308,7 @@ void ApproxCloseness::runOnPivot(index i, const std::vector<node> &pivot, const
         if (samples[pivot[u]] == u) { // u belongs to the sampled nodes, compute its score exactly
             index j = pivot[u];
             last[j] = i;
-            dist[j] = d;
+            dist[j] = static_cast<node>(d);
```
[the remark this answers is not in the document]
angriman: `dist` is of type node, I assume it would be better to use `count`.

networkit/cpp/algebraic/CSRMatrix.cpp
fabratu: I also thought about that one in the review for a moment. In general it is more versatile when using iterator arithmetic, so that seems ok in terms coding-style.

angriman: And the cast is needed because the second parameter is of type `std::iterator_traits<InputIt>::difference_type` which is signed.

networkit/cpp/centrality/GroupClosenessGrowShrinkImpl.hpp
angriman: Ok that's done.

networkit/cpp/algebraic/CSRMatrix.cpp
avdgrinten: And why is `std::next` needed?

angriman: Hm I think I misinterpreted a warning, the only fix needed was the cast, not using `std::next`. I relaced the usages of `std::next` with just the `+` operator.

fabratu (dismissed):
LTGM. I think it is not critical to solve every narrowing-conversion for now. Rather have this PR merged sooner than later to have CI back on track.

networkit/cpp/centrality/ApproxCloseness.cpp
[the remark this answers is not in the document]
hmeyerhenke: dist should rather be of type edgeweight, shouldn't it?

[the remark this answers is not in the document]
angriman: Yes after a closer look it seems more reasonable to use `edgeweight`.

avdgrinten (approved):

avdgrinten:
Merging to unbreak the CI.

PR #759, opened 2021-05-26
Make CI fail if new files are not compliant with coding style

angriman:
Because all C++ files in NetworKit do not comply with our coding style, we cannot automatically enforce new files to be compliant unless the developer adds the `// networkit-format` line in all new files.
Most of the times this does not happen and makes us waste a lot of time in the review. This PR adds a new check that makes CI fail if newly added files do not comply with our coding style even if the `// networkit-format` line is not added.


avdgrinten:
Not sure if pulling the repo to detect old files is the best strategy. What if a file is moved?

angriman:
If a non-formatted file is moved then I think it is ok to also require to format it (I did that in past PRs). Otherwise, detecting which files are new or just moved/renamed does not seem trivial to me.

fabratu:
I would propose to use a list produced by `git diff`(based only on new files in the PR) and filter this by file-endings. Have no experience though, whether this also tracks moved files.

angriman:
Now I am getting all the newly added files with `git diff --name-status master`.

.github/workflows/ci.yml
```diff
@@ -437,4 +437,9 @@ jobs:
       - name: Prepare environment and run checks
         run:  |
           set -e
+          git remote add networkit "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY"
+          git fetch networkit
+          export CUR_BRANCH=$(git branch --show-current)
+          git checkout --track networkit/master
```
michitux, line 443: Doesn't this break/fail if the current branch is named `master`, too (`fatal: A branch named 'master' already exists.`)? E.g., if I push to the master branch of my fork and then open a pull request from there or if the CI runs on the master branch? Wouldn't it be much simpler to use `networkit/master` in the diff command?

extrafiles/tooling/nktooling/__init__.py
```diff
@@ -109,6 +109,12 @@ def getCXXFiles(includeSources = True, includeHeaders = True):
 
 	return files
 
+def getNewCXXFiles():
+	""" Returns a list of all new C++ files added to the repository."""
+	out = subprocess.Popen(("git", "diff", "--name-status", "master"), stdout=subprocess.PIPE)
+	out = subprocess.Popen(("grep", "^A.*\.[ch]pp$"), stdin=out.stdout, stdout=subprocess.PIPE)
+	return subprocess.check_output(("awk", "{ print $2 }"), stdin=out.stdout).decode("utf8").split("\n")[:-1]
```
michitux, line 116: Wouldn't it be cleaner and possibly less error-prone to do the post-processing in Python? Further, `pygit2` could be used to directly use the git repository in Python, but that's minor nit-picking. I'm not saying that this must be changed, after all this is just a helper for CI where we are in a relatively controlled environment.

michitux (commented):
This seems to contain a lot of code to check new files that do not contain `// networkit-format` on-the-fly once during the PR. If, e.g., a second PR modifies these files, they won't be checked again. I think the goal should be that new files contain `// networkit-format`, not that they are compliant without containing `// networkit-format`. My suggestion would be to instead simply require that new files contain `// networkit-format` as a separate check and to leave the remaining checks unmodified.

.github/workflows/ci.yml
michitux: Another thought: locally, this will fail of course unless you have a remote named `networkit`. Maybe you could first check if `networkit/master` exists and if not compare to `master`?

avdgrinten:
All these solutions with git diffs do not sound very appealing to me. I'd suggest to keep a list of files that do *not* have `// networkit-format` instead; this list can shrink over time. That approach is more robust and allows us to also move / rename files. Yes, having such a list in the repo is ugly, but it's a pragmatic solution.

In the future (if almost all files are formatted), we could replace the inverse oif `// networkit-format` by `// no-networkit-format` and get rid of the list.

michitux:
If we create such a list in the first place, why don't we get rid of `// networkit-format` immediately and instead rely on that list? Otherwise, that information is somehow redundant and we would also need to check that the list is consistent with the actual files. On the other hand, what speaks against making the switch to `// no-networkit-format` right now? Automatically adding that line and removing the old lines shouldn't be much work but seems much more natural as new files would be formatted automatically.

avdgrinten:
Making the switch to `// no-networkit-format` right now is also a possibility.

avdgrinten (approved):
LGTM.

Issue #740, opened 2021-04-20
Friends classes of Graph should be removed


avdgrinten:
This blocks moving to a different graph implementation. The current `friend`s should use `unsafe` methods instead.

PR #760, opened 2021-05-27
Remove friendship classes from Graph and replace with Unsafe functions instead.

Relux-the-Relux:
Addressing #740.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1414,30 +1414,6 @@ public:
     count numberOfNodes() const noexcept { return n; }
 
     /**
-     * Return the number of nodes in the graph.
-     * @return The number of nodes.
-     */
-    count numberOfNodes(Unsafe) const noexcept { return outEdges.size(); }
-
-    /**
-     * Return the number of nodes in the graph.
-     * @return The number of nodes.
-     */
-    count numberOfInNodes(Unsafe) const noexcept { return inEdges.size(); }
-
-    /**
-     * Return the number of nodes in the graph.
-     * @return The number of nodes.
-     */
-    count numberOfWeightedNodes(Unsafe) const noexcept { return outEdgeWeights.size(); }
-
-    /**
-     * Return the number of nodes in the graph.
-     * @return The number of nodes.
-     */
-    count numberOfWeightedInNodes(Unsafe) const noexcept { return inEdgeWeights.size(); }
-
-    /**
      * Return the number of edges in the graph.
      * @return The number of edges.
      */
@@ -1507,6 +1483,15 @@ public:
     void setWeight(node u, node v, edgeweight ew);
 
     /**
+     * Set the weight to the i-th neighbour of u.
+     *
+     * @param[in]	u	endpoint of edge
+     * @param[in]	i	index of the nexight
+     * @param[in]	weight	edge weight
+     */
+    void setWeightAtIthNeighbor(Unsafe, node u, index i, edgeweight ew);
+
+    /**
      * Increase the weight of an edge. If the edge does not exist,
      * it will be inserted.
      *
@@ -1515,15 +1500,6 @@ public:
      * @param[in]	weight	edge weight
      */
     void increaseWeight(node u, node v, edgeweight ew);
-
-    /**
-     * Set the weight of i-th edge in the adjecency array of u.
-     *
-     * @param[in]	u	endpoint of edge
-     * @param[in]	i	index of edge in the adjacency array for u.
-     * @param[in]	weight	edge weight
-     */
-    void setWeightAtIthNeighbor(Unsafe, node u, index i, edgeweight ew);
 
     /* SUMS */
 
@@ -1607,6 +1583,10 @@ public:
 
     /**
      * Returns the index of node v in the array of outgoing edges of node u.
+     *
+     * @param u Node
+     * @param v Node
+     * @return index of node v in the array of outgoing edges of node u.
      */
     index indexOfNeighbor(node u, node v) const { return indexInOutEdgeArray(u, v); }
 
@@ -1644,7 +1624,7 @@ public:
      * @param u Node.
      * @param i index; should be in [0, degreeOut(u))
      * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
-     * edge id, or @c defaultEdgeWeight if unweighted.
+     * edge weight, or @c defaultEdgeWeight if unweighted.
      */
     std::pair<node, edgeweight> getIthNeighborWithWeight(node u, index i) const {
         if (!hasNode(u) || i >= outEdges[u].size())
@@ -1658,7 +1638,7 @@ public:
      * @param u Node.
      * @param i index; should be in [0, degreeOut(u))
      * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
-     * edge id, or @c defaultEdgeWeight if unweighted.
+     * edge weight, or @c defaultEdgeWeight if unweighted.
      */
     std::pair<node, edgeweight> getIthNeighborWithWeight(Unsafe, node u, index i) const {
         if (!isWeighted())
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -96,10 +96,6 @@ class CurveballMaterialization;
  */
 class Graph final {
 
-    friend class ParallelPartitionCoarsening;
-    friend class GraphBuilder;
-    friend class CurveballDetails::CurveballMaterialization;
-
     // graph attributes
     //!< current number of nodes
     count n;
@@ -1041,6 +1037,28 @@ public:
      */
     void preallocateDirected(node u, size_t outSize, size_t inSize);
 
+    /**
+     * Reserves memory in the node's edge containers for directed graphs.
+     *
+     * @param u the node memory should be reserved for
+     * @param outSize the amount of memory to reserve for out edges
+     *
+     * This function is thread-safe if called from different
+     * threads on different nodes.
+     */
+    void preallocateDirectedOutEdges(node u, size_t outSize);
+
+    /**
+     * Reserves memory in the node's edge containers for directed graphs.
+     *
+     * @param u the node memory should be reserved for
+     * @param inSize the amount of memory to reserve for in edges
+     *
+     * This function is thread-safe if called from different
+     * threads on different nodes.
+     */
+    void preallocateDirectedInEdges(node u, size_t inSize);
+
     /** EDGE IDS **/
 
     /**
@@ -1396,6 +1414,30 @@ public:
     count numberOfNodes() const noexcept { return n; }
 
     /**
+     * Return the number of nodes in the graph.
+     * @return The number of nodes.
+     */
+    count numberOfNodes(Unsafe) const noexcept { return outEdges.size(); }
+
+    /**
+     * Return the number of nodes in the graph.
+     * @return The number of nodes.
+     */
+    count numberOfInNodes(Unsafe) const noexcept { return inEdges.size(); }
+
+    /**
+     * Return the number of nodes in the graph.
+     * @return The number of nodes.
+     */
+    count numberOfWeightedNodes(Unsafe) const noexcept { return outEdgeWeights.size(); }
+
+    /**
+     * Return the number of nodes in the graph.
+     * @return The number of nodes.
+     */
+    count numberOfWeightedInNodes(Unsafe) const noexcept { return inEdgeWeights.size(); }
+
+    /**
      * Return the number of edges in the graph.
      * @return The number of edges.
      */
@@ -1473,6 +1515,15 @@ public:
      * @param[in]	weight	edge weight
      */
     void increaseWeight(node u, node v, edgeweight ew);
+
+    /**
+     * Set the weight of i-th edge in the adjecency array of u.
+     *
+     * @param[in]	u	endpoint of edge
+     * @param[in]	i	index of edge in the adjacency array for u.
+     * @param[in]	weight	edge weight
+     */
+    void setWeightAtIthNeighbor(Unsafe, node u, index i, edgeweight ew);
 
     /* SUMS */
 
@@ -1555,6 +1606,11 @@ public:
     }
 
     /**
+     * Returns the index of node v in the array of outgoing edges of node u.
+     */
+    index indexOfNeighbor(node u, node v) const { return indexInOutEdgeArray(u, v); }
+
+    /**
      * Return the i-th (outgoing) neighbor of @a u.
      *
      * @param u Node.
@@ -1580,6 +1636,34 @@ public:
         if (!hasNode(u) || i >= outEdges[u].size())
             return nullWeight;
         return isWeighted() ? outEdgeWeights[u][i] : defaultEdgeWeight;
+    }
+
+    /**
+     * Get i-th (outgoing) neighbor of @a u and the corresponding edge weight.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
+     * edge id, or @c defaultEdgeWeight if unweighted.
+     */
+    std::pair<node, edgeweight> getIthNeighborWithWeight(node u, index i) const {
+        if (!hasNode(u) || i >= outEdges[u].size())
+            return {none, none};
+        return getIthNeighborWithWeight(unsafe, u, i);
+    }
+
+    /**
+     * Get i-th (outgoing) neighbor of @a u and the corresponding edge weight.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
+     * edge id, or @c defaultEdgeWeight if unweighted.
+     */
+    std::pair<node, edgeweight> getIthNeighborWithWeight(Unsafe, node u, index i) const {
+        if (!isWeighted())
+            return {outEdges[u][i], defaultEdgeWeight};
+        return {outEdges[u][i], outEdgeWeights[u][i]};
     }
 
     /**
--- a/networkit/cpp/graph/GraphBuilder.cpp
+++ b/networkit/cpp/graph/GraphBuilder.cpp
@@ -173,7 +173,6 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
 }
 
 void GraphBuilder::toGraphDirectSwap(Graph &G) {
-    assert(is G.empty());
 
     for (index u = 0; u < outEdges.size(); u++){
         for (index j = 0; j < outEdges[u].size(); j++){
--- a/networkit/cpp/graph/GraphBuilder.cpp
+++ b/networkit/cpp/graph/GraphBuilder.cpp
@@ -145,11 +145,11 @@ void GraphBuilder::increaseInWeight(node u, node v, edgeweight ew) {
 
 Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
     Graph G(n, weighted, directed);
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
+    assert(n == G.upperNodeIdBound());
+    #ifdef NETWORKIT_SANITY_CHECKS
+    assert(G.checkConsistency());
+    #endif
+    
     // copy edges and weights
     if (autoCompleteEdges) {
         if (parallel) {
@@ -160,13 +160,11 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
     } else {
         toGraphDirectSwap(G);
     }
-
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
-    G.m = numberOfEdges(G);
+    G.setEdgeCount(unsafe, numberOfEdges(G));
+    assert(n == G.upperNodeIdBound());
+    #ifdef NETWORKIT_SANITY_CHECKS
+    assert(G.checkConsistency());
+    #endif
     G.shrinkToFit();
 
     reset();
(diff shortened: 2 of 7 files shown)


include/networkit/graph/Graph.hpp
```diff
@@ -1441,6 +1633,15 @@ class Graph final {
...
     void increaseWeight(node u, node v, edgeweight ew);
 
+    /**
+     * Increase the weight of i-th edge in the adjecency array of u.
+     *
+     * @param[in]	u	endpoint of edge
+     * @param[in]	i	index of edge in the adjacency array for u.
+     * @param[in]	weight	edge weight
+     */
+    void increaseWeightAtItNeighbour(Unsafe, node u, index i, edgeweight ew);
```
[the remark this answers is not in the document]
Relux-the-Relux: oh yeah

include/networkit/graph/Graph.hpp
```diff
@@ -1362,6 +1523,37 @@ class Graph final {
      */
     count numberOfNodes() const noexcept { return n; }
 
+    /**
+     * Returns the size of the outEdges vector. This most is likely the number of nodes but might
+     * differ if nodes have been deleted.
+     * @return The size of outEdges.
+     */
+    count numberOfNodes(Unsafe) const noexcept { return outEdges.size(); }
```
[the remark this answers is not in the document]
Relux-the-Relux: the problem with `numberOfWeightedNodes` and `numberOfWeightedInNodes` needs to be 0 in the unweighted/undirected when `GraphBuilder` checks for the consistency of the graph.

maybe we could simply make these private and add the checks to `checkConsistency()` and remove the assertions from `GraphBuilder` replacing them only with the `checkConsistency()` call.

networkit/cpp/randomization/CurveballImpl.cpp
```diff
@@ -154,29 +154,35 @@ void CurveballMaterialization::toGraphParallel(Graph &G) {
...
+    for (index u = 0; u < new_outEdges.size(); u++) {
+        for (index j = 0; j < new_outEdges[u].size(); j++) {
+            G.addPartialEdge(unsafe, u, new_outEdges[u][j]);
+        }
+    }
 
     // Reserve the space
-    G.parallelForNodes([&](node v) { G.outEdges[v].reserve(outDeg[v] + missingEdgesCounts[v]); });
+    G.parallelForNodes(
+        [&](node v) { G.preallocateUndirected(v, outDeg[v] + missingEdgesCounts[v]); });
```
avdgrinten, line 166: The `addPartialEdge` loop can be moved into the parallelForNodes below (after preallocation).

networkit/cpp/randomization/CurveballImpl.cpp
```diff
@@ -154,29 +154,35 @@ void CurveballMaterialization::toGraphParallel(Graph &G) {
...
+            const node u = G.getIthNeighbor(v, neighbor_id);
+            G.addPartialEdge(unsafe, u, v);
         }
     });
 
     // TODO: is the networkit adjacency list even sorted for the neighbours? if
     // not omit this
     // Sort neighbours
-    G.parallelForNodes([&](node v) { std::sort(G.outEdges[v].begin(), G.outEdges[v].end()); });
+    // G.parallelForNodes([&](node v) { std::sort(G.outEdges[v].begin(), G.outEdges[v].end()); });
```
avdgrinten, line 179: Remove this chunk (as the comment states).

networkit/cpp/randomization/CurveballImpl.cpp
```diff
@@ -206,29 +212,34 @@ void CurveballMaterialization::toGraphSequential(Graph &G) {
...
+    G.removeAllEdges();
+    for (index u = 0; u < new_outEdges.size(); u++) {
+        for (index j = 0; j < new_outEdges[u].size(); j++) {
+            G.addPartialEdge(unsafe, u, new_outEdges[u][j]);
+        }
+    }
 
     // Reserve the space
-    G.forNodes([&](node v) { G.outEdges[v].reserve(outDeg[v] + missingEdgesCounts[v]); });
+    G.forNodes([&](node v) { G.preallocateUndirected(v, outDeg[v] + missingEdgesCounts[v]); });
```
avdgrinten, line 223: See above.

networkit/cpp/randomization/CurveballImpl.cpp
```diff
@@ -206,29 +212,34 @@ void CurveballMaterialization::toGraphSequential(Graph &G) {
...
+            const node u = G.getIthNeighbor(v, neighbor_id);
+            G.addPartialEdge(unsafe, u, v);
         }
     });
 
     // TODO: is the networkit adjacency list even sorted for the neighbours? if
     // not omit this
     // Sort neighbours
-    G.forNodes([&](node v) { std::sort(G.outEdges[v].begin(), G.outEdges[v].end()); });
+    // G.forNodes([&](node v) { std::sort(G.outEdges[v].begin(), G.outEdges[v].end()); });
```
avdgrinten, line 236: See above.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -145,11 +145,8 @@ void GraphBuilder::increaseInWeight(node u, node v, edgeweight ew) {
 
 Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
     Graph G(n, weighted, directed);
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
+    assert(G.checkConsistency(n));
```
avdgrinten, line 148: `checkConsistency` is expensive and should be behind `NETWORKIT_SANITY_CHECKS`.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -161,12 +158,8 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
         toGraphDirectSwap(G);
     }
 
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
-    G.m = numberOfEdges(G);
+    assert(G.checkConsistency(n));
```
avdgrinten, line 161: Likewise.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -213,24 +226,26 @@ void GraphBuilder::toGraphParallel(Graph &G) {
                                         // undirected graphs
                 inEdgesPerThread[tid][u].push_back(v);
                 if (weighted) {
-                    edgeweight ew = outEdgeWeights[v][i];
-                    inWeightsPerThread[tid][u].push_back(ew);
+                    G.addPartialEdge(unsafe, v, u, outEdgeWeights[v][i]);
+                    inWeightsPerThread[tid][u].push_back(outEdgeWeights[v][i]);
+                } else {
+                    G.addPartialEdge(unsafe, v, u);
                 }
```
avdgrinten, line 233: Is adding edges safe here? I assume no two threads access the same `v` at the same time?

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -213,24 +226,26 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
             }
             if (u == v) {
                 numberOfSelfLoopsPerThread[tid]++;
+                if(!directed){
+                    inEdgesPerThread[tid][u].push_back(v);
+                    if (weighted) inWeightsPerThread[tid][u].push_back(outEdgeWeights[v][i]);
+                }
```
avdgrinten, line 240: Why is this addition necessary?

avdgrinten: The formatting of the `if` is broken.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -239,60 +254,75 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
-                G.inEdgeWeights[v].reserve(inDeg);
-            }
-        } else {
-            G.outEdges[v].reserve(outDeg + inDeg);
-            if (weighted) {
-                G.outEdgeWeights[v].reserve(outDeg + inDeg);
-            }
+            G.preallocateDirectedInEdges(v, inDeg);
+            } else {
+            G.preallocateUndirected(v, outDeg + inDeg);
```
avdgrinten, line 259: Formatting is broken.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -239,60 +254,75 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
         if (directed) {
-            for (int tid = 0; tid < maxThreads; tid++) {
-                copyAndClear(inEdgesPerThread[tid][v], G.inEdges[v]);
-            }
             if (weighted) {
                 for (int tid = 0; tid < maxThreads; tid++) {
-                    copyAndClear(inWeightsPerThread[tid][v], G.inEdgeWeights[v]);
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                    G.addPartialInEdge(unsafe, v, inEdgesPerThread[tid][v][j], inWeightsPerThread[tid][v][j]);
+                    }
```
avdgrinten, line 268: Formatting is broken.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -239,60 +254,75 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
         } else {
-            for (int tid = 0; tid < maxThreads; tid++) {
-                copyAndClear(inEdgesPerThread[tid][v], G.outEdges[v]);
-            }
             if (weighted) {
                 for (int tid = 0; tid < maxThreads; tid++) {
-                    copyAndClear(inWeightsPerThread[tid][v], G.outEdgeWeights[v]);
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                    G.addPartialEdge(unsafe, v, inEdgesPerThread[tid][v][j], inWeightsPerThread[tid][v][j]);
+                    }
```
avdgrinten, line 282: Formatting is broken.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -239,60 +254,75 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
-                    copyAndClear(inWeightsPerThread[tid][v], G.inEdgeWeights[v]);
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                    G.addPartialInEdge(unsafe, v, inEdgesPerThread[tid][v][j], inWeightsPerThread[tid][v][j]);
+                    }
+                }
+            } else {
+                for (int tid = 0; tid < maxThreads; tid++) {
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                        G.addPartialInEdge(unsafe, v, inEdgesPerThread[tid][v][j]);
+                    }
```
avdgrinten, line 274: Why do we need a new else case here?

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -239,60 +254,75 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
-                    copyAndClear(inWeightsPerThread[tid][v], G.outEdgeWeights[v]);
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                    G.addPartialEdge(unsafe, v, inEdgesPerThread[tid][v][j], inWeightsPerThread[tid][v][j]);
+                    }
+                }
+            } else {
+                for (int tid = 0; tid < maxThreads; tid++) {
+                    for (index j = 0; j < inEdgesPerThread[tid][v].size(); j++){
+                        G.addPartialEdge(unsafe, v, inEdgesPerThread[tid][v][j]);
+                    }
```
avdgrinten, line 288: Why do we need a new else case here?

include/networkit/graph/Graph.hpp
```diff
@@ -1441,6 +1459,15 @@ class Graph final {
...
     void increaseWeight(node u, node v, edgeweight ew);
 
+    /**
+     * Increase the weight of i-th edge in the adjecency array of u.
+     *
+     * @param[in]	u	endpoint of edge
+     * @param[in]	i	index of edge in the adjacency array for u.
+     * @param[in]	weight	edge weight
+     */
+    void increaseWeightAtIthNeighbor(Unsafe, node u, index i, edgeweight ew);
```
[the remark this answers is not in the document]
Relux-the-Relux: It was more a question of speed that I implemented as a separate function, to avoid having to call the weight everytime we want to increase it.

include/networkit/graph/Graph.hpp
```diff
@@ -1521,6 +1548,15 @@ class Graph final {
...
     }
 
+    /**
+     * Returns the index of node v in the array of outgoing edges of node u.
+     *
+     * @param u Node
+     * @param v Node
+     * @return index of node v in the array of outgoing edges of node u.
+     */
+    index indexOfNeighbor(Unsafe, node u, node v) const { return indexInOutEdgeArray(u, v); }
```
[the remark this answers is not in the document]
Relux-the-Relux: you're right it shouldn't be

networkit/cpp/graph/GraphBuilder.cpp
Relux-the-Relux: It is so we don't try to access the weight array when we don't have any weights, which would cause a segmentation fault

include/networkit/graph/Graph.hpp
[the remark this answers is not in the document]
avdgrinten: But aside from safety checks (which are omitted by the `Unsafe` variant anyway), that is equivalent, right? Surely the compiler is able to inline the call (at least, if we define the function in the header) and will generate identical assembly?

networkit/cpp/graph/GraphBuilder.cpp
avdgrinten: Makes sense. How did the old code deal with that?

Relux-the-Relux: it just replaced both arrays in graph, since it would just put an array of size 0 into in the unweighted case.

include/networkit/graph/Graph.hpp
```diff
@@ -1535,6 +1571,36 @@ class Graph final {
...
+     * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
+     * edge weight, or @c defaultEdgeWeight if unweighted.
+     */
+    std::pair<node, edgeweight> getIthNeighborWithWeight(node u, index i) const {
+        if (!hasNode(u) || i >= outEdges[u].size())
+            return {none, none};
+        if (!isWeighted())
+            return {outEdges[u][i], defaultEdgeWeight};
+        return {outEdges[u][i], outEdgeWeights[u][i]};
+    }
```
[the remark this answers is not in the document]
Relux-the-Relux: Yes it is used in graphbuilder

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -161,12 +158,8 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
...
     }
 
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
-    G.m = numberOfEdges(G);
+    assert(G.checkConsistency(n));
+    G.setEdgeCount(unsafe, numberOfEdges(G));
```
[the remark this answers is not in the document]
Relux-the-Relux: Because it only checks for number for the node upper bound, the number of edges is not relevant for cosistency.

networkit/cpp/graph/GraphBuilder.cpp
Relux-the-Relux: yes it is safe

include/networkit/graph/Graph.hpp
```diff
@@ -88,9 +88,9 @@ class CurveballMaterialization;
  */
 class Graph final {
 
-    friend class ParallelPartitionCoarsening;
-    friend class GraphBuilder;
-    friend class CurveballDetails::CurveballMaterialization;
+    // friend class ParallelPartitionCoarsening;
+    // friend class GraphBuilder;
+    // friend class CurveballDetails::CurveballMaterialization;
```
avdgrinten, line 93: This needs to be fixed.

include/networkit/graph/Graph.hpp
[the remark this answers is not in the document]
avdgrinten: What's the progress here? I really don't want an "increase" function as a basic building block.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -145,11 +145,11 @@ void GraphBuilder::increaseInWeight(node u, node v, edgeweight ew) {
...
     Graph G(n, weighted, directed);
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
+    #ifdef NETWORKIT_SANITY_CHECKS
+    assert(n == G.upperNodeIdBound());
+    assert(G.checkConsistency());
+    #endif
```
avdgrinten, line 151: The cheap `n == G.upperNodeIdBound()` assertion can stay outside of NETWORKIT_SANITY_CHECKS.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -160,13 +160,11 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
...
-    assert(G.outEdges.size() == n);
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
-    G.m = numberOfEdges(G);
+    #ifdef NETWORKIT_SANITY_CHECKS
+    assert(n == G.upperNodeIdBound());
+    assert(G.checkConsistency());
+    #endif
```
avdgrinten, line 166: Likewise.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -213,24 +231,26 @@ void GraphBuilder::toGraphParallel(Graph &G) {
...
+                    inWeightsPerThread[tid][u].push_back(outEdgeWeights[v][i]);
+                } else {
+                    G.addPartialEdge(unsafe, v, u);
                 }
             }
             if (u == v) {
                 numberOfSelfLoopsPerThread[tid]++;
+                if (!directed) {
+                    inEdgesPerThread[tid][u].push_back(v);
+                    if (weighted) inWeightsPerThread[tid][u].push_back(outEdgeWeights[v][i]);
```
avdgrinten, line 244: Coding style.

networkit/cpp/randomization/CurveballImpl.cpp
```diff
@@ -206,29 +206,31 @@ void CurveballMaterialization::toGraphSequential(Graph &G) {
         }
     });
 
-    G.outEdges.swap(new_outEdges);
+    G.removeAllEdges();
+    for (index u = 0; u < new_outEdges.size(); u++) {
+    }
```
avdgrinten, line 211: Huh?

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -160,13 +160,11 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
...
-    assert(G.outEdgeWeights.size() == (weighted ? n : 0));
-    assert(G.inEdges.size() == (directed ? n : 0));
-    assert(G.inEdgeWeights.size() == ((weighted && directed) ? n : 0));
-
-    G.m = numberOfEdges(G);
+    #ifdef NETWORKIT_SANITY_CHECKS
+    assert(n == G.upperNodeIdBound());
+    assert(G.checkConsistency());
+    #endif
+    G.setEdgeCount(unsafe, numberOfEdges(G));
```
avdgrinten, line 167: `checkConsistency` should be adjusted to also check the edge count. (And hence `setEdgeCount()` should be called before `checkConsistency()`.)

avdgrinten (commented):
Looks mostly good now. Please do a final pass to resolve the comments and:
* Cleanup the code by removing the commented out parts. There are also `if`s that have the `then-case on the same lines; these should be fixed.
* Tidy up the commit structure, by removing the commits that fix earlier commits. It might make sense to squash everything except for the documentation additions into a single commit.

michitux:
Do you have any idea what impact these changes have on performance and memory usage? From a short look, this should for example duplicate the memory usage of `METISGraphReader` as the graph is now copied from `GraphBuilder` into the graph instead of swapping the edge data structures which is imho not acceptable. Further, the method name `toGraphDirectSwap` is now completely misleading, as it is no longer a direct swap but a fully sequential copy that does not even pre-allocate the vectors. It should be renamed, which is possible as it is a private method. As mentioned above, I would suggest to eliminate all uses of `GraphBuilder` that has `autoCompleteEdges` set to `false`.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -175,14 +173,34 @@ Graph GraphBuilder::toGraph(bool autoCompleteEdges, bool parallel) {
 }
 
 void GraphBuilder::toGraphDirectSwap(Graph &G) {
-    G.outEdges = std::move(outEdges);
-    G.outEdgeWeights = std::move(outEdgeWeights);
-    G.inEdges = std::move(inEdges);
-    G.inEdgeWeights = std::move(inEdgeWeights);
+    G.removeAllEdges();
```
michitux, line 176: This is a private method that is only called with an empty graph. An assert that the graph is empty would be more appropriate imho.

Relux-the-Relux: But i don't think the Graph would be empty. DirectSwap we expect the graph to not be empty.

michitux: `toGraphDirectSwap` is only called in `toGraph` (it is private). There, the graph object is initialized with `Graph G(n, weighted, directed);`. I cannot see any code there that could add edges to the graph that is called before `toGraphDirectSwap`.

avdgrinten:
I am aware that this is a performance pessimization. Michael is probably right about rewriting the users of GraphBuilder that do not need edge completion (these can directly use the unsafe methods). GraphBuilder itself can be re-written to use a better algorithm (there are smarter transpose algorithms than what we implement right now). This can probably also half the memory consumption and get it back closer to the original memory consumption (by only storing the half of the edges that still requires completion and directly inserting the other half).

I think that should be done in a separate PR though (to avoid making this PR more complicated). This means that we will not merge this PR before the next release (which is scheduled for June 30th) since it doesn't make sense to merge a performance pessimization right before the release. We should still complete this PR now, immediately merge it after the release and then optimize `GraphBuilder`.

avdgrinten (approved):

Issue #764, opened 2021-06-11
Optimize memory usage for LAMG and ConjugateGradient


kisasexypantera94:
Currently ```LAMG``` and ```ConjugateGradient``` needs to store all RHSs and results at the same time, which is unnecessary. I suggest using lambda loaders and processors like so:
```c++
const auto rhs_loader = [&](size_t i)
{
    nk::Vector Y(n);
    ...
    return Y;
};

const auto result_processor = [&](const nk::Vector& result)
{
    ...
};

size_t n_rhs = n;
size_t m_rhs = m;
lamg.parallelSolve(rhs_loader, result_processor, n_rhs, m_rhs);
```

avdgrinten:
What about passing a (mutable) reference to a nk::Vector to the RHS producer function? That way we could avoid memory allocations.

This would be a good addition, if we can re-implement the old API on top of the new one (and thus preserve backwards compatibility).

kisasexypantera94:
Now that #765 is in the master, the question arises of how to use the new API in places where it is required to store all the solutions, as in ```CommuteTimeDistance```. Storing all solutions allows to quickly answer requests for the distance between any two nodes, but requires storing a huge (probably dense) matrix. In addition, the user may be interested in the distance between specific nodes. Maybe allow the user to pass a list of from-to node pairs as an argument and compute the distances for them in advance? Though the internals of the classes can get a little messy, there will be two arrays - ```solutions``` and ```distances```, the contents of which depend on the arguments.
I did not mention RHS vectors, they don't need to be stored anyway.

Issue #761, opened 2021-06-04
Effective Resistance


hamidehhhs:
I wonder if it is  possible to compute  or approximate effective resistance between two pairs of nodes in networkit ?

hmeyerhenke:
If you go to the Documentation on the homepage, then search for "commute". Commute time and resistance distance are the same except for the volume(G) factor.

hamidehhhs:
Thanks for your reply.
I have tested that function on a network with around 400*10^6 edges but it failed because of too much memory consumption. 
I wished there is any alternative!

hmeyerhenke:
There is. Use CG to set up and solve the corresponding linear system (see numerics folder). For a few pairs this should work within a few minutes, I would assume. LAMG as solver is probably too space-consuming on such a graph, depending on the RAM you have...

PR #765, opened 2021-06-12
Optimize memory usage in LAMG and ConjugateGradient

kisasexypantera94:
This PR adds an optimized parallel solution loop to ```LAMG``` and ```ConjugateGradient``` as proposed in #764. It will also solve the problem described in #761.
Changes in ```AlgebraicSpanningEdgeCentrality``` show how the new API can be used

### DIFF
--- a/include/networkit/algebraic/algorithms/AlgebraicSpanningEdgeCentrality.hpp
+++ b/include/networkit/algebraic/algorithms/AlgebraicSpanningEdgeCentrality.hpp
@@ -9,6 +9,7 @@
 #define NETWORKIT_ALGEBRAIC_ALGORITHMS_ALGEBRAIC_SPANNING_EDGE_CENTRALITY_HPP_
 
 #include <cmath>
+#include <atomic>
 
 #include <networkit/centrality/Centrality.hpp>
 #include <networkit/numerics/LAMG/Lamg.hpp>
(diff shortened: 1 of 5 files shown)


include/networkit/algebraic/algorithms/AlgebraicSpanningEdgeCentrality.hpp
```diff
@@ -82,34 +83,35 @@ void AlgebraicSpanningEdgeCentrality<Matrix>::runApproximation() {
...
+        });
+        return yRow;
+    };
+
+    std::vector<std::atomic<double>> scoreDataAtom(m);
+    const auto resultProcessor = [&](count i, const Vector& zRow) {
+        G.forEdges([&](node u, node v, edgeid e) {
+            double diff = (zRow[u] - zRow[v]);
+            diff *= diff;
+            for (double old = scoreDataAtom[e]; !scoreDataAtom[e].compare_exchange_strong(old, old + diff););
```
kisasexypantera94, line 103: not sure about this

include/networkit/numerics/LAMG/Lamg.hpp
```diff
@@ -243,6 +269,46 @@ void Lamg<Matrix>::parallelSolve(const std::vector<Vector>& rhs, std::vector<Vec
...
+
+            for (index i = 0; i < (index) numThreads; ++i) {
+                compSolvers.push_back(SolverLamg<Matrix>(compHierarchies[0], smoother));
+            }
+        }
+
+        count n = rhsSize.first;
+        count m = rhsSize.second;
+        std::vector<Vector> results(numThreads, Vector(m));
+        std::vector<Vector> RHSs(numThreads, Vector(m));
```
kisasexypantera94, line 291: solvers are responsible for resources. An alternative would be to give this control to user, but I don't see much benefit in this

include/networkit/numerics/LinearSolver.hpp
```diff
@@ -84,6 +85,31 @@ class LinearSolver {
...
+    /**
+     * RHS loader. Accepts index of RHS column and mutable reference to properly sized row Vector, which is not necessary zero-valued.
+     * Returns a reference to RHS vector.
+     */
+    using RHSLoader = std::function<Vector&(count, Vector&)>;
+
+    /**
+     * Result processor. Accepts index of result column and immutable reference to result Vector.
+     */
+    using ResultProcessor = std::function<void(count, const Vector&)>;
```
avdgrinten, line 99: Instead of using `std::function` here (which has a high overhead), I would turn the parallelSolve function into a template (that requires that we implement it in the header which is a bit annoying, but not an issue).

kisasexypantera94: there is a problem with ```parallelSolve``` being virtual. I thought of doing type-erasure but that would require std::any, which only appeared in C++17

avdgrinten: Hm, one option would be to not make it virtual. It is a new overload anyway, so maybe that is a reasonable trade-off?

avdgrinten: Callers that do not know the type of the solver that they are using could still use the old `parallelSolve`.

kisasexypantera94: > Hm, one option would be to not make it virtual. It is a new overload anyway, so maybe that is a reasonable trade-off?

I think the same. Also there seems to be only one place where LinearSolver polymorphism is used (MaxentStress), so in most cases callers will know the type of solver.

include/networkit/algebraic/algorithms/AlgebraicSpanningEdgeCentrality.hpp
avdgrinten: Can't we use `fetch_add`? Alternatively, it might be clear to use a matrix that is indexed by `(thread ID) x node` such that we can aggregate over all threads after finishing the computation (and thus we can drop the atomics here).

avdgrinten (commented):
Looks mostly good to me, except for one comment. The commit history is a bit messy, please squash the fixes into earlier commits.

include/networkit/algebraic/algorithms/AlgebraicSpanningEdgeCentrality.hpp
kisasexypantera94: sadly std::atomic<double>::fetch_add is specialized only in c++20. Sure using thread matrix would be much more intuitive, but I think it might be worth sacrificing readability here in favor of saving memory

kisasexypantera94: I thought again, and it will actually not be very intuitive, the profit will only be in getting rid of the atomics

avdgrinten: Okay, but if we keep the atomics, we can at least reduce their memory ordering requirements and pass `std::memory_order_relaxed` to the `compare_exchange_strong` and the atomic load in the loop that sets `scoreData` below.

After this change is done, the PR can be merged.

avdgrinten (approved):
LGTM now, thank you again for the PR!

PR #767, opened 2021-06-15
Fixes pre- and post-increment of NeighborIterator + minor fixes/additions

fabratu:

- Fix pre- and post-increment of `NeighborIterator` (fixes error on CI with #734)
- Change to return-by-reference for pre-increment operator of all graph-iterator types
- Add missing Cython-extension for `NeighborWeightRange`

### DIFF
--- a/networkit/graph.pyx
+++ b/networkit/graph.pyx
@@ -612,6 +612,42 @@ cdef class Graph:
 			yield dereference(it)
 			preincrement(it)
 
+	def iterNeighborsWeights(self, u):
+		"""
+		Iterates over a range of the neighbors of a node including the edge weights.
+		The iterator is not safe to use with unweighted graphs. To avoid unsafe behavior
+		a runtime error will be thrown.
+
+		Parameters:
+		-----------
+		u : Node
+		"""
+		if not self._this.isWeighted():
+			raise RuntimeError("iterNeighborsWeights: Use this iterator only on weighted graphs.")
+		
+		it = self._this.weightNeighborRange(u).begin()
+		while it != self._this.weightNeighborRange(u).end():
+			yield dereference(it)
+			preincrement(it)		
+
+	def iterInNeighborsWeights(self, u):
+		"""
+		Iterates over a range of the in-neighbors of a node including the edge weights.
+		The iterator is not safe to use with unweighted graphs. To avoid unsafe behavior
+		a runtime error will be thrown.
+
+		Parameters:
+		-----------
+		u : Node
+		"""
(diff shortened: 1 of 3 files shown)


avdgrinten (approved):
LGTM.

michitux (changes requested):
The new Python interface triggers a segmentation fault when called on an unweighted graph. At the very least an exception should be raised and this should be documented as this is totally unexpected as, e.g., `iterEdgesWeights()` works just fine. In the Python interface, a fix should also be possible without introducing a lot of overhead.

fabratu:
Hm, that is a bit surprising.

There are asserts in C++, which test for possible traps - in this case whether the graph is weighted or not. Now that I am testing this, it seems that also other asserts like `exists[u]` are not correctly triggered (tested with a non-existing node in `iterNeighbors(u)`). Currently these cases are not covered by CI.

michitux:
Asserts are only triggered when they are enabled, which is normally not the case in release builds which are used in the Python interface. Further, asserts just terminate the program instead of throwing an exception so they are not the right tool to check input parameters from Python for validity. Or did I misunderstood where these asserts should have been triggered?

fabratu:
Ok, maybe it was just me who was surprised. After a bit more reading on the subject "asserts vs. exceptions" it became clear to me, that I misunderstood the usage of `assert`s (at least in C++). Nevermind my first comment then ...

avdgrinten:
I consider it a serious mistake by CMake to disable assertions in release builds (other build system, for exampe meson, don't do that). But in any case, assertions should not be used to validate input, but to check preconditions.

Error handling in the Graph class is certainly a bit messy. I think we should have a discussion what mistakes the Graph class should protect against. I would be fine with assertions in the C++ code and exceptions in Python (and the `unsafe` C++ methods do not need to do any error checking).

For now (since we did not have such a discussion yet), I would prefer to just catch these mistakes at the Python boundary.

networkit/graph.pyx
```diff
@@ -612,6 +612,37 @@ cdef class Graph:
...
+			preincrement(it)		
+
+	def iterInNeighborsWeights(self, u):
+		"""
+		Iterates over a range of the in-neighbors of a node including the edge weights.
+
+		Parameters:
+		-----------
+		u : Node
+		"""
```
michitux, line 640: Here, the same check should be added - or is this handled in C++?

fabratu: Yes, this also needs it.

michitux (approved):

PR #780, opened 2021-07-01
Refactor: use vector comparator in Aux

angriman:
This PR replaces custom vector-based comparators with the ones implemented in `Aux`.


fabratu (approved):
LGTM.

PR #784, opened 2021-07-02
Feature/plm with custom initialization

jukaradayi:
This is a suggestion to add a partition as an optional parameter for the PLM algorithm. This partition can then be used to initialize the PLM.
This is particularly useful (in my case) when dealing with dynamic graphs, to check the influence of new links on current communities.

The PLM run can still be called without a partition and would run as usual; the CI is unaffected.

### DIFF
--- a/include/networkit/community/PLM.hpp
+++ b/include/networkit/community/PLM.hpp
@@ -34,7 +34,7 @@ public:
      * @param[in] recurse use recursive coarsening, see http://journals.aps.org/pre/abstract/10.1103/PhysRevE.89.049902 for some explanations (default: true)
      *
      */
-    PLM(const Graph& G, Partition zeta=Partition(0), bool refine=false, double gamma = 1.0, std::string par="balanced", count maxIter=32, bool turbo = true, bool recurse = true);
+    PLM(const Graph& G, bool refine=false, double gamma = 1.0, std::string par="balanced", count maxIter=32, bool turbo = true, bool recurse = true, Partition zeta=Partition(0));
 
     PLM(const Graph& G, const PLM& other);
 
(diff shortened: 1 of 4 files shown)


networkit/cpp/community/PLM.cpp
```diff
@@ -51,7 +55,7 @@ void PLM::run() {
     std::vector<double> volCommunity(o, 0.0);
     zeta.parallelForEntries([&](node u, index C) { // set volume for all communities
         if (C != none)
-            volCommunity[C] = volNode[u];
+            volCommunity[C] += volNode[u];
```
michitux, line 58: This needs to be an atomic increment as this in a parallel loop.

hmeyerhenke: Or rather an OpenMP reduction...?

jukaradayi: Good point, I updated the loop, same way as the loop on line 267, thank you !

include/networkit/community/PLM.hpp
```diff
@@ -33,7 +34,7 @@ class PLM final : public CommunityDetectionAlgorithm {
      * @param[in] recurse use recursive coarsening, see http://journals.aps.org/pre/abstract/10.1103/PhysRevE.89.049902 for some explanations (default: true)
      *
      */
-    PLM(const Graph& G, bool refine=false, double gamma = 1.0, std::string par="balanced", count maxIter=32, bool turbo = true, bool recurse = true);
+    PLM(const Graph& G, bool refine=false, double gamma = 1.0, std::string par="balanced", count maxIter=32, bool turbo = true, bool recurse = true, Partition zeta=Partition(0));
```
angriman, line 37: Adding an initial partition looks like a good idea. However, the parameter list is already quite long and setting the partition this way seems inconvenient. What about adding a `setInitialPartition` method? This method could also be overloaded to also take an rvalue ref, so that we can avoid unnecessary copies.

PR #790, opened 2021-07-29
Collection of small fixes / improvements for CMake and CI

fabratu:

- Fixing the documentation build by pinning sphinx to `<4.1` due to an incompatibility with breathe. This is currently worked on: https://github.com/michaeljones/breathe/pull/711 and https://github.com/sphinx-doc/sphinx/issues/9433
- Linking `Python3_libraries` for cython-extensions is only necessary for Windows. While there seems to be no problem with Linux, on macOS this can lead to segmentation faults on certain compiler/lib-combinations (e.g. `conda`) when importing NetworKit. Therefore it is removed for the other OSes.
- Github Actions: 
  - Raising minimum version of gcc to version 6. Also switching from Ubuntu 16.04 to 18.04, since packages are natively available in this distro (universe repo) and 16.04 will deprecate in September 2021.
  - Raising latest gcc to 11. For Ubuntu-images this is done by importing repos from Hirsute, since there is no PPA or gcc-11 in Github runners or base 20.04.
  - Removing unnecessary repos from Linux-builds
  - Adding a timeout to the googletest-step from the coverage-build. Occasionally single test-steps takes too long and get killed by CI (only for combination debug + coverage). Normally it takes 20-30 min, adding a timeout of 60 min should give enough space and seems to diminish this problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,6 +311,7 @@ jobs:
       - name: Prepare environment and run checks
         run:  ${{ github.workspace }}/.github/workflows/scripts/core_sanitizers_coverage.sh
         shell: bash
+        timeout-minutes: 60
       - name: Update Coveralls
         if: github.event_name == 'push' && github.ref == 'refs/heads/master'
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,7 +140,7 @@ jobs:
     strategy:
       matrix:
         os: ['10.15']  # 11.0 temp. removed due to errors in preview image / git actions pipelines
-        compiler: ['clang-12', 'gcc-10']
+        compiler: ['clang-12', 'gcc-11']
     continue-on-error: ${{ matrix.os == '11.0' }}
     steps:
       - name: Install prerequisites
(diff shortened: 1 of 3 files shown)


avdgrinten:
Which tests taste more than 20min? Even with coverage collection enabled, they shouldn't be that slow.

fabratu:
All tests together (~1000) takes more than 20 minutes (so the single-run step from CI). The list of tests, which take longer than the average, includes for example `FiniteEmbeddingTest.testFiniteEmbeddingOnSmallGraph` and `DynBetweennessGTest.runDynVsStaticCaseInsertUndirected`. Have not yet checked thoroughly for a complete list of tests.

avdgrinten:
Okay. (Regardless of this PR, we should identify long-running tests and adjust them if they take excessive amounts of time.)

What does the new timeout achieve? When does GitHub Actions kill the step?

fabratu:
> When does GitHub Actions kill the step?

That's hard to tell. I couldn't find anything on our particular case. In principal CI has a total total (for the complete job) of 6 hours and doesn't care about a more fine-granular stalling. In our case, where the shell-script triggers the googletests, for long running tests, there is no output for a certain time. I would guess, that missing new output is related to triggering a kill.

Example: 
```
1: [----------] 1 test from FiniteEmbeddingTest
1079
1: [ RUN      ] FiniteEmbeddingTest.testFiniteEmbeddingOnSmallGraph
1080
1/2 Test #1: networkit_tests ..................Subprocess killed***Exception: 653.17 sec
```

> What does the new timeout achieve?

This is not confirmed officially, but my guess is that by setting a timeout (which is none-existing for steps by default) progression on a step is not anymore monitored. After adding the timeouts, the following runs did not trigger a kill anymore (which is of course no proof).

angriman (approved):

PR #791, opened 2021-07-29
Add difub for computing the diameter of directed strongly connected networks

noahares:
Compute the diameter of directed graphs with one strongly connected component faster than with the current naive approach.
Implemented as the practical part of @michitux lecture on networks.
Difub: http://link.springer.com/chapter/10.1007/978-3-642-30850-5_10


networkit/cpp/distance/Diameter.cpp
```diff
@@ -52,38 +65,124 @@ std::pair<count, count> Diameter::getDiameter() const {
...
+
+    numBFS = 2;
+
+    // difub
+    for (; i > 0; --i) {
+        handler.assureRunning();
+        count size_f = i < distancesF.size() ? distancesF[i].size() : 0;
+        count size_b = i < distancesB.size() ? distancesB[i].size() : 0;
+#pragma omp parallel for
+        for (count j = 0; j < size_f + size_b; ++j) {
```
michitux, line 141: This needs to be `omp_index` which is a signed integer when using MSVC as - at lest with the parameters NetworKit uses - MSVC doesn't support OpenMP loops with unsigned integers.

noahares: Thanks, I fixed it. Changing `j` to `omp_index` should be enough? In `cpp/distance/APSP.cpp` the loop has a `count` in its condition.

noahares:
Master branch seems to pass CI now, merged it and now this branch should pass it as well.

include/networkit/graph/BFS.hpp
```diff
@@ -37,7 +37,7 @@ auto callBFSHandle(F &f, node u, count) -> decltype(f(u)) {
  * @param handle Takes a node as input parameter.
  */
 template <class InputIt, typename L>
-void BFSfrom(const Graph &G, InputIt first, InputIt last, L handle) {
+void BFSfrom(const Graph &G, InputIt first, InputIt last, L handle, bool reverse = false) {
```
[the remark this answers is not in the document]
noahares: Thank you for the feedback. This sounds like a good way to do it. I will implement it like this.

PR #798, opened 2021-08-12
Update DevGuide

angriman:
Update style guide in the DevGuide.


avdgrinten:
Add a sentence: "No new code files should use `// no-networkit-format`."

PR #800, opened 2021-08-13
Remove GraphBuilder calls when it is is unneeded and remove the GraphBuilder directswap option.

Relux-the-Relux:
DirectSwap doesn't do what its name implies and is unneeded. This PR removes it and makes all classes that used it to handle the Graph directly to avoid the GraphBuilder overhead. So that only classes that make use of the Edge completion feature use the GraphBuilder.


include/networkit/graph/GraphBuilder.hpp
```diff
@@ -166,7 +166,7 @@ class GraphBuilder {
     /**
      * Generates a Graph instance. The graph builder will be reseted at the end.
      */
-    Graph toGraph(bool autoCompleteEdges, bool parallel = false);
+    Graph toGraph(bool parallel = false);
```
avdgrinten, line 169: We should probably not change the signature of the function silently. Instead, let it still take the autoCompleteEdges parameter but throw an exception if it is set to false. Also deprecate the function.

We can add another `Graph completeGraph()` function that does not take the parameter. (And `toGraph()` can simply call `completeGraph()`.

avdgrinten:
Please do add the new `completeGraph()` function mentioned above and mark the old function as deprecated.

avdgrinten (approved):
LGTM.

PR #802, opened 2021-08-16
Added overloaded function that returns a reference instead of copying in Graphcoarsening

CxVercility:
As discussed in https://github.com/networkit/networkit/issues/801
Avoids copies, and allows the graph to be moved.

Perhaps there should be a check that makes sure the graph hasnt been moved or changed if a call to `getCoarseToFineNodeMapping` is made, but im not sure how that would work.

### DIFF
--- a/networkit/cpp/coarsening/GraphCoarsening.cpp
+++ b/networkit/cpp/coarsening/GraphCoarsening.cpp
@@ -14,20 +14,25 @@ GraphCoarsening::GraphCoarsening(const Graph& G) : Algorithm(), G(&G) {
 
 }
 
-Graph GraphCoarsening::getCoarseGraph() const {
-    if(!hasRun) {
-        throw std::runtime_error("Call run()-method first.");
-    }
+const Graph& GraphCoarsening::getCoarseGraph() const {
+    assureFinished();
     return Gcoarsened;
 }
 
-std::vector<node> GraphCoarsening::getFineToCoarseNodeMapping() const {
-    if(!hasRun) {
-        throw std::runtime_error("Call run()-method first.");
-    }
+Graph& GraphCoarsening::getCoarseGraph() {
+    assureFinished();
+    return Gcoarsened;
+}
+
+const std::vector<node>& GraphCoarsening::getFineToCoarseNodeMapping() const {
+    assureFinished();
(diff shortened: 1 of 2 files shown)


CxVercility:
Also changed the const function to return `const Graph&` instead of `Graph`
As discussed in https://github.com/networkit/networkit/issues/801 again.

networkit/cpp/coarsening/GraphCoarsening.cpp
```diff
@@ -14,7 +14,14 @@ GraphCoarsening::GraphCoarsening(const Graph& G) : Algorithm(), G(&G) {
 
 }
 
-Graph GraphCoarsening::getCoarseGraph() const {
+const Graph& GraphCoarsening::getCoarseGraph() const {
+    if(!hasRun) {
+        throw std::runtime_error("Call run()-method first.");
+    }
```
angriman, line 20: Here you can use `assureFinished`

CxVercility: It's really the same except for a slightly different message isnt it? 
Shouldnt the other methods be using `assureFinished()` then too?

angriman: Yes, it's the standard way of assuring that the algorithm was executed. It was probably introduced later than some classes so we still have some instances of `if (!hasRun) throw ...`

CxVercility: Alright, i replaced it in all three functions.

angriman:
Thank you, the changes look good to me. Can you please squash the three commits into a single one?

CxVercility:
> Thank you, the changes look good to me. Can you please squash the three commits into a single one?

I actually just realized that the same issue exists for the `getFineToCoarseNodeMapping` while benchmarking my algorithm.
Unless there's a reason not to, i'd also include a non-const overloaded function for this and make the existing const one return `const vector&`
Ill squash all of them together then.

angriman:
I think that's reasonable.

CxVercility:
> I think that's reasonable.

Done. Thanks for the quick responses.

angriman (approved):

PR #803, opened 2021-08-17
clang-format: algebraic/algorithms

angriman:
As discussed, we should slowly adapt non-compliant files to our coding style. The idea is to format a few files per PR, none of which is being modified by other PRs. This PR formats files in `algebraic/algorithms`


avdgrinten (approved):
LGTM.

PR #804, opened 2021-08-19
Refactor: use vector comparator in Dijkstra

angriman:
Replace custom vector-based comparator in `Dijkstra` with the one implemented in `Aux`.


fabratu (approved):
LGTM.

PR #805, opened 2021-08-20
Add citation file

angriman:
GitHub recently introduced enhanced support for citations [1]. This PR adds a new `CITATION.cff` for NetworKit. To me, the solution that seems to work best is to add the NetworKit journal paper as the `preferred-citation` (see [2]), which can later be updated with more recent NetworKit papers. If needed, we can add further citations as `references`.

[1] https://github.blog/2021-08-19-enhanced-support-citations-github/
[2] https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md


fabratu (approved):
LGTM.

PR #808, opened 2021-08-31
Fix to call subgraphFromNodes

muik:
`subgraphFromNodes()` function was moved from the `graph`-module to the `graphtools`-module [in NetworKit 8.0](https://github.com/networkit/networkit/blob/master/CHANGES.md#changes-in-python-interface). But there was one thing missing, so I fixed it.


angriman (approved):
Thank you, LGTM

PR #810, opened 2021-09-28
Refactor PageRank: Add sink handling for directed graphs, fix norm-PageRank

fabratu:
This PR includes the following concerning PageRank:

- Adds sink-handling in order to properly fulfill the Markov chain model. See (among others): https://dl.acm.org/doi/pdf/10.1145/1151087.1151090
- Fixes an error with the normalized PageRank #720 (NPR). The result of NPR was wrong due to missing sink-handling - so this fix is more a drive-by.
- Adds documentation for new PageRank behavior and results
- Adjusting gtests accordingly

### DIFF
--- a/include/networkit/centrality/PageRank.hpp
+++ b/include/networkit/centrality/PageRank.hpp
@@ -18,7 +18,19 @@ namespace NetworKit {
 
 /**
  * @ingroup centrality
- * Compute PageRank as node centrality measure.
+ * Compute PageRank as node centrality measure. In the default mode this computation is in line
+ * with the original paper "The PageRank citation ranking: Bringing order to the web." by L. Brin et
+ * al (1999). In later publications ("PageRank revisited." by M. Brinkmeyer et al. (2005) amongst
+ * others) sink-node handling was added for directed graphs in order to comply with the theoretical
+ * assumptions by the underlying Markov chain model. This can be activated by setting the matching
+ * parameter to true. By default this is disabled, since it is an addition to the original
+ * definition.
+ *
+ * Page-Rank values can also be normalized by post-processed according to "Comparing Apples and
+ * Oranges: Normalized PageRank for Evolving Graphs" by Berberich et al. (2007). This decouples
+ * the PageRank values from the size of the input graph. To enable this, set the matching parameter
+ * to true. Note that sink-node handling is automatically activated if normalization is used.
+ *
  * NOTE: There is an inconsistency in the definition in Newman's book (Ch. 7) regarding
  * directed graphs; we follow the verbal description, which requires to sum over the incoming
  * edges (as opposed to outgoing ones).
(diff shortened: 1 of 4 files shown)


avdgrinten:
Deviations from the original paper should probably be gated behind flags.

hmeyerhenke:
It's more of a model fix than a real deviation. But I see your point. Still, I think it would be correct to use this as the default approach. @fabratu, are there any significant performance implications?

fabratu:
Didn't do a thorough measurement, Since sink-handling is done after each iteration, results can of course differ quite a bit for different graphs. For a smaller graph (`foodweb-baydry`) the time roughly doubles from 30μs to 60μs.

fabratu:
In any case, for the normalized PageRank the sink-handling has to be default case to be in line if it's original paper.

avdgrinten:
It's not clear to me whether special casing sink nodes it required to make the measure meaningful. As far as I can see, the special handling yields a different measure but not necessarily a more "correct" one.

fabratu:
> It's not clear to me whether special casing sink nodes it required to make the measure meaningful. As far as I can see, the special handling yields a different measure but not necessarily a more "correct" one.

All right - I changed the default behavior back to "classical" PageRank with an additional parameter. The normalized PageRank is done with sink handling in any case due to its original paper.

include/networkit/centrality/PageRank.hpp
```diff
@@ -34,17 +46,21 @@ class PageRank final : public Centrality {
...
      * @param[in] damp Damping factor of the PageRank algorithm.
      * @param[in] tol Error tolerance for PageRank iteration.
+     * @param[in] distributeSinks Set to distribute PageRank values for sink nodes. (default =
+     * false)
+     * @param[in] normalized Set to normalize Page-Rank values in order to decouple it from the
+     * network size. (default = false)
      */
-    PageRank(const Graph &G, double damp = 0.85, double tol = 1e-8, bool normalized = false);
+    PageRank(const Graph &G, double damp = 0.85, double tol = 1e-8, bool distributeSinks = false,
+             bool normalized = false);
```
avdgrinten, line 55: This changes the meaning of a call such as `PageRank(G, 0.5, 1e-3, true)` (which now enables distributeSinks instead of normalization).

Please make `distributeSinks` the last parameter. Please also make it use an `enum class SinkHandling { NO_SINK_HANDLING, DISTRIBUTE_SINKS }` or something similar.

fabratu: Added an enum. Currently it seems, that `enum class` is not directly supported in Cython (will be supported in 0.3.0). I was not able to successfully apply any of the known workarounds. Also tried some type-casting. In the end everytime is boils down to some of those:
```
...
/Users/fbt/Documents/maks/networkit/main/fork2/networkit/centrality.cpp:39886:132: error: invalid operands to binary expression ('enum NetworKit::PageRank::SinkHandling' and 'int')
                            return (enum NetworKit::PageRank::SinkHandling) (((((enum NetworKit::PageRank::SinkHandling)digits[1]) << PyLong_SHIFT) | (enum NetworKit::PageRank::SinkHandling)digits[0]));
...
```

This also happens if other enums are changed to `enum class`.

avdgrinten (approved):
LGTM now.

PR #811, opened 2021-10-01
Add Parallel Leiden Algorithm

CxVercility:
I implemented a parallel version of the Leiden algorithm for my [bachelor thesis](https://zenodo.org/record/5564953)
This code is basically a stripped version of my [own repository](https://github.com/CxVercility/ParallelLeiden) (since it was cluttered with benchmark code etc) which also contains sequential versions and a second parallel version for both local moving and refinement, though the alternative parallel versions haven't been tested thoroughly so I wouldnt include them.

This was my first C++ project ever and I didnt really put a lot of effort into making the code look nice so Im not expecting this to go through asit is. Feel free to change or add anything you like, if you want only minor changes done i'll be happy to do those as well. 

Greets.

### DIFF
--- a/include/networkit/community/ParallelLeiden.hpp
+++ b/include/networkit/community/ParallelLeiden.hpp
@@ -1,0 +1,90 @@
+#ifndef NETWORKIT_COMMUNITY_PARALLEL_LEIDEN_HPP_
+#define NETWORKIT_COMMUNITY_PARALLEL_LEIDEN_HPP_
+
+#include <atomic>
+#include <condition_variable>
+#include <iomanip>
+#include <iostream>
+#include <mutex>
+#include <omp.h>
+#include <thread>
+#include <tlx/unused.hpp>
+#include <networkit/Globals.hpp>
+#include <networkit/auxiliary/Parallel.hpp>
+#include <networkit/auxiliary/Parallelism.hpp>
+#include <networkit/auxiliary/SignalHandling.hpp>
+#include <networkit/auxiliary/Timer.hpp>
+#include <networkit/coarsening/ParallelPartitionCoarsening.hpp>
+#include <networkit/community/CommunityDetectionAlgorithm.hpp>
+#include <networkit/community/Modularity.hpp>
+#include <networkit/community/PLM.hpp>
+#include <networkit/graph/Graph.hpp>
+#include <networkit/structures/Partition.hpp>
+
+namespace NetworKit {
+
+class ParallelLeiden final : public CommunityDetectionAlgorithm {
+public:
+    /**
+     *
+     * @param graph A networkit graph
+     * @param iterations Number of Leiden Iterations to be run
+     * @param randomize Randomize node order?
+     * @param gamma Resolution parameter
+     */
+    explicit ParallelLeiden(const Graph &graph, int iterations = 3, bool randomize = true,
+                            double gamma = 1);
+
+    void run() override;
+
+    int VECTOR_OVERSIZE = 10000;
+
+private:
+    inline double modularityDelta(double cutD, double degreeV, double volD) const {
+        return cutD - gamma * degreeV * volD * inverseGraphVolume;
+    };
+
+    inline double modularityThreshold(double cutC, double volC, double degreeV) const {
+        return cutC - gamma * (volC - degreeV) * degreeV * inverseGraphVolume;
+    }
+
+    static inline void lockLowerFirst(index a, index b, std::vector<std::mutex> &locks) {
+        if (a < b) {
+            locks[a].lock();
+            locks[b].lock();
+        } else {
+            locks[b].lock();
+            locks[a].lock();
+        }
+    }
+
+    void flattenPartition();
+
+    void calculateVolumes(const Graph &graph);
+
+    void parallelMove(const Graph &graph);
+
+    Partition parallelRefine(const Graph &graph);
+
+    double inverseGraphVolume; // 1/vol(V)
+
+    std::vector<double> communityVolumes;
+
+    std::vector<std::vector<node>> mappings;
+
+    static constexpr int WORKING_SIZE = 1000;
+
+    double gamma; // Resolution parameter
+
+    bool changed;
+
+    int numberOfIterations;
+
+    Aux::SignalHandler handler;
+
+    bool random;
+};
+
+} // namespace NetworKit
+
+#endif // NETWORKIT_COMMUNITY_PARALLEL_LEIDEN_HPP_
--- a/networkit/community.pyx
+++ b/networkit/community.pyx
@@ -418,6 +418,34 @@ cdef class PLM(CommunityDetector):
 	def prolong(Graph Gcoarse, Partition zetaCoarse, Graph Gfine, vector[node] nodeToMetaNode):
 		return Partition().setThis(PLM_prolong(Gcoarse._this, zetaCoarse._this, Gfine._this, nodeToMetaNode))
 
+cdef extern from "<networkit/community/ParallelLeiden.hpp>":
+
+	cdef cppclass _ParallelLeiden "NetworKit::ParallelLeiden"(_CommunityDetectionAlgorithm):
+		_ParallelLeiden(_Graph _G) except +
+		_ParallelLeiden(_Graph _G, int iterations, bool_t randomize, double gamma) except +
+
+cdef class ParallelLeiden(CommunityDetector):
+	""" Parallel Leiden Algorithm
+
+		Parameters:
+		-----------
+		G : networkit.Graph
+			A graph.
+		randomize : bool, optional
+			Whether to randomize the node order or not
+		iterations : count
+            Maximum count of Leiden runs
+		gamma : double
+			Multi-resolution modularity parameter:
+			1.0 -> standard modularity
+	 		0.0 -> one community
+	 		2m 	-> singleton communities
+	"""
+
+	def __cinit__(self, Graph G not None, int iterations = 3, bool_t randomize = True, double gamma = 1):
+		self._G = G
+		self._this = new _ParallelLeiden(G._this,iterations,randomize,gamma)
+
 cdef extern from "<networkit/community/LouvainMapEquation.hpp>":
 	cdef cppclass _LouvainMapEquation "NetworKit::LouvainMapEquation"(_CommunityDetectionAlgorithm):
 		_LouvainMapEquation(_Graph, bool, count, string ) except +
--- a/networkit/cpp/community/ParallelLeiden.cpp
+++ b/networkit/cpp/community/ParallelLeiden.cpp
@@ -1,0 +1,536 @@
+#include <networkit/community/ParallelLeiden.hpp>
+
+namespace NetworKit {
+ParallelLeiden::ParallelLeiden(const Graph &graph, int iterations, bool randomize, double gamma)
+    : CommunityDetectionAlgorithm(graph), gamma(gamma), numberOfIterations(iterations),
+      random(randomize) {
+    this->result = Partition(graph.numberOfNodes());
+    this->result.allToSingletons();
+}
+
+void ParallelLeiden::run() {
+    if (VECTOR_OVERSIZE < 1) {
+        throw std::invalid_argument("VECTOR_OVERSIZE cant be smaller than 1");
+    }
+    auto totalTime = Aux::Timer();
+    totalTime.start();
+    do { // Leiden iteration
+        INFO(numberOfIterations, " Leiden iteration(s) left");
+        numberOfIterations--;
+        changed = false;
+        const Graph *currentGraph = G;
+        Graph coarse;
+        Partition refined;
+        calculateVolumes(*currentGraph);
+        do {
+            handler.assureRunning();
+            parallelMove(*currentGraph);
+            // If each community consists of exactly one node we're done, i.e. when |V(G)| = |P|
+            if (currentGraph->numberOfNodes() != result.numberOfSubsets()) {
+                break;
+            }
+            handler.assureRunning();
+            refined = parallelRefine(*currentGraph);
+            handler.assureRunning();
+            ParallelPartitionCoarsening ppc(*currentGraph, refined); // Aggregate graph
+            ppc.run();
+            auto temp = std::move(ppc.getCoarseGraph());
+            auto map = std::move(ppc.getFineToCoarseNodeMapping());
+            // Maintain Partition, add every coarse Node to the community its fine Nodes were in
+            //  unlike in louvain, 2 coarse Nodes can belong to the same community
+            Partition p(temp.numberOfNodes());
+            p.setUpperBound(result.upperBound());
+            currentGraph->parallelForNodes([&](node u) { p[map[u]] = result[u]; });
+
+            mappings.emplace_back(std::move(map));
+            result = std::move(p);
+            std::swap(temp, coarse);
+            currentGraph = &coarse;
+        } while (true);
+        flattenPartition();
+        INFO("Leiden iteration done, took ", totalTime.elapsedTag(), "so far");
+    } while (changed && numberOfIterations > 0);
+    hasRun = true;
+}
+
+void ParallelLeiden::calculateVolumes(const Graph &graph) {
+    auto timer = Aux::Timer();
+    timer.start();
+    // thread safe reduction. Avoid atomic calculation of total graph volume for unweighted graphs.
+    // Vol(G) is then 2*|E|
+    communityVolumes.clear();
+    communityVolumes.resize(result.upperBound() + VECTOR_OVERSIZE);
+    if (graph.isWeighted()) {
+        std::vector<double> threadVolumes(omp_get_max_threads());
+        graph.parallelForNodes([&](node a) {
+            {
+                edgeweight ew = graph.weightedDegree(a, true);
+#pragma omp atomic
+                communityVolumes[result[a]] += ew;
+                threadVolumes[omp_get_thread_num()] += ew;
+            }
+        });
+        for (const auto vol : threadVolumes) {
+            inverseGraphVolume += vol;
+        }
+        inverseGraphVolume = 1 / inverseGraphVolume;
+    } else {
+        inverseGraphVolume = 1.0 / (2 * graph.numberOfEdges());
+        graph.parallelForNodes([&](node a) {
+            {
+#pragma omp atomic
+                communityVolumes[result[a]] += graph.weightedDegree(a, true);
+            }
+        });
+    }
+    TRACE("Calculating Volumes took " + timer.elapsedTag());
+}
+
+void ParallelLeiden::flattenPartition() {
+    auto timer = Aux::Timer();
+    timer.start();
+    if (mappings.empty()) {
+        return;
+    }
+    // Create a new partition with size |V(G)| (the fine/bigger Graph)
+    Partition flattenedPartition(G->numberOfNodes());
+    flattenedPartition.setUpperBound(result.upperBound());
+    int i = mappings.size() - 1;
+    std::vector<node> &lower = mappings[i--];
+    while (i >= 0) {
+        // iteratively "resolve" (i.e compose) mappings. Let "lower" be a mapping thats
+        // below "higher" in the hierarchy (i.e. of a later aggregation)
+        // If higher[index] = z and lower[z] = x then set higher[index] = x
+        std::vector<node> &upper = mappings[i--];
+        for (auto &idx : upper) {
+            idx = lower[idx];
+        }
+        lower = upper;
+    }
+    G->parallelForNodes([&](node a) { flattenedPartition[a] = lower[a]; });
+    flattenedPartition.compact(true);
+    result = flattenedPartition;
+    mappings.clear();
+    TRACE("Flattening partition took " + timer.elapsedTag());
+}
+
+void ParallelLeiden::parallelMove(const Graph &graph) {
+    DEBUG("Local Moving : " + std::to_string(graph.numberOfNodes()) + " Nodes ");
+    std::vector<count> moved(omp_get_max_threads(), 0);
+    std::vector<count> totalNodesPerThread(omp_get_max_threads(), 0);
+    std::atomic_int singleton(0);
+    //^^^ Debug stuff
+
+    // Only insert nodes to the queue when they're not already in it.
+    std::vector<std::atomic_bool> inQueue(graph.upperNodeIdBound());
+    std::queue<std::vector<node>> queue;
+    std::mutex qlock;                      // queue lock
+    std::condition_variable workAvailable; // waiting/notifying for new Nodes
+
+    std::atomic_bool resize(false);
+    std::atomic_int waitingForResize(0);
+    std::atomic_int waitingForNodes(0);
+
+    std::vector<int> order;
+    int tshare;
+    int tcount;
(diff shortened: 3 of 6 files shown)


CxVercility:
Is it on purpose that unsigned index counts as an error in the windows build?
there is a flag to change that and i dont see how it would be relevant https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/#new-openmpllvm-switch

avdgrinten:
OpenMP support on Windows is a lot better now than a few years ago. However, let's stick to the current compiler version and command line switches for now. It shouldn't be a lot of work to cast loop indices to signed numbers.

avdgrinten:
Out of curiosity: do you have any benchmarks against NetworKit's parallel Louvain implementation?

CxVercility:
> OpenMP support on Windows is a lot better now than a few years ago. However, let's stick to the current compiler version and command line switches for now. It shouldn't be a lot of work to cast loop indices to signed numbers.

Probably not, it's just weird to have two casts within one for loop, since I have to cast graph.numberOfNodes to signed as well, otherwise it breaks Linux builds due to comparison between different signedness. Ill try doing that later.

As for PLM, I only bench'ed it on the uk2007 graph (ran out of time for more) but the differences between the algorithms makes it hard to compare.

The tl:dr is that plm is faster and shows slightly higher speed ups in my benchmark. Though I didnt turn refinement on, and my leiden Implementation was always run three times. I dont know whether two times would have been enough to achieve similar quality or not.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,577 @@
...
+    result = flattenedPartition;
+    mappings.clear();
+    PLTRACE("Flattening partition took " << timer.elapsedMilliseconds() / 1000.0 << "s")
+}
+
+void ParallelLeiden::parallelMove(const Graph &graph) {
+#ifdef PLFINE
+    cout << "Local Moving : " << graph.numberOfNodes() << " Nodes " << endl;
+    std::vector<long> moved(omp_get_max_threads(), 0);
+    std::vector<long> totalNodesPerThread(omp_get_max_threads(), 0);
```
fabratu, line 142: Instead of using `long` (signed datatype) as vector-items, better use `count` (alias for `uint64_t`) to avoid overflow for huge number of nodes.  In addition -  if I understand your code correctly, the items are always positive anyways.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,577 @@
...
+             << totalNodesPerThread[omp_get_thread_num()] << "Nodes and moved "
+             << moved[omp_get_thread_num()] << endl;
+#endif
+    }
+    result.setUpperBound(upperBound);
+    assert(queue.empty());
+    assert(waitingForNodes == tcount);
+#ifdef PLFINE
+    long totalMoved = accumulate(moved.begin(), moved.end(), 0L);
+    long totalWorked = accumulate(totalNodesPerThread.begin(), totalNodesPerThread.end(), 0L);
```
fabratu, line 365: If you change `moved` and `totalNodesPerThread` datatypes, also apply it here.

fabratu (changes requested):
Sry for the waiting time. Thanks again for your contribution! 

Overall the code is in a good shape, so the review does not contain any major roadblock. Note that for some review-comments I did not mark all occasions.

Concerning the column limit: I think we have a pretty standard line length, popular code-styles like Google, Micrsoft and LLVM range between 80 and 120.  Of course that is always depending on personal taste (my setup is tree-view + two split window on a WQHD screen). Feel free to open a idea thread in "Discussions" so we can take a look at this.

include/networkit/community/ParallelLeiden.hpp
```diff
@@ -0,0 +1,102 @@
...
+private:
+    inline double modularityDelta(double cutD, double degreeV, double volD) const {
+        return cutD - gamma * degreeV * volD * inverseGraphVolume;
+    };
+
+    inline double modularityThreshold(double cutC, double volC, double degreeV) const {
+        return cutC - gamma * (volC - degreeV) * degreeV * inverseGraphVolume;
+    }
+
+    static inline void lockLowerFirst(index a, index b, std::vector<std::mutex> &locks) {
```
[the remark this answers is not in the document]
CxVercility: Mostly a "doesn't hurt so why not" thing. IDE kept bugging me about it.

CxVercility:
Some errors that didn't up last time, weird. Also accidentally messed up my commit, I'll get to it when I get off of work tomorrow. Or maybe some other day.

include/networkit/community/ParallelLeiden.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <networkit/auxiliary/SignalHandling.hpp>
+#include <networkit/auxiliary/Timer.hpp>
+#include <networkit/coarsening/ParallelPartitionCoarsening.hpp>
+#include <networkit/community/CommunityDetectionAlgorithm.hpp>
+#include <networkit/community/Modularity.hpp>
+#include <networkit/community/PLM.hpp>
+#include <networkit/graph/Graph.hpp>
+#include <networkit/structures/Partition.hpp>
+
+using namespace NetworKit;
```
angriman, line 23: `ParallelLeiden` is already defined within the `NetworKit` namespace, this line should not be necessary.

angriman (changes requested):

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,529 @@
+#include <networkit/community/ParallelLeiden.hpp>
+
+#define WORKING_SIZE 1000
```
angriman, line 3: It's cleaner to define this a `static constexpr` member variable.

include/networkit/community/ParallelLeiden.hpp
```diff
@@ -0,0 +1,81 @@
...
+using namespace std;
```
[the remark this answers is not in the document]
avdgrinten: Same holds for the NetworKit namespace.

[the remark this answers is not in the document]
CxVercility: Didn't mean to, just massively messed up my working directory somewhere along the way and lost changes without noticing. I'll get to it later today.

CxVercility:
I'm just too stupid for git, I swear. What even is rebasing.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,530 @@
...
+    mappings.clear();
+    TRACE("Flattening partition took " + timer.elapsedTag());
+}
+
+void ParallelLeiden::parallelMove(const Graph &graph) {
+    DEBUG("Local Moving : " + std::to_string(graph.numberOfNodes()) + " Nodes ");
+    std::vector<count> moved(omp_get_max_threads(), 0);
+    std::vector<count> totalNodesPerThread(omp_get_max_threads(), 0);
+    std::atomic_int singleton(0);
+    //^^^ Debug stuff
```
angriman, line 120: Are these only needed for debug purposes right? Why don't you use them only in debug mode with `#ifndef NDEBUG`?

CxVercility: Mainly because of readability. I used to have these enclosed by ifndef's but I'd also have to enclose all the increments in the middle of the code.

If you don't care about that, I'll enclose them again.

CxVercility: Namely, these two:
```if (0 > modThreshold || maxDelta > modThreshold) {
                    moved[omp_get_thread_num()]++;
                    if (0 > maxDelta) { // move node to empty community
                        singleton++;
```
Thats why i used to have a macro that only inserted that code when ndebug wasn't present (which made it look like a function call / a one liner, instead of the one line + two additional ones)

angriman: I see the readability argument. It might be a good idea to add a similar macro to NetworKit instead of defining it for just one class.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,530 @@
...
+        int end = (1 + order[omp_get_thread_num()]) * tshare;
+
+        for (int i = start; i < end; i++) {
+            if (graph.hasNode(i)) {
+                currentNodes.push_back(i);
+                atomic_init(&inQueue[i], true);
+            }
+        }
+        if (random)
+            shuffle(currentNodes.begin(), currentNodes.end(), mt);
```
angriman, line 168: Not sure why this compiles. Don't you want to use `std::shuffle` here? There are two other occurrences of this.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,530 @@
...
+                if (0 > modThreshold || maxDelta > modThreshold) {
+                    moved[omp_get_thread_num()]++;
+                    if (0 > maxDelta) { // move node to empty community
+                        singleton++;
+                        bestCommunity = upperBound++;
+                        if (bestCommunity >= communityVolumes.capacity()) {
+                            // Chances are this will never happen. Ever. Seriously...
+                            bool expected = false;
+                            if (resize.compare_exchange_strong(expected, true)) {
+                                vectorSize += 10000;
```
angriman, line 234: Maybe store `10000` in a member variable, it's more expressive and you avoid magic numbers.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,530 @@
...
+    std::vector<std::mutex> locks(refined.upperBound());
+    std::vector<node> nodes(graph.upperNodeIdBound(), none);
+#pragma omp parallel
+    {
+        std::vector<index> neighComms;
+        // Keeps track of relevant Neighbor communities. Needed to reset the clearlist fast
+        std::vector<double> cutWeights(refined.upperBound()); // cut from Node to Communities
+        auto &mt = Aux::Random::getURNG();
+#pragma omp for
+        for (long u = 0; u < static_cast<long>(graph.upperNodeIdBound()); u++) {
```
angriman, line 355: Here you can use the `omp_index` type, it is defined in `Globals.hpp`.

networkit/community.pyx
```diff
@@ -418,6 +418,34 @@ cdef class PLM(CommunityDetector):
 	def prolong(Graph Gcoarse, Partition zetaCoarse, Graph Gfine, vector[node] nodeToMetaNode):
 		return Partition().setThis(PLM_prolong(Gcoarse._this, zetaCoarse._this, Gfine._this, nodeToMetaNode))
 
+cdef extern from "<networkit/community/ParallelLeiden.hpp>":
+
+	cdef cppclass _ParallelLeiden "NetworKit::ParallelLeiden"(_CommunityDetectionAlgorithm):
+		_ParallelLeiden(_Graph _G) except +
+		_ParallelLeiden(_Graph _G, int iterations, bool_t randomize, double gamma) except +
+
+cdef class ParallelLeiden(CommunityDetector):
```
angriman, line 427: Please add a simple Python test (mainly to test the Python interface).

networkit/cpp/community/test/CommunityGTest.cpp
```diff
@@ -174,6 +175,28 @@ TEST_F(CommunityGTest, testPLM) {
...
+    EXPECT_TRUE(GraphClusteringTools::isProperClustering(G, zeta));
+
+    ParallelLeiden plnr(G,3,false);
+    plnr.run();
+    Partition zeta2 = plnr.getPartition();
+
+    DEBUG("number of clusters: " , zeta2.numberOfSubsets());
+    DEBUG("modularity: " , modularity.getQuality(zeta2, G));
+    EXPECT_TRUE(GraphClusteringTools::isProperClustering(G, zeta2));
+}
```
angriman, line 198: The algorithm seems quite complex but you have just one test. Does this cover all the relevant parts of the algorithm?

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: Perhaps, This would probably belong to a new pull request though. 
Where do you suggest it to be put then, and it's name?
Simply DEBUGONLY ?

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: Good question. I used to have Std as a global namespace there, I don't know why this didn't explode. I'll add the std:: anyway.

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: Will do.

networkit/community.pyx
CxVercility: Is there a similar one for plm?
Truth is, I haven't touched python at all so all I'm really doing there is copying plm and eyeballing which changes are needed lol.

networkit/cpp/community/ParallelLeiden.cpp
angriman: Yes, this should be implemented in another PR. The details about how to implement this can be discussed in a new issue/discussion.

networkit/cpp/community/test/CommunityGTest.cpp
CxVercility: Not at all. I used to "test" my algorithm simply by comparing the results and whether or not the algorithm crashed. 
I didn't really come up with any good ideas to test individual parts of it due to its non-deterministic behavior. It's also unclear when a result is found to be good or not, since higher modularity doesn't necessarily mean a better result. I also had no necessity for them, as my code isn't even evaluated as part of my Bachelor thesis.

The only other simple tests I can come up with on the spot is checking that every node is assigned to a community. One could also check that every node in the refined partition is still within its host community, which would be a little more effort though.

networkit/cpp/community/ParallelLeiden.cpp
avdgrinten: It works by chance since the iterator is also in namespace std and this triggers ADL (argument dependent lookup).

networkit/cpp/community/ParallelLeiden.cpp
avdgrinten: The entire concurrent resizing logic is a bit hard to read. Can you add a comment why this is safe? I assume that it is because no threads access `cutWeights` and `communityVolumes` while the resize operation is in progress? Why is this guaranteed to hold?

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: The type has to be signed because of openmp restrictions

angriman: Only if you use MSVC; in that case, `omp_index` is signed.

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: cutweights is thread-local and communityvolumes is only resized when all threads have yielded so no access can occur.

Ive added this in the code.

networkit/cpp/community/ParallelLeiden.cpp
CxVercility: Oh, ok

networkit/community.pyx
angriman: Yes, here is an example: https://github.com/networkit/networkit/blob/master/networkit/test/test_algorithms.py#L307

networkit/cpp/community/ParallelLeiden.cpp
avdgrinten: Thank you, that clears it up!

networkit/cpp/community/test/CommunityGTest.cpp
avdgrinten: For non-deterministic algos, it's OK to simply test that they don't crash.

If possible (in reasonable time), it would be nice if the "algo does not crash" tests are exhaustive though (= all cases of the algorithm are covered). Does the current test trigger all code paths (more-or-less)?

CxVercility: No, the resize logic is only triggered if the shared communityVolumes vector becomes too small. This doesn't even happen on the largest graph I  found (3b edges, 100m nodes) so it'd require making the default oversizing of the vectors smaller for the test specifically.

Since im not aware of any methods to change private variables (like reflection in Java) this would mean making the oversize variable public / adding a setter.

Apart from that every line of code should be triggered, unless the graph is small enough to make the algorithm stop after a single local moving.

networkit/cpp/community/ParallelLeiden.cpp
angriman: As discussed in #839, I take back my comment. It should not be too costly anyway to have those vectors in release mode as well.

networkit/cpp/community/test/CommunityGTest.cpp
avdgrinten: Making the variables public and writable sounds fine to me.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,534 @@
...
+      random(randomize) {
+    this->result = Partition(graph.numberOfNodes());
+    this->result.allToSingletons();
+}
+
+void ParallelLeiden::run() {
+    auto totalTime = Aux::Timer();
+    totalTime.start();
+    do { // Leiden iteration
+        INFO(std::to_string(numberOfIterations) + " Leiden iteration(s) left");
```
angriman, line 15:
```suggestion
        INFO(numberOfIterations, " Leiden iteration(s) left");
```

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,534 @@
...
+        bool done; // have we moved any nodes? If yes, start local moving phase again. Else stop
+        const Graph *currentGraph = G;
+        Graph coarse;
+        Partition refined;
+        calculateVolumes(*currentGraph);
+        do {
+            handler.assureRunning();
+            parallelMove(*currentGraph);
+            // If each community consists of exactly one node we're done, i.e. when |V(G)| = |P|
+            done = currentGraph->numberOfNodes() == result.numberOfSubsets();
```
angriman, line 27: Maybe this is a bit simpler:
```suggestion
            if (currentGraph->numberOfNodes() == result.numberOfSubsets())
                break;
```

networkit/cpp/community/test/CommunityGTest.cpp
CxVercility: Do you want me to make the other constexpr variable public then as well? (The one WORKING_SIZE one. Specifies how many nodes are collected in a thread until the vector is pushed to a shared queue)

avdgrinten: You can probably just expose whatever is necessary for the test (unless that leads to a very ugly interface or so). If there are no invariants (or only straightforward ones that can be checked at the top of `run()`), I'd prefer public members over setters + getters.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,536 @@
...
+                    }
+                }
+                uniqueLock.unlock(); // Notified and all done, stop.
+                workAvailable.notify_all();
+                break;
+            }
+        } while (true);
+        TRACE("Thread " + std::to_string(omp_get_thread_num()) + " worked "
+              + std::to_string(totalNodesPerThread[omp_get_thread_num()]) + "Nodes and moved "
+              + std::to_string(moved[omp_get_thread_num()]));
```
angriman, line 329: `TRACE` accepts variadic arguments, so you can avoid manually concatenating strings by passing multiple arguments to `TRACE`. For example `TRACE("Thread ", omp_get_thread_num(), " worked ", ...) `. If you do that, you can also avoid using `std::to_string`, as this is already done automatically by `TRACE`.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,536 @@
...
+    }
+    result.setUpperBound(upperBound);
+    assert(queue.empty());
+    assert(waitingForNodes == tcount);
+    if (Aux::Log::isLogLevelEnabled(Aux::Log::LogLevel::debug)) {
+        count totalMoved = accumulate(moved.begin(), moved.end(), (count)0);
+        count totalWorked =
+            accumulate(totalNodesPerThread.begin(), totalNodesPerThread.end(), (count)0);
+        std::cout << "[DEBUG] Total worked: " + std::to_string(totalWorked)
```
angriman, line 338: Same as above, here you can use the `DEBUG` macro instead of `std::cout`.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,536 @@
...
+              + std::to_string(totalNodesPerThread[omp_get_thread_num()]) + "Nodes and moved "
+              + std::to_string(moved[omp_get_thread_num()]));
+    }
+    result.setUpperBound(upperBound);
+    assert(queue.empty());
+    assert(waitingForNodes == tcount);
+    if (Aux::Log::isLogLevelEnabled(Aux::Log::LogLevel::debug)) {
+        count totalMoved = accumulate(moved.begin(), moved.end(), (count)0);
+        count totalWorked =
+            accumulate(totalNodesPerThread.begin(), totalNodesPerThread.end(), (count)0);
```
angriman, line 337: Here you should use `std::accumulate`.

networkit/cpp/community/ParallelLeiden.cpp
```diff
@@ -0,0 +1,536 @@
...
+    }
+    result.setUpperBound(upperBound);
+    assert(queue.empty());
+    assert(waitingForNodes == tcount);
+    if (Aux::Log::isLogLevelEnabled(Aux::Log::LogLevel::debug)) {
+        count totalMoved = std::accumulate(moved.begin(), moved.end(), (count)0);
+        count totalWorked =
+            accumulate(totalNodesPerThread.begin(), totalNodesPerThread.end(), (count)0);
+        (void) totalMoved; //get around unused variable
+        (void) totalWorked;
```
angriman, line 339: You can use `tlx::unused`.

CxVercility:
A "final" note from me:
As previously mentioned, I included a second version of both local moving and refinement in another repository (linked in the first post). 
The "experimental" refinement version as it is denoted in my code is *probably* significantly lower on low thread counts (<4) but *might* scale worse. The experimental is basically a primitive version in which parallelism is implemented by working communities in parallel, but seperately (so each community is worked by one thread only). Since this doesn't require locking, its pretty intuitive that this would perform better with low thread counts. 
I have *not* tested this version as often as the version included in this pull request. I just wanted to mention it in case anyone cares and wants to look into it in the future.

I have only analyzed this very briefly (2 graphs tested) since i ran out of time for my thesis and it wasn't of much importance for the thesis itself.

angriman:
The PR looks good, I pushed some minor stylistic changes myself so that we can include this algorithm in the upcoming release.

angriman (approved):

fabratu (approved):

avdgrinten:
Thanks again for the PR! Glad that this is now merged :)

PR #812, opened 2021-10-04
Fix Googletest Primer link in devGuide docs.

Kropekk:
The old version pointed to a dead location - Googletest Primer has a new address.


angriman (approved):
Thank you, LGTM

angriman:
This fix is trivial, can be merged right away.

PR #813, opened 2021-10-07
Add cibuildwheel script to create wheel-artifacts on GHA

fabratu:
This PR adds the following:

- Github actions based wheel-builder for Linux (x86_64 CPython 3.6-3.9, aarch64 CPython 3.9) based on Manylinux2010 for x86_64 and Manylinux2014 for aarach64, macOS (x86_64 CPython 3.6-3.9) based on toolchain 10.9 and Windows (amd64 CPython 3.9). 
- Upload as an artifact on the CI-pipeline. This makes the files available as download in the current CI-build. Downloads will be available for ~~14~~ 7 days on PR and 1 month on master.
- Pinning Python executable by providing it as a parameter from setuptools to cmake. This is necessary for Manylinux-images to work properly. Otherwise cmake and setuptools work on different Python-environments.

Remarks:
- Thanks to @nomanrz (see #451) for the initial work. I adapted the idea and especially the inside into Manylinux-images was helpful. 
- Manylinux2010 builds can be installed with pip from version 19.0, which is per default only available from Python 3.7 (ships with 20.0 for virtualenv). We still support Python 3.6, but this goes EOL in 2 months - so I didn't saw a point in still building on top of Manylinux1 (which is also EOL).
- I had to drop arm64_darwin builds for now. While AppleClang supports arm64/universal2 builds on GA virtual runners 10.15 and 11 (XCode versions >12.2), the problem remains, that OpenMP and arm-based AppleClang create segmentation faults. Tried several approaches to recreate a cross-compilation with the working miniforge/conda-forge environment based on custom build clang-compilers. However GA images do have issues when working with conda-environments. Couldn't get either the C++ or the Python-toolchain to properly compile the source. Solutions should only be a matter of time though (same for other data science packages like numpy). So this will be added in a future PR.

### DIFF
--- a/.github/workflows/build_wheels.yaml
+++ b/.github/workflows/build_wheels.yaml
@@ -1,0 +1,154 @@
+name: Wheel builder
+
+on: [push, pull_request]
+
+jobs:
+  cpython-linux-x86_64:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        cibw_archs: ["x86_64"]
+        manylinux_image: ["manylinux2010"]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.9'
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.1.1
+        env:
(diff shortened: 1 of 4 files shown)


normanrz:
Thanks for taking over my old PR! Will the wheels also be published through pypi for easy pip installation?

PR #820, opened 2021-10-19
GHA: Fix windows-builds by passing Python-version from setup.py to CMake

fabratu:
This change fixes the currently failing Windows-builds. For GHA-builds (could also apply for other setups), there are several Python-distributions installed by default. `FindPython` from CMake uses the most recent found by default. To override this, we either have to supply the Python-libraries by hand (needed for disabling `FindPython`) or pin the Python version. Since the former can be a challenging task on it's own - I decided to pin the Python version to the interpreter, which calls `setup.py`.


avdgrinten:
Can't Python tell us where its libraries reside? Via that sysconfig module or so?

Issue #104, opened 2018-01-29
Counting number of shortest paths in SSSP algos


avdgrinten:
While looking at the fix of #98, I noticed the following bug: In Dijkstra.cpp we optionally count the number of shortest paths to all nodes. We do this by checking 

```
distances[v] == distances[u] + w
```

this is actually wrong if the weights are not integer (e.g. 0.1 + 0.2 does not compare equal to 0.3 for double-precision floats).

I do not think there is an obvious fix for this problem. Things that we could do are:

- Keep the code. It is simple and works for integers. Document that the code does not work for floating point numbers and/or emit a warning if we encounter this case.

- Perform an epsilon comparison (i.e. consider two paths equal if their length only differs by some epsilon). This will require more complex changes to the case where the distance to a node is updated: We can no longer throw away all predecessors as we have to keep those that are still within epsilon of the new distance. This needs non-constant time. Furthermore, epsilon errors can accumulate along a path.

I am not sure how important path counting is and if there are any users of this feature (especially for floating point weights). Please comment on if you think that a real fix to this problem is worthwhile.

hmeyerhenke:
I have a bad feeling with not supporting floating point weights. At least this problem must be documented properly.

Do you see an elegant way for a dual solution in which the user can choose between the integer and the floating point version? Without much code duplication and only negligible performance impact?

clintg6:
I just tested networkit's implementation against both igraph and networkx using floating point edge weights and got the exact same paths.

PR #832, opened 2021-11-11
Refactor Vector

angriman:
- The inner product function is already implemented in `std`, this PR replaces the custom implementation in `Vector` with `std::inner_product`.
- The `==` operator of `std::vector` compares the content of two vectors, so we can use that instead of manually comparing each element.
- Computing the mean can be parallelized using a parallel reduction (if available).

### DIFF
--- a/networkit/cpp/algebraic/Vector.cpp
+++ b/networkit/cpp/algebraic/Vector.cpp
@@ -51,13 +51,7 @@ double Vector::mean() const {
 }
 
 bool Vector::operator==(const Vector &other) const {
-    if (getDimension() != other.getDimension() || isTransposed() != other.isTransposed()) return false;
-
-    for (index i = 0; i < getDimension(); i++) {
-        if (values[i] != other[i]) return false;
-    }
-
-    return true;
+    return isTransposed() == other.isTransposed() && values == other.values;
 }
 
 bool Vector::operator!=(const Vector &other) const {
--- a/networkit/cpp/algebraic/Vector.cpp
+++ b/networkit/cpp/algebraic/Vector.cpp
@@ -61,12 +61,15 @@ bool Vector::operator!=(const Vector &other) const {
 
 double Vector::innerProduct(const Vector &v1, const Vector &v2) {
     assert(v1.getDimension() == v2.getDimension());
-    double scalar = 0.0;
-    for (index i = 0; i < v1.getDimension(); ++i) {
-        scalar += v1[i] * v2[i];
-    }
-
-    return scalar;
+    double inner_prod = 0.;
+#ifndef NETWORKIT_OMP2
+#pragma omp parallel for reduction(+ : inner_prod)
+    for (omp_index i = 0; i < static_cast<omp_index>(v1.getDimension()); ++i)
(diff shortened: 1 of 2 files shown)


networkit/cpp/algebraic/Vector.cpp
```diff
@@ -37,36 +37,29 @@ double Vector::length() const {
...
 
 bool Vector::operator==(const Vector &other) const {
-    if (getDimension() != other.getDimension() || isTransposed() != other.isTransposed()) return false;
-
-    for (index i = 0; i < getDimension(); i++) {
-        if (values[i] != other[i]) return false;
-    }
-
-    return true;
+    return values == other.values;
```
fabratu, line 54: Hm, it is not 100% the same - but in principle we have here the same problem (also before the proposed change) with comparison of doubles (see #104). 

This would be more expensive, but we could add an epsilon-comparison: https://stackoverflow.com/questions/17333/what-is-the-most-effective-way-for-float-and-double-comparison

networkit/cpp/algebraic/Vector.cpp
```diff
@@ -37,36 +37,29 @@ double Vector::length() const {
...
 
 double Vector::innerProduct(const Vector &v1, const Vector &v2) {
-    assert(v1.getDimension() == v2.getDimension());
-    double scalar = 0.0;
-    for (index i = 0; i < v1.getDimension(); ++i) {
-        scalar += v1[i] * v2[i];
-    }
-
-    return scalar;
+    return std::inner_product(v1.values.begin(), v1.values.end(), v2.values.begin(), 0.0);
```
fabratu, line 62: Is this maybe a candidate for parallelization?

networkit/cpp/algebraic/Vector.cpp
```diff
@@ -37,36 +37,29 @@ double Vector::length() const {
...
+#else
+    sum = std::accumulate(values.begin(), values.end(), 0.);
+#endif
 
-    return sum / (double) this->getDimension();
+    return sum / (double)n;
 }
 
 bool Vector::operator==(const Vector &other) const {
-    if (getDimension() != other.getDimension() || isTransposed() != other.isTransposed()) return false;
```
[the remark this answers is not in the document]
angriman: Actually I should keep the check for `isTransposed()`.

networkit/cpp/algebraic/Vector.cpp
angriman: That's a good idea.

networkit/cpp/algebraic/Vector.cpp
```diff
@@ -37,36 +37,29 @@ double Vector::length() const {
...
-    return true;
+    return values == other.values;
 }
 
 bool Vector::operator!=(const Vector &other) const {
     return !(*this == other);
 }
 
 double Vector::innerProduct(const Vector &v1, const Vector &v2) {
-    assert(v1.getDimension() == v2.getDimension());
```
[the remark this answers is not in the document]
angriman: I restored it.

fabratu (approved):
LGTM.

PR #833, opened 2021-11-12
Refactor GroupCloseness

angriman:
- Replace copy of `Graph` with `const Graph *`
- Remove unnecessary references to `G` since it is already a member variable
- Use algorithms implemented in `Traversal::` in `scoreOfGroup` instead of re-implementing them
- Bugfix: the algorithm discards low-degree nodes, which leads to incorrect results in small networks.
- Further performance improvements (mainly removing unnecessary variables)


fabratu (approved):
LGTM.

PR #835, opened 2021-11-14
spgemm

valentin-noske:


avdgrinten:
This PR has lots of files that should not be commit to the repo (IDE configuration?).

Regarding the implementation: using std::unordered_map is likely faster than std::map for the SPA. The implementation does not use any parallelism; it is probably a good idea to add that (the most simple way to do that would be having a symbolic phase that computes the row pointers before the numeric phase).

PR #836, opened 2021-11-15
Refactor CSRMatrix

angriman:
As the last step, the constructor of `CSRMatrix` restores the row indices to their correct values. However, this is equivalent to rotating the vector to the right by one and resetting its first value to zero, which can be implemented with just two lines of code instead of writing another loop.


fabratu (approved):
LGTM.

Issue #830, opened 2021-11-10
Inconsistency between edge ID ordering and iterEdges()


mcogoni:
Hi,
I was comparing the output of NetworkX and Networkit for EBC and I wasn't able to find anywhere in the docs the order in which edgeScores is written. After looking into the sources I found out that I should use the function edgeId to go from (origin, destination) to the edgeId. A much better solution would be to follow the same order produced by iterEdges() that sorts by origin node instead of destination node as followed by edgeId. A nice addition would be a ranking method for the EBC and not only for the BC, that way there would be no mistakes...

Maybe this issue is already clarified in the docs, but I really tried without success!

Thanks!
marco

avdgrinten:
This issue has been reported before.

One possibility is guaranteeing that the edge ID ordering matches the iterEdges() ordering, but that is not done by the current implementation.

mcogoni:
For a start, I'd mention the necessity to call edgeID in the documentation...

Strato75:
Hi @avdgrinten, maybe making the ```edgeScores()``` method returns a dictionary where the key is a tuple (origin node, destination node) and the value the EBC could be a solution, what about it?
Also adding a method to the ```nxadapter``` class to get the mapping between networkx and networkit nodes would be cool!

PR #837, opened 2021-11-15
Update IterEdges() documentation

Relux-the-Relux:
Makes clear that it does not follow the order of edges Ids addressing #830 .

### DIFF
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -129,7 +129,8 @@ cdef class Betweenness(Centrality):
 
 
 	def edgeScores(self):
-		""" Get a vector containing the betweenness score for each edge in the graph.
+		""" Get a vector containing the betweenness score for each edge in the graph
+		in ascending edge ID order.
 
 		Returns:
 		--------
@@ -2079,7 +2080,8 @@ cdef class SpanningEdgeCentrality(Algorithm):
 		return (<_SpanningEdgeCentrality*>(self._this)).runParallelApproximation()
 
 	def scores(self):
-		""" Get a vector containing the SEC score for each edge in the graph.
+		""" Get a vector containing the SEC score for each edge in the graph
+		in ascending edge ID order.
 
 		Returns:
 		--------
@@ -2176,7 +2178,8 @@ cdef class ApproxSpanningEdge(Algorithm):
 
 	def scores(self):
 		"""
-		Return the spanning edge approximation for each edge of the graph.
+		Return the spanning edge approximation for each edge of the graph
+		in ascending edge ID order.
 
 
 		Returns:
--- a/networkit/graph.pyx
+++ b/networkit/graph.pyx
@@ -571,6 +571,12 @@ cdef class Graph:
 	def iterEdges(self):
 		"""
 		Iterates over the edges of the graph.
+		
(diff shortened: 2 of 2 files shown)


networkit/graph.pyx
```diff
@@ -571,6 +571,11 @@ cdef class Graph:
 	def iterEdges(self):
 		"""
 		Iterates over the edges of the graph.
+		
+		The iteration is done from node to node, iterating over all edges
+		of a node before passing to edges of the next node in the order of node Ids.
```
angriman, line 576: Maybe rephrase with "For each node u in the graph in ascending node id order, the iterator yields the edges incident to u".

avdgrinten:
In addition to this change, we should update the docs of the edge centrality algos to state that their result is indexed by edge ID.

avdgrinten:
This is also underspecified for undirected graphs. Which direction of an edge is returned?

angriman:
Yes, we should specify that for undirected graphs we return edges `(u, v)` where `u < v` (to avoid iterating twice over the same edge) while for directed graphs we return out-edges.

networkit/graph.pyx
```diff
@@ -571,6 +571,12 @@ cdef class Graph:
 	def iterEdges(self):
 		"""
 		Iterates over the edges of the graph.
+		
+		For each node u in the graph in ascending node id order,
+		the iterator yields the out-edges of u in directed Graphs
+		and the edges (u,v) in which u < v for undirected Graphs.
```
angriman, line 577: Graphs -> graphs

angriman (approved):

networkit/centrality.pyx
```diff
@@ -129,7 +129,8 @@ cdef class Betweenness(Centrality):
 
 
 	def edgeScores(self):
-		""" Get a vector containing the betweenness score for each edge in the graph.
+		""" Get a vector containing the betweenness score for each edge in the graph
+		in ascending edge id order.
```
avdgrinten, line 133: `edge ID`, not `id`.

avdgrinten: Maybe also add `(as returned by Graph.edgeId())`.

networkit/graph.pyx
```diff
@@ -571,6 +571,12 @@ cdef class Graph:
 	def iterEdges(self):
 		"""
 		Iterates over the edges of the graph.
+		
+		For each node u in the graph in ascending node id order,
```
avdgrinten, line 575: `id` -> `ID`.

networkit/graph.pyx
```diff
@@ -571,6 +571,12 @@ cdef class Graph:
...
+		the iterator yields the out-edges of u in directed graphs
+		and the edges (u,v) in which u < v for undirected graphs.
+		
+		It does NOT follow the order of edgeIds (if present).
```
avdgrinten, line 579: Do not capitalize `NOT`. `edgeIds` -> `edge IDs`.

avdgrinten (changes requested):

networkit/centrality.pyx
angriman: @Relux-the-Relux this comment is yet to be addressed.

avdgrinten (approved):
LGTM.

PR #839, opened 2021-11-16
Debug_only Macro

CxVercility:
As described in #811 this can be used to avoid bloating code with ifndef and endifs to maintain readability.

Where to put it, and how to name it is up for discussion.

### DIFF
--- a/include/networkit/Globals.hpp
+++ b/include/networkit/Globals.hpp
@@ -23,6 +23,14 @@ using omp_index = int64_t;
 using omp_index = index;
 #endif // _MSC_VER
 
+#ifdef NDEBUG
+#define DEBUG_ONLY(x)
+#else
+#define DEBUG_ONLY(x) = x;
+#endif
+
+
+
 using count = uint64_t;    ///< more expressive name for an integer quantity
 using node = index;        ///< node indices are 0-based
 using edgeweight = double; ///< edge weight type


include/networkit/Globals.hpp
```diff
@@ -23,6 +23,14 @@ using omp_index = int64_t;
 using omp_index = index;
 #endif // _MSC_VER
 
+#ifdef NDEBUG
+#define DEBUG_ONLY(x)
+#else
+#define DEBUG_ONLY(x) = x;
+#endif
```
avdgrinten, line 30: This should be called `NETWORKIT_DEBUG_ONLY` to keep the macro namespace clean. The `!NDEBUG` case also seems to be broken (what is the `=` for?).

avdgrinten:
I am not convinced that this is a good idea. It is probably better to avoid diverging code between `NDEBUG` and `!NDEBUG`. Why isn't `assert` sufficient? What do we need to disable at compile time? Isn't a simple `if` or `#ifdef` enough to disable computations that only happen in debugging mode? Usually, one can just do the computations both in debugging and in release mode and let the compiler optimize the debugging code away in release mode.

CxVercility:
each #ifdef and #endif incurs 2 whole additional lines which makes the code pretty unreadable in my case because multiple debug only lines are close to each other

This is just an idea up for discussion.

include/networkit/Globals.hpp
CxVercility: I typed this from memory since it was only a suggestion for now.

avdgrinten:
Then let's just keep the computation in the release code unless the performance impact is really measurable. This might require adding `(void)x;` for variables `x` that are only read by assertions (to get around unused variable warning), but this is more pretty than a `NETWORKIT_DEBUG_ONLY` macro.

For the rare cases where it is measurable, we can use `#ifdef` (or use a plain `if` on a `constexpr` variable that is true/false depending on `NDEBUG`; C++17 will make this a bit smoother with `if constexpr`).

CC @angriman

PR #840, opened 2021-11-17
Add topology sort for both C++ and Python interface

fabratu:
This PR adds a sequential Topology Sort algorithm as requested/discussed here: https://github.com/networkit/networkit/discussions/829

### DIFF
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -421,6 +421,18 @@ Graph restoreGraph(const std::vector<node> &invertedIdMap, const Graph &G);
 void sortEdgesByWeight(Graph &G, bool decreasing = false);
 
 /**
+ * Given a directed graph G, the topology sort algorithm creates one valid topology order of nodes.
+ * Undirected graphs are not accepted as input, since a topology sort is a linear ordering of
+ * vertices such that for every edge u -> v, node u comes before v in the ordering.
+ *
+ * This is a helper function. Instead of calling it via GraphTools, it is also possible to create a
+ * TopologicalSort-object from the base-class.
+ * @param   G           Directed input graph
+ * @return              A vector of node-ids sorted according to their topology.
+ */
+std::vector<node> topologicalSort(const Graph &G);
+
+/**
  * Rename nodes in a graph using a callback which translates each old id to a new one.
  * For each node u in input graph, oldIdToNew(u) < numNodes.
  *
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -5,6 +5,7 @@
 #include <networkit/auxiliary/Random.hpp>
 #include <networkit/graph/Graph.hpp>
 #include <networkit/graph/GraphTools.hpp>
+#include <networkit/graph/TopologicalSort.hpp>
 
 namespace NetworKit {
 
@@ -502,6 +503,13 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
                 return e1.v < e2.v;
             return e1.weight < e2.weight;
         });
+}
+
+std::vector<node> topologicalSort(const Graph &G) {
+    TopologicalSort topSort(G);
+    topSort.run();
+
+    return topSort.getResult();
 }
 
 } // namespace GraphTools
--- a/networkit/cpp/graph/TopologicalSort.cpp
+++ b/networkit/cpp/graph/TopologicalSort.cpp
@@ -1,0 +1,66 @@
+/*
+ * TopologicalSort.cpp
+ *
+ *  Created on: 22.11.2021
+ *      Author: Fabian Brandt-Tumescheit
+ */
+#include <networkit/graph/TopologicalSort.hpp>
+
+namespace NetworKit {
+
+TopologicalSort::TopologicalSort(const Graph &G) : G(&G) {
+    if (!G.isDirected())
+        throw std::runtime_error("Topological sort is defined for directed graphs only.");
+}
+
+void TopologicalSort::run() {
+    reset();
+
+    std::stack<node> nodeStack;
+
+    G->forNodes([&](node u) {
+        if (topSortMark[u] == NodeMark::PERM)
+            return;
+
+        nodeStack.push(u);
+        do {
+            node v = nodeStack.top();
+            if (topSortMark[v] != NodeMark::NONE) {
+                nodeStack.pop();
+                if (topSortMark[v] == NodeMark::TEMP) {
+                    topSortMark[v] = NodeMark::PERM;
+                    topology[current] = v;
+                    current--;
+                }
+            } else {
+                topSortMark[v] = NodeMark::TEMP;
+                G->forNeighborsOf(v, [&](node w) {
+                    if (topSortMark[w] == NodeMark::NONE)
+                        nodeStack.push(w);
+                    else if (topSortMark[w] == NodeMark::TEMP)
+                        throw std::runtime_error("Error: the input graph has cycles.");
+                });
+            }
+        } while (!nodeStack.empty());
+    });
+
+    hasRun = true;
+}
+
+void TopologicalSort::reset() {
+    // Reset node marks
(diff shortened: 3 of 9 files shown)


networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1023,4 +1023,36 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+    G.addEdge(0, 1);
+    G.addEdge(0, 2);
+    G.addEdge(2, 1);
+    G.addEdge(1, 3);
+    G.addEdge(4, 2);
+
+    if (directed()) {
+        std::vector<node> res = GraphTools::topologySort(G);
+        std::vector<node> test = {4, 0, 2, 1, 3};
+        EXPECT_TRUE(res == test);
```
opatut, line 1048: This test looks rather unstable, as there are multiple possible results, and the actual returned order depends on the implementation, the actual choice from the possible results is undefined behaviour. 

The test could instead iterate through all edges *(u, v)* and assert that u's index in the result is smaller than v's index.

opatut (commented):
Thanks for implementing this so quickly. Not sure what the requirements are for tests, but I see a case here for a little improvement for future test stability and correctness.

Otherwise, awesome work ;) (Haven't tested it yet though, no idea how to compile it yet).

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -504,5 +504,45 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
...
+    count current = G.numberOfNodes() - 1;
+    std::function<void(node)> exploreNode;
+
+    exploreNode = [&](node u) {
+        if (topSortMark[u] == nodeMark::PERM) {
+            return;
+        }
+        if (topSortMark[u] == nodeMark::TEMP) {
+            throw std::runtime_error(
+                "Error: the input graph has cycles. Therefore no valid topology can be computed.");
```
angriman, line 525:
```suggestion
                "Error: the input graph has cycles.");
```

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -504,5 +504,45 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
...
+        }
+
+        topSortMark[u] = nodeMark::TEMP;
+
+        for (auto v : G.neighborRange(u)) {
+            exploreNode(v);
+        }
```
angriman, line 532:
```suggestion
        G.forNeighbors(u, exploreNode);
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -395,7 +395,7 @@ std::unordered_map<node, node> getRandomContinuousNodeIds(const Graph &graph);
  * @param 	nodeIdMap	The mapping from node ids with gaps to continuous node ids (i.e. from
  * @getContinuousNodeIds)
  * @param 	G 			The compacted graph (currently only needed for the upper node id bound)
- * @return 				A vector of nodes id where the index is the node id of the compacted graph
+ * @return 				A vector of node-ids where the index is the node id of the compacted graph
```
angriman, line 398: This change is not related to the PR.

include/networkit/graph/GraphTools.hpp
```diff
@@ -420,6 +420,15 @@ Graph restoreGraph(const std::vector<node> &invertedIdMap, const Graph &G);
...
 void sortEdgesByWeight(Graph &G, bool decreasing = false);
 
+/**
+ * Given a directed graph G, the topology sort algorithm creates one valid topology order of nodes.
+ * Undirected graphs are not accepted as input, since a topology sort is a linear ordering of
+ * vertices such that for every edge u -> v, node u comes before v in the ordering.
+ * @param   G           Directed input graph
+ * @return              A vector of node-ids sorted according to their topology.
+ */
+std::vector<node> topologySort(const Graph &G);
```
angriman, line 430:
```suggestion
std::vector<node> topologicalSort(const Graph &G);
```

networkit/graphtools.pyx
```diff
@@ -43,6 +43,7 @@ cdef extern from "<networkit/graph/GraphTools.hpp>" namespace "NetworKit::GraphT
 	unordered_map[node,node] getContinuousNodeIds(_Graph G) nogil except +
 	unordered_map[node,node] getRandomContinuousNodeIds(_Graph G) nogil except +
 	void sortEdgesByWeight(_Graph G, bool_t) nogil except +
+	vector[node] topologySort(_Graph G) nogil except +
```
angriman, line 46: Please add a simple Python test.

avdgrinten:
In general, I would prefer to have this as an `Algorithm` subclass. This has the advantage that the state (= the `std::vector`s etc.) can be put into the class and re-used if multiple runs of the algorithm are required.

We can still have a helper function but it should internally just construct an object of the subclass and call `run()`.

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -504,5 +504,45 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
         });
 }
 
+std::vector<node> topologySort(const Graph &G) {
+    if (!G.isDirected()) {
+        throw std::runtime_error(
+            "Sorting according to topology is only defined for directed graphs.");
```
angriman, line 510:
```suggestion
            "Topological sort is defined for directed graphs only.");
```

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -504,5 +504,43 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
...
+            "Topological sort is defined for directed graphs only.");
+    }
+
+    enum nodeMark { NONE, TEMP, PERM };
+    std::vector<nodeMark> topSortMark(G.numberOfNodes(), nodeMark::NONE);
+    std::vector<node> result(G.numberOfNodes(), 0);
+    count current = G.numberOfNodes() - 1;
+    std::function<void(node)> exploreNode;
+
+    exploreNode = [&](node u) {
```
avdgrinten, line 519: This still uses `std::function` (which has a high overhead) even though that is not necessary:
```suggestion
    auto exploreNode = [&](node u) {
```

avdgrinten: Ah wait, you're doing this to enable the recursion. It's probably better to use a normal function in this case (and not a lambda).

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void TopologicalSort::run() {
+    if (hasRun)
+        reset();
+
+    std::stack<node> nodeStack;
+
+    G->forNodes([&](node u) {
+        if (topSortMark[u] != NodeMark::PERM) {
```
angriman, line 28: Just a minor thing here: to avoid indenting the whole function, I would write something like
```
if (topSortMark[u] == NodeMark::PERM)
    return;
// Rest of the code
```

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -0,0 +1,70 @@
...
+            nodeStack.push(u);
+            do {
+                node v = nodeStack.top();
+                if (topSortMark[v] != NodeMark::NONE) {
+                    nodeStack.pop();
+                    if (topSortMark[v] == NodeMark::TEMP) {
+                        topSortMark[v] = NodeMark::PERM;
+                        topology[current] = v;
+                        current--;
+                        assert(current >= 0);
```
angriman, line 38: This assertion is not needed, `current` is an unsigned integer.

networkit/cpp/graph/test/TopologicalSortGTest.cpp
```diff
@@ -0,0 +1,76 @@
+/*
+ * TopologicalSort.hpp
```
angriman, line 2: The file name is different

angriman (approved):

PR #841, opened 2021-11-17
Fixed author in citation of `ApproxCloseness`

danielw2904:
Here the link to arXive https://arxiv.org/abs/1409.0035


fabratu (approved):
LGTM - good catch.

PR #842, opened 2021-11-20
Remove unused CSRMatrix source file

angriman:
When we merged the templated CSRMatrix implementation we should have removed `CSRMatrix.cpp`. We probably forgot about that (this file is not even compiled).


avdgrinten (approved):
LGTM.

PR #844, opened 2021-11-20
Refactor CSRMatrix sort() and diagonal()

angriman:
Our CSRMatrix class uses simple quicksort and binary search implementations. This PR:
- In `sort()`: uses `std::sort` to compute a permutation of column indices and uses `ArrayTools::applyPermutation` to sort the values and column indices
- Simplifies the checks at the end of `sort()` with `std::is_sorted` and enables them only if NetworKit sanity checks are enabled, as originally intended
- Adds a simple test for `sort()`
- Uses `std::lower_bound` to search for diagonal elements in `diagonal()`

### DIFF
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -429,25 +429,12 @@ public:
         if (sorted()) {
 #pragma omp parallel for
             for (omp_index i = 0; i < static_cast<omp_index>(diag.getDimension()); ++i) {
-                if (rowIdx[i] == rowIdx[i + 1])
-                    continue; // no entry in row i
-                index left = rowIdx[i];
-                index right = rowIdx[i + 1] - 1;
-                index mid = (left + right) / 2;
-                while (left <= right) {
-                    if (columnIdx[mid] == i) {
-                        diag[i] = nonZeros[mid];
-                        break;
-                    }
 
-                    if (columnIdx[mid] < i) {
-                        left = mid + 1;
-                    } else {
-                        right = mid - 1;
-                    }
+                const auto it = std::lower_bound(columnIdx.begin() + rowIdx[i],
+                                                 columnIdx.begin() + rowIdx[i + 1], i);
 
-                    mid = (left + right) / 2;
-                }
+                if (it != columnIdx.end() && *it == static_cast<index>(i))
+                    diag[i] = nonZeros[it - columnIdx.begin()];
             }
         } else {
 #pragma omp parallel for
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -379,23 +379,14 @@ public:
                                                   nonZeros.begin() + endOfRow, permutation.begin());
             }
         }
-
-#ifndef NETWORKIT_SANITY_CHECKS
-        bool sorted = true;
-        tlx::unused(sorted);
-#pragma omp parallel for
-        for (omp_index i = 0; i < static_cast<omp_index>(nRows); ++i) {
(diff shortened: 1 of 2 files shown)


include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -395,25 +354,42 @@ class CSRGeneralMatrix {
...
-                    sorted = false;
-                    break;
-                }
-            }
-        }
+        for (omp_index i = 0; i < static_cast<omp_index>(nRows); ++i)
+            if (sorted
+                && !std::is_sorted(columnIdx.begin() + rowIdx[i],
+                                   columnIdx.begin() + rowIdx[i + 1]))
+                sorted = false;
```
avdgrinten, line 392: I would simply `if(sorted) assert(std::is_sorted(...))` here instead of modifying the `sorted` member.

angriman:
Another idea to improve `sort()` would be to apply counting-sort instead of `std::sort()`. If the row is "dense" enough (it has many non-zero elements), counting-sort should be faster. However, we should run some experiments to understand when one algorithm is more convenient than the other.

include/networkit/algebraic/CSRGeneralMatrix.hpp
angriman: Agree, I addressed this.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -395,29 +354,40 @@ class CSRGeneralMatrix {
...
+                const index startOfRow = rowIdx[i], endOfRow = rowIdx[i + 1];
+                const count nonZerosInRow = endOfRow - startOfRow;
+                if (nonZerosInRow <= 1
+                    || std::is_sorted(columnIdx.begin() + startOfRow, columnIdx.begin() + endOfRow))
+                    continue;
+
+                permutation.resize(nonZerosInRow);
+                std::iota(permutation.begin(), permutation.end(), index{0});
+                std::sort(permutation.begin(), permutation.end(),
+                          [&columnIdx = columnIdx, startOfRow = startOfRow](index x, index y) {
```
avdgrinten, line 371: Just capture by `[&]` here (more specific captures are largely useless).

avdgrinten (commented):
One more minor thing, afterwards, this can be merged.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -395,29 +354,39 @@ class CSRGeneralMatrix {
...
+#pragma omp parallel
+        {
+            std::vector<index> permutation(nCols);
+#pragma omp for schedule(guided)
+            for (omp_index i = 0; i < static_cast<omp_index>(nRows); ++i) {
+                const index startOfRow = rowIdx[i], endOfRow = rowIdx[i + 1];
+                const count nonZerosInRow = endOfRow - startOfRow;
+                if (nonZerosInRow <= 1
+                    || std::is_sorted(columnIdx.begin() + startOfRow, columnIdx.begin() + endOfRow))
+                    continue;
```
avdgrinten, line 366: Why do we check whether the row is sorted first? Did you benchmark this? I would expect the `std::sort` implementation to work well on already sorted data.

angriman: Apart from `std::sort`, the idea is to avoid the two useless `applyPermutation` calls (under the assumption that `std::is_sorted` is faster than two calls `applyPermutation` calls). However, I do not have benchmark data to confirm that.

avdgrinten: Okay, makes sense.

avdgrinten (approved):
LGTM.

PR #848, opened 2021-11-24
Bug fix: group closeness local search

angriman:
The local-search approximation algorithm should evaluate all swaps. However, the current implementation excludes leaves of the DAG, which leads to potentially incorrect results and random failures of `CentralityGTest.testGroupClosenessLocalSearch`.


avdgrinten:
Does this affect grow-shrink etc? What is the DAG here?

angriman:
These changes do not affect GS, but the error probably affects the non-local GS. I have to look at GS more carefully.

angriman:
Non-extended GS should consider leaves of the DAG as candidates for swaps as well, so I also applied the fix to non-extended GS.

avdgrinten (approved):
LGTM.

PR #852, opened 2021-12-07
GHA: Update wheelbuilder to support CPython 3.10. Remove EOL CPython 3.6 and 3.7.

fabratu:


.github/workflows/build_wheels.yml
```diff
@@ -98,92 +94,157 @@ jobs:
...
-          name: networkit-cp39-win-amd64.whl
-          path: ./wheelhouse/*cp39*.whl
-          retention-days: ${{ env.ARTIFACT_RETENTION }}
\ No newline at end of file
+          path: ./dist/*networkit*.tar.gz
+          retention-days: ${{ env.ARTIFACT_RETENTION }}
+
+  # This job checks whether the current workflow is triggered by a new tag with the following form: INT.INT or INT.INT.INT (for example: 1.0, 1.0.1, ...)
+  # The result is saved in a variable and used as a conditional variable when uploading packages.
+  check_tag:
```
avdgrinten, line 216: This job lacks a `name:`.

avdgrinten (changes requested):

.github/workflows/build_wheels.yml
```diff
@@ -98,92 +94,158 @@ jobs:
...
+        id: check-tag
+        run: |
+          if [[ ${{ github.ref }} =~ ^refs\/tags\/release\/[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
+            echo "::set-output name=run_jobs::true"
+          else
+            echo "::set-output name=run_jobs::false"
+          fi
+
+  upload-packages:
+    if: needs.check_tag.outputs.run_other_jobs == 'true' && github.repository == 'networkit/networkit'
```
avdgrinten, line 232: `run_other_jobs` does not match the output of `echo` (which is `run_jobs`), so that won't work here. Maybe something like `is_release` is more appropriate anyway.

avdgrinten (approved):
LGTM now.

PR #860, opened 2021-12-21
Implementation of the PGD {3,4}-graphlets counting algorithm

RobinPetit:
As proposed in the discussion #850, here is an implementation of the PGD algorithm for graphlet counting.

There are still 3 things that are unclear to me:
1. As I am no expert in openmp, should I use `#pragma omp <...>` in a lambda function given to `Graph::parallelForEdges` (e.g. [here](https://github.com/networkit/networkit/commit/8f61673cef8c6a058aa9a0070fd3657386a29ee1#diff-490cf76d323b89e05914d6f317f5f194bdab64c703cde7fb2ed9132edda7a5bbR120))?
2. `GraphletsCounter::getGraphletsCounts` returns a `const std::vector<count>&` so far because I did not see some sort of a `np`-compatible C++/Cython/Python class that would allow no data copy in the Cython interface. Is this the intended networkit way or did I just miss it?
3. The PGD algorithm gives the 3-graphlets for free when counting the 4-graphlets so the code "appends" them at the start of the vector. Is this a good idea though?

I will write a notebook after point 2 is resolved.


angriman:
Hi, thank you for the PR! About your questions:

1. Yes that is fine
2. Same, we do that as well in other parts of the codebase
3. I guess that's fine as long as it is documented

I see that some builds are still failing. Are you working on that? We will provide a full review once all builds pass.

RobinPetit:
> I see that some builds are still failing. Are you working on that?

Sure. I don't have much time right now but I'll soon fix what's left to fix.

PR #863, opened 2022-01-05
Move to C++17

angriman:
This PR raises the NetworKit C++ standard to 17. The oldest clang version that supports C++17 is 5.0 and it was released on 7.9.17, so I guess that this PR should not be merged before 7.9.22.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ on: [push, pull_request]
 # Default parameters for all builds.
 env:
   OMP_NUM_THREADS: 2
-  CXX_STANDARD: 14
+  CXX_STANDARD: 17
   MONOLITH: ON
   TLX_PATH: /home/runner/work/networkit/networkit/tlx
   TLX_PATH_WIN: /d/a/networkit/networkit/tlx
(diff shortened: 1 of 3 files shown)


avdgrinten:
What about GCC?

angriman:
On [https://gcc.gnu.org/projects/cxx-status.html#cxx17](https://gcc.gnu.org/projects/cxx-status.html#cxx17) it's written that "C++17 features are available since GCC 5". However, it seems to me that the oldest GCC version that fully supports C++17 is GCC 7 (released on May 2, 2017).

avdgrinten:
This PR has conflicts.

avdgrinten:
Since this passes CI, does that mean that all our supported compilers also support this flag now? (@fabratu)

fabratu:
Technically no. Clang 4 is missing two C++17 features (https://clang.llvm.org/cxx_status.html#cxx17), GCC 6 some more (https://gcc.gnu.org/projects/cxx-status.html#cxx17). Full support would be in Clang 8 and GCC 8. 

We can switch to GCC 7 (released 02. May 2017), where only one feature is missing. As @angriman initally suggested, Clang 5 can be used from 07. September 2022 on, which would also support most of the features.

I am not sure of the impact of the missing features are in practice though.

avdgrinten:
It's not necessary that all features are supported (we can simply avoid the unsupported ones), but the flag should be supported. Clang 4 doesn't seem to be in our CI matrix anyway, so I guess it's not a supported compiler?

fabratu:
Clang 4 at least doesn't support the flag - so in this sense it is not supported. According to their doc, for clang 4 one should use `-std=c++1z` for C++17.

avdgrinten (approved):
LGTM.

PR #864, opened 2022-01-11
Remove deprecated code

angriman:
In NetworKit 8.1 we deprecated `Algorithm::toString()` and `Algorithm::isParallel()`. In the next release cycle, we can remove these functions.


avdgrinten (approved):
LGTM.

PR #855, opened 2021-12-13
Documentation: add missing parameters in community-module, add missing csbridge-module

fabratu:
- Addresses  #838. By adding the function header as the first line of the doc, this enables rendering of missing parameters. This PR only fixes this for the `community`-module. For the PRs to become not too big, other modules will be addressed in other PRs.
- Fix several documentation strings, which are not compliant with sphinx-automodule (`community`-module)
- Add missing function documentation (`community`-module)
- Add missing function documentation in C++ counterparts
- Add missing `csbridge`-module to Python-documentation

PR #865, opened 2022-01-12
Documentation: Fix input/output-parameter description for Python community-module

fabratu:
This is a follow-up to #855. While the parameters are shown by no in the documentation, the description for input and output-parameters is not shown in many cases. This is due to having ':' or too long scores for 'Returns' or 'Parameters' headings. Also fixed some minor inconsistencies in parameter-types and tags.

I did overlook this in the first PR -.- Unfortunately the documentation for `autodoc/automodule` doesn't really say something about how these headings are allowed to look like. Good news: documentation now looks as intended. Note that now most parameter-types can also be clicked.

Example from the modified documentation:
Bildschirmfoto 2022-01-12 um 13 07 43

Once this is approved/merged I would do the same treatment for the other modules.

### DIFF
--- a/networkit/community.pyx
+++ b/networkit/community.pyx
@@ -78,8 +78,8 @@ cdef class CommunityDetector(Algorithm):
 		
 		Returns a partition of the clustering.
 
-		Returns:
-		--------
+		Returns
+		-------
 		networkit.Partition:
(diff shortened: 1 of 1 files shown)


avdgrinten (approved):
LGTM.

PR #869, opened 2022-01-27
Refactor: use STL algorithms

angriman:
I spotted several places of our codebase that can be simplified by using appropriate STL algorithms and range-based `for` loops. This PR fixes a few of them, in particular:
- uses `std::max/min_element` and `std::accumulate` where applicable
- uses `std::min/max` where applicable
- changes index-based for-loops with range-based for-loops
- fixes initialization bugs (i.e., `index i = -1` yields +inf, not -1)

### DIFF
--- a/include/networkit/centrality/KadabraBetweenness.hpp
+++ b/include/networkit/centrality/KadabraBetweenness.hpp
@@ -175,11 +175,7 @@ class KadabraBetweenness : public Algorithm {
 
     count maxAllocatedFrames() const {
         assureFinished();
-        count maxNFrames = 0;
-        for (auto nFrames : maxFrames) {
-            maxNFrames = std::max(nFrames, maxNFrames);
-        }
-        return maxNFrames;
+        return *std::max_element(maxFrames.begin(), maxFrames.end());
     }
 
   protected:
--- a/networkit/cpp/distance/Diameter.cpp
+++ b/networkit/cpp/distance/Diameter.cpp
@@ -268,11 +268,7 @@ edgeweight Diameter::estimatedVertexDiameter(const Graph& G, count samples) {
         edgeweight vd = estimateFrom(u);
         DEBUG("sampled vertex diameter from node ", u, ": ", vd);
         #pragma omp critical
-        {
-            if (vd > vdMax) {
-                vdMax = vd;
-            }
-        }
+        vdMax = std::max(vdMax, vd);
     }
 
     return vdMax;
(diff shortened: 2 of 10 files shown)


networkit/cpp/distance/Diameter.cpp
```diff
@@ -304,22 +298,21 @@ edgeweight Diameter::estimatedVertexDiameterPedantic(const Graph& G) {
...
-                largest_comp_size = it->second;
-            }
-        }
-        INFO("Largest component size: ", largest_comp_size);
-        vd = largest_comp_size;
+        const auto compSizes = cc.getComponentSizes();
+        vd = std::max_element(
+                 compSizes.begin(), compSizes.end(),
+                 [](const auto &c1, const auto &c2) -> bool { return c1.second < c2.second; })
+                 ->second;
```
avdgrinten, line 313: This is UB if compSizes is empty.

include/networkit/centrality/KadabraBetweenness.hpp
```diff
@@ -175,11 +175,7 @@ class KadabraBetweenness : public Algorithm {
 
     count maxAllocatedFrames() const {
         assureFinished();
-        count maxNFrames = 0;
-        for (auto nFrames : maxFrames) {
-            maxNFrames = std::max(nFrames, maxNFrames);
-        }
-        return maxNFrames;
+        return *std::max_element(maxFrames.begin(), maxFrames.end());
```
avdgrinten, line 178: This faults if maxFrames is empty.

networkit/cpp/distance/Diameter.cpp
angriman: If the graph is not empty (there is a check before) then `compSizes` cannot be empty right?

include/networkit/centrality/KadabraBetweenness.hpp
angriman: `maxFrames` is resized with `omp_get_max_threads()` before each execution, it should never be empty right?

networkit/cpp/distance/Diameter.cpp
avdgrinten: Yes, makes sense.

avdgrinten (approved):
LGTM.

include/networkit/centrality/KadabraBetweenness.hpp
avdgrinten: Where is that done? In the constructor?

angriman: It's done in `init()` (always called by `run()`)
https://github.com/networkit/networkit/blob/eec604c774d0bf98ce6c5829f3eb73ac0358d16e/networkit/cpp/centrality/KadabraBetweenness.cpp#L280

Here we are guaranteed that `run()` was called by `assureFinished()`.

PR #870, opened 2022-01-27
 Documentation: Fix missing/non-visible doc for Python modules beginning with 'c'

fabratu:
This PR is a follow-up and fixes missing/non-visible doc for modules: clique, coarsening, coloring, components, correlation, csbridge (remaining modules beginning with letter 'c').

### DIFF
--- a/networkit/coarsening.pyx
+++ b/networkit/coarsening.pyx
@@ -22,6 +22,8 @@ cdef extern from "<networkit/coarsening/GraphCoarsening.hpp>":
 		map[node, vector[node]] getCoarseToFineNodeMapping() except +
 
 cdef class GraphCoarsening(Algorithm):
+	""" Abstract base class for coarsening measures"""
+
 	cdef Graph _G
 
 	def __init__(self, *args, **namedargs):
@@ -29,12 +31,42 @@ cdef class GraphCoarsening(Algorithm):
 			raise RuntimeError("Error, you may not use GraphCoarsening directly, use a sub-class instead")
 
 	def getCoarseGraph(self):
+		"""
+		getCoarseGraph()
+
+		Returns the coarse graph
+
+		Returns
+		-------
+		networkit.Graph
+			Coarse graph
+		"""
 		return Graph(0).setThis((<_GraphCoarsening*>(self._this)).getCoarseGraph())
 
 	def getFineToCoarseNodeMapping(self):
+		"""
+		getCoarseGraph()
+
+		Returns the coarse graph
+
+		Returns
+		-------
+		list
+			List containing mapping, whereas index represents node u from fine graph and value represents node v from coarse graph  
+		"""
 		return (<_GraphCoarsening*>(self._this)).getFineToCoarseNodeMapping()
 
 	def getCoarseToFineNodeMapping(self):
+		"""
+		getCoarseGraph()
+
+		Returns the coarse graph
+
+		Returns
+		-------
+		map[node, list]
+			Map containing node from coarse graph and all its mappings from finer graph
+		"""
 		return (<_GraphCoarsening*>(self._this)).getCoarseToFineNodeMapping()
 
 
@@ -45,6 +77,20 @@ cdef extern from "<networkit/coarsening/ParallelPartitionCoarsening.hpp>":
 
 
 cdef class ParallelPartitionCoarsening(GraphCoarsening):
+	"""
+	ParallelPartitionCoarsening(G, zeta, parallel = True)
+	
+	Coarsens graph according to a partition.
+ 	
+ 	Parameters
+ 	----------
+ 	G : networkit.Graph
+		The graph.
+	zeta : networkit.Partition
+		The partition, which is used for coarsening.
+ 	parallel : bool, optional
+		If true, algorithm runs in parallel
+	"""
 	def __cinit__(self, Graph G not None, Partition zeta not None, parallel = True):
 		self._this = new _ParallelPartitionCoarsening(G._this, zeta._this, parallel)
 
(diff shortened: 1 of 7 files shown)


networkit/coarsening.pyx
```diff
@@ -22,19 +22,51 @@ cdef extern from "<networkit/coarsening/GraphCoarsening.hpp>":
...
 
 	def getCoarseGraph(self):
+		"""
+		getCoarseGraph()
+
+		Returns the coarse graph
+
+		Returns
+		-------
+		G
```
angriman, line 41: Should you specify the return type here? In this case, `networkit.Graph`.

networkit/coloring.py
```diff
@@ -23,6 +55,20 @@ def valid(self, color):
 		return True
 
 	def split(self, color, depth=0):
+		"""
+		split(color, depth=0)
+
+		Helper method for computing spectral coloring. 
+		This function does not need to be called, it is sufficient to call run() method
```
angriman, line 62: Either `call run()` or `call the run() method`

networkit/coloring.py
```diff
@@ -38,6 +84,13 @@ def split(self, color, depth=0):
 			self.split(otherColor, depth=depth+1)
 
 	def buildReverseDict(self):
+		"""
+		buildReverseDict()
+
+		Helper method for computing spectral coloring. 
+		This function does not need to be called, it is sufficient to call run() method
```
angriman, line 91: As above

networkit/components.pyx
```diff
@@ -45,78 +45,99 @@ cdef class ConnectedComponents(Algorithm):
...
-		--------
-		map:
+		""" 
+		getComponentSizes()
+		
+		Get the component sizes.
+
+		Returns
+		-------
+		map
```
angriman, line 99: I think that the return type here should be `dict`, not `map`.

networkit/components.pyx
```diff
@@ -219,22 +278,26 @@ cdef class StronglyConnectedComponents:
...
+		getPartition()
+
 		Returns a Partition object representing the strongly connected components.
 
-		Returns:
-		--------
-		Partition
+		Returns
+		-------
+		networkit.partition
```
angriman, line 287: This should be `networkit.Partition` (capital `P`)

networkit/components.pyx
```diff
@@ -219,22 +278,26 @@ cdef class StronglyConnectedComponents:
...
+		numberOfComponents()
+		
 		Returns the number of strongly connected components of the graph.
 
-		Returns:
-		--------
-		int
+		Returns
+		-------
+		count
```
angriman, line 300: I am not sure if we should do this. In the Python docs we often mix C++ types (e.g., `NetworKit::count`) and Python types (e.g., `list`) and this is confusing. However, this issue should be addressed in another PR. For now we should probably avoid using C++ types in Python docs.

networkit/components.pyx
```diff
@@ -243,34 +306,39 @@ cdef class StronglyConnectedComponents:
...
+		Parameters
+		----------
 		u : node
 			A node in the graph.
 
-		Returns:
-		int
+		Returns
+		-----------
+		count
```
angriman, line 316: As above

networkit/components.pyx
fabratu: Yes, this makes more sense than mixing both types. Also agree to defer this to another (set of) PR(s). However I will directly use Python types for previously missing doc-strings.

fabratu:
Took care of the comments - thanks for the review!

networkit/components.pyx
```diff
@@ -45,78 +45,99 @@ cdef class ConnectedComponents(Algorithm):
...
+		Get the component sizes.
 
-		Returns:
-		--------
-		map:
-			The map from component to size.
+		Returns
+		-------
+		dict
+			A dict containing the component and their sizes.
```
angriman, line 100: Maybe it is clearer something like "A dict of 'int: int', mapping component ids to their size."

networkit/components.pyx
```diff
@@ -301,42 +372,59 @@ cdef class WeaklyConnectedComponents(Algorithm):
...
+		Returns the map from component to size.
 
-			Returns:
-			--------
-			map[index, count]
-			 	A map that maps each component to its size.
+		Returns
+		-------
+		dict
+			A dict that maps each component to its size.
```
angriman, line 414: Each component id.

networkit/components.pyx
```diff
@@ -367,32 +457,41 @@ cdef class BiconnectedComponents(Algorithm):
...
+		Returns the map from component to size.
 
-			Returns:
-			--------
-			map[count, count]
-			A map that maps each component to its size.
+		Returns
+		-------
+		dict
+			A dict that maps each component to its size.
```
angriman, line 481: Component id to its size.

networkit/components.pyx
```diff
@@ -423,64 +525,87 @@ cdef class DynConnectedComponents(Algorithm):
...
+		Returns the map from component to size.
 
-			Returns:
-			--------
-			map[index, count]
-			 	A map that maps each component to its size.
+		Returns
+		-------
+		dict
+			A dict that maps each component to its size.
```
angriman, line 567: Component id, as above.

networkit/components.pyx
```diff
@@ -423,64 +525,87 @@ cdef class DynConnectedComponents(Algorithm):
...
+		Returns all the components, each stored as (unordered) set of nodes.
 
-			Returns:
-			--------
-			vector[vector[node]]
-				A vector of vectors. Each inner vector contains all the nodes inside the component.
+		Returns
+		-------
+		list[list[int]]
+			A list of lists. Each inner list contains all the nodes inside the component.
```
angriman, line 580: In some cases we also use `vector[vector[int]]`. I would recommend using `list` as `vector` is not a Python type.

networkit/components.pyx
```diff
@@ -515,66 +643,88 @@ cdef class DynWeaklyConnectedComponents(Algorithm):
...
+		Returns the map from component to size.
 
-			Returns:
-			--------
-			map[index, count]
-			 	A map that maps each component to its size.
+		Returns
+		-------
+		dict
+			A dict that maps each component to its size.
```
angriman, line 685: Component id, as above.

networkit/components.pyx
fabratu: Yeah, we are also not really consistent concerning listing types for lists/dicts. I would maybe prefer something like:
```
dict(int : int)
    A dict, mapping component ids to their size.
```
I am not sure the `,` belongs there btw.

angriman: Yes, this is even better, I think. I would not write a comma there.

networkit/components.pyx
fabratu: Hm, based on the dict-comment above - `list(list(int))` or `list[list[int]]`. The later one would be more in line with Python lists, however in that case we should use `dict{int : int}`, right?

fabratu: I think I am leaning more towards using `(` and `)` for naming types of composite data types.

angriman: I agree with both.

fabratu: So far we had one occasion where `[]` was used (community module). I added this fix to this PR.

PR #871, opened 2022-01-28
Refactor: use STL algorithms (2)

angriman:
As in #869: use STL algorithms where applicable.

### DIFF
--- a/networkit/cpp/distance/NeighborhoodFunction.cpp
+++ b/networkit/cpp/distance/NeighborhoodFunction.cpp
@@ -34,10 +34,10 @@ void NeighborhoodFunction::run() {
             nf[tid][dist] += 1;
         });
     });
-    count size = 0;
-    for (index i = 0; i < max_threads; ++i) {
-        size = std::max(size, (count)nf[i].size());
-    }
+    const count size =
+        std::max_element(nf.begin(), nf.end(), [](const auto &m1, const auto &m2) -> bool {
+            return m1.size() < m2.size();
+        })->size();
     result = std::vector<count>(size-1, 0);
     for (const auto& local_nf : nf) {
         for (const auto& elem : local_nf) {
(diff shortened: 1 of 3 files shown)


networkit/cpp/distance/NeighborhoodFunction.cpp
```diff
@@ -46,9 +46,10 @@ void NeighborhoodFunction::run() {
             }
         }
     }
-    for (index i = 1; i < size-1; ++i) {
-        result[i] += result[i-1];
-    }
+
+    if (result.size() > 2)
+        std::partial_sum(result.begin(), result.end() - 1, result.begin());
```
avdgrinten, line 51: I think the `- 1` is broken here. The previous code operated on the whole array but this replacement skips the last element.

networkit/cpp/distance/Volume.cpp
```diff
@@ -51,21 +51,14 @@ std::vector<double> Volume::volume(const Graph &G, std::vector<double> rs, count
...
+                return nodeDist.second <= rs[i];
+            });
     }
     std::vector<double> ys;
-    ys.reserve(xs.size());
-    for (const auto &x : xs) {
-        ys.push_back(x / samples);
-    }
+    std::transform(xs.cbegin(), xs.cend(), std::back_inserter(ys),
+                   [samples](double x) { return x / static_cast<double>(samples); });
```
avdgrinten, line 61: Not sure if this is really simpler than the original code. Anyway, we should keep the `.reserve()`.

networkit/cpp/distance/NeighborhoodFunction.cpp
angriman: Ah true, the vector has size `size - 1`.

avdgrinten (approved):

fabratu (approved):
LGTM.

PR #872, opened 2022-01-29
Refactor: GraphEvent

angriman:
Currently, `GraphEvent` implements two static functions `equal` and `compare`, which should be replaced with proper comparison operators. This PR
- implements comparison operators for `GraphEvent`
- deprecates `GraphEvent::equal` and `GraphEvent::compare`
- makes `GraphEvent` operators available in Python


fabratu (approved):
LGTM.

PR #873, opened 2022-01-31
Fix: Spanning Forest and KruskalMSF

angriman:
- `SpanningForest` inherits from `Algorithm`, which removes redundant code.
- `KruskalMSF` for unweighted graphs uses `SpanningForest` (and not the more expensive Kruskal algorithm).
- Add a simple Python test for `SpanningForest`.


avdgrinten (approved):
LGTM. Although I'm not a fan of `std::transform`, the original `push_back` loop looks just easier to read here (same for `std::for_each`).

PR #874, opened 2022-01-31
Fix warning (narrowing conversion)

angriman:
Fixes a narrowing conversion warning from clang-tidy that appears for every `.cpp` file that uses `Graph`.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -2283,7 +2283,11 @@ void Graph::sortEdges(Lambda lambda) {
         if (adjList.size() > indices.size())
             indices.resize(adjList.size());
 
-        const auto indicesEnd = indices.begin() + adjList.size();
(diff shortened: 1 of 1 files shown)


fabratu (dismissed):

avdgrinten:
What is the exact warning that is fixed here? Why `long` and not some other type?

avdgrinten:
Technically, to make this equal to the type of `operator+` for the iterator, we'd have to cast to `std::iterator_traits<X>::difference_type` (where `X` is the type of the iterator, see https://en.cppreference.com/w/cpp/iterator/iterator_traits). In practice, this is probably always `ptrdiff_t`.

angriman:
Thanks, now it is casted to `std::iterator_traits<std::vector<index>::const_iterator>::difference_type`

fabratu (approved):
LGTM and thanks for @avdgrinten for pointing out.

PR #875, opened 2022-01-31
Fix: use of undeclared identifier

angriman:
FIxes `use of undeclared identifier 'validate'` while compiling NetworKit with `RelWithDebInfo` (clang 10).
Shouldn't we test this configuration in our CI system?


fabratu (approved):

avdgrinten:
I consider CMake to be broken in this regard -- there is no good reason to set NDEBUG (it could arguably be even omitted in release builds by default). Unfortunately, I have not seen an easy way to fix that.

PR #877, opened 2022-02-03
Python-API: Support for 3D-coordinates with Maxent-Stress layout.

fabratu:
While the C++ code supports arbitrary dimensions, this is not easily exposed via Cython currently and require more work. This approach just adds 3D coordinates in addition to the already supported 2D coordinates. As a drive-by it fixes some wrong doc-strings. Complete doc-overhaul will be added as part of the current doc-improvement PR series.

### DIFF
--- a/networkit/viz.pyx
+++ b/networkit/viz.pyx
@@ -18,6 +18,7 @@ cdef extern from "<networkit/viz/Point.hpp>" namespace "NetworKit" nogil:
 	cdef cppclass Point[T]:
 		Point()
 		Point(T x, T y)
+		count getDimensions()
 		T& operator[](const index i) except +
 		T& at(const index i) except +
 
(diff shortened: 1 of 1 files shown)


networkit/viz.pyx
```diff
@@ -60,21 +61,30 @@ cdef class GraphLayoutAlgorithm:
...
 		return self._this.numEdgeCrossings()
 
 	def getCoordinates(self):
-		""" Computes approximation (in parallel) of the Spanning Edge Centrality. """
+		""" Get node coordinates for computed layout."""
 		if self._this == NULL:
 			raise RuntimeError("Error, object not properly initialized")
-		cdef pair[double, double] pr = pair[double, double](0, 0)
+		cdef vector[double] pr3d = [0, 0, 0]
+		cdef vector[double] pr2d = [0, 0]
```
angriman, line 74: Can't we create `pr3d` or `pr2d` depending on whether `num_dim` is 3 or 2 (by moving them inside the `if` statement below)?

angriman (dismissed):

fabratu:
Since the code is changed, I thought of including the documentation fixes for the viz-module in this PR. Or better put this on another later PR?

angriman:
I think it's better to have the documentation fixes in a separate PR.

fabratu:
Ok, removed the doc-fixes.

angriman (approved):

fabratu:
Merging this earlier, since it is only a small fix and is needed for #891 .

PR #878, opened 2022-02-05
Refactor: Math libraries

angriman:
- Remove usage of deprecated `math.h` C library
- the "ttmath" library is added as a git submodule


avdgrinten:
`math.h` is not really deprecated, is it? It's just the C version of `<cmath>`.

avdgrinten:
Maybe the 2nd and 3rd commit should be squashed together, such that we do not end up in a situation where NetworKit does not build.

angriman:
Ah true, I misinterpreted the documentation about "C compatibility headers" https://en.cppreference.com/w/cpp/header

avdgrinten (approved):
LGTM.

PR #879, opened 2022-02-05
Remove usage of functions from math.h

angriman:
Replace usage of math functions from the C standard library `math.h`  with C++ standard math functions from `cmath`.
I generated these changes automatically (of course I checked them before opening this PR); the only exception is:
https://github.com/networkit/networkit/compare/master...angriman:refactor/drop-math-h?expand=1#diff-d450828fc2762048c7b60b0267df545b4f96f5624e489fbc8293e87eb8f351a8R109
which does not compile without `static_cast<int>` because we are trying to call `std::min` with `double` and `int` parameters and template type deduction fails.


avdgrinten (approved):
LGTM.

angriman:
@avdgrinten or @fabratu could you please review this again? I had to force-push to fix a conflict.

fabratu (approved):

PR #880, opened 2022-02-08
New Feature: Node Attributes

fidus58:


include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+    class NodeAttributeStorageBase {
+    public:
+        NodeAttributeStorageBase(const Graph *graph, std::string name, std::type_index type)
+            : name{std::move(name)}, type{type}, theGraph{graph} {}
+
+        virtual ~NodeAttributeStorageBase() = default;
+
+        virtual void invalidateAttributes() = 0;
+
+        std::string getName() { return name; }
```
angriman, line 160:
```suggestion
        const std::string &getName() const noexcept { return name; }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+        std::type_index getType() { return type; }
```
angriman, line 162:
```suggestion
        std::type_index getType() const noexcept { return type; }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+        bool isValid(index i) { return i < valid.size() && valid[i]; }
```
angriman, line 164:
```suggestion
        bool isValid(index i) const noexcept { return i < valid.size() && valid[i]; }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                valid[i] = false;
+                --validElements;
+            }
+        }
+
+    protected:
+        void markValid(index i) {
+            if (!theGraph->hasNode(i)) {
+                throw std::runtime_error("This node does not exist");
+            }
```
angriman, line 178: I think that we should expect the user to pass correct indices and replace these bound checks with assertions.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            if (i >= valid.size()) {
+                valid.resize(i + 1);
+            }
+            valid[i] = true;
+            ++validElements;
+        }
+
+        void checkIndex(index i) {
```
angriman, line 186:
```suggestion
        void checkIndex(index i) const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            if (!theGraph->hasNode(i)) {
+                throw std::runtime_error("This node does not exist");
+            }
+            if (!isValid(i)) {
+                throw std::runtime_error("Invalid attribute value");
+            }
```
angriman, line 192: Same here. It seems to me that this check is useful for debugging, so we should use assertions instead of runtime checks.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+    template <typename T>
+    class NodeAttributeStorage : public Graph::NodeAttributeStorageBase {
+    public:
+        NodeAttributeStorage(const Graph *theGraph, std::string name)
+            : NodeAttributeStorageBase{theGraph, std::move(name), typeid(T)} {}
+
+        ~NodeAttributeStorage() override { invalidateAttributes(); }
+
+        void invalidateAttributes() override {
+            for (auto att : attrSet)
```
angriman, line 216: Isn't `att` a copy?
```suggestion
            for (auto &att : attrSet)
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                att->invalidateAttribute();
+        }
+
+        void resize(index i) {
+            if (i >= values.size()) {
+                values.resize(i + 1);
+            }
+        }
+
+        auto size() { return validElements; }
```
angriman, line 226:
```suggestion
        auto size() const noexcept { return validElements; }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+        void set(index i, T v) {
+            markValid(i);
+            resize(i);
+            values[i] = std::move(v);
```
angriman, line 231: For this to work, shouldn't `v` be an rvalue ref? `void set(index i, T &&v)`.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+        }
+
+        // instead of returning an std::optional (C++17) we provide these
+        // C++14 options
+        // (1) throw an exception when invalid:
+        T get(index i) { // may throw
```
angriman, line 237:
```suggestion
        T get(index i) const { // may throw
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            checkIndex(i);
+            return values[i];
+        }
+
+        // (2) give default value when invalid:
+        T get(index i, T defaultT) {
```
angriman, line 243:
```suggestion
        T get(index i, T defaultT) const noexcept {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                    }
+                    ++idx;
+                }
+                return *this;
+            }
+
+            Iterator &operator++() {
+                if (!storage) {
+                    throw std::runtime_error("Invalid attribute iterator");
+                }
```
angriman, line 300: As above, we should expect the user to use the iterator correctly. I recommend replacing this with an assertion.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                ++idx;
+                return nextValid();
+            }
+
+            auto operator*() {
+                if (!storage) {
+                    throw std::runtime_error("Invalid attribute iterator");
+                }
```
angriman, line 308: As above.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                return std::make_pair(idx, storage->values[idx]);
+            }
+
+            //          auto nodeValuePair() { return std::make_pair(idx, **this); }
```
angriman, line 312: This line can be removed.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+            bool operator==(Iterator const &iter) {
```
angriman, line 314:
```suggestion
            bool operator==(Iterator const &iter) const noexcept {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                if (storage == nullptr && iter.storage == nullptr) {
+                    return true;
+                }
+                return storage == iter.storage && idx == iter.idx;
+            }
+
+            bool operator!=(Iterator const &iter) { return !(*this == iter); }
```
angriman, line 321:
```suggestion
            bool operator!=(Iterator const &iter) const noexcept { return !(*this == iter); }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            NodeAttributeStorage<T> *storage;
+            index idx;
+        }; // class Iterator
+    private:
+        class IndexProxy {
+        public:
+            IndexProxy(NodeAttributeStorage<T> *storage, index idx) : storage{storage}, idx{idx} {}
+
+            // reading at idx
+            operator T() {
```
angriman, line 333:
```suggestion
            operator T() const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            }
+
+            bool operator!=(Iterator const &iter) { return !(*this == iter); }
+
+        private:
+            NodeAttributeStorage<T> *storage;
+            index idx;
+        }; // class Iterator
+    private:
+        class IndexProxy {
```
angriman, line 328: It's not clear what this class is doing, could you add a brief description?

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+            // reading at idx
+            operator T() {
+                storage->checkIndex(idx);
+                return storage->values[idx];
+            }
+
+            // writing at idx
+            T &operator=(T const &other) {
+                storage->set(idx, std::move(other));
```
angriman, line 340: Since `other` is a const ref, `std::move` does nothing here right?

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            if (owned_storage) {
+                owned_storage->attrSet.erase(this);
+            }
+        }
+
+        auto begin() { return Iterator(owned_storage.get()).nextValid(); }
+
+        auto end() { return Iterator(nullptr); }
+
+        auto size() { return owned_storage->size(); }
```
angriman, line 388: These can all be `const`

angriman: `size` can also be `noexcept`

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+
+        void set(index i, T v) {
+            checkAttribute();
+            owned_storage->set(i, std::move(v));
+        }
+
+        auto get(index i) {
```
angriman, line 395:
```suggestion
        auto get(index i) const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            checkAttribute();
+            return owned_storage->get(i);
+        }
+
+        auto get(index i, T defaultT) {
```
angriman, line 400:
```suggestion
        auto get(index i, T defaultT) const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            checkAttribute();
+            return owned_storage->get(i, defaultT);
+        }
+
+        IndexProxy operator[](index i) {
```
angriman, line 405:
```suggestion
        IndexProxy operator[](index i) const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            checkAttribute();
+            return IndexProxy(owned_storage.get(), i);
+        }
+
+        void checkAttribute() {
+            if (!valid) {
+                throw std::runtime_error("Invalid attribute");
+            }
```
angriman, line 413: As already mentioned, replace with assertion.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+        }
+
+        void write(const std::string &filename) {
+            std::ofstream out(filename);
+            if (!out)
+                std::cerr << "cannot open '" << filename << "' for writing\n";
```
angriman, line 419: You could use a macro from `Log.hpp`, for example `ERROR("cannot open...")`

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+                auto v = pair.second;
+                out << n << "\t" << v << "\n";
+            }
+            out.close();
+        }
+
+        void read(const std::string &filename) {
+            std::ifstream in(filename);
+            if (!in) {
+                std::cerr << "cannot open '" << filename << "' for reading\n";
```
angriman, line 433: As above, you can use `ERROR(...)`

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            while (true) {
+                int n;
+                T t;
+                std::string line;
+                std::getline(in, line);
+                if (!in)
+                    break;
+                std::istringstream istring(line);
+                istring >> n >> t;
+                std::cout << "got: " << n << " " << t << "\n";
```
angriman, line 444: The standard output print can be removed.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            if (!in) {
+                std::cerr << "cannot open '" << filename << "' for reading\n";
+            }
+            while (true) {
+                int n;
+                T t;
+                std::string line;
+                std::getline(in, line);
+                if (!in)
+                    break;
```
angriman, line 441: Maybe it's cleaner to have
```cpp
std::string line;
while (std::getline(in, line))
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+    class NodeAttributeMap {
+        friend Graph;
+        const Graph *theGraph;
+
+    public:
+        std::unordered_map<std::string, std::shared_ptr<NodeAttributeStorageBase>> attrMap;
+
+        NodeAttributeMap(const Graph *g) : theGraph{g} {}
+
+        auto find(std::string const &name) {
```
angriman, line 466:
```suggestion
        auto find(std::string const &name) const {
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            auto it = attrMap.find(name);
+            if (it == attrMap.end()) {
+                throw std::runtime_error("No such attribute");
+            }
+            return it;
```
angriman, line 471: This can be simplified with `return attrMap.at(name);`

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+        }
+
+        template <typename T>
+        auto attach(const std::string &name) {
+            auto ownedPtr = std::make_shared<NodeAttributeStorage<T>>(theGraph, std::string{name});
+            auto insertResult = attrMap.insert(std::make_pair(ownedPtr->getName(), ownedPtr));
```
angriman, line 477:
```suggestion
            auto insertResult = attrMap.emplace(ownedPtr->getName(), ownedPtr);
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+            if (it->second.get()->getType() != typeid(T))
+                throw std::runtime_error("Type mismatch in nodeAttributes().get()");
+            return NodeAttribute<T>{std::static_pointer_cast<NodeAttributeStorage<T>>(it->second)};
+        }
+
+    }; // class NodeAttributeMap
+    NodeAttributeMap nodeAttributeMap;
+
+public:
+    auto &nodeAttributes() { return nodeAttributeMap; }
```
angriman, line 505:
```suggestion
    auto &nodeAttributes() const noexcept { return nodeAttributeMap; }
```

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,389 @@ class Graph final {
...
+    auto attachNodeStringAttribute(const std::string &name) {
+        nodeAttributes().theGraph = this;
+        return nodeAttributes().attach<std::string>(name);
+    }
+
+    void detach(const std::string &name) { nodeAttributes().detach(name); }
+
+    // using NodeStringAttribute = NodeAttribute<std::string>;
+    using NodeIntAttribute = NodeAttribute<int>;
+    // using NodeDoubleAttribute = NodeAttribute<double>;
```
angriman, line 528: Unused code should be removed

angriman (changes requested):
Thank you for providing this very useful feature! In addition to my comments, there are a few minor things to fix:
- The PR modifies files that are not related to node attributes (e.g., `Log`, `StringBuilder`, etc.). Please remove those changes.
- It's not very clear how the ownership of node attributes works. In the beginning, I was confused by the usage of raw pointers, but it seems to me that the attributes are owned by the `NodeAttribute` class via the shared pointer `owned_storage`. It would be helpful to add a brief description of what each class is doing and add documentation to the functions of `NodeAttribute` and `NodeAttributeStorage`.
- Please avoid snake_case for variable names, our coding style is to use camelCase.

include/networkit/graph/Graph.hpp
avdgrinten: The outcome of the discussion here: https://github.com/networkit/networkit/discussions/769 is that `Graph` should use exceptions, not assertion to check for broken indices.

include/networkit/graph/Graph.hpp
avdgrinten: Still still works (and moves from the local `v` to `values`). It just doesn't move from the caller's argument.

include/networkit/graph/Graph.hpp
avdgrinten: See above (https://github.com/networkit/networkit/discussions/769).

include/networkit/graph/Graph.hpp
avdgrinten: See above (https://github.com/networkit/networkit/discussions/769).

include/networkit/graph/Graph.hpp
avdgrinten: See above (https://github.com/networkit/networkit/discussions/769).

include/networkit/graph/Graph.hpp
avdgrinten: But the error message is a bit more helpful this way.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,395 @@ class Graph final {
...
+    class NodeAttributeStorageBase {
+    public:
+        NodeAttributeStorageBase(const Graph *graph, std::string name, std::type_index type)
+            : name{std::move(name)}, type{type}, theGraph{graph} {}
+
+        virtual ~NodeAttributeStorageBase() = default;
+
+        virtual void invalidateAttributes() = 0;
+
+        const std::string getName() const noexcept { return name; }
```
angriman, line 165:
```suggestion
        const std::string &getName() const noexcept { return name; }
```

networkit/graph.pyx
```diff
@@ -676,7 +737,7 @@ cdef cppclass EdgeCallBackWrapper:
 			(<object>callback)(u, v, w, eid)
 		except Exception as e:
 			error = True
-			message = stdstring("An Exception occurred, aborting execution of iterator: {0}".format(e))
+			message = stdstring("An Exception occureed, aborting execution of iterator: {0}".format(e))
```
angriman, line 740: This change is probably unintentional

angriman:
I removed changes unrelated to node attributes and addressed further minor issues. The PR looks good to me now, there are just two things missing that I forgot to point out in my first review:
- Add simple Python tests (just to ensure that the Python attributes work as expected)
- Add an example notebook that shows how to use node attributes in Python. We could either add a new `Node Attributes.ipynb` in the `notebook` directory, or modify `GraphNotebook.ipynb`.

avdgrinten:
It should be possible to remove the `attrSet` in `NodeAttributeStorage`, right? (I think I discussed that with Klaus in person.)

fidus58:
> It should be possible to remove the `attrSet` in `NodeAttributeStorage`, right? (I think I discussed that with Klaus in person.)

currently copied attributes share their storage - so when detaching an attribute from a graph (storage is destructed) all attribute copies are must be invalidated.

we could decide, not to allow for copies at all (as it was in the first place)?

avdgrinten:
We can store whether the attribute storage is valid within the storage itself. There is no need to invalidate views, we can simply invalidate the entire storage.

fidus58:
> On 17. Feb 2022, at 12:28, Alexander van der Grinten ***@***.***> wrote:
> 
> 
> We can store whether the attribute storage is valid within the storage itself. There is no need to invalidate views, we can simply invalidate the entire storage.
> 
> 

And then throw, when accessing an attribute without storage?

avdgrinten:
Well, the storage will still exist, right? (Because of the shared_ptr in the view.) It will just be invalid. Yes, in that case, we can throw.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,394 @@ class Graph final {
...
+
+        virtual ~NodeAttributeStorageBase() = default;
+
+        virtual void invalidateAttributes() = 0;
+
+        const std::string &getName() const noexcept { return name; }
+
+        std::type_index getType() const noexcept { return type; }
+
+        bool isValid(index i) const noexcept { return i < valid.size() && valid[i]; }
```
angriman, line 169: Since `i` is a node, we should use the `node` type.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,394 @@ class Graph final {
...
+
+        // Called by Graph when node is deleted.
+        void invalidate(index i) {
```
angriman, line 172: As above

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,394 @@ class Graph final {
...
+            if (isValid(i)) {
+                valid[i] = false;
+                --validElements;
+            }
+        }
+
+    protected:
+        void markValid(index i) {
```
angriman, line 180: As above

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,394 @@ class Graph final {
...
+                throw std::runtime_error("This node does not exist");
+            }
+            if (i >= valid.size()) {
+                valid.resize(i + 1);
+            }
+            valid[i] = true;
+            ++validElements;
+        }
+
+        void checkIndex(index i) const {
```
angriman, line 191: As above

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,394 @@ class Graph final {
...
+            : NodeAttributeStorageBase{theGraph, std::move(name), typeid(T)} {}
+
+        ~NodeAttributeStorage() override { invalidateAttributes(); }
+
+        void invalidateAttributes() override {
+            for (auto &att : attrSet)
+                att->invalidateAttribute();
+        }
+
+        void resize(index i) {
```
angriman, line 225: Here as well, and also for other methods that take a node as an input parameter.

angriman (approved):
I added further C++ and Python tests, and also a short example in the graph notebook. Looks good to me now. Since this is an important feature, I encourage other maintainers to take a look at this PR.

include/networkit/graph/Graph.hpp
```diff
@@ -139,6 +145,379 @@ class Graph final {
...
+    template <typename T>
+    class NodeAttribute;
+
+    template <typename T>
+    class NodeAttributeStorage : public NodeAttributeStorageBase {
+    public:
+        NodeAttributeStorage(const Graph *theGraph, std::string name)
+            : NodeAttributeStorageBase{theGraph, std::move(name), typeid(T)} {}
+
+        void resize(node i) {
```
fabratu, line 221: Is there a reason why `resize` and `size` are implemented in the derived class and not in the base class?

networkit/graph.pyx
```diff
@@ -665,6 +665,186 @@ cdef class Graph:
 			yield dereference(it)
 			preincrement(it)
 
+	def attachNodeAttribute(self, name, ofType):
+		"""
+		Attaches a node attribute to the graph and returns it
```
fabratu, line 670: This and the following comments refer to the documentation. While the build on CI is successful, we currently not check (and thats not that easy) the correct syntax and needed doc-strings for rendering. When I build the branch locally, the following (errornous) output is generated:

Bildschirmfoto 2022-02-25 um 12 32 45
Bildschirmfoto 2022-02-25 um 12 33 07

Add `attachNodeAttribute(name, ofType)` as the first line of the doc (enables declaration visibility in rendered doc).

networkit/graph.pyx
```diff
@@ -665,6 +665,186 @@ cdef class Graph:
...
+
+		A = G.attachNodeAttribute("attributeIdentifier", ofType)
+
+		all values are initially undefined
+
```
fabratu, line 675: Unneccesary empty line (multiple occassions)

networkit/graph.pyx
```diff
@@ -665,6 +665,186 @@ cdef class Graph:
...
+		A[node] = value
+
+		and get by
+
+		value = A[node]
+
+		getting undefined values raises a ValueError
+		removing a node makes all its attributes undefined
+
+		Parameters:
```
fabratu, line 687: Remove `:` (and one `-` below). Otherwise Sphinx automodule will not render the parameter correctly.

networkit/graph.pyx
```diff
@@ -665,6 +665,186 @@ cdef class Graph:
...
+		-----------
+		name   : a distinguished name for this attribute
```
fabratu, line 689: Format for parameters is:
```
name : str
	A distinguished name for this attribute.
ofType : int or float or str
	The type of this attribute.
```
(Again, this is needed in order to preserve correct rendering of doc)

networkit/graph.pyx
```diff
@@ -665,6 +665,186 @@ cdef class Graph:
...
+		if type(name)!=str:
+			raise Exception("Attribute name should be a string")
+		if ofType==int:
+			return NodeAttribute(NodeIntAttribute().setThis(self._this.attachNodeIntAttribute(name.encode('UTF-8')), &self._this), int)
+		if ofType==float:
+			return NodeAttribute(NodeDoubleAttribute().setThis(self._this.attachNodeDoubleAttribute(name.encode('UTF-8')), &self._this), float)
+		if ofType==str:
+			return NodeAttribute(NodeStringAttribute().setThis(self._this.attachNodeStringAttribute(name.encode('UTF-8')), &self._this), str)
+
+	def detach(self, name):
```
fabratu, line 701: As above (see comments for doc) for both declaration + parameters.

include/networkit/graph/Graph.hpp
angriman: Because the attribute values are stored in the derived class (`values` vector).

networkit/graph.pyx
angriman: Thanks for testing the docs, now it should be fine.

include/networkit/graph/Graph.hpp
fabratu: Ah, right - somehow overlooked this. And the whole approach of having a base class is because of the generalized `NodeAttributeMap`. Ok, nevermind then.

networkit/graph.pyx
fabratu: Sorry for being nitpicky, but still looks a bit odd. (I also could have been clearer with my comments)

Bildschirmfoto 2022-02-25 um 13 31 11

Besides the wrong look description for `attachNodeAttribute`, we also have missing punctuation marks and the parameter description of `detach` is still wrong.

fabratu: Also in general - would it make sense to rename `detach` to `detachNodeAttribute` to be in line with `attachNodeAttribute`?

fabratu:
For the Python API. Maybe this is not the intended way to use it, but the output of the attributes can be a bit unexpected?
```
>>> res = G.attachNodeAttribute("my int attr2", int)
>>> res.attr[0] = 1
>>> print(res.attr[0])
1
>>> print(res.attr[1])
Traceback (most recent call last): ...
>>> for val in res.attr:
...     print(val)
...
(0, 1)
```

Why does iterating over the attribute gives two values instead of one?

Also the return type is currently `networkit.graph.NodeAttribute`, however the structure or behavior is not documented anywhere. Could you add a doc-string for `class NodeAttribute`?

networkit/graph.pyx
```diff
@@ -665,6 +665,220 @@ cdef class Graph:
...
+		"""
+		if not isinstance(name, str):
+			raise Exception("Attribute name has to be a string")
+
+		if ofType == int:
+			return NodeAttribute(NodeIntAttribute().setThis(self._this.attachNodeIntAttribute(name.encode('UTF-8')), &self._this), int)
+		elif ofType == float:
+			return NodeAttribute(NodeDoubleAttribute().setThis(self._this.attachNodeDoubleAttribute(name.encode('UTF-8')), &self._this), float)
+		elif ofType == str:
+			return NodeAttribute(NodeStringAttribute().setThis(self._this.attachNodeStringAttribute(name.encode('UTF-8')), &self._this), str)
```
angriman, line 709: Instead of `.encode('UTF-8')` we can use the `stdstring` function from `helpers`, I will push a fix for this.

networkit/graph.pyx
```diff
@@ -665,6 +665,220 @@ cdef class Graph:
...
+		Detaches a node attribute from the graph.
+
+		Parameters
+		----------
+		name : str
+			The distinguished name for the attribute to detach.
+		"""
+		if not isinstance(name, str):
+			raise Exception("Attribute name has to be a string")
+		self._this.detachNodeAttribute(name.encode('UTF-8'))
```
angriman, line 724: Same here

angriman (approved):
Thanks for improving the documentation. Unless other issues arise, I think that this is ready to be merged.

networkit/graph.pyx
fabratu: Thanks! This is something, which is done in several modules actually (re-implementing instead of using `helpers`). Already on my agenda to fix this in an upcoming PR :)

PR #882, opened 2022-02-09
Format headers of matching and numerics

angriman:
This is a continuation of #803. Changes do not create conflicts with existing PRs and are automatically generated with the exception of:
- move a few comments on a new line for better readability
- add missing `#include`; this was necessary because changing the order of some `#include`s broke the build

### DIFF
--- a/include/networkit/algebraic/MatrixTools.hpp
+++ b/include/networkit/algebraic/MatrixTools.hpp
@@ -11,6 +11,7 @@
 #include <atomic>
 #include <cmath>
 
+#include <networkit/algebraic/AlgebraicGlobals.hpp>
 #include <networkit/graph/Graph.hpp>
 
 namespace MatrixTools {
--- a/include/networkit/matching/Matcher.hpp
+++ b/include/networkit/matching/Matcher.hpp
@@ -1,4 +1,3 @@
-// no-networkit-format
 /*
  * Matcher.hpp
  *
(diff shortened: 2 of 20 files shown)


avdgrinten (changes requested):
Non-automated parts of the PR should be in their own commits (for example, it's impossible to review which #include was added).

angriman:
Yes good point, I moved those changes to separate commits. The missing include is fixed before the formatting to avoid commits where NetworKit would not build successfully.

include/networkit/algebraic/MatrixTools.hpp
```diff
@@ -11,7 +11,7 @@
 #include <atomic>
 #include <cmath>
 
-#include <networkit/algebraic/Globals.hpp>
+#include <networkit/algebraic/AlgebraicGlobals.hpp>
```
avdgrinten, line 14: Hm? Why is this changed? (Instead of doing this in the first commit?)

avdgrinten (commented):
When you're fixing comments, you can also turn them into proper sentences with capitalization and punctuation. Other than that, this LGTM.

include/networkit/algebraic/MatrixTools.hpp
angriman: Ah sorry, instead of using `git commit --fixup` to fix the first commit I used `git commit --amend`, I'll fix this now.

fabratu (approved):
LGTM.

angriman:
Thanks for the review. I need another one from @avdgrinten due to the requested changes.

avdgrinten (approved):

PR #884, opened 2022-02-13
Refactor: replace set with unordered set

angriman:
- Replace `std::set` with `std::unordered_set` in places where order is not important
- Remove unused instance of `std::unordered_set` from release mode

### DIFF
--- a/include/networkit/sparsification/SimmelianScore.hpp
+++ b/include/networkit/sparsification/SimmelianScore.hpp
@@ -9,8 +9,12 @@
 #ifndef NETWORKIT_SPARSIFICATION_SIMMELIAN_SCORE_HPP_
 #define NETWORKIT_SPARSIFICATION_SIMMELIAN_SCORE_HPP_
 
-#include <networkit/edgescores/EdgeScore.hpp>
 #include <set>
+#include <unordered_set>
+
+#include <tlx/define/deprecated.hpp>
+
+#include <networkit/edgescores/EdgeScore.hpp>
 
 namespace NetworKit {
 
@@ -81,13 +85,19 @@ public:
             const std::vector<RankedNeighbors>& neighbors,
             const count& maxRank);
 
-    void matchNeighbors(node ego, node alter, bool reciprocityCheck,
+    void TLX_DEPRECATED(matchNeighbors(node ego, node alter, bool reciprocityCheck,
                         std::vector<RankedEdge>::const_iterator &egoIt,
                         const RankedNeighbors &egoNeighbors, std::set<node> &egoNeighborsUnmatched,
-                        std::set<node> &alterNeighborsUnmatched, count rank, count &overlap);
+                        std::set<node> &alterNeighborsUnmatched, count rank, count &overlap));
 
 protected:
     const std::vector<count>* triangles;
+
+    void matchNeighbors(node ego, node alter, bool reciprocityCheck,
+                        std::vector<RankedEdge>::const_iterator &egoIt,
+                        const RankedNeighbors &egoNeighbors, std::unordered_set<node> &egoNeighborsUnmatched,
+                        std::unordered_set<node> &alterNeighborsUnmatched, count rank, count &overlap);
+
 
 };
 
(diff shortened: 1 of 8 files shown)


include/networkit/sparsification/SimmelianScore.hpp
```diff
@@ -83,8 +84,8 @@ class SimmelianScore : public EdgeScore<double> {
 
     void matchNeighbors(node ego, node alter, bool reciprocityCheck,
                         std::vector<RankedEdge>::const_iterator &egoIt,
-                        const RankedNeighbors &egoNeighbors, std::set<node> &egoNeighborsUnmatched,
-                        std::set<node> &alterNeighborsUnmatched, count rank, count &overlap);
+                        const RankedNeighbors &egoNeighbors, std::unordered_set<node> &egoNeighborsUnmatched,
+                        std::unordered_set<node> &alterNeighborsUnmatched, count rank, count &overlap);
```
avdgrinten, line 88: Doesn't this affect users of the function? If yes, deprecation (or maybe just providing two functions) is necessary.

networkit/cpp/linkprediction/AdjustedRandIndex.cpp
```diff
@@ -6,15 +6,14 @@
...
+#include <unordered_set>
 
 #include <networkit/linkprediction/AdjustedRandIndex.hpp>
 
 namespace NetworKit {
 
 double AdjustedRandIndex::runImpl(node u, node v) {
-  std::set<node> uNeighbors;
-  std::set<node> vNeighbors;
+  std::unordered_set<node> uNeighbors, vNeighbors;
```
michitux, line 16: Running `set_intersection` (which needs a sorted range) on an unordered set looks wrong. I think this is introducing a bug. Using `std::sort` on a `std::vector` might be more efficient, though. Reading the code further, I think there should be a set difference instead of a union, though after knowing the size of the intersection, it should be possible to calculate all of these numbers without further set operations.

michitux (changes requested):

angriman:
Set union and intersection need indeed a sorted range, I restored the file to its original version.

networkit/cpp/io/SNAPEdgeListPartitionReader.cpp
```diff
@@ -51,7 +53,9 @@ Cover SNAPEdgeListPartitionReader::read(const std::string &path,
         } else {
             std::stringstream linestream(line);
             while (linestream >> current) {
+#ifndef NDEBUG
                 uniqueIDs.insert(current);
+#endif
```
avdgrinten, line 58: What does this do?

avdgrinten: If NDEBUG is set, it does nothing. Otherwise, it inserts the element into a set, but even if it does, there is no fail state?

michitux (approved):

networkit/cpp/io/SNAPEdgeListPartitionReader.cpp
michitux: There seems to be a debug log using `uniqueIDs` below, I think letting this depend on `NDEBUG` is therefore not the best option.

avdgrinten: You cannot assume that `DEBUG` is only compiled in when NDEBUG is not set. Why does this pass CI?

angriman: That's why I asked what to do. We could either add the `#ifndef NDEBUG` to the log message (so that we do not compile it in release mode), or we just remove it so that the set is not needed anymore.

avdgrinten: I guess the message was done for a reason, so putting it into `#ifndef` sounds reasonable.

angriman: Agree, I put the debug message within the `#ifndef`.

avdgrinten (approved):

PR #885, opened 2022-02-14
Refactor: Matching

angriman:
- Remove redundant `MyEdge` struct from `LocalMaxMatcher` (we can use `WeightedEdge` already defined in `Graph`)
- Ensure that `run` is called before `Matcher::getMatching`
- Add missing `hasRun = true` in `LocalMaxMatcher`


avdgrinten (approved):
LGTM.

PR #887, opened 2022-02-16
Documentation fix for Python-modules: dynamic, dynamics, embedding, engineering, flow

fabratu:
This PR fixes documentation for Python module `dynamic, dynamics, embedding, engineering, flow`. This also includes conversion of data types to Python native data types as discussed in https://github.com/networkit/networkit/pull/870.

I also introduced another change for enum-blocks. Until now all enum-definitions were also part of the doc, but only listed without additional context on how to correctly use them and how they are tied to other functions. Example from `centrality`:

Bildschirmfoto 2022-02-16 um 10 02 17

For `GraphEvent` this behavior is changed, so that the enum is listed as a parameter. In addition ` :undoc-members:` from automodule has to be removed, so that the original enum-blocks are gone. Example:

Bildschirmfoto 2022-02-16 um 10 04 47

### DIFF
--- a/networkit/dynamics.pyx
+++ b/networkit/dynamics.pyx
@@ -4,12 +4,72 @@ from .base cimport _Algorithm
 from .base cimport Algorithm
 from .graph cimport _Graph, Graph
 
+def graphFromStream(stream, weighted, directed):
+	""" 
+	graphFromStream(stream, weighted, directed)
+
+	Convenience function for creating a new graph from a stream of graph events
+
+	Parameters
+	----------
+	stream : list(networkit.dynamics.GraphEvent)
+		Event stream
+	weighted : bool
+		Produce a weighted or unweighted graph
+	directed : bool
+		Produce a directed or undirected graph
+	"""
+	G = Graph(0, weighted, directed)
+	gu = GraphUpdater(G)
+	gu.update(stream)
+	return G
+
 def stdstring(pystring):
-	""" convert a Python string to a bytes object which is automatically coerced to std::string"""
+	""" 
+	stdstring(pystring)
+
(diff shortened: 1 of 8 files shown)


networkit/dynamics.pyx
```diff
@@ -62,6 +150,21 @@ cdef extern from "<networkit/dynamics/DGSStreamParser.hpp>":
...
+
+	Create a DGSStreamParser, handling graph streams encoded in DGS-based files.
+	For documentation about DGS, see: https://graphstream-project.org/doc/Advanced-Concepts/The-DGS-File-Format/
+
+	Parameters
+	----------
+	path : str
+		Filename including path for DGS-file.
+	mapped : bool, optional
+		Indicate whether file includes mapping. Default: True
```
angriman, line 164: Maybe `Indicates` or just drop the word.

networkit/dynamics.pyx
```diff
@@ -62,6 +150,21 @@ cdef extern from "<networkit/dynamics/DGSStreamParser.hpp>":
...
+	baseIndex : int, optional
+		Indicate whether a fixed base index should be set. Default: 0
```
angriman, line 166: As above.

networkit/flow.pyx
```diff
@@ -48,42 +50,48 @@ cdef class EdmondsKarp(Algorithm):
...
+		getMaxFlow()
+
+		Returns the value of the maximum flow from source to sink.
 
-		Returns:
-		--------
-		edgeweight
+		Returns
+		-------
+		double
```
angriman, line 59:
```suggestion
		float
```

PR #888, opened 2022-02-18
Documentation fix for Python-modules: generators, GEXFIO, gephi, globals

fabratu:
Follow up PR of #870.


networkit/generators.pyx
```diff
@@ -370,24 +513,97 @@ For a temperature of 0, the model resembles a unit-disk model in hyperbolic spac
...
+		capacity : int
+			Tuning parameter.
+		"""
 		(<_HyperbolicGenerator*>(self._this)).setLeafCapacity(capacity)
 
 	def setBalance(self, balance):
+		"""
+		setBalance(balance)
+
+		Set the balance of the quad tree. Value should be between 0.0 and 1.0.
```
angriman, line 532:
```suggestion
		Set the balance of the quadtree. Value should be between 0.0 and 1.0.
```

networkit/generators.pyx
```diff
@@ -370,24 +513,97 @@ For a temperature of 0, the model resembles a unit-disk model in hyperbolic spac
...
 	def setTheoreticalSplit(self, theoreticalSplit):
+		"""
+		setTheoreticalSplit(theoreticalSplit)
+
+		When using a theoretically optimal split, the quadtree will be flatter, 
+		but running time usually longer.
+
+		Parameters
+		----------
+    	theoreticalSplit : bool
```
angriman, line 551: Indentation looks wrong here.

networkit/generators.pyx
```diff
@@ -743,23 +1008,56 @@ cdef class LFRGenerator(Algorithm):
...
+		Parameters
+		----------
+		path : str
+			String containing a path.
+		"""
 		cls.paths["refImplDir"] = path
 
 
 	@classmethod
 	def fit(cls, Graph G, scale=1, vanilla=False, communityDetectionAlgorithm=PLM, plfit=False):
```
angriman, line 1038: It's a bit weird to have `cls` instead of `self`. Should we change it? Maybe on a separate PR.

fabratu: I also wondered - maybe indeed split that into another PR.

networkit/gephi/pyclient.py
```diff
@@ -33,85 +33,220 @@
...
+		"""
+		self.unflushedDumps = self.unflushedDumps + 1
+		if self.unflushedDumps > self.autoflush:
+			self.flush()
+			self.unflushedDumps = 0
+	def _send(self, data):
+		print('passing')
+		pass
+
+	def add_node(self, id, flush=True, **attributes):
```
angriman, line 91: These function names could also be changed to camelCase

fabratu: Right, should I deprecate the previous ones (since it is only a helper class)?

angriman: Yes, but better to do that in a separate PR

angriman (approved):

PR #890, opened 2022-02-21
Documentation fix for Python-modules: helpers, independentset, linkprediction, …

fabratu:
... matching, nxadapter, partitioning, plot, profiling.

Follow up from #870.


networkit/helpers.pyx
```diff
@@ -41,18 +53,60 @@ def ranked(sample):
 	return result
 
 def sorted(sample):
```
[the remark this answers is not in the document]
fabratu: I think the idea is to use the sort-functionality from C++ (`auxiliary/Parallel.hpp`). Haven't checked that, but it should be faster in most cases?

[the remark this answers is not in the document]
angriman: This uses `std::sort` on a copy of the input array, and I wouldn't be so sure that this implementation is faster than using `sorted()`. I think we don't need this function because `sorted()` in Python is already pretty fast, and we don't have benchmark data that shows the benefit of this additional function.

networkit/independentset.pyx
```diff
@@ -6,31 +6,84 @@ from libcpp.string cimport string
...
+	Luby's parallel maximal independent set algorithm.
+	"""
+
+	def __cinit__(self, *args):
+		self._this = new _Luby()
+
+	def __dealloc__(self):
+		if self._this is not NULL:
+			del self._this
+			self._this = NULL
```
angriman, line 89: This probably leads to segfault due to a double free (`__dealloc__` is already implemented in `IndependentSetFinder`). I would recommend writing a simple Python test for this class.

fabratu: I think the double free is somehow avoided, at least it doesn't happen locally. I removed it nevertheless and added a small test. Should we do the same for linkprediction?

fabratu: (More a general question/remark ... since only a few modules are covered by Python tests).

angriman: I think that we should have simple tests for the Python frontend. However, this should be done in separate PRs.

networkit/test/test_independentset.py
```diff
@@ -0,0 +1,34 @@
...
+		G.addEdge(0, 2)
+		G.addEdge(1, 2)
+		G.addEdge(2, 3)
+
+		luby = nk.independentset.Luby()
+		res = luby.run(G)
+		count = 0
+		for id in res:
+			if id == True:
+				count = count +1
```
angriman, line 18: A simpler version could be:
```suggestion
		count = sum(res)
```

networkit/test/test_independentset.py
```diff
@@ -0,0 +1,34 @@
...
+		self.assertGreaterEqual(count, 1)
+		self.assertLessEqual(count, 2)
+
+		G.addEdge(0, 3)
+		G.addEdge(1, 3)
+		res = luby.run(G)
+		count = 0
+		for id in res:
+			if id == True:
+				count = count +1
```
angriman, line 29: Same as above.

fabratu: Thanks - good catch.

angriman (approved):

PR #891, opened 2022-02-23
Add new module vizbridges: Create visualization widgets in 2D and 3D with Cytoscape and Plotly.

fabratu:
This PR adds the following:

- Deprecation of `csbridge.py` in favor of a more generalized module for visualization widgets.
- New module does support creating 2D plots based on Cytoscape or 3D plots based on Plotly, whereas the calling convention is nearly the same as before for `csbridge.py`.
- Changing variables and function names to camelCase.
- A notebook with examples on how to use the new module + unit-tests for CI.

### DIFF
--- a/networkit/csbridge.py
+++ b/networkit/csbridge.py
@@ -1,6 +1,7 @@
 # local imports
 from .support import MissingDependencyError
 from .structures import Partition
+from .vizbridges import widgetFromGraph
 
 # external imports
 try:
@@ -21,11 +22,15 @@ def widget_from_graph(G, node_scores = None, node_partition = None, node_palette
 	""" 
 	widget_from_graph(G, node_scores = None, node_partition = None, node_palette = None, show_ids = True)
 
+	Note
+	----
+	DEPRECATED. Use networkit.vizbridges.widgetFromGraph() instead.
+
 	Creates a ipycytoscape-widget from a given graph. The returned widget already contains
 	all nodes and edges from the graph. The graph is colored using an array of norm. rgb-values
 	based on seaborn perceptually uniform color map (rocket) or a user given custom color array. 
 	See matplotlib color maps for the correct formatting: 
-	https://matplotlib.org/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap 
+	https://matplotlib.org/api/_as_gen/matplotlib.colors.Colormap.html#matplotlib.colors.Colormap
 
 	Parameters
 	----------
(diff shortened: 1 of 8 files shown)


networkit/csbridge.py
```diff
@@ -21,132 +22,9 @@ def widget_from_graph(G, node_scores = None, node_partition = None, node_palette
...
-		e = ipycytoscape.Edge(data={"source": u, "target": v, "classes": edge_class })
-		edges.append(e)
-
-	# It is much faster to add edges and nodes in bulk.
-	cytoWidget.graph.add_nodes(nodes)
-	cytoWidget.graph.add_edges(edges, G.isDirected())
 
-	cytoWidget.set_style(s)
-	cytoWidget.set_layout(name='cose')
-	return cytoWidget
```
angriman, line 152: I recommend adding a deprecation warning here, users will probably not read the code/docs.

networkit/vizbridges.py
```diff
@@ -0,0 +1,282 @@
...
+				'css': {
+					'background-color': 'data(color)',
+					'content': 'data(id)'
+				}}]
+	else:
+		s = [{
+			'selector': 'node',
+				'css': {
+					'background-color': 'data(color)'
+				}}]
```
angriman, line 154: Indentation looks strange here

fabratu:
Even though CI completes, there are errors on the notebook check. I assume there is something wrong in the setup, will need to check (both on the error and why it is ignored).

fabratu:
Fixed the error for CI in general, however this needs the changes from #877 order to function properly (for 3D).

angriman (approved):

fabratu:
Based on feedback for the widget, I added an option for forcing Plotly also for 2D graphs. In addition helper code for computation of color-scales are moved out of the main function.

networkit/vizbridges.py
```diff
@@ -0,0 +1,335 @@
...
+			e = ipycytoscape.Edge(data={"source": u, "target": v, "classes": edgeClass })
+			edges.append(e)
+
+		# It is much faster to add edges and nodes in bulk.
+		graphWidget.graph.add_nodes(nodes)
+		graphWidget.graph.add_edges(edges, G.isDirected())
+
+		# Set layout
+		graphWidget.set_style(s)
+		graphWidget.set_layout(name='cose')
```
angriman, line 201: Wouldn't be better to allow the user to choose the layout?

fabratu: Yeah I thought about that ... however the number of parameters are already long. Since the returned widget supports the same functions, the user can always change the layout (and other functions for both plugins) afterwards.

angriman: Ah I see

angriman (approved):

PR #894, opened 2022-02-24
Documentation fix for Python-modules: randomization to workflows

fabratu:
Follow up of #870.

### DIFF
--- a/networkit/sampling.py
+++ b/networkit/sampling.py
@@ -3,25 +3,43 @@
 __author__ = "Elisabetta Bergamini"
 
 def bfsSample(G, source=None, k = 50):
-    """ Start a BFS from source node, return node-induced subgraph of the first k nodes discovered"""
-    if not source:
-        source = nk.graphtools.randomNode(G)
-    n = G.numberOfNodes()
-    visited = [False]*n
-    Q = [source]
-    closest = set([source])
-    global found
-    found = 0
-    while len(Q) > 0 and found < k:
-        u = Q.pop(0)
-        def enqueue(u,v,weight, eid):
-            global found
-            if not visited[v] and found < k:
-                found += 1
-                visited[v] = True
-                Q.append(v)
-                closest.add(v)
-        G.forEdgesOf(u, enqueue)
-    print("found {0} nodes".format(len(closest)))
-    G1 = nk.graphtools.subgraphFromNodes(G, closest)
-    return G1
+	""" 
+	bfsSample(G, source=None, k=50)    
+
+	Start a BFS from source node, return node-induced subgraph of the first k nodes discovered.
+
+	Parameters
+	----------
+	G : networkit.Graph
+		The input graph.
+	source : int, optional
+		The starting node. If none is choosen, then a random node is picked. Default: None
+	k : int, optional
+		Stop after k nodes are discovered. Default: 50
+
+	Returns
+	-------
+	networkit.Graph
+		Subgraph based on bfsSample.
+	"""
+	if not source:
+		source = nk.graphtools.randomNode(G)
+	n = G.numberOfNodes()
+	visited = [False]*n
+	Q = [source]
+	closest = set([source])
+	global found
+	found = 0
+	while len(Q) > 0 and found < k:
+		u = Q.pop(0)
+		def enqueue(u,v,weight, eid):
+			global found
+			if not visited[v] and found < k:
+				found += 1
+				visited[v] = True
+				Q.append(v)
+				closest.add(v)
+		G.forEdgesOf(u, enqueue)
+	print("found {0} nodes".format(len(closest)))
+	G1 = nk.graphtools.subgraphFromNodes(G, closest)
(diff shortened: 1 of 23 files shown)


networkit/sampling.py
```diff
@@ -3,25 +3,43 @@
 __author__ = "Elisabetta Bergamini"
 
 def bfsSample(G, source=None, k = 50):
```
angriman, line 5: It's strange to have this implemented in Python. In a separate PR we should implement this in C++ and move it to a dedicated module.

networkit/simulation.pyx
```diff
@@ -31,26 +31,41 @@ cdef extern from "<networkit/simulation/EpidemicSimulationSEIR.hpp>":
...
-	transP : double
-		transmission probability
-	eTime : count
-		exposed time
-	iTime : count
-		infectious time
-	zero : node
-		starting node
+ 	tMax : int
+ 		Max. number of timesteps.
```
angriman, line 43:
```suggestion
 		Maximum number of timesteps.
```

networkit/sparsification.pyx
```diff
@@ -46,6 +46,23 @@ cdef class EdgeScore(Algorithm):
...
+		score(u, v)
+
+		Get a vector containing the score for a certain edge connecting node u and v.
+
+		Parameters
+		----------
+		u : int
+			The first node.
+		v : int
+			The second node.
```
angriman, line 59: Aren't `optional` and `default: None` missing here?

networkit/sparsification.pyx
```diff
@@ -46,6 +46,23 @@ cdef class EdgeScore(Algorithm):
 		self._G = None # just to be sure the graph is deleted
 
 	def score(self, u, v = None):
+		"""
+		score(u, v)
```
angriman, line 50:
```suggestion
		score(u, v = None)
```

networkit/stats.pyx
```diff
@@ -16,7 +16,19 @@ cdef extern from "<networkit/auxiliary/Parallel.hpp>" namespace "Aux::Parallel":
 
 def gini(values):
```
angriman, line 17: Not sure if we need this feature at all, it's probably already implemented in other external packages. I think we should deprecate it.

networkit/stopwatch.py
```diff
@@ -8,7 +8,8 @@
 
 import time
 
-"""stopwatch is a very simple Python module for measuring time.
```
angriman, line 11: Same here, do we need NetworKit to implement a Python stopwatch? I think this is another feature we should deprecate.

networkit/viz.pyx
```diff
@@ -60,13 +60,29 @@ cdef class GraphLayoutAlgorithm:
...
 
 	def getCoordinates(self):
-		""" Computes approximation (in parallel) of the Spanning Edge Centrality. """
+		""" 
+		getCoordinates()
+		
+		Computes approximation (in parallel) of the Spanning Edge Centrality.
+		
+		Returns
+		-------
```
angriman, line 84: Return parameters are missing

networkit/viztasks.py
```diff
@@ -14,11 +14,39 @@
 	have_nx = True
 
 def save(name, dir="."):
```
angriman, line 16: This should be deprecated, it's out of NetworKit's scope

networkit/viztasks.py
```diff
@@ -30,8 +58,18 @@ def coloringToColorList(G, coloring):
 
 
 def drawGraph(G, **kwargs):
-	""" Draws a graph via networkX. Passes additional arguments beyond the graph to networkx.draw(...).
```
angriman, line 33: Since we have our own cytoscape-based tool, we don't need to rely on networkx' graph drawing features anymore. This can be deprecated as well, I think.

networkit/viztasks.py
```diff
@@ -44,8 +82,20 @@ def drawGraph(G, **kwargs):
 	networkx.draw(nxG, **kwargs)
 
 def drawCommunityGraph(G, zeta, **kwargs):
```
angriman, line 84: Same here

networkit/workflows.py
```diff
@@ -14,16 +14,18 @@
 
 def extractLargestComponent(G):
```
angriman, line 15: These functions have a specific use case and I think are out of NetworKit's scope. We should deprecate them.
`extractLargestConnectedComponent` can be removed since it was deprecated more than two releases cycles ago.

fabratu:
Thanks for the review. I pushed an update for all doc-related issues. I think the rest should go into another PR. All comments are valid and we should aim for a more consistent, which includes removing redundant functions, variable naming and deprecation of certain functions.

angriman:
Yes, I agree. This PR should stick with documentation fixes, deprecations and removals should be addressed in a separate PR.

fabratu:
Will merge this earlier since it only changes documentation strings.

PR #895, opened 2022-02-27
Refactor: Untested Code

angriman:
- Rename tests so that they are not filtered out by our GTest filter (see the `Unittests-X.cpp` file).
- Add missing tests for matrices iterators
- Add missing tests for Vector
- Removal of unused default constructor and assignment operators. I am not so sure about this; from the covered lines on coveralls, these functions seem to never be used. I would appreciate any feedback about this.


fabratu (approved):
LGTM - not 100% about the deletion of explicit constructors/destructors for `DynamicMatrix` and `Vector` though. If I did not miss anything from C++ doc, these default definitions in these cases should be generated anyways.

PR #896, opened 2022-02-28
Return by const ref when possible

angriman:
Return by const ref instead of by value when possible.


avdgrinten:
All these functions return const refs to member variables, right?

angriman:
Yes

fabratu:
The Cython interface should also be changed to reflect the new return type, right? Maybe this has also implications for `cdef` definitions.

angriman:
I am not sure about that, in other places we return by copy as well (e.g. `scores()` in `Centrality`). Don't we need to convert the C++ vector to a Python list anyway?

angriman:
Returning by reference in Cython seems to work, so I changed it accordingly. Topological sort is the only exception; the C++ implementation has its own class in the `graph` module but, in Python, we implement it as a _static_ function within `graphtools`. Since it is a static function, we cannot return the result by reference. Why do we have this inconsistency?

avdgrinten:
What happens on the Cython side if we return a const ref? Does Python still do a copy when converting to a list? If not, this would lead to a lifetime issue, wouldn't it?

angriman:
I think that Python copies when converting to a list. When printing the addresses of the C++ vector and of the Python list we get two different addresses:
```
...
>>> d = apsp.getDistances()
Address: 0x7f55380029f8 <- This is the address of the C++ distance vector
>>> hex(id(d))
'0x7f5585ec2440'
```

fabratu:
> Returning by reference in Cython seems to work, so I changed it accordingly. Topological sort is the only exception; the C++ implementation has its own class in the `graph` module but, in Python, we implement it as a _static_ function within `graphtools`. Since it is a static function, we cannot return the result by reference. Why do we have this inconsistency?

Hm, good question Maybe it is a better idea to put this in `graph.pyx` as the other graph-classes. If you see fit, than you could add this to the PR. Otherwise I would suggest handle the inconsistency in another PR.

angriman:
I would recommend fixing `TopologicalSort` in a separate PR.

networkit/distance.pyx
```diff
@@ -95,7 +95,7 @@ cdef extern from "<networkit/distance/SSSP.hpp>":
 
 	cdef cppclass _SSSP "NetworKit::SSSP"(_Algorithm):
 		_SSSP(_Graph G, node source, bool_t storePaths, bool_t storeNodesSortedByDistance, node target) except +
-		vector[edgeweight] getDistances() except +
+		vector[edgeweight] &getDistances() except +
```
fabratu, line 98: While at it, it would be a good idea to also change `SPSP`.

fabratu (changes requested):
I think `PLM` and `PLP` both have a inspector `getTiming`-method with internal member variable, which could also benefit from const ref treatment.

PR #897, opened 2022-02-28
Refactor: components

angriman:
Classes computing connected components can be simplified by using inherited functions from `ComponentDecomposition`. This also allows to clean up a lot of duplicate Python code.


avdgrinten (commented):
Why are the functions made virtual? Can't the derived classes use the existing infrastructure in the base class?

In general, it is quite ugly if inheritance is used to change the behavior of the base class.

angriman:
With the exception of `getPartition()` (not virtual anymore), every derived class implements these functions differently, some of them do not even use the `component` member variable to partition vertices in components. To do something cleaner as you say we need to move some member variables to the base class and implement these functions there.

angriman:
I think that this should be done as well. I'll convert this PR to a draft and re-open it once ready.

angriman:
Now all derived classes use the `Partition` object as the main data structure so we don't need virtual functions.
Since the `Dyn(Weakly)ConnectedComponents` classes were very messy, I also simplified them, in particular:
- Component ids are managed with the `Partition` object and not using a queue
- The edge ids map uses the `Edge` object instead of `std::pair<node, node>`
- Removal of unnecessary helper functions and member variables.

The implementation of the two classes can now be merged in a single one using pImpl (as we did for `(Weakly)ConnectedComponents`. But this is something for a separate PR.

angriman:
While writing some more extensive tests for `Dyn(Weakly)ConnectedComponents` I also spotted a bug (in `reverseBFS`, the `level` variable does not represent the distance from `v`), so I thought that rewriting those two classes was a better idea.
Now, these two classes use the same (and much simpler) private implementation in `DynConnectedComponentsImpl`, exactly as we are already doing for `(Weakly)ConnectedComponents`. Further changes:
- more extensive tests that fail with the original implementation
- deprecate `runSequential()` in `ParallelConnectedComponents`, which is already implemented in `ConnectedComponents`.

angriman:
`ComponentDecomposition` holds the data structure and implements APIs that are shared by all algorithms (e.g., `numberOfComponents`, `componentOfNode`, etc). The classes now implement the main algorithm only. For `BiconnectedComponents` this is not possible because a node can be in two biconnected components at the same time.

`Dyn(Weakly)ConnectedComponents` had two separate implementations. Since the algorithm is the same, we now have a single private implementation in `DynConnectedComponentsImpl` which implements dynamic algorithms for both connected and weakly connected components. Note that the private implementation holds a pointer to the `Partition` object.

avdgrinten:
Why are there .cpp files for the `...Impl` classes? They are templates anyway, they can be implemented in the `...Impl.hpp` files.

angriman:
It's just to separate interface and implementation. The template parameter is bool so we can instantiate it explicitly into the `.cpp` file.

avdgrinten:
Sure, but I don't really see how that is an advantage as the entire interface is internal anyway (and the cost is that one has to browse through two files to understand the implementation).

angriman:
Makes sense, I moved the implementation to the header file.

include/networkit/components/BiconnectedComponents.hpp
```diff
@@ -44,22 +44,15 @@ class BiconnectedComponents final : public Algorithm {
...
      *	@return The number of biconnected components.
      */
-    count numberOfComponents();
-
-    /*
-     * Get the size of each component.
-     *
-     * @return Map from component index to size.
-     */
-    std::map<count, count> getComponentSizes();
```
avdgrinten, line 54: Why is this function removed? That's an API break.

avdgrinten: Okay, I guess it's added back in a later commit.

avdgrinten (approved):
LGTM.

PR #898, opened 2022-03-10
Python: Marks current node attributes API as experimental.

fabratu:
Based on a discussion with @avdgrinten - since it not clear yet how to support arbitrary node attributes, maybe it is a good idea to mark the Python-API for node attributes as experimental. This leaves more flexibility in term of changing the API without the need for deprecation.


avdgrinten (approved):
LGTM. I guess this also applies to the C++ side. We should also mention this in the release notes.

PR #907, opened 2022-03-22
Fix Unreachable ttmath Library

angriman:
The `ttmath` library [1] is unreachable at the moment. For better reliability, I updated the submodule to a fork of ttmath within the NetworKit organization.

[1] https://gitea.ttmath.org/tomasz.sowa/ttmath


fabratu:
Merging this early because otherwise all other CI-tests fail.

PR #911, opened 2022-03-25
Extend SPSP

angriman:
This PR provides a feature requested in our mailing list, i.e., to extend SPSP so that the user can specify target nodes. In particular:
- SPSP has now another constructor and a function to set target nodes
- In the case of target nodes, SPSP uses new `MultiTargetBFS/Dijkstra` algorithms that stop the exploration once all target nodes have been visited -- I preferred not implementing this within `BFS` and `Dijkstra` because `SSSP` would then implement too many functions relative to different features and I find this confusing for the users.

Additional changes:
- Member variables that store the distances were moved from derived classes to the base class `STSP` so that we can "un-virtualize" `STSP` functions.
- Parametrization of the tests of the `distance` module and fix of bugged tests.
- Python tests for the `distance` module are moved to another file.
- Minor fixes for the documentation of `BFS`.


networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -160,41 +178,31 @@ TEST_F(DistanceGTest, testBidirectionalBFS) {
...
-    Graph gen = ErdosRenyiGenerator(n, 0.05, false).generate();
-    Graph gen1 = ErdosRenyiGenerator(n, 0.05, true).generate();
-    gen.forEdges([&](node u, node v) {
-        G.setWeight(u, v, Aux::Random::probability());
-    });
-    gen1.forEdges([&](node u, node v) {
-        G1.setWeight(u, v, Aux::Random::probability());
-    });
+    auto G = generateERGraph(500, isDirected() ? 0.05 : 0.02);
+    const auto infDist = std::numeric_limits<edgeweight>::max();
```
fabratu, line 203: Only minor: the value of `infDist` is used for several tests. Maybe it is a good idea to change it to be accessible by the complete class.

fabratu (changes requested):
Hm, a general comment (likely outside the scope of this PR): I think the way the shortest path algorithms are organized can be improved. `SSSP` and `STSP` are abstract classes and cannot be used directly, but `SPSP` and `APSP` can. From a user perspective this can lead to confusion, especially if one is not familiar with inheritance and abstract classes. 

Another thing is the naming scheme with respect to the current feature set. From the perspective of a inexperienced user, it is not so clear what interface to use. Do I use `SPSP` or a child of `STSP`; what exactly differentiates `SSSP` and `STSP` (due to the same interface)?

If possible, one way around it would be to hide the abstract base classes in Python interface (can look into it) or document properly what subclasses exist (for example when spawning a `SSSP` and `STSP` object).

networkit/cpp/distance/MultiTargetBFS.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    auto visitNode = [&](node u) -> bool {
+        if (visited[u]) // Already visited, nothing to do
+            return false;
+
+        // Check if we are visiting a target node
+        const auto it = targetsSet.find(u);
+        if (it != targetsSet.end()) {
+            targetIdx.emplace(u, distances.size());
+            distances.push_back(static_cast<edgeweight>(level));
```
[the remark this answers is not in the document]
angriman: I agree, but this might not be possible in C++ (e.g., when the targets are stored in an `unordered_set`).

include/networkit/distance/AStarGeneral.hpp
```diff
@@ -67,13 +67,14 @@ class AStarGeneral final : public STSP {
         do {
             top = heap.extract_top();
             if (top == target) {
+                distance = distFromSource[target];
```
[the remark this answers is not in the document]
angriman: Good catch, now `distance` is set to infinity by default at the beginning of the algorithm.

angriman:
Thanks for the review. I agree that distance algorithms need to be restructured and simplified; since this PR brings a new feature, we should do that separately. I addressed all the comments.

networkit/distance.pyx
```diff
@@ -32,9 +32,13 @@ cdef extern from "cython_helper.h":
 cdef extern from "<networkit/distance/STSP.hpp>":
 	cdef cppclass _STSP "NetworKit::STSP"(_Algorithm):
 		_STSP(_Graph G, node source, node target, bool_t storePred) except +
```
fabratu, line 34: This doesn't seem to be necessary, but I would suggest to also add the second constructor for `STSP`.

networkit/cpp/distance/MultiTargetBFS.cpp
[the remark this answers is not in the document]
fabratu: Right, I wasn't aware of `unordered_set`. In that case I would suggest to document the behavior for C++ (the need to iterate through `getTargetIndexMap()` to get correct target value). For the Python interface I think it is a better solution to return a correctly indexed distance list (since Cython mapping is `list` <-> `vector`).

networkit/cpp/distance/BidirectionalBFS.cpp
```diff
@@ -15,7 +15,7 @@ void BidirectionalBFS::run() {
     if (G->isWeighted())
         WARN("Treating the graph as unweighted!");
 
-    stDist = 0;
+    distance = 0;
```
fabratu, line 18: If BidirectionalBFS is constructed using the new iterator-based target-list constructor, the algorithm would crash (due to `target=none`)?

include/networkit/distance/STSP.hpp
```diff
@@ -69,17 +90,79 @@ class STSP : public Algorithm {
...
+
+    /**
+     * Sets the target nodes.
+     *
+     * @param targetsFirst,targetsLast Range of target nodes.
+     */
+    template <class InputIt>
+    void setTargets(InputIt targetsFirst, InputIt targetsLast) {
+        assert(std::all_of(targetsFirst, targetsLast, [&](auto u) { return G->hasNode(u); }));
+        targets.assign(targetsFirst, targetsLast);
```
fabratu, line 141: Should `target` be set back to `none`? Or otherwise should we drop variable `target` and use `targets`-vector in all cases (with implied changes for BidirectionalBFS/Dijkstra)?

networkit/cpp/distance/BidirectionalDijkstra.cpp
```diff
@@ -15,7 +15,7 @@ void BidirectionalDijkstra::run() {
     if (!G->isWeighted())
         WARN("The graph is unweighted, use BidirectionalBFS for better "
              "efficiency!");
-    stDist = 0.;
+    distance = 0.;
```
fabratu, line 18: Same as for `BidirectionalBFS`.

fabratu (changes requested):

include/networkit/distance/STSP.hpp
angriman: I think that none of the possibilities is satisfactory. It's weird for a class dealing with a single target to have it stored in a vector, while `target` is never used by `MultiTargetBFS/Dijkstra` -- so there is no point in setting it to `none`. I think that this issue can be solved by restructuring these algorithms.

networkit/cpp/distance/BidirectionalBFS.cpp
angriman: Good point, now both `BidirectionalBFS/Dijkstra` have their own constructor that takes a single target node.

networkit/distance.pyx
angriman: Unfortunately, it does not seem to be possible to have a template constructor in Cython. Let me know if I am wrong.

networkit/cpp/distance/MultiTargetBFS.cpp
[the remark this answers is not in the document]
angriman: This is now implemented by keeping a copy of the targets list, which is used to sort the distances.

networkit/distance.pyx
fabratu: Hm, I was just thinking to have the same as for `SPSP` (no general template version): 

```
_STSP(_Graph G, node source, vector[node].iterator targetsFirst, vector[node].iterator targetsLast, bool_t) except +
```

But it's really only for consistency with the other base classes.

include/networkit/distance/STSP.hpp
fabratu: Agreed.

networkit/distance.pyx
angriman: Ah, I see. I added it.

fabratu (dismissed):
LGTM.

fabratu (approved):
LGTM.

angriman:
Merging now because this PR has been approved for more than 7 days before I force-pushed to resolve a trivial conflict.

PR #915, opened 2022-04-01
Introduce function to randomize weights in a graph

Relux-the-Relux:


angriman:
We could have a more generic function that takes a distribution:
```cpp
template <class Distribution>
void GraphTools:generateRandomWeights(Graph &G, Distribution distr = std::uniform_real_distribution<edgeweight>{0, 1}) {
    if (!G.isWeighted())
        G = GraphTools::toWeighted(G);
    G.parallelForEdges([&](node u, node v) { G.setWeight(u, v, distr(Aux::getURNG()); });
}
```

This also allows having a unique function instead of re-implementing it for each distribution we want.

This solution will probably not work in Python -- I don't see a `random.pyx` available in Cython. We could either provide a function that generates edge weights between 0 and 1 (i.e., we use the default uniform real distribution), or we write some wrapper function in `GraphTools` that allows the user to set the upper/lower bounds of the distribution.

include/networkit/graph/GraphTools.hpp
```diff
@@ -406,6 +406,24 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
 
+/**
+ * Given a  grah G, randomizes the weights in the graph. The weights are distributed in
+ * the range [0, 1] by default, unless a distribution is provided. This occurs in-place
+ * if the graph is already weighted, but creates a weighted copy before randomizing weights
+ * if it wasn't.
+ *
+ * @param   G           Directed input graph
+ * @param   distr       Random distribution
+ * @return              A graph with random edge weights
```
angriman, line 417: The function returns void, you can remove this line.

include/networkit/graph/GraphTools.hpp
```diff
@@ -406,6 +406,24 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+ */
+template <class Distribution = std::uniform_real_distribution<edgeweight>>
+void randomizeWeights(Graph &G, Distribution distr = std::uniform_real_distribution<edgeweight>{
+                                    0, std::nexttoward(1.0, 2.0)}) {
+    if (!G.isWeighted())
+        G = toWeighted(G);
+    G.forEdges([&](node u, node v) { G.setWeight(u, v, distr(Aux::Random::getURNG())); });
```
angriman, line 424: This can run in parallel, which is important for large graphs. However, the result will not be deterministic. To have deterministic results we should either use a fixed random seed for each edge or run the algorithm sequentially.

Relux-the-Relux: Does the result not being deterministic actually matter here? None of the previous uses it is substituting bothered making it sequential or setting a seed for each edge.

angriman: The previous uses were internal, this is now a public API so I think it matters.

Relux-the-Relux: It seems `forEdges` was inherently sequential. Made so that it will run parallel by default with the option to run sequentially.

avdgrinten: Having a sequential flag is not really a great API. Call it `deterministic` instead or so. I'd also prefer if it was the last parameter.

include/networkit/graph/GraphTools.hpp
```diff
@@ -406,6 +406,38 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+    Graph &G,
+    Distribution distr = std::uniform_real_distribution<edgeweight>{0, std::nexttoward(1.0, 2.0)},
+    bool deterministic = false) {
+    if (!G.isWeighted())
+        G = toWeighted(G);
+    if (!deterministic) {
+        G.parallelForEdges(
+            [&](node u, node v) { G.setWeight(u, v, distr(Aux::Random::getURNG())); });
+    } else {
+        G.forEdges([&](node u, node v) { G.setWeight(u, v, distr(Aux::Random::getURNG())); });
```
angriman, line 432: I think we can make this parallel by using a generator for each thread and by setting a deterministic random seed for each edge. A simple idea (I am not proposing to use this exact code) could be:
```cpp
G.parallelForEdges([&](node u, node v) {
   std::mt19937 gen;
   gen.seed(std::hash<node>()(u) ^ std::hash<node>()(v) ^ Aux::Random::getSeed());
   G.setWeight(u, v, distr(gen));
});
```

Relux-the-Relux: Are the edges I don't think we can set a generator for each thread considering the way `parallelForEdges` is implemented. Since getSeed isn't thread-dependant we can't use that, as which seed each thread gets is still non-deterministic.

So would need to be something like just ` gen.seed(std::hash<node>()(u) ^ std::hash<node>()(v));`, since URNG says it is thread-local, but it actually isn't (the `localSeedGeneration` is never used, it is just set to the global seed and `getSeed` just gets the global seed), so making any calls to NetworKit's seeds would make it non-deterministic.

Relux-the-Relux:
Maybe if we added a parallelForEdges that divides the chunks deterministically among threads.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +418,42 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  */
 std::vector<node> topologicalSort(const Graph &G);
 
+/**
+ * Given a  grah G, randomizes the weights in the graph. The weights are distributed in
```
rionda, line 422: I would suggest adding "uniformly" after "distributed".

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1033,12 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
     }
 }
+
+TEST_F(GraphToolsGTest, testEdgesRandomizer) {
+    Aux::Random::setSeed(1, true);
+    DorogovtsevMendesGenerator generator(1000);
+    Graph G1 = generator.generate();
+    Graph G(G1, true, false);
+    GraphTools::randomizeWeights(G);
+}
```
angriman, line 1043: Once the deterministic version works in parallel, we should also provide meaningful tests. Here we are only checking that the function runs, not that it actually works.

include/networkit/graph/GraphTools.hpp
avdgrinten: Well, we can just use `hash(u) * someLargePrime + hash(v)` or something similar.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +418,27 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+ */
+template <class Distribution = std::uniform_real_distribution<edgeweight>>
+void randomizeWeights(Graph &G, Distribution distr = std::uniform_real_distribution<edgeweight>{
+                                    0, std::nexttoward(1.0, 2.0)}) {
+    if (!G.isWeighted())
+        G = toWeighted(G);
+    std::vector<std::mt19937> gens(omp_get_max_threads());
+    for (std::mt19937 gen : gens) {
+        gen.seed(Aux::Random::getSeed());
+    }
```
angriman, line 437: Shouldn't you use a reference here?
```suggestion
    for (auto &gen : gens) {
        gen.seed(Aux::Random::getSeed());
    }
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,53 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+#pragma omp parallel
+    {
+        std::mt19937 gen;
+        auto baseSeed = Aux::Random::integer();
+
+#pragma omp for schedule(dynamic)
+        for (omp_index u = 0; u < G.upperNodeIdBound(); u++) {
+            gen.seed(std::hash<node>()(u) + 0x9e3779b9 + (baseSeed << 6) + (baseSeed >> 2));
+            for (index j = 0; j < G.degree(u); j++) {
+                node v = G.getIthNeighbor(u, j);
```
angriman, line 444: This is quadratic in the number of neighbors. Can't we use `G.forOutEdgesOf` and keep a separate index?

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,53 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+                if (G.isDirected() || u <= v) {
```
angriman, line 445: I'd recommend to avoid too many indentations. Use:
```suggestion
                if (!G.isDirected() && u > v)
                    continue;
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,53 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+                    edgeweight ew = distr(gen);
+                    G.setWeightAtIthNeighbor(Unsafe{}, u, j, ew);
+                    if (G.isDirected()) {
+                        for (index k = 0; k < G.degreeIn(v); k++) {
+                            if (G.getIthInNeighbor(v, k) == u) {
```
angriman, line 450: As above, this is quadratic in the number of in-neighbors. We could use `G.forInEdgesOf` and keep a separate index `k`.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,53 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+                                G.setWeightAtIthInNeighbor(Unsafe{}, v, k, ew);
+                                break;
+                            }
+                        }
+                    } else if (u != v) {
+                        for (index k = 0; k < G.degreeOut(v); k++) {
+                            if (G.getIthNeighbor(v, k) == u) {
```
angriman, line 457: Ditto.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+    edgeweight average = 0.0;
+    edgeweight squareDiffSum = 0.0;
+    G.forEdges([&](node, node, edgeweight w) {
+        edgesChecked++;
+        edgeweight diff = (w - average);
+        average += diff / edgesChecked;
+        squareDiffSum += diff * (w - average);
+    });
+    EXPECT_NEAR(1.0 / 12.0, squareDiffSum / edgesChecked,
+                1e-2); // variance for the uniform distributions
```
angriman, line 1056:
```suggestion
     // variance for the uniform distributions
    EXPECT_NEAR(1.0 / 12.0, squareDiffSum / edgesChecked, 1e-2);
```

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
         EXPECT_TRUE(G.checkConsistency());
     }
 }
+
+TEST_P(GraphToolsGTest, testEdgesRandomizer) {
+    if (!weighted())
+        return;
+    Aux::Random::setSeed(1, true);
+    ErdosRenyiGenerator generator(3000, 0.33, directed());
+    Graph G1 = generator.generate();
```
angriman, line 1041:
```suggestion
    auto G1 = ErdosRenyiGenerator{3000, 0.33, directed()}.generate();
```

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+    Aux::Random::setSeed(1, true);
```
angriman, line 1042: Why do we have to set the seed twice?

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+    Graph G(G1, true, directed());
+    GraphTools::randomizeWeights(G);
+
+    count edgesChecked = 0;
```
angriman, line 1046: Is this needed? Why wouldn't `G.numberOfEdges()` work?

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+    edgeweight average = 0.0;
+    edgeweight squareDiffSum = 0.0;
+    G.forEdges([&](node, node, edgeweight w) {
+        edgesChecked++;
+        edgeweight diff = (w - average);
+        average += diff / edgesChecked;
```
angriman, line 1052: This does not seem an average. Could you add some explanation about what are you computing here?

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1039,4 +1032,35 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+        squareDiffSum += diff * (w - average);
```
angriman, line 1053: This does not seem to be right, `average` is not the actual average.

networkit/cpp/graph/test/GraphToolsGTest.cpp
angriman: Could you add some explanation about why do we expect this result?

networkit/cpp/graph/test/GraphToolsGTest.cpp
Relux-the-Relux: it is th online algorithm for the average, it is indeed the average. `avg = avg + (newWeight - avg)/samples`;

networkit/cpp/graph/test/GraphToolsGTest.cpp
Relux-the-Relux: Probably not necessary.

networkit/cpp/graph/test/GraphToolsGTest.cpp
hmeyerhenke: Why is an online algorithm necessary? Looks like this leads to more arithmetic operations than a simple sum and one final division.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,59 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+            index k;
+            edgeweight ew;
+            G.forEdgesOf(u, [&](node v) {
+                if (!G.isDirected() && u > v) {
+                    j++;
+                    return;
+                }
+                ew = distr(gen);
+                G.setWeightAtIthNeighbor(Unsafe{}, u, j, ew);
+                j++;
```
angriman, line 453: If you don't need to read the variable before it is incremented, use pre-increment.
```suggestion
                ++j;
```
https://www.youtube.com/watch?v=ObVRSNvGitE

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,59 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+    if (!G.isWeighted())
+        G = toWeighted(G);
+#pragma omp parallel
+    {
+        std::mt19937 gen;
+        auto baseSeed = Aux::Random::integer();
+
+#pragma omp for schedule(dynamic)
+        for (omp_index u = 0; u < G.upperNodeIdBound(); u++) {
+            gen.seed(std::hash<node>()(u) + 0x9e3779b9 + (baseSeed << 6) + (baseSeed >> 2));
```
fabratu, line 442: Leave a comment what is done here (and why).

networkit/graphtools.pyx
```diff
@@ -705,3 +706,19 @@ cdef class GraphTools:
...
 		result = createAugmentedGraph(G._this)
 		return Graph().setThis(result.first), result.second
+
+	@staticmethod
+	def randomizeWeights(Graph G):
+		"""
+		randomizeWeights(G)
+
+		Given a graph G, it randomizes the weights in the graph. The weights are distributed in
+		the range [0, 1]. This occurs in-place if the graph is already weighted,
```
fabratu, line 716: Same as for C++ doc string.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,59 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  */
 std::vector<node> topologicalSort(const Graph &G);
 
+/**
+ * Given a  grah G, randomizes the weights in the graph. The weights are distributed in
```
angriman, line 423:
```suggestion
 * Randomizes the weights of the given graph. The weights are uniformly distributed in
```

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,59 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+ * the range [0, 1] by default, unless a distribution is provided. This occurs in-place
+ * if the graph is already weighted, but creates a weighted copy before randomizing weights
+ * if it wasn't.
+ * @param   G           Directed input graph
+ * @param   distr       Random distribution
+ */
+template <class Distribution = std::uniform_real_distribution<edgeweight>>
+void randomizeWeights(Graph &G, Distribution distr = std::uniform_real_distribution<edgeweight>{
+                                    0, std::nexttoward(1.0, 2.0)}) {
```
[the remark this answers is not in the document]
Relux-the-Relux: It is so that 1.0 is still included, if we used 1.0 here instead it would be [0,1) and not [0,1].

include/networkit/graph/GraphTools.hpp
fabratu: Thanks for inserting comments - however the remark was about what the ingredients in the seed actual do. I think the combination of magic number and bit shift is taken from boost::combine_hash?

Also the `baseSeed` makes the result not deterministic if the number of threads is bigger >1, right?

fabratu (changes requested):

include/networkit/graph/GraphTools.hpp
Relux-the-Relux: > Also the baseSeed makes the result not deterministic if the number of threads is bigger >1, right?

I tested and the results are deterministic

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,63 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+                    ++j;
+                    return;
+                }
+                ew = distr(gen);
+                G.setWeightAtIthNeighbor(Unsafe{}, u, j, ew);
+                ++j;
+                k = 0;
+                // we need to set the other direction of the edge to the same weight
+                if (G.isDirected()) {
+                    G.forInEdgesOf(v, [&](node w) {
```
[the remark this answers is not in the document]
Relux-the-Relux: The problem is that we care about the node ids and not the edge ids to decided when to skip an edge.

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1047,4 +1040,43 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
         EXPECT_TRUE(G.checkConsistency());
     }
 }
+
+TEST_P(GraphToolsGTest, testEdgesRandomizer) {
+    if (!weighted())
+        return;
+    Aux::Random::setSeed(1, true);
+    ErdosRenyiGenerator generator(2000, 0.3, directed());
+    Graph G1 = generator.generate();
```
fabratu, line 1049: You can merge both lines to `auto G1 = ErdosRenyiGenerator{2000, 0.3, directed()}.generate();`

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -1047,4 +1040,43 @@ TEST_P(GraphToolsGTest, testEdgesSortedByWeight) {
...
+                           [&](node u, node v, edgeweight ew) { EXPECT_EQ(G.weight(v, u), ew); });
+        });
+}
+
+TEST_P(GraphToolsGTest, testEdgesRandomizerDeterminism) {
+    if (!weighted())
+        return;
+    Aux::Random::setSeed(1, true);
+    ErdosRenyiGenerator generator(2000, 0.3, directed());
+    Graph G1 = generator.generate();
```
fabratu, line 1073: As above

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,63 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+ */
+template <class Distribution = std::uniform_real_distribution<edgeweight>>
+void randomizeWeights(Graph &G, Distribution distr = std::uniform_real_distribution<edgeweight>{
+                                    0, std::nexttoward(1.0, 2.0)}) {
+    if (!G.isWeighted())
+        G = toWeighted(G);
+#pragma omp parallel
+    {
+        std::mt19937 gen;
+        auto baseSeed = Aux::Random::integer();
```
angriman, line 439: Doesn't this make the function non-deterministic? I suggest to use `Aux::Random::getSeed()`.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,63 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+
+#pragma omp for schedule(static)
+        for (omp_index u = 0; u < G.upperNodeIdBound(); u++) {
```
angriman, line 442: Why not `G.parallelForNodes()`? This would also fail if the graph has deleted nodes.

include/networkit/graph/GraphTools.hpp
```diff
@@ -418,6 +419,63 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
+            // we use the hash of the origin node so that the weights are independent
+            // of  processing order
+            gen.seed(std::hash<node>()(u) + 0x9e3779b9 + (baseSeed << 6) + (baseSeed >> 2));
+            index j = 0;
+            index k;
+            edgeweight ew;
```
angriman, line 448: You can move these two inside the loop, there is no advantage in declaring them here.

networkit/graphtools.pyx
```diff
@@ -705,3 +706,20 @@ cdef class GraphTools:
...
+		the range [0, 1] by default, unless a different distribution is provided. However it
+		is only strictly in-place for already weighted graphs. For unweighted graphs a copy is
+		created before randomizing weights.
+
+		Parameters
+		----------
+		G : networkit.Graph
+			The directed input graph.
+		"""
+		randomizeWeights(G._this)
```
angriman, line 725: Please add some additional checks, e.g., that the graph is weighted and the edge weights are within the expected values.

include/networkit/graph/GraphTools.hpp
Relux-the-Relux: No, we est fr non-derminism with testEdgesRandomizerDeterminism

include/networkit/graph/GraphTools.hpp
Relux-the-Relux: because it wouldn't be deterministic. Also, it doesn't fail with deleted nodes because `outEdges` of a node is not deleted when a node is removed but simply has its edges removed.

networkit/graphtools.pyx
Relux-the-Relux: Why is this necessary? If the graph is unweighted then the graph will be made to be weighted and the previous edge weights are ignored anyway.

include/networkit/graph/GraphTools.hpp
angriman: Still I don't understand what is the reason to use a random number as the base seed while we have `Aux::Random::getSeed()`.

include/networkit/graph/GraphTools.hpp
angriman: Alright, then I only a minor recommendation:
```suggestion
#pragma omp for schedule(static)
        for (omp_index u = 0; u < G.upperNodeIdBound(); ++u) {
```

networkit/graphtools.pyx
angriman: Ah sorry, I thought that this was a test.

include/networkit/graph/GraphTools.hpp
Relux-the-Relux: We are starting one random number generator per thread. If all threads used the same seed we will have lots of duplicate weights.

angriman: Good point. However, that's true only if (i) [`globalSeedGeneration`](https://github.com/networkit/networkit/blob/master/networkit/cpp/GlobalState.cpp#L9) is not 0 and if (ii) [`seedUseThreadId`](https://github.com/networkit/networkit/blob/master/networkit/cpp/GlobalState.cpp#L10) is false, see the [`getSeed` implementation](https://github.com/networkit/networkit/blob/master/networkit/cpp/auxiliary/Random.cpp#L26).

If (i) does not hold, then `getSeed()` behaves exactly as you did here. If (i) holds and (ii) does not, then each thread will be assigned a different seed. So my recommendation is to use `Aux::Random::integer()` only when both (i) and (ii) hold. Doing so, we avoid assigning the same seed to every thread, and the user can choose the base seed if they want.

Also, `baseSeed` can be `const`.
```cpp
const auto baseSeed = (GlobalState::getGlobalSeed() != 0 && !GlobalState::getSeedUseThreadId()) ?
  Aux::Random::integer() : Aux::Random::getSeed();
```

angriman: @Relux-the-Relux do you agree on this? I think that once this is done we can merge the PR.

fabratu: @Relux-the-Relux @angriman Hopefully that is ok - I agree with the suggested change and updated the PR accordingly. If the PR is ok now, we can integrate it in the release then.

angriman (approved):
Thanks, LGTM.

angriman:
@fabratu this cannot be merged without your approval.

PR #917, opened 2022-04-06
Add option to return distances as numpy arrays.

tillahoffmann:
Returning distances (or other lists of values) from `cython` can be a significant part of the computational cost in `networkit`. For example, consider the following snippet that gets all pairs shortest paths.

```python
def target(graph):
    apsp = nk.distance.APSP(graph).run()
    apsp.getDistances()
    
%lprun -f target [target(nk.generators.ErdosRenyiGenerator(500, 0.1).generate()) for _ in range(1000)]
```

About a third of the time is spent on *getting* distances rather than *evaluating* the distances.

```
Timer unit: 1e-06 s

Total time: 11.8565 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     1                                           def target(graph):
     2      1000    7833515.0   7833.5     66.1      apsp = nk.distance.APSP(graph).run()
     3      1000    4022974.0   4023.0     33.9      apsp.getDistances()
```

This PR adds an optional argument `asarray` to the `getDistances` such that a numpy array is returned if `asarray=True`.

```python
def target(graph):
    apsp = nk.distance.APSP(graph).run()
    apsp.getDistances(asarray=True)
    
%lprun -f target [target(nk.generators.ErdosRenyiGenerator(500, 0.1).generate()) for _ in range(1000)]
```

This speeds up returning values by a factor of 11 and the `target` function by a factor of about 1.4 (in this particular example).

```
Timer unit: 1e-06 s

Total time: 8.38477 s

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     1                                           def target(graph):
     2      1000    8041571.0   8041.6     95.9      apsp = nk.distance.APSP(graph).run()
     3      1000     343195.0    343.2      4.1      apsp.getDistances(asarray=True)
```

This PR ended up touching many more files than I anticipated because, for the fused types to work properly, all types had to "originate from the same source", i.e. we couldn't have multiple definitions of `ctypedef uint64_t index` throughout the code base but needed one `ctypedef` that was imported into other modules. This change should also help with maintainability, however.

Running more careful benchmarks (rather than just using the line profiler) indicates that these changes can give a 20x speed up compared with returning a list of lists of values and a >50x speed up compared with converting a list of lists of values to a numpy array for subsequent analysis. See [here](https://gist.github.com/tillahoffmann/ee02e08fdb1f628b39f35a61694481d7) for details.

~Apologies for the whitespace changes; my editor was a bit overeager.~

### DIFF
--- a/networkit/distance.pyx
+++ b/networkit/distance.pyx
@@ -178,17 +178,17 @@ cdef class SSSP(Algorithm):
 		if type(self) == SSSP:
 			raise RuntimeError("Error, you may not use SSSP directly, use a sub-class instead")
 
-	def getDistances(self, bint asarray=False):
+	def getDistances(self, asarray=None):
 		"""
-		getDistances(asarray=False)
+		getDistances(asarray=None)
 
 		Returns a list of weighted distances from the source node, i.e. the
  	 	length of the shortest path from the source node to any other node.
 
 		Parameters
 		----------
-		asarray : bool, optional
-			Return the result as a numpy array.
+		asarray : optional
+			Return the result as a numpy array. Default: Falsy.
 
  	 	Returns
  	 	-------
@@ -1182,16 +1182,16 @@ cdef class APSP(Algorithm):
 	def __dealloc__(self):
 		self._G = None
 
-	def getDistances(self, bint asarray=False):
+	def getDistances(self, asarray=None):
 		"""
-		getDistances(asarray=False)
+		getDistances(asarray=None)
 
 		Returns a vector of vectors of distances between each node pair.
 
 		Parameters
 		----------
-		asarray : bool, optional
-			Return the result as a numpy array.
+		asarray : optional
+			Return the result as a numpy array. Default: Falsy.
 
 		Returns
 		-------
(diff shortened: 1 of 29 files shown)


avdgrinten:
Thank you for the PR. This request is certainly reasonable and adding better numpy support would be desirable. I haven't had time to review everything yet. Unfortunately, there is a large number of whitespace changes. Could you either remove them from the PR or extract them into their own commit?

tillahoffmann:
Hey, if the whitespace changes are a bother from a review perspective, here are the changes without whitespace showing up: https://github.com/networkit/networkit/pull/917/files?w=1 But let me know if you'd prefer having separate commits for a cleaner commit history, and I'll have a look at how to untangle the changes.

tillahoffmann:
Here are the changes without the whitespace. Hope `git diff -b` didn't brake anything in the process.

networkit/distance.pyx
```diff
@@ -1253,23 +1259,34 @@ cdef class SPSP(Algorithm):
 	def __dealloc__(self):
 		self._G = None
 
-	def getDistances(self):
-		""" 
-		getDistances()
-		
+	def getDistances(self, bint asarray=False):
```
avdgrinten, line 1262: Let's use `None` as default here, that gives us the possibility to switch to NumPy arrays by default in the future.

avdgrinten: I'm not sure if `asarray` is really a good name here. Would `as_np` or `np_array` be better? Or will most users get what `asarray` is supposed to do anyway?

avdgrinten (commented):
LGTM except for one comment.

networkit/distance.pyx
tillahoffmann: Good point, will update to `None` as default. I borrowed [`asarray`](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html) because it's a commonly used function in numpy to ensure one gets an array. Open to naming suggestions though.

avdgrinten: Alright, then let's keep `asarray`. We can merge the PR as soon as the `None` default is updated.

avdgrinten (approved):
Alright, I don't like the "falsy" wording too much but that can be fixed independently. LGTM.

tillahoffmann:
We could also just stick to `False` and `True` and say "defaults to `False`". Probably, the user won't care about the internals?

PR #921, opened 2022-04-08
Fix #916

### LINKED ISSUE #916: PageRank example is not correct
I think the pageRank example in the notebook is not proper. It requires to specify "tol" like below:
nk.centrality.PageRank(K, damp = 0.85, tol = 1e-6)

Right now the example is like below (I hope here 1e-6 is supposed to be the value of "tol", but it is considering damp = 1e-6 actually):
pr = nk.centrality.PageRank(K, 1e-6)

angriman:
This PR fixes #916 by removing the tolerance parameter from the PageRank example notebook -- which is read instead as the damping factor.

### DIFF
--- a/notebooks/Centrality.ipynb
+++ b/notebooks/Centrality.ipynb
@@ -685,7 +685,7 @@
    "outputs": [],
    "source": [
     "# PageRank\n",
-    "pr = nk.centrality.PageRank(K, 1e-6)\n",
+    "pr = nk.centrality.PageRank(K, damp=0.85, tol=1e-9)\n",
     "pr.run()\n",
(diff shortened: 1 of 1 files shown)


fabratu (dismissed):

avdgrinten:
Why don't we fix this by using a named parameter? The new code is less instructive than the old one, isn't it?

angriman:
I just don't see the advantage of setting a different tolerance value, the results on the considered graph remain the same.

avdgrinten:
We'll it's a tutorial. The purpose is showing the user how to do it.

angriman:
Then it is probably a good idea to also show how to set the dampening factor.

avdgrinten (approved):

PR #932, opened 2022-05-02
Fix Bug in DynDijstrika related to previously visited nodes.

Relux-the-Relux:
Fix a bug that occurred when a previously visited node was reached with a lower updated distance than their current distance, which caused sometimes paths to be ignored.

### DIFF
--- a/networkit/cpp/distance/test/DynSSSPGTest.cpp
+++ b/networkit/cpp/distance/test/DynSSSPGTest.cpp
@@ -12,7 +12,6 @@
 #include <networkit/distance/BFS.hpp>
 #include <networkit/distance/DynDijkstra.hpp>
 #include <networkit/distance/Dijkstra.hpp>
-#include <networkit/io/METISGraphReader.hpp>
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/generators/DorogovtsevMendesGenerator.hpp>
 #include <networkit/generators/ErdosRenyiGenerator.hpp>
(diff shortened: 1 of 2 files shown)


avdgrinten:
What bug was fixed here? The commit message is quite vague.

angriman (changes requested):
Thank you for the fix. Could you isolate the bug fix in another commit? The changes to `DynSSSPGTest` do not seem related to the bug.

Also, could you make the tests that generate random graphs and/or edge updates reproducible by setting a fixed random seed? I guess that the random failures we were getting before were also due to unreproducible experiments.

avdgrinten:
Just to be sure: the bug is now fixed for any random seed, right? We're just fixing the seed for reproducibility, not to make the test work, right?

Relux-the-Relux:
> Just to be sure: the bug is now fixed for any random seed, right? We're just fixing the seed for reproducibility, not to make the test work, right?

Yes, it should be working on any random seed. Seed 1 was one of the seeds it was failing on. However, I am unsure why is still failing on macOS. Will try to figure it out.

Relux-the-Relux:
This just now truly fix it, the second bug it was that when a number of paths for previously visited nodes changed it wasn't updating them for the successors.

networkit/cpp/distance/test/DynSSSPGTest.cpp
```diff
@@ -348,43 +348,37 @@ TEST_F(DynSSSPGTest, testDynamicBFSGeneratedGraphEdgeDeletion) {
...
+             v1 = GraphTools::randomNode(G), v2 = GraphTools::randomNode(G);
+
+        DEBUG("Adding edge number ", i);
+        G.addEdge(v1, v2);
+        DEBUG("Running update with dynamic dijkstra");
+        dyn_dij.update(GraphEvent(GraphEvent::EDGE_ADDITION, v1, v2, defaultEdgeWeight));
+        DEBUG("Running from scratch with dijkstra");
+        dij.run();
+        G.forNodes([&] (node i) {
+            EXPECT_EQ(dyn_dij.distance(i), dij.distance(i));
```
angriman, line 374: Since we are comparing floating-point numbers, I recommend using this
```suggestion
            EXPECT_DOUBLE_EQ(dyn_dij.distance(i), dij.distance(i));
```

angriman (approved):

PR #934, opened 2022-05-05
Deprecate Python module 'partitioning'. Move functionality to 'community'.

fabratu:
This PR also adds a second version of `community.getImbalance` for both C++ and Python. Not sure, whether the former definition in the community module should be kept.

### DIFF
--- a/networkit/cpp/community/GraphClusteringTools.cpp
+++ b/networkit/cpp/community/GraphClusteringTools.cpp
@@ -7,15 +7,25 @@ namespace NetworKit {
 
 namespace GraphClusteringTools {
 
+namespace {
+float computeMaxClusterSize(const Partition &zeta) {
+    std::vector<count> clusterSizes = zeta.subsetSizes();
+    return (float)*std::max_element(clusterSizes.begin(), clusterSizes.end());
+}
+} // namespace
+
 float getImbalance(const Partition &zeta) {
     float avg = std::ceil(
         (float)zeta.numberOfElements()
         / (float)zeta.numberOfSubsets()); // TODO number of nodes and not number of elements
-    std::vector<count> clusterSizes = zeta.subsetSizes();
-    float maxClusterSize = (float) *std::max_element(clusterSizes.begin(),
-            clusterSizes.end());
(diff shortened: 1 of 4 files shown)


networkit/cpp/community/GraphClusteringTools.cpp
```diff
@@ -14,8 +14,17 @@ float getImbalance(const Partition &zeta) {
...
+}
+
+float getImbalance(const Partition &zeta, const Graph &graph) {
+    float avg = std::ceil(
+        (float)graph.numberOfNodes()
+        / (float)zeta.numberOfSubsets());
+    std::vector<count> clusterSizes = zeta.subsetSizes();
+    float maxClusterSize = (float) *std::max_element(clusterSizes.begin(),
+            clusterSizes.end());
+    return static_cast<float>(maxClusterSize / avg);
```
angriman, line 27: Both variables are float, the cast should not be necessary
```suggestion
    return maxClusterSize / avg;
```

networkit/cpp/community/GraphClusteringTools.cpp
```diff
@@ -14,8 +14,17 @@ float getImbalance(const Partition &zeta) {
     std::vector<count> clusterSizes = zeta.subsetSizes();
     float maxClusterSize = (float) *std::max_element(clusterSizes.begin(),
             clusterSizes.end());
-    float imbalance = maxClusterSize / avg;
-    return imbalance;
+    return static_cast<float>(maxClusterSize / avg);
```
angriman, line 17:
```suggestion
    return maxClusterSize / avg;
```

networkit/cpp/community/GraphClusteringTools.cpp
angriman: The two functions are very similar, the only difference is the computation of `avg`. I suggest having a single private function (e.g., `computeImbalance`) and using it in the two public `getImbalace`.

angriman (approved):

fabratu:
Merging this early since it should be part of the upcoming release.

PR #935, opened 2022-05-05
Use std::sample

angriman:
Use the `std::sample` algorithm introduced in C++17 to sample random nodes.

The failures in the tests for dynamic betweenness do not seem related to my changes. However, the test looks wrong to me: it uses as a baseline an inexact algorithm instead of the exact one. Changing the baseline with the exact algorithm seems to resolve the issue -- see the second commit.

### DIFF
--- a/include/networkit/distance/NeighborhoodFunctionHeuristic.hpp
+++ b/include/networkit/distance/NeighborhoodFunctionHeuristic.hpp
@@ -47,7 +47,6 @@ private:
     std::vector<count> result;
 
     /* selection schemes implemented as private functions */
-    std::vector<node> random(const Graph &G, count nSamples);
     std::vector<node> split(const Graph &G, count nSamples);
 };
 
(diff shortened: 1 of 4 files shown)


fabratu (approved):
LGTM. Do we have any implications concerning the performance of std::sample?

angriman:
I did not benchmark it. For these general-purpose algorithms, however, the implementation provided by the STL is in general better (more efficient/reliable) than a custom implementation.

fabratu:
Would suggest the same, thats why I approved it. Just curiosity on whether there are more data points. A quick benchmarking with generated graphs showed for me very similar performance.

hmeyerhenke:
Please make sure that such changes do not degrade the performance significantly. Some benchmarking should be in place in general. For small/local changes they don't have to be comprehensive. Trust alone can be problematic, though.

avdgrinten (changes requested):
There should definitely be benchmarks for this PR. *Most* STL implementations are good but there are lots of infamous counter examples. Plus, the existing code already goes into some effort to optimize this function.

angriman:
It seems that `GraphTools::randomNodes` is indeed faster than `std::sample`, here is a speedup plot: [speed.pdf](https://github.com/networkit/networkit/files/8935095/speed.pdf). I considered graphs with up to 100M nodes from which we sample 10% to 90% of their nodes. Running times for this experiment are between 1e-5 (for the smallest graphs) and 14 seconds (for the largest graphs).

PR #936, opened 2022-05-05
Use node iterator to sort nodes by degree

angriman:
Use node iterator to simplify code that creates a vector of nodes sorted by their degree.


fabratu (approved):
LGTM.

PR #937, opened 2022-05-05
clang-format 'algebraic' and 'clique' sources

angriman:
And also global sources.


fabratu (approved):
LGTM.

PR #940, opened 2022-05-06
Unify naming scheme and implementation of enums

fabratu:
Besides unifying the definition I also tried to establish a consistent naming scheme. In order to maintain backwards compatibility, I left some of the older enum-names.

### DIFF
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -7,7 +7,6 @@ from libcpp.utility cimport pair
 from libcpp cimport bool as bool_t
 
 import math
-from enum import Enum
 
 from .base cimport _Algorithm, Algorithm
 from .dynamics cimport _GraphEvent, GraphEvent
@@ -756,13 +755,15 @@ cdef class DynBetweennessOneNode:
 
 cdef extern from "<networkit/centrality/Closeness.hpp>" namespace "NetworKit::ClosenessVariant":
 
-	cdef enum _ClosenessVariant  "NetworKit::ClosenessVariant":
-		standard,
-		generalized
+	cpdef enum _ClosenessVariant  "NetworKit::ClosenessVariant":
+		STANDARD,
+		GENERALIZED
 
-class ClosenessVariant:
-	Standard = _ClosenessVariant.standard
-	Generalized = _ClosenessVariant.generalized
+class ClosenessVariant(object):
+	STANDARD = _ClosenessVariant.STANDARD
+	GENERALIZED = _ClosenessVariant.GENERALIZED
+	Standard = STANDARD # this + following added for backwards compatibility
+	Generalized = GENERALIZED
 
 cdef extern from "<networkit/centrality/Closeness.hpp>":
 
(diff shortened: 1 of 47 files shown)


networkit/centrality.pyx
```diff
@@ -1956,13 +1949,15 @@ cdef class KPathCentrality(Centrality):
...
-		InEdges = 1
+	cdef enum _EdgeDirection "NetworKit::EdgeDirection":
+		InEdges,
+		OutEdges
 
 class EdgeDirection(object):
-	inEdges = InEdges
-	outEdges = OutEdges
+	inEdges = _EdgeDirection.InEdges #added for backwards compatibility
+	outEdges = _EdgeDirection.OutEdges #added for backwards compatibility
```
angriman, line 1958: Should we deprecate these two options and raise warnings if they are used?

networkit/centrality.pyx
```diff
@@ -2416,23 +2411,23 @@ cdef class EigenvectorCentrality(Centrality):
...
+		NO_SINK_HANDLING,
 		DISTRIBUTE_SINKS
 
 class Norm(object):
-	l1norm = L1Norm
-	l2norm = L2Norm
+	L1Norm = _Norm.L1Norm
+	L2Norm = _Norm.L2Norm
+	l1norm = _Norm.L1Norm #added for backwards compatibility
+	l2norm = _Norm.L2Norm #added for backwards compatibility
```
angriman, line 2426: As above.

fabratu:
I personally like the `CamelCase` and it is already what we used mostly in the past. Any suggestions? 

Official Python [docs](https://docs.python.org/3/library/enum.html) use the screaming snake case, while C++ leans towards `camelCase` (treat them as normal variables). Following the argument of consistency, we could either use screaming snake case or follow C++ naming conventions.

fabratu:
Same style makes sense. After looking at all C++ enums, it seems most of the members are  `SCREAMING_SNAKE_CASE` (16x) (others: `camelCase` (7x), `CamelCase` (4x)). Also for some enums there are different styles for the identifier and the members :/
That makes `SCREAMING_SNAKE_CASE` the best candidate, since it is also used in Python whenever the enum was redefined as variable of the `cdef class`.

avdgrinten:
Sounds good.

fabratu:
Currently in the process of renaming everything, will take a bit of more time. Actually find it hard to deprecate enums without triggering deprecated warning during building (and therefore creating CI-errors for some builds). One example is `ApproxCloseness.hpp`, where the enum is part of the constructor.

avdgrinten:
Why do we need to deprecate enums?

fabratu:
Tried to convert `enum` to `enum class`, however with Cython <3.0 (currently in alpha state) scoped enums are not officially supported. There are suggested workarounds (basically different ways of wrapping with explicit casting), but this does not work with an `enum class` used as a function parameter. There are varying errors for different compilers. The only way around it would be to cast the enum-parameter to primitives explicitly, which would lead to more code changes on the C++ side.

I would suggest to wait with converting the C++ enums until 3.0 is out and established. Tried out the alpha and this makes it work ootb (also removes some duplicate but currently necessary code).

avdgrinten (approved):
LGTM. Seems that we already used the `readability-identifier-naming.EnumCase` check for clang-tidy before. Why didn't this check warn about ALL_CAPS enum names?

fabratu:
Yeah - I also recognized that part during refactoring ;) Currently we only set the options for `readability-identifier-naming`, however it is not listed as part of the checks. I would say, it is OOS of this PR to change this behavior.

avdgrinten:
Alright, but that is something that we should consider in a future PR.

PR #941, opened 2022-05-06
Refactor Tests for Dynamic Betweenness

angriman:
`DynBetweennessGTest` has a lot of useless/duplicate code. This PR simplifies the tests and sets fixed random seeds to make the tests reproducible.

### DIFF
--- a/networkit/cpp/centrality/test/CentralityGTest.cpp
+++ b/networkit/cpp/centrality/test/CentralityGTest.cpp
@@ -175,6 +175,16 @@ TEST_F(CentralityGTest, runApproxBetweennessSmallGraph) {
     DEBUG("scores: ", bc);
 }
 
+TEST_F(CentralityGTest, runApproxBetweenness) {
+    DorogovtsevMendesGenerator generator(100);
+    Graph G1 = generator.generate();
+    Graph G(G1, true, false);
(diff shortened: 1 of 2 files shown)


networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -175,6 +175,16 @@ TEST_F(CentralityGTest, runApproxBetweennessSmallGraph) {
...
 }
 
+TEST_F(CentralityGTest, runApproxBetweenness) {
+    DorogovtsevMendesGenerator generator(100);
+    Graph G1 = generator.generate();
+    Graph G(G1, true, false);
+    ApproxBetweenness bc(G, 0.1, 0.1);
+    bc.run();
+    ApproxBetweenness bc1(G1, 0.1, 0.1);
+    bc1.run();
```
fabratu, line 185: Should the results from the weighted and the unweighted also be compared to each other?

angriman: Good point, here I just moved the test to another file, my changes only involve tests for dynamic betweenness. This could be added in a separate PR.

fabratu (approved):
LGTM.

PR #943, opened 2022-05-15
clang-format sources of coarsening, components, correlation, and dynamics

angriman:
This is a continuation of #803.


fabratu (approved):
LGTM.

PR #944, opened 2022-05-18
Raise version to 10.0. 

fabratu:
Also updated the package classifier and credits.


fabratu:
Merging this early since it is needed for the upcoming release.

PR #947, opened 2022-05-25
Increase timeout for coverage build

fabratu:
Recently the coverage-run on GHA had several timeouts due to variatons in the CI runner performance. Therefore the timeout is increased.


fabratu:
Thx for the review! I will merge this early, since it is a trivial change (for CI only) and another PR is depending on it.

Issue #942, opened 2022-05-13
ParallelConnectedComponents cause Segmentation fault


academe-01:
Hi,
any idea what's the problem?

```!python
import networkit as nk

x = nk.Graph(6)
x.addEdge(0, 1)
x.addEdge(1, 2)
x.addEdge(3, 4)
x.addEdge(4, 5)
cc = nk.components.ParallelConnectedComponents(x)
cc.run()

cc.hasFinished()
>> True
cc.numberOfComponents()
>> 2

cc.getComponents()
Segmentation fault (core dumped)
```

angriman:
I replicated your issue and, after some testing, it seems that `ParallelConnectedComponents` is bugged. Our tests are not failing because`getComponents()` is never tested for this algorithm.

academe-01:
> My recommendation is to use `ConnectedComponents`, even though it is not parallel, the algorithm is quite fast.

I need to process few billions of CC, unfortunately single threaded ```ConnectedComponents``` not working in my case....

fabratu:
Closing this issue. We merged a PR that fixes this problem. This will be part of the next (possibly bug-fix) release. You can use it right now by installing from the master-branch (or downloading a wheel artifact from CI).

PR #951, opened 2022-06-06
Fix segmentation fault in ParallelConnectedComponents

Relux-the-Relux:
this addresses #942 

It was previously just initializing everything to the `none` which caused just for the label `none` to be passed around in the algorithm, not assigning anything. So now all nodes are now assigned to their own partition. As none of the nodes were being assigned it would fail the `assert(component[u] != none);` from `ComponentDecomposition::componentOfNode`.

Also, the partition needed to be compacted at the end because the partition IDs would not be continuous and the number of partitions is not updated. Non-existing node entries also need to be removed from the partition in the case where the Graph node Ids are not continuous so that the number of components is correct, otherwise, they would be seen as single node components(the vectors for those partitions would be empty in the result from `getComponents()`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -486,7 +486,12 @@ if (NETWORKIT_BUILD_TESTS)
 					networkit_state
 			)
 		endif()
-
+		
+		if(WIN32)
+			set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /wd4996")
+		else()
+			set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} -Wno-deprecated-declarations")
+		endif()
 		set_target_properties(networkit_tests PROPERTIES
 			CXX_STANDARD ${NETWORKIT_CXX_STANDARD}
 			COMPILE_FLAGS "${NETWORKIT_CXX_FLAGS}"
--- a/networkit/cpp/components/test/ConnectedComponentsGTest.cpp
+++ b/networkit/cpp/components/test/ConnectedComponentsGTest.cpp
@@ -87,9 +87,14 @@ TEST_F(ConnectedComponentsGTest, testParallelConnectedComponents) {
         count seqNum = cc.numberOfComponents();
         ParallelConnectedComponents ccPar(G);
         ccPar.run();
-        count parNum = cc.numberOfComponents();
+        count parNum = ccPar.numberOfComponents();
         DEBUG("Number of components: ", seqNum);
         EXPECT_EQ(seqNum, parNum);
+        G.forNodes([&](node u) { EXPECT_EQ(cc.componentOfNode(u), ccPar.componentOfNode(u)); });
+        ccPar.runSequential();
+        parNum = ccPar.numberOfComponents();
+        EXPECT_EQ(seqNum, parNum);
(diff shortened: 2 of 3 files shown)


Relux-the-Relux:
> Why does this fix the issue? Can you add a comment? Why is compaction of component IDs needed here?

updated my comment with the explanation.

avdgrinten:
Where (and why) do we get a segfault if there are unused node IDs in the partition?

networkit/cpp/components/test/ConnectedComponentsGTest.cpp
```diff
@@ -93,6 +93,22 @@ TEST_F(ConnectedComponentsGTest, testParallelConnectedComponents) {
...
+    G.addEdge(1, 2);
+    G.addEdge(3, 4);
+    G.addEdge(4, 5);
+    ConnectedComponents cc(G);
+    cc.run();
+    count seqNum = cc.numberOfComponents();
+    ParallelConnectedComponents ccPar(G);
+    ccPar.run();
+    count parNum = ccPar.numberOfComponents();
+    EXPECT_EQ(seqNum, parNum);
```
angriman, line 108:
```suggestion
    ParallelConnectedComponents ccPar(G);
    ccPar.run();
    EXPECT_EQ( cc.numberOfComponents(), ccPar.numberOfComponents());
```

Could you also:
- add another test with a bigger graph from `input`
- check that the components computed by `ParallelConnectedComponents` in the next test are the same as those computed by `ConnectedComponents` (as you are already doing below).

Relux-the-Relux:
I think #942 failed to correctly transcribe what they actually got., for when testing I also only got 1 parallel component.

Relux-the-Relux:
I think #942 simply mistakenly wrote the wrong number of PCC that they got. Might be a typo. Since I also only observed 1 PCC.

academe-01:
> I think #942 simply mistakenly wrote the wrong number of PCC that they got. Might be a typo. Since I also only observed 1 PCC.

wasn't a typo, everything been copy/pasted from interpreter

networkit/cpp/components/test/ConnectedComponentsGTest.cpp
```diff
@@ -90,6 +90,12 @@ TEST_F(ConnectedComponentsGTest, testParallelConnectedComponents) {
         count parNum = cc.numberOfComponents();
         DEBUG("Number of components: ", seqNum);
         EXPECT_EQ(seqNum, parNum);
+        G.forNodes([&](node u) { EXPECT_EQ(cc.componentOfNode(u), ccPar.componentOfNode(u)); });
+        ccPar.runSequential();
+        parNum = cc.numberOfComponents();
```
fabratu, line 95: `parNum = ccPar.numberOfComponents();`

CMakeLists.txt
```diff
@@ -486,10 +486,15 @@ if (NETWORKIT_BUILD_TESTS)
 					networkit_state
 			)
 		endif()
-
+		
+		if(WIN32)
+			set(NETWORKIT_TESTS_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /wd4996")
```
fabratu, line 491: Only minor, but no need to introduce a new variable. Just reuse `NETWORKIT_CXX_FLAGS`. We do so in other parts of the CMake-file.

fabratu (approved):
LGTM now.

Issue #747, opened 2021-04-23
Do not pass 0 for non-existing edge ids


michitux:
Currently, when there are no edge ids, `0` is passed, e.g., in `forEdges` or `forNeighbors`. I think passing `0` if there are no edge ids is dangerous as it might lead to unexpected bugs if users have an edge attribute and just always read the first value because no edge ids are present in the graph. Unless there is a good reason to keep 0, I would suggest to change this to `none` so at least you notice that the value is not a valid edge id. This also affects #734 as it introduces new functionality also passing `0` as edge id.

hmeyerhenke:
I agree with @michitux.

PR #957, opened 2022-06-20
Make non-existing edge ids return as none instead of 0

Relux-the-Relux:
Previously non-existing edge ids were returned as `0` which could be misleading. Now they return as `none` to be clear that the edge id doesn't exist.

This addresses #747


avdgrinten (approved):
LGTM.

PR #960, opened 2022-06-23
Avoid unnecessary std::fill in sampleNodes

angriman:
When we sample more than n/2 nodes, we can avoid `std::fill` if we mark as selected non-pivot nodes.


avdgrinten:
LGTM. Are there any tests that verify the behavior of sampleNodes?

angriman:
Yes, all three cases are handled here: https://github.com/networkit/networkit/blob/228cbf8832b97538097b21fd9383dd763cff0efe/networkit/cpp/graph/test/GraphToolsGTest.cpp#L224

avdgrinten (approved):

PR #962, opened 2022-06-29
clang-format sources of io

angriman:

### DIFF
--- a/networkit/cpp/io/BinaryEdgeListPartitionReader.cpp
+++ b/networkit/cpp/io/BinaryEdgeListPartitionReader.cpp
@@ -1,21 +1,21 @@
-// no-networkit-format
 #include <fstream>
 
 #include <networkit/io/BinaryEdgeListPartitionReader.hpp>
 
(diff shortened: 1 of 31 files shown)


angriman:
Windows builds seem to be broken, the error message is: `C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\um\winnt.h(169): fatal error C1189: #error:  "No Target Architecture"
`

angriman:
Actually, the error is because clang-format changed the order of the `#include`s. Some files have `#include <windows.h>`, which must be placed before other headers that depend on `windows.h`, I'll fix that.

fabratu (approved):
LGTM.

avdgrinten:
Won't this be reverted by running clang-format again?

angriman:
> Tried it locally and nothing is changed. I think the additional new line is preventing this.

As @fabratu wrote, I added an empty line between `#include<windows.h>` and the other headers; this prevents clang-format from moving it below other headers.

PR #964, opened 2022-07-06
Add option to add edges from a scipy.sparse.coo_matrix and numpy cython support

Relux-the-Relux:
Adding the option to read edges from `coo_array`s by using the Cython NumPy API which should be significantly faster than just using the regular Python API.

There seems to be a problem in the CI where it doesn't find `numpy` for `setup.py`


avdgrinten:
It would be good to some simple benchmarks to check if the performance increased substantially.

Also, it would be nice to support COO arrays directly (in case the user does not use Scipy). That is almost no extra work.

Relux-the-Relux:
> Also, it would be nice to support COO arrays directly (in case the user does not use Scipy). That is almost no extra work.

scipy is currently a requirement to install NetworKit, so I don't think users won't have scipy. But I can do it if that is still desirable.

avdgrinten:
Well, it's just a matter of adding a function that access 3 numpy arrays instead of one scimy matrix. Once that is added, the scipy-based function can call this new one.

.github/workflows/build_wheels.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.1.1
         env:
-          CIBW_BEFORE_BUILD: pip install cython
+          CIBW_BEFORE_BUILD: pip install cython numpy
```
avdgrinten, line 28: Why do we need to install `numpy` explicitly? It's already specified as a dependency in `setup.py`.

Relux-the-Relux: Since it a depency of `setup.py` we need to install it before we run `setup.py` so that it can pass the correct numpy path to cmake. When did install numpy explictly I got a import error. Before this `setup.py` didn't need numpy because we weren't using Cython's numpy API.

avdgrinten: I think this can be solved properly by adding a `pyproject.toml`.

fabratu: For this PR, I think we should only add a minimal `pyproject.toml`. It would also be a good starting point in migrating from the (more or less) deprecated `setup.py` to `pyproject.toml/setup.cfg` build format later on.

This should suffice (with a drive by fix of adding Cython as depency):

```
[build-system]
requires = ["cython", "numpy", "setuptools"]
```

This minimal file (+ changes to the build commands in CI-configuration) would at least eliminate the need to install numpy explicitely, where we don't call `setup.py` with additional parameters. To trigger the installation of build requirements, we have to use PEP517 conform calling mechanisms (like `pip`, `wheel` or `build`) and remove explicit calls to `setup.py build_ext`.

Example (`source-distribution` job in `build_wheels.yml`):

```
#Before
pip install cython numpy
python3 setup.py build_ext
python3 setup.py sdist

# After
pip install build
python3 -m build --sdist .
```

For other cases like `python setup.py build_ext --inplace --external-tlx=...` we have to write a wrapper in order to pass custom options to `pip install`. However, I would propose to leave this out of this PR.

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
 			raise Exception("Attribute name has to be a string")
 		self._this.detachNodeAttribute(stdstring(name))
 
+	def from_coordinates(self, np.ndarray[DINT_t, ndim=1] row, np.ndarray[DINT_t, ndim=1] col, np.ndarray[DDOUBLE_t, ndim=1] data):
+		"""
+		from_coo_matrix(input)
```
fabratu, line 913: Typo? 

Also I think the name `from_coordinates` can be improved, cause it might be misleading. How about the originally suggested `from_coo_arrays`?

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
...
+		----------
+		input : scipy.sparse.coo_matrix
+			Edges in the coo_matrix format.
+		"""
+		cdef np.ndarray[DINT_t, ndim=1] row, col
+		cdef np.ndarray[DDOUBLE_t, ndim=1] data
+		row = input.row.astype(np.int64)
+		col = input.col.astype(np.int64)
+		data =input.data.astype(np.float64)
+		if not isinstance(input, coo_matrix):
```
fabratu, line 946: This should likely be tested at the beginning of the function?

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
...
+		Parameters
+		----------
+		input : scipy.sparse.coo_matrix
+			Edges in the coo_matrix format.
+		"""
+		cdef np.ndarray[DINT_t, ndim=1] row, col
+		cdef np.ndarray[DDOUBLE_t, ndim=1] data
+		row = input.row.astype(np.int64)
+		col = input.col.astype(np.int64)
+		data =input.data.astype(np.float64)
```
fabratu, line 945: `data = input.data.astype(np.float64)`

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
...
+		the columns are the edge targets and the data are the edge weights.
+
+		Parameters
+		----------
+		row : numpy.ndarray
+			Sources of the edges
+		col : numpy.ndarray
+			Targets of the edges
+		data :
+			Weight of the eges
```
fabratu, line 925: `edges.`

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
...
+
+		Inserts edges from coordinate input, where the rows are the edge sources, and
+		the columns are the edge targets and the data are the edge weights.
+
+		Parameters
+		----------
+		row : numpy.ndarray
+			Sources of the edges
+		col : numpy.ndarray
+			Targets of the edges
```
fabratu, line 923: Missing punctuation.

networkit/graph.pyx
```diff
@@ -901,6 +908,47 @@ cdef class Graph:
...
+		"""
+		from_coo_matrix(input)
+
+		Inserts edges from coordinate input, where the rows are the edge sources, and
+		the columns are the edge targets and the data are the edge weights.
+
+		Parameters
+		----------
+		row : numpy.ndarray
+			Sources of the edges
```
fabratu, line 921: Missing punctuation.

fabratu (changes requested):
In addition to benchmark the speed-up in respect to pure-Python, can you also test the current `range`-loop against a classic C-loop? In principle modern Cython versions should automatically optimize this (but maybe not). If not, better use C-loops of course.

fabratu:
Some speed-up measurements:

```
G = nk.generators.BarabasiAlbertGenerator(10, 100000).generate()
row = np.zeros(G.numberOfEdges(), dtype=np.uint)
col = np.zeros(G.numberOfEdges(), dtype=np.uint)
data = np.zeros(G.numberOfEdges(), dtype=np.double)

i = 0
for u,v in G.iterEdges():
    row[i] = u
    col[i] = v
    data[i] = random.uniform(1.0, 100.0)

n = G.numberOfNodes()

S = sc.sparse.coo_matrix((data, (row, col)), dtype = np.double)
start = time.time()
for i in range(100):
    nk.GraphFromCoo(S, n)
end = time.time()

print("Use Cython+Numpy interface with matrix (calling C++ addEdge directly): ", end-start)

start = time.time()
for i in range(100):
    nk.GraphFromCoo(S)
end = time.time()

print("Use Cython+Numpy interface with matrix (calling Python addEdge): ", end-start)

start = time.time()
for i in range(100):
    nk.GraphFromCoo((data,(row,col)), n)
end = time.time()

print("Use Cython+Numpy interface (calling C++ addEdge directly): ", end-start)

start = time.time()
for i in range(100):
    nk.GraphFromCoo((data,(row,col)))
end = time.time()

print("Use Cython+Numpy interface (calling Python addEdge): ", end-start)

start = time.time()
for i in range(100):
    G2 = nk.Graph(n, weighted = True)
    for j in range(len(row)):
        G2.addEdge(row[j], col[j], data[j])
end = time.time()

print("Use Python: ", end-start)
```

Results (so basically time for creating and add edges on 100x graphs with 100k nodes each):

```
Use Cython+Numpy interface with matrix (calling C++ addEdge directly):  0.6958749294281006
Use Cython+Numpy interface with matrix (calling Python addEdge):  6.129072904586792
Use Cython+Numpy interface (calling C++ addEdge directly):  0.5570602416992188
Use Cython+Numpy interface (calling Python addEdge):  6.020819902420044
Use Python:  25.36305522918701
```

networkit/graph.pyx
```diff
@@ -414,6 +423,75 @@ cdef class Graph:
...
+					col = inputData[1].view(dtype = np.uint)
+					data = np.ones(len(row), dtype = np.double)
+				except (TypeError, ValueError) as e:
+					raise TypeError('invalid input format') from e				
+		else:
+			raise TypeError('invalid input format')
+
+		cdef int numEdges = np.shape(row)[0]
+
+		if addMissing:
```
bernlu, line 484: I am not sure why we need to call the python `addEdge` here? We could just add some code to add missing nodes instead? But I do not know the use cases for this `addMissing` parameter - some context might be useful to understand this.

networkit/graph.pyx
```diff
@@ -930,6 +1008,50 @@ cdef class Graph:
 			raise Exception("Attribute name has to be a string")
 		self._this.detachNodeAttribute(stdstring(name))
 
+def GraphFromCoo(inputData, n=0, bool_t weighted=False, bool_t directed=False, bool_t edgesIndexed=False):
+	"""
+	graphFromInputData(inputData, n):
```
bernlu, line 1013: function name and doc string are inconsistent (and also missing some parameters)

notebooks/GraphNotebook.ipynb
```diff
@@ -395,6 +395,105 @@
     "Note, in an undirected graph, like we have here, the [forEdgesOf](https://networkit.github.io/dev-docs/python_api/networkit.html?highlight=foredgesof#networkit.Graph.forEdgesOf) iterator returns all edges of a node. When dealing with a directed graph only the out edges are returned. The rest of the edges can be accessed using the [forInEdgesOf](https://networkit.github.io/dev-docs/python_api/networkit.html?highlight=forinedges#networkit.Graph.forInEdgesOf) iterator."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Graph from Pandas, Scipy and Numpy data\n",
+    "\n",
+    "In addition to manually fill the graph datastructure, it is also possible to create a graph based on given input data in COO (coordinate) format by using `nk.GraphFromCoo(...)`. The parameter syntax is related to [scipy.sparse.coo_array](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_array.html). We start by manually define `row`, `col` and `data` arrays (whereas `A[row[k], col[k]] = data[k]`) and use them as a constructed tuple in the form `(data, (row, col))`."
```
bernlu, line 404: ... manually filling ...

notebooks/GraphNotebook.ipynb
```diff
@@ -395,6 +395,105 @@
...
+    "data = np.array([1.0, 1.0, 1.0])\n",
+    "\n",
+    "GData = nk.GraphFromCoo((data,(row,col)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For speedup purposes, it is possible to also pass the number of expected nodes as a parameter. Due to the tiny size of our example graph, this makes not much difference in this case. However, for the majority of use cases, defining `n` results in a **much** faster creation of the graph."
```
bernlu, line 428: Maybe change this to 
> Due to the tiny size of our example graph, the difference is very small in this case. However, for the majority of use cases, providing `n` results in a **much** faster creation of the graph.

bernlu (changes requested):
There are a few weird diffs/changes at the last lines of files, where I do not see any changes at all.

networkit/graph.pyx
fabratu: From my understanding, this is a convenient function/parameter. Basically what it does is adding the missing nodes instead (with all cases, including dynamic graphs with previously deleted nodes). I was trying to mirror the behavior of `addEdge`, which is supporting `addMissing`.

The other possibilities would be to either omit `addMissing` or re-implement the function. Do you prefer one of those?

networkit/graph.pyx
```diff
@@ -1026,7 +1026,9 @@ def GraphFromCoo(inputData, n=0, bool_t weighted=False, bool_t directed=False, b
 
 	Note
 	----
-	If only pairs of row and column indices (i,j) are given, each edge is given weight 1.0 (even in case of a weighted graph).
+	- If only pairs of row and column indices (i,j) are given, each edge is given weight 1.0 (even in case of a weighted graph).
+	- There is no check if :code:`n` is the correct size. If the parameter is used, make sure that is at least the
```
bernlu, line 1030: "make sure that **it** is at least"

fabratu (approved):

bernlu (approved):

PR #965, opened 2022-07-07
Clang format sources of linkprediction, matching, and numerics

angriman:


fabratu:
Yes, I noticed this error also on another PR. Currently looking into it.

Issue #931, opened 2022-05-02
Inconsistent weights after GraphTools::toUndirected


angriman:
If we pass a directed weighted graph `Gd` with bidirectional edges such as `(u, v, w1)` and `(v, u, w2)` with `w1 != w2` to `GraphTools::toUndirected()`, the resulting undirected graph `Gu` will have inconsistent edge weights. If we traverse the edge `e = (u, v)` in `Gu`, the weight will be `w1`, but if we traverse `e` in the other direction, the weight will be `w2`, which is obviously wrong since `Gu` is undirected.

This happens because the copy constructor of `Graph` (used by `GraphTools::toUndirected`) simply copies the edge weights to the new graph.

Relux-the-Relux:
Which behavior would be desirable here? Should it then become a multigraph or should the edge have a weight of `w1+w2`?

angriman:
I don't think we should generate a multigraph, they are not well-supported in NetworKit. What about taking as input a binary operator? We could use `std::plus` by default.

PR #966, opened 2022-07-07
Fix the undirected Graph copy constructor so that it doesn't generate multigraphs

Relux-the-Relux:
This addresses #931 and should fix the inconsistent results from `GraphTools::toUndirected`.

Now when inserting the inEdges of the previous graph we check if the edge going in the opposite directions is already present in the Graph.


networkit/cpp/graph/Graph.cpp
```diff
@@ -106,44 +107,67 @@ Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
         // G is directed, but we want an undirected graph
         // so we need to combine the out and in stuff for every node
         outEdges.resize(z);
+        if (weighted)
```
fabratu, line 110: If you change the order of loops, pls check performance changes for typical use cases.

Rearranging the loops leads to a different memory layout concerning the involved edge vectors. In the current scheme `edges`, `weights` and `ids` are contiguous each (with respect to iterating over nodes). After rearranging memory is sorted by nodes (`[edges_0, weights_0, ids_0, edges_1, weights_1, ...]`).

networkit/cpp/graph/Graph.cpp
```diff
@@ -106,44 +107,67 @@ Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
...
+            count edgeSurplus = 0;
+            for (count i = 0; i < G.inEdges[u].size(); i++) {
+                node v = G.inEdges[u][i];
+                bool alreadyPresent = false;
+                for (count j = 0; j < G.outEdges[u].size(); j++) {
+                    if (v == G.outEdges[u][j]) {
+                        alreadyPresent = true;
+                        if (u != v) {
+                            edgeSurplus++;
+                            if (weighted && G.isWeighted())
```
fabratu, line 148: This can be simplified by using one `if` for `weighted` and a conditional operator.

networkit/cpp/graph/Graph.cpp
```diff
@@ -106,44 +107,67 @@ Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
...
+                                outEdgeWeights[u][j] =
+                                    f(G.inEdgeWeights[u][i], G.outEdgeWeights[u][j]);
+                            else if (weighted)
+                                outEdgeWeights[u][j] = outEdgeWeights[u][j] =
```
fabratu, line 152: Typo?

fabratu (changes requested):

networkit/cpp/graph/Graph.cpp
Relux-the-Relux: The reason I changed the order of the loops is that when doing the weights and Ids we would have to check in the `out`/`inEdges` array to check if there is a collision between an inedge and an outedge anyway so that we wouldn't have continuous access of memory there even if we inserted each of them in their separate loop.

networkit/cpp/graph/Graph.cpp
```diff
@@ -56,7 +56,8 @@ Graph::Graph(std::initializer_list<WeightedEdge> edges) : Graph(0, true) {
     }
 }
 
-Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
+Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed,
+             std::function<edgeweight(edgeweight, edgeweight)> f)
```
avdgrinten, line 60: We can't use `std::function` in performance critical code, `std::function` has a high per-call overhead. Use a template parameter instead.

angriman:
It is also important to add some documentation, currently there is none. In particular, we should describe how the constructor behaves when we try to build an undirected graph out of a directed graph.

networkit/cpp/graph/Graph.cpp
```diff
@@ -106,45 +107,68 @@ Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
...
             }
         }
+        for (node u = 0; u < z; u++) {
+            // this is necessary to avoid multi edges, because both u -> v and v -> u can exist in G
+            count edgeSurplus = 0;
+            for (count i = 0; i < G.inEdges[u].size(); i++) {
+                node v = G.inEdges[u][i];
+                bool alreadyPresent = false;
+                for (count j = 0; j < G.outEdges[u].size(); j++) {
+                    if (v == G.outEdges[u][j]) { // the edge already exists as an out edge
```
angriman, line 145: We can remove some indentations and make the code more readable
```suggestion
                    if (v != G.outEdges[u][j])
                        continue;
```

networkit/cpp/graph/Graph.cpp
```diff
@@ -106,45 +107,68 @@ Graph::Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed)
...
-                                     G.inEdgeIds[u].end());
             }
         }
+        for (node u = 0; u < z; u++) {
+            // this is necessary to avoid multi edges, because both u -> v and v -> u can exist in G
+            count edgeSurplus = 0;
+            for (count i = 0; i < G.inEdges[u].size(); i++) {
+                node v = G.inEdges[u][i];
+                bool alreadyPresent = false;
+                for (count j = 0; j < G.outEdges[u].size(); j++) {
```
angriman, line 144: Avoid post-increments unless they are necessary
```suggestion
                for (count j = 0; j < G.outEdges[u].size(); ++j) {
```

networkit/cpp/graph/test/GraphToolsGTest.cpp
```diff
@@ -855,17 +855,28 @@ TEST_P(GraphToolsGTest, testToUndirected) {
...
     auto testGraphs = [&](const Graph &G, const Graph &G1) {
+        // we need this because we lose edges due to some nodes having both an in edge and an out
+        // edge to the same node.
+        count edgesLost = 0;
+        if (G.isDirected()) {
+            G.forEdges([&](node u, node v, edgeweight) {
+                if (G.hasEdge(v, u) && u != v)
+                    edgesLost++;
+            });
+        }
```
angriman, line 866:
```suggestion
        if (!G.isDirected())
            return;
        const count edgesLost = G.parallelSumForEdges([&](node u, node v, edgeweight) {
            return (u != v && G.hasEdge(v, u)) ? 1 : 0;
        });
```

networkit/test/test_graphtools.py
```diff
@@ -297,15 +297,19 @@ def testToUndirected(self):
 		def testGraphs(G, G1):
 			self.assertEqual(G.numberOfNodes(), G1.numberOfNodes())
 			self.assertEqual(G.upperNodeIdBound(), G1.upperNodeIdBound())
-			self.assertEqual(G.numberOfEdges(), G1.numberOfEdges())
+			edgesLost = 0;
+			for e in G.iterEdges():
+				if G.hasEdge(e[1], e[0]) and e[1] is not e[0]:
```
angriman, line 302:
```suggestion
				if G.hasEdge(e[1], e[0]) and e[1] != e[0]:
```

networkit/test/test_graphtools.py
```diff
@@ -297,15 +297,19 @@ def testToUndirected(self):
...
+					edgesLost = edgesLost + 1
```
angriman, line 303:
```suggestion
					edgesLost += 1
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,11 @@ class Graph final {
      */
     Graph(count n = 0, bool weighted = false, bool directed = false, bool edgesIndexed = false);
 
-    Graph(const Graph &G, bool weighted, bool directed, bool edgesIndexed = false);
+    Graph(
+        const Graph &G, bool weighted, bool directed, bool edgesIndexed = false,
+        edgeweight (*edgeMerger)(edgeweight, edgeweight) = [](edgeweight w1, edgeweight w2) {
+            return w1 + w2;
+        });
```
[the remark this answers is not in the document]
Relux-the-Relux: It won't compile under windows, since we have a default case.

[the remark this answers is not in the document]
angriman: What error do you get?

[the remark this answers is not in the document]
angriman: Ah sorry, I just noticed that you already fixed this.

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                    // undirected
+                    inEdgeWeights = G.inEdgeWeights;
+                    outEdgeWeights = G.outEdgeWeights;
+                } else {
+                    // G has no weights, set defaultEdgeWeight for all edges
+                    if (directed) {
+                        inEdgeWeights.resize(z);
+                        for (node u = 0; u < z; u++) {
+                            inEdgeWeights[u] =
+                                std::vector<edgeweight>(G.inEdges[u].size(), defaultEdgeWeight);
```
angriman, line 1399:
```suggestion
                            inEdgeWeights[u].resize(G.inEdges[u].size(), defaultEdgeWeight);
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                        }
+                    }
+
+                    outEdgeWeights.resize(z);
+                    for (node u = 0; u < z; u++) {
+                        outEdgeWeights[u] =
+                            std::vector<edgeweight>(outEdges[u].size(), defaultEdgeWeight);
```
angriman, line 1406:
```suggestion
                        outEdgeWeights[u].resize(outEdges[u].size(), defaultEdgeWeight);
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                        // same for weights
+                        outEdgeWeights[u].reserve(G.outEdgeWeights[u].size()
+                                                  + G.inEdgeWeights[u].size());
+                        outEdgeWeights[u].insert(outEdgeWeights[u].end(),
+                                                 G.outEdgeWeights[u].begin(),
+                                                 G.outEdgeWeights[u].end());
+                    } else {
+                        // we are undirected, so no need to write anything into inEdgeWeights
+                        outEdgeWeights[u] =
+                            std::vector<edgeweight>(outEdges[u].size(), defaultEdgeWeight);
```
angriman, line 1437:
```suggestion
                        outEdgeWeights[u].resize(outEdges[u].size(), defaultEdgeWeight);
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                    }
+                }
+                if (G.hasEdgeIds() && edgesIndexed) {
+                    // copy both out and in edges ids into our new outEdgesIds
+                    outEdgeIds[u].reserve(G.outEdgeIds[u].size() + G.inEdgeIds[u].size());
+                    outEdgeIds[u].insert(outEdgeIds[u].end(), G.outEdgeIds[u].begin(),
+                                         G.outEdgeIds[u].end());
+                }
+            }
+            for (node u = 0; u < z; ++u) {
```
angriman, line 1447: Can we use a parallel loop here? This operation seems quite expensive and we only modify the out-edges of `u`.
```suggestion
            G.parallelForNodes([&](node u) {
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                    if (!alreadyPresent) { // an equivalent out edge wasn't present so we add it
+                        outEdges[u].push_back(v);
+                        if (weighted)
+                            outEdgeWeights[u].push_back(G.isWeighted() ? G.inEdgeWeights[u][i]
+                                                                       : defaultEdgeWeight);
+                        if (G.hasEdgeIds() && edgesIndexed)
+                            outEdgeIds[u].push_back(G.inEdgeIds[u][i]);
+                    }
+                }
+                m -= edgeSurplus;
```
angriman, line 1480: If we use a parallel loop, this can be updated using an atomic operation or an omp reduction.

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                if (G.isWeighted()) {
+                    inEdgeWeights = G.outEdgeWeights;
+                    outEdgeWeights = G.outEdgeWeights;
+                } else {
+                    // initialize both inEdgeWeights and outEdgeWeights with the
+                    // defaultEdgeWeight
+                    inEdgeWeights.resize(z);
+                    for (node u = 0; u < z; u++) {
+                        inEdgeWeights[u] =
+                            std::vector<edgeweight>(inEdges[u].size(), defaultEdgeWeight);
```
angriman, line 1497:
```suggestion
                        inEdgeWeights[u].resize(inEdges[u].size(), defaultEdgeWeight);
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,155 @@ class Graph final {
...
+                    }
+                    outEdgeWeights.resize(z);
+                    for (node u = 0; u < z; u++) {
+                        outEdgeWeights[u] =
+                            std::vector<edgeweight>(outEdges[u].size(), defaultEdgeWeight);
```
angriman, line 1502:
```suggestion
                        outEdgeWeights[u].resize(outEdges[u].size(), defaultEdgeWeight);
```

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,153 @@ class Graph final {
...
+        } else if (G.isDirected()) {
+            // G is directed, but we want an undirected graph
+            // so we need to combine the out and in stuff for every node
+            outEdges.resize(z);
+            if (weighted)
+                outEdgeWeights.resize(z);
+            if (G.hasEdgeIds() && edgesIndexed)
+                outEdgeIds.resize(z);
+#pragma omp parallel for
+            for (omp_index u = 0; u < static_cast<omp_index>(z); ++u) {
```
angriman, line 1421: Why do we use these loops and not `parallelForNodes`? If `G` has deleted nodes then we do useless iterations, which are automatically skipped by graph iterators.

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,151 @@ class Graph final {
...
+            }
+        } else if (G.isDirected()) {
+            // G is directed, but we want an undirected graph
+            // so we need to combine the out and in stuff for every node
+            outEdges.resize(z);
+            if (weighted)
+                outEdgeWeights.resize(z);
+            if (G.hasEdgeIds() && edgesIndexed)
+                outEdgeIds.resize(z);
+            G.balancedParallelForNodes([&](node u) {
```
angriman, line 1420: I'd recommend using `parallelForNodes`. The workload depends on the degree of `u` and with the `guided` omp schedule we would incur the same inefficiencies as described in #819.

include/networkit/graph/Graph.hpp
```diff
@@ -1363,7 +1363,151 @@ class Graph final {
...
+                    }
+                }
+                if (G.hasEdgeIds() && edgesIndexed) {
+                    // copy both out and in edges ids into our new outEdgesIds
+                    outEdgeIds[u].reserve(G.outEdgeIds[u].size() + G.inEdgeIds[u].size());
+                    outEdgeIds[u].insert(outEdgeIds[u].end(), G.outEdgeIds[u].begin(),
+                                         G.outEdgeIds[u].end());
+                }
+            });
+            G.balancedParallelForNodes([&](node u) {
```
angriman, line 1444: Same here.

include/networkit/graph/Graph.hpp
Relux-the-Relux: I benchmarked both the `parallelForNodes` and `balancedParallelForNodes` and `balancedParallelForNodes` ended being 20-40% faster.

angriman: Could you provide further details about those benchmarks? As I did in #819, It would be helpful to know the size of the graphs, their degree distribution, the running time of the two loops, and the machine you used.

angriman: Thanks. Compared to #819 these are rather small graphs and the balanced version saves us less than 2s in the best case. Can you run the same experiment on some larger graphs (e.g., up to a few billion edges) on a more powerful server and see if the results change?

Relux-the-Relux: I tried with some bigger graphs on the gruenau1 server and the trend of `balancedParallelForLoops` being faster continues. I will try with graphs bigger than 1 billion edges next to see if it changes something.

| graph name | Nodes | Edges | Time with parallelForLoops in us| Time with balancedParallelForLoops |
| :---         |     :---:      |      :---:      |          ---: |          ---: |
| [web-indochina-2004-all](https://networkrepository.com/web-indochina-2004-all.php)   | 7.5M     | 194M   |1 342 921 834|66 780 383|
| [web-uk-2002-all](https://networkrepository.com/web-uk-2002-all.php)     | 18.5M    | 298M   |40 839 508| 29 361 124|

angriman: Thanks for running the experiments. I forgot to mention that in #819 the recommendation is to use the `dynamic` schedule. `parallelForNodes` uses the default static schedule which, as your experiments are showing, seems to be worse than the `guided` one for this use case. I think that for this PR it is fine to use `balancedParallelForNodes`.

angriman (approved):

PR #969, opened 2022-07-13
Update MIT license for 2022.

fabratu:
Updated the license. @hmeyerhenke Not sure, whether legally the involved universites should be named.


avdgrinten (approved):

avdgrinten:
We could also update the meta data in version.py (but probably in another PR).

Issue #953, opened 2022-06-08
Wrong TriangleEdgeScore on multi-graphs


darabos:
For example,
```python
import networkit as nk
G = nk.graph.Graph(2)
G.addEdge(0, 1)
G.addEdge(0, 1)
s = nk.sparsification.TriangleEdgeScore(G)
G.indexEdges()
s.run()
print(s.scores())
```
prints
```
[18446744073709551615, 18446744073709551615]
```

ChibaNishizekiTriangleEdgeScore is also wrong, it just outputs zeros for the duplicate edges.
```python
G = nk.graph.Graph(3)
G.addEdge(0, 1)
G.addEdge(0, 1)
G.addEdge(2, 0)
G.addEdge(2, 1)
s = nk.sparsification.ChibaNishizekiTriangleEdgeScore(G)
G.indexEdges()
s.run()
print(s.scores())
```
prints
```
[2, 0, 2, 2]
```

I'm not sure what would be the right fix, just wanted to let you know. Thanks!

hmeyerhenke:
Multigraphs are not guaranteed to be handled correctly throughout many algorithms. (Early on we had to take a decision and agreed not to support multigraphs since they complicate the code and make it slower.) This is not documented well, I agree.

fabratu:
Closing this issue now. We merged a PR that improves the documentation and enables checking for multi edges when inserting edges. Feature will be part of the next release. You can use it right now by installing from the master-branch (or downloading a wheel artifact from CI).

PR #972, opened 2022-07-15
Modify addEdge to check for multi-edges

fabratu:
This PR add an additional (opt.) parameter to addEdge in C++/Python for checking whether an edge is already present in the graph. The documentation/tutorials are updated accordingly.

Also checked for other places, where a warning about unsupported multi-edges might be missing in the documentation. Maybe we could also use the modified addEdge for reader/writer or converter to ensure a warning is triggered when inserting multi-edges (like in #953).

Another approach would be an additional method like `addEdgeSafe`, where multi-edges are checked first.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1655,52 +1655,76 @@ public:
      * Insert an edge between the nodes @a u and @a v. If the graph is
      * weighted you can optionally set a weight for this edge. The default
      * weight is 1.0. Note: Multi-edges are not supported and will NOT be
-     * handled consistently by the graph data structure.
+     * handled consistently by the graph data structure. It is possible to check
+     * for multi-edges by enabling parameter "checkForMultiEdges". If already present,
+     * the new edge is not inserted. Enabling this check increases the complexity of the function
+     * to O(max(deg(u), deg(v))).
      * @param u Endpoint of edge.
      * @param v Endpoint of edge.
      * @param weight Optional edge weight.
+     * @param checkMultiEdge If true, this enables a check for a possible multi-edge.
+     * @return @c true if edge has been added, false otherwise (in case checkMultiEdge is set to
+     * true and the new edge would have been a multi-edge.)
      */
-    void addEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
+    bool addEdge(node u, node v, edgeweight ew = defaultEdgeWeight, bool checkMultiEdge = false);
 
     /**
      * Insert an edge between the nodes @a u and @a v. Unline the addEdge function, this function
      * does not not add any information to v. If the graph is weighted you can optionally set a
      * weight for this edge. The default weight is 1.0. Note: Multi-edges are not supported and will
-     * NOT be handled consistently by the graph data structure.
+     * NOT be handled consistently by the graph data structure. It is possible to check
+     * for multi-edges by enabling parameter "checkForMultiEdges". If already present,
+     * the new edge is not inserted. Enabling this check increases the complexity of the function
+     * to O(max(deg(u), deg(v))).
      * @param u Endpoint of edge.
      * @param v Endpoint of edge.
      * @param weight Optional edge weight.
      * @param ew Optional edge weight.
      * @param index Optional edge index.
+     * @param checkMultiEdge If true, this enables a check for a possible multi-edge.
+     * @return @c true if edge has been added, false otherwise (in case checkMultiEdge is set to
+     * true and the new edge would have been a multi-edge.)
      */
-    void addPartialEdge(Unsafe, node u, node v, edgeweight ew = defaultEdgeWeight,
-                        uint64_t index = 0);
+    bool addPartialEdge(Unsafe, node u, node v, edgeweight ew = defaultEdgeWeight,
+                        uint64_t index = 0, bool checkForMultiEdges = false);
 
     /**
      * Insert an in edge between the nodes @a u and @a v in a directed graph. If the graph is
      * weighted you can optionally set a weight for this edge. The default
      * weight is 1.0. Note: Multi-edges are not supported and will NOT be
-     * handled consistently by the graph data structure.
+     * handled consistently by the graph data structure. It is possible to check
+     * for multi-edges by enabling parameter "checkForMultiEdges". If already present,
+     * the new edge is not inserted. Enabling this check increases the complexity of the function
+     * to O(max(deg(u), deg(v))).
      * @param u Endpoint of edge.
      * @param v Endpoint of edge.
      * @param ew Optional edge weight.
      * @param index Optional edge index.
+     * @param checkMultiEdge If true, this enables a check for a possible multi-edge.
+     * @return @c true if edge has been added, false otherwise (in case checkMultiEdge is set to
+     * true and the new edge would have been a multi-edge.)
      */
-    void addPartialInEdge(Unsafe, node u, node v, edgeweight ew = defaultEdgeWeight,
-                          uint64_t index = 0);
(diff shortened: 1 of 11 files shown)


networkit/cpp/graph/Graph.cpp
```diff
@@ -591,12 +591,20 @@ edgeweight Graph::weightedDegreeIn(node u, bool countSelfLoopsTwice) const {
...
-void Graph::addEdge(node u, node v, edgeweight ew) {
+void Graph::addEdge(node u, node v, edgeweight ew, bool checkForMultiEdges) {
     assert(u < z);
     assert(exists[u]);
     assert(v < z);
     assert(exists[v]);
 
+    if(checkForMultiEdges) {
+        if (std::find(outEdges[u].begin(), outEdges[u].end(), v) != outEdges[u].end() || std::find(outEdges[v].begin(), outEdges[v].end(), u) != outEdges[v].end()) {
+            std::cerr << "Not adding edge " << u << "-" << v << " since it is already present.\n";
```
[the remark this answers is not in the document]
fabratu: I had `WARN` in first place, however this does not trigger a warning when called from Python.

include/networkit/graph/Graph.hpp
```diff
@@ -1655,12 +1655,15 @@ class Graph final {
      * Insert an edge between the nodes @a u and @a v. If the graph is
      * weighted you can optionally set a weight for this edge. The default
      * weight is 1.0. Note: Multi-edges are not supported and will NOT be
-     * handled consistently by the graph data structure.
+     * handled consistently by the graph data structure. It is possible to check
+     * for multi-edges by enabling parameter "checkForMultiEdges". If detected, the new
+     * edge is not inserted. This increases the runtime of the function by O(Δ(G)).
```
angriman, line 1660:
```suggestion
     * for multi-edges by enabling parameter "checkForMultiEdges". If already present, the new
     * edge is not inserted. Enabling this check increases the complexity of the function to O(max(deg(u), deg(v))).
```

networkit/cpp/graph/Graph.cpp
```diff
@@ -591,12 +591,20 @@ edgeweight Graph::weightedDegreeIn(node u, bool countSelfLoopsTwice) const {
...
-void Graph::addEdge(node u, node v, edgeweight ew) {
+void Graph::addEdge(node u, node v, edgeweight ew, bool checkForMultiEdges) {
     assert(u < z);
     assert(exists[u]);
     assert(v < z);
     assert(exists[v]);
 
+    if (checkForMultiEdges) {
+        if (std::find(outEdges[u].begin(), outEdges[u].end(), v) != outEdges[u].end()
+            || std::find(outEdges[v].begin(), outEdges[v].end(), u) != outEdges[v].end()) {
```
angriman, line 602: This check doesn't look right. If we have a directed graph with the edge `(v, u)`, then we cannot add the edge `(u, v)`.
Can't we simply use `if (hasEdge(u, v))`?

notebooks/GraphNotebook.ipynb
```diff
@@ -544,9 +600,14 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.10.5"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "b0fa6594d8f4cbf19f97940f81e996739fb7646882a419484c72d19e05852a7e"
+   }
```
angriman, line 608: Not sure if these changes to the notebook are intended.

networkit/cpp/graph/Graph.cpp
fabratu: True, that makes more sense.

notebooks/GraphNotebook.ipynb
fabratu: No, they are not - thanks for the review!

networkit/cpp/graph/Graph.cpp
```diff
@@ -591,12 +591,19 @@ edgeweight Graph::weightedDegreeIn(node u, bool countSelfLoopsTwice) const {
...
 
-void Graph::addEdge(node u, node v, edgeweight ew) {
+void Graph::addEdge(node u, node v, edgeweight ew, bool checkForMultiEdges) {
     assert(u < z);
     assert(exists[u]);
     assert(v < z);
     assert(exists[v]);
 
+    if (checkForMultiEdges) {
+        if (hasEdge(u, v)) {
```
angriman, line 601:
```suggestion
    if (checkForMultiEdges && hasEdge(u, v)) {
```

hmeyerhenke:
Make sure that the performance implications are negligible. Our documentation should be clear enough that multigraphs are not supported.

fabratu:
I did a small comparison (300x inserting 1 mil. edges into a graph with 100k nodes) with the original code vs. new code (wo. multi check) vs. new code (w. multi check):

original: 14.88s +- 0.1
new (wo. multi check): 14.90 +- 0.1
new (w. multi check): 16.21 +- 0.11

So the performance loss due to the additional conditional seems negligible (when `checkMultiEdge` is `False`).

Of course the overhead by performing the multi check grows with increasing ratio of edges vs. nodes. Doing 300x 1mil. edges into graph with 100 nodes:

new (w. multi check): 28.22 +- 0,25

hmeyerhenke:
The default should be not to check. But if someone wants to check, they can.

fabratu:
I agree that we should probably not do the multi checks per default in performance critical situations. For some modules like `io` and `generators` this might be a good idea to use it. 

Other parts of the code-base currently using `addEdge` include forest- and algebraic-datastructures + dynamic algorithms.

hmeyerhenke:
Well, we do not support multi-edges. Why should we devote significant cycles to such tests? To make users aware, okay. But that is sufficient in io, not other parts. Unless a generator is not guaranteed to produce simple graphs.

avdgrinten (changes requested):
Okay, seems that the current design has a consensus.

However, instead of letting `addEdge` emit an error message, we should let `addEdge` (and `addPartial{In,Out}Edge`) return a `bool` to indicate whether the edge was added or not.

The callers should deal with error reporting; they should use NetworKit's logging facility instead of `std::cerr`.

networkit/cpp/io/DibapGraphReader.cpp
```diff
@@ -161,15 +161,15 @@ Graph DibapGraphReader::read(const std::string &path) {
         for (index v = 0; v < (count)V; ++v) {
             for (index e = of[v]; e < (index)of[v + 1]; ++e) {
                 if (v <= (index)to[e]) {
-                    graph.addEdge(v, to[e], ew[e]);
+                    graph.addEdge(v, to[e], ew[e], true);
```
angriman, line 164: This (and in other files below) changes the behavior of the readers right? Since we do not support multi-edges, I think it is ok to discard them by default. However, we should properly document this in the notes of the next release.

avdgrinten (commented):
The approach looks good now. Are there any generators that also need to be adjusted?

fabratu:
> The approach looks good now. Are there any generators that also need to be adjusted?

Hm, I am not in-depth familiar with every generator. Some at least have explicit checks already, some don't create multi-edges due to their underlying model. Also most generators use `GraphBuilder.addHalfEdge`, which includes asserts for multi-edge checks. During CI-builds these should show up.

For two generators I am not sure though: Regular Ring Lattice and Stochastic Block Model. Based on the code, it seems they can create graphs with multi-edges.

avdgrinten:
Alright. I think it would be fine to merge the PR as-is. If we later find that these generators still need to be adjusted, we can do it in a separate PR.

avdgrinten (approved):

PR #983, opened 2022-08-22
Refactor PLP

angriman:
- Fix a potential bug in PLP (a variable was updated non-atomically in a parallel loop)
- Replace `std::map` with `std::unordered_map` because the order of the elements is not needed
- Minor stylistic changes

### DIFF
--- a/networkit/cpp/community/PLP.cpp
+++ b/networkit/cpp/community/PLP.cpp
@@ -61,8 +61,7 @@ void PLP::run() {
      * iteration.
      */
 
-    std::vector<bool> activeNodes(z); // record if node must be processed
-    activeNodes.assign(z, true);
+    std::vector<bool> activeNodes(z, true); // record if node must be processed
 
     Aux::Timer runtime;
 
--- a/networkit/cpp/community/PLP.cpp
+++ b/networkit/cpp/community/PLP.cpp
@@ -42,8 +42,8 @@ void PLP::run() {
         updateThreshold = (count)(n / 1e5);
     }
 
-    count nUpdated; // number of nodes which have been updated in last iteration
-    nUpdated = n;   // all nodes have new labels -> first loop iteration runs
+    // number of nodes which have been updated in last iteration
+    count nUpdated = n; // all nodes have new labels -> first loop iteration runs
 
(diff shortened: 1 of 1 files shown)


fabratu:
Since the remaining errors occur on Windows, do you need assistance in finding the error? I could check here locally.

angriman:
That would be helpful, I don't have much time to debug that error at the moment 😄 thank you!

fabratu:
Using different version of MSVC I could not reproduce the error locally. Also the failing does not take more time than expected. After increasing the log-level on the CI-run, it seems that a `bad alloc` is raised (full error: `unknown file: error: C++ exception with description "bad allocation" thrown in the test body.`).

SInce the run also takes ~45min I would suggest that the VM memory is indeed running out. For other non-MS based runners, there is no unproportinal increase in running time.

Bottom line: not sure why this happens, maybe related to allocating memory for the `unordered_map` during the loop.

fabratu:
Hm, not sure what changed code-wise, but the error still remains the same (timeout on Windows machines). If the `std::unordered_map` is the culprit here, I found an older SO-post which might give some hints (if still applicable): https://stackoverflow.com/questions/39313820/stdunordered-map-does-not-release-memory

It is more about `erase/reset` then temporary objects, but maybe the mentioned memory deallocation of pages has some influence. Their [current STL-implementation](https://github.com/microsoft/STL/blob/2f8342a3a57fb157d881c6a2d42a917d20e413f8/stl/inc/unordered_map) of `unordered_map` seems to have changed with respect to the SO-post, since currently they use a bidirectional linked list instead of vectors (if I read it correctly). You could try to build with C++13 for MSVC and see whether this changes something. Another way would be to reuse the datastructure and call reset at the end of each loop-run.

angriman:
Seems ok now. I just replaced the `[]` operator with `find`. I still don't get why `[]` did not work.

fabratu:
> Seems ok now. I just replaced the `[]` operator with `find`. I still don't get why `[]` did not work.

Good question, at least the `unordered_map` was somehow related to the error. LGTM.

fabratu (approved):

PR #984, opened 2022-08-22
Fix crash of SPSP in case of unreachable targets

angriman:
SPSP crashes if the user passes unreachable targets (the node -> distance map lookup fails for unreachable targets because they are never added to the map). This can easily be fixed by setting to infinity the distance to those targets.
This PR also adds a new test with an unreachable target that fails with the previous implementation of SPSP.

### DIFF
--- a/networkit/cpp/distance/MultiTargetBFS.cpp
+++ b/networkit/cpp/distance/MultiTargetBFS.cpp
@@ -1,6 +1,7 @@
 #include <networkit/distance/BFS.hpp>
 #include <networkit/distance/MultiTargetBFS.hpp>
 
+#include <limits>
 #include <queue>
 #include <unordered_set>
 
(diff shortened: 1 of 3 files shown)


networkit/cpp/distance/MultiTargetBFS.cpp
```diff
@@ -61,6 +62,14 @@ void MultiTargetBFS::run() {
         toNextLevel();
     }
 
+    // Mark targets not reached as unreachable
+    while (!targetsSet.empty()) {
```
fabratu, line 66: Why not just iterating over targetsSet? This would remove the need for erasing the iterator in each loop run.

fabratu, line 66: Why not just iterating over targetsSet? This would remove the need for erasing the iterator in each loop run.

networkit/cpp/distance/MultiTargetDijkstra.cpp
```diff
@@ -41,6 +41,14 @@ void MultiTargetDijkstra::run() {
         });
     } while (!heap.empty());
 
+    // Mark targets not reached as unreachable
+    while (!targetsSet.empty()) {
```
fabratu, line 45: Same here.

fabratu (approved):
LGTM now.

PR #988, opened 2022-09-02
Initialize edgeID datastructure when constr. graph with indexed edges.

### LINKED ISSUE #986: Core dumped when adding weighted edges with edgesIndexed=True
Hello all, first issue here.

Decided I would report the following:
I have a dict mapping strings to node ids as well as a list of edges

    import networkit as nk

    nodes = {'Joe':0, 'Bob': 1, 'Sue': 2}  # 60 nodes approx.
    edges = [
        {"from":0, "to":1, "weight":  0.1231232329}, 
        {"from":1 ,"to": 2, "weight:" 123.413123627)
    ] # 2-3k edges approx.

    G = nk.Graph(n=len(nodes), weighted=True, directed=True, edgesIndexed=True)
    
    for edge in edges:
        G.addEdge(u=edge["from"], v=edge["to"], w=round(edge["weight"], 4))

results in:
    Segmentation fault (core dumped)

Tested the same logic with edgesIndexed=False and haven't yet encountered the core dump.
Also tried groupbyng my data by from, to in order to avoid multiEdges but still receive the error.

Perhaps there is some elementary logical error I'm making due to my lack of understanding in c++ and low-level memory management. Furthermore, this is my first ever issue so please be lenient.

Happy to provide more data as needed.

fabratu:
Fixes #986.

### DIFF
--- a/networkit/cpp/graph/test/GraphGTest.cpp
+++ b/networkit/cpp/graph/test/GraphGTest.cpp
@@ -132,6 +132,21 @@ void GraphGTest::SetUp() {
 
 /** CONSTRUCTORS **/
 
+TEST(GraphGTest, testDefConstructorWithUndirIndex) {
+    // Test indexed + undirected graph
+    Graph GUndir(3, false, false, true);
+    GUndir.addEdge(0, 1);
+    EXPECT_EQ(GUndir.edgeId(0, 1), 0);
+    EXPECT_EQ(GUndir.edgeId(1, 0), 0);
+}
+
+TEST(GraphGTest, testDefConstructorWithDirIndex) {
+    // Test indexed + directed graph
+    Graph GDir(3, false, true, true);
+    GDir.addEdge(0, 1);
+    EXPECT_EQ(GDir.edgeId(0, 1), 0);
+}
+
 TEST_P(GraphGTest, testCopyConstructor) {
     Graph G = Graph(this->Ghouse, false, false);
     Graph GW = Graph(this->Ghouse, true, false);
--- a/networkit/cpp/graph/Graph.cpp
+++ b/networkit/cpp/graph/Graph.cpp
@@ -37,7 +37,8 @@ Graph::Graph(count n, bool weighted, bool directed, bool edgesIndexed)
       outgoing edges, for undirected graphs outEdges stores the adjacency list of
       undirected edges*/
       outEdges(n), inEdgeWeights(weighted && directed ? n : 0), outEdgeWeights(weighted ? n : 0),
-      inEdgeIds(), outEdgeIds(), nodeAttributeMap(this) {}
+      inEdgeIds(edgesIndexed && directed ? n : 0), outEdgeIds(edgesIndexed ? n : 0),
+      nodeAttributeMap(this) {}
 
 Graph::Graph(std::initializer_list<WeightedEdge> edges) : Graph(0, true) {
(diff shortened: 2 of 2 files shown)


networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -132,6 +132,12 @@ void GraphGTest::SetUp() {
 
 /** CONSTRUCTORS **/
 
+TEST_F(GraphGTest, testDefaultConstructorIndexed) {
+    Graph G = Graph(3, false, false, true);
```
angriman, line 136:
```suggestion
    Graph G(3, false, false, true);
```

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -132,6 +132,12 @@ void GraphGTest::SetUp() {
...
+    G.addEdge(0, 1);
+    ASSERT_EQ(0, G.edgeId(0, 1));
```
angriman, line 138: In this case it is better to use `EXPECT_EQ` (see https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html). Also, avoid yoda conditions (https://en.wikipedia.org/wiki/Yoda_conditions)
```suggestion
    EXPECT_EQ(G.edgeId(0, 1), 0);
```

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -132,6 +132,12 @@ void GraphGTest::SetUp() {
 
 /** CONSTRUCTORS **/
 
+TEST_F(GraphGTest, testDefaultConstructorIndexed) {
```
angriman, line 135: It seems that you have to use parametric tests:
```suggestion
TEST_P(GraphGTest, testDefaultConstructorIndexed) {
```

fabratu: Yeah, I am not sure why though. We are mixing TEST_F and TEST_P also in other test files (same syntax). Trying to find a way around it, otherwise the same test is repeated 4x.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -132,6 +132,12 @@ void GraphGTest::SetUp() {
 
 /** CONSTRUCTORS **/
 
+TEST(GraphGTest, testDefaultConstructorIndexed) {
+    Graph G(3, false, false, true);
```
angriman, line 136: Maybe test it for directed graphs as well?

fabratu: Good point.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -132,6 +132,18 @@ void GraphGTest::SetUp() {
 
 /** CONSTRUCTORS **/
 
+TEST(GraphGTest, testDefaultConstructorIndexed) {
```
angriman, line 135: I'd recommend to split this into two tests, one for directed and one for undirected graphs.
Test isolation is something that we could improve across the whole library.

fabratu: You are right. In my mind the test was too small to seperate it, but this more a decision about how we want to structure the tests.

angriman (approved):

PR #997, opened 2022-10-04
Allow single % and integer weights with MatrixMarketReader

Relux-the-Relux:
First, the Matrix Market Format allows for both either a single or double `%` in the first line(https://networkrepository.com/mtx-matrix-market-format.html) so I fixed it in our reader.

Also currently we don't accept integer weights but integer weights and doubles would only not be able to match up for numbers larger than 2^52. Since most weights are smaller than that I think it is better if we accept the weights and just display a warning that for weights larger than 4.5*10^15 the stored weights will not match the given weights perfectly as otherwise, we're just causing an annoyance of having to go back and edit the files to say "real" instead of `integer`.

Lastly, I added tests for the MatrixmMarketRead as there were none.


networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -63,7 +64,12 @@ CSRMatrix MatrixMarketReader::read(std::istream &in) {
             if (data == "pattern") {
                 weighted = false;
             } else if (data != "real") {
-                throw std::runtime_error("Unsupported data type: " + data);
+                if (data == "integer")
+                    WARN(
+                        "Weights in NetworKit are stored as double precision floating point number while the given file uses integer weights.\
+                This means that weights bigger than 4.5e15 the weights will be recorded imprecisely.");
```
fabratu, line 70: This means that weights bigger than 4.5*e^15 will be recorded imprecisely.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1225,4 +1227,68 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
     }
 }
 
+TEST_F(IOGTest, testMatrixMarketReader) {
```
fabratu, line 1230: Better have seperate tests for all cases (faster to track down if there are errors).

fabratu (changes requested):

fabratu (changes requested):
Please also update the readme for inputfiles -> `https://github.com/networkit/networkit/blob/master/input/README.md`.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1225,4 +1227,79 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
     }
 }
 
+TEST_F(IOGTest, testMatrixMarketReaderUnweightedUndirected) {
+    MatrixMarketReader reader;
+
+    CSRGeneralMatrix csr = reader.read("input/chesapeake.mtx");
```
angriman, line 1233: If the elements in this matrix are `double`s you can simply use `CSRMatrix`

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1225,4 +1227,79 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
     }
 }
 
+TEST_F(IOGTest, testMatrixMarketReaderUnweightedUndirected) {
+    MatrixMarketReader reader;
+
+    CSRMatrix csr = reader.read("input/chesapeake.mtx");
```
angriman, line 1233:
```suggestion
    MatrixMarketReader reader;

    CSRMatrix csr = MatrixMarketReader{}.read("input/chesapeake.mtx");
```
Also in other tests below.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1225,4 +1227,79 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
...
+    EXPECT_EQ(csr(17, 13), 1.0);
+}
+
+TEST_F(IOGTest, testMatrixMarketReaderWeightedUndirected) {
+    MatrixMarketReader reader;
+
+    CSRMatrix csr = reader.read("input/LFAT5.mtx");
+    EXPECT_EQ(csr.numberOfRows(), 14);
+    EXPECT_EQ(csr.nnz(),
+              30 * 2 - 14); // double due to being undirected but we don't double count self loops
```
angriman, line 1266:
```suggestion
    // double due to being undirected but we don't double count self loops
    EXPECT_EQ(csr.nnz(), 30 * 2 - 14);
```

networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -9,13 +9,14 @@
...
 
+#include <networkit/auxiliary/Log.hpp>
 #include <networkit/io/MatrixMarketReader.hpp>
 
 namespace NetworKit {
 
 namespace {
 static constexpr char COMMENT_CHAR = '%';
-static const std::string MAGIC = "%%matrixmarket";
+static const std::string MAGIC = "%matrixmarket";
```
angriman, line 19:
```suggestion
static constexpr std::string MAGIC = "%matrixmarket";
```

networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -54,7 +55,7 @@ CSRMatrix MatrixMarketReader::read(std::istream &in) {
             format = tolower(format);
             data = tolower(data);
             qualifier = tolower(qualifier);
-            if (magic != MAGIC) {
+            if (magic != MAGIC && magic != ("%" + MAGIC)) {
```
angriman, line 58:
```suggestion
            if (magic != MAGIC1 && magic != MAGIC2)) {
```

networkit/cpp/io/MatrixMarketReader.cpp
```diff
@@ -54,16 +56,21 @@ CSRMatrix MatrixMarketReader::read(std::istream &in) {
...
                 throw std::runtime_error("Unsupported format: " + format);
             }
             if (data == "pattern") {
                 weighted = false;
             } else if (data != "real") {
-                throw std::runtime_error("Unsupported data type: " + data);
+                if (data == "integer")
+                    WARN(
+                        "Weights in NetworKit are stored as double precision floating point number while the given file uses integer weights.\
+                This means that weights bigger than 4.5e15 the weights will be recorded imprecisely.");
```
fabratu, line 71:
```suggestion
                This means that weights bigger than 4.5e15 will be recorded imprecisely.");
```

PR #998, opened 2022-10-21
Remove possibility to ignore unformatted files

angriman:
Unformatted files can be skipped by adding the line `//no-networkit-format` at the top of cpp files. Since all cpp files are
now formatted, this option is not needed anymore.

PR #1001, opened 2022-10-27
CI: Raise llvm toolkit to v15.0 (clang, clang-format, clang-tidy)

fabratu:
This updates the tooling for llvm-based builds / tests. GCC compiler is not touched, since there are still problems with GCC 12+, which needs to be resolved beforehand.

This PR also solves problems on CI, since OpenMP is compatible with -/+ one major version. For the most recent GHA images, the version deviates by two (as seen on #998).

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -4,7 +4,6 @@ Checks: >
     bugprone-copy-constructor-init,
     bugprone-dangling-handle,
     bugprone-dynamic-static-initializers,
-    bugprone-exception-escape,
     bugprone-fold-init-type,
     bugprone-forward-declaration-namespace,
     bugprone-forwarding-reference-overload,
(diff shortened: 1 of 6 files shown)


.clang-tidy
```diff
@@ -43,6 +42,7 @@ Checks: >
     readability-avoid-const-params-in-decls,
     readability-const-return-type,
     readability-container-size-empty
+# temp. disabled: bugprone-exception-escape
```
angriman, line 45: Is this still necessary? Or did you manage to address the errors with your changes in `KadabraBetweenness` and `DynBFS`?

Relux-the-Relux: Apparently, there too many errors so they will be addressed in a second PR by me, so that we can merge this quickly.

Relux-the-Relux (approved):
LGTM.

.clang-tidy
angriman: Alright makes sense.

fabratu: Yes, the changed files are handling `modernize-use-emplace`, the other errors are more technical (and more numerous). Since we need the update to fix CI, fixing the remaining tidy issues is imho out of scope from this PR.

fabratu:
I will merge this early, since it is needed for fixing CI pipeline.

PR #1003, opened 2022-10-27
SpanningTree: bugfix suggestion and tests

### LINKED ISSUE #1002: Kruskal Spanning Tree Bug
The spanning tree code in `KruskalMSF` seems to have a bug for weighted graphs (intoduced with the merge from #873 ?).
As the testcases only cover unweighted graphs, none of them fails.
`sortedEdges` is empty when std::transform tries to run resulting in a segfault.

CharJon:
Fixes #1002

### DIFF
--- a/networkit/cpp/graph/KruskalMSF.cpp
+++ b/networkit/cpp/graph/KruskalMSF.cpp
@@ -30,9 +30,9 @@ void KruskalMSF::run() {
         UnionFind uf(z);
 
         // sort edges in decreasing weight order
-        std::vector<MyEdge> sortedEdges; // (m);
+        std::vector<MyEdge> sortedEdges(G->numberOfEdges());
         std::transform(G->edgeWeightRange().begin(), G->edgeWeightRange().end(),
-                       std::back_inserter(sortedEdges), [](const auto &edge) -> MyEdge {
+                       sortedEdges.begin(), [](const auto &edge) -> MyEdge {
                            return {edge.u, edge.v, edge.weight};
                        });
         Aux::Parallel::sort(sortedEdges.begin(), sortedEdges.end());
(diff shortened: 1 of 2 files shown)


networkit/cpp/graph/KruskalMSF.cpp
```diff
@@ -32,7 +32,7 @@ void KruskalMSF::run() {
         // sort edges in decreasing weight order
         std::vector<MyEdge> sortedEdges; // (m);
         std::transform(G->edgeWeightRange().begin(), G->edgeWeightRange().end(),
-                       sortedEdges.begin(), [](const auto &edge) -> MyEdge {
+                       std::back_inserter(sortedEdges), [](const auto &edge) -> MyEdge {
```
fabratu, line 35: Since transform should work on already allocated memory, it might be beneficial performance-wise to pre-allocate the vector on definition.

fabratu (changes requested):
Thanks for the PR! I have a minor change request, otherwise this looks fine.

networkit/cpp/graph/KruskalMSF.cpp
CharJon: Makes a lot of sense. Will take care of this.

CharJon:
Is there something I could change to stop the pipeline from failing? I do not see why the checks pass on the main branch, but not on this one.

fabratu:
> Is there something I could change to stop the pipeline from failing? I do not see why the checks pass on the main branch, but not on this one.

Most of the problems will be fixed with #1001 (plan is to have it in master shortly). I am not sure about the failing `DynBetweennessGTest.runDynVsStatic` test. However I would suggest re-running the tests after merge/rebase.

fabratu:
#1001 is merged, you can now rebase.

PR #1004, opened 2022-11-02
Refactor TopCloseness, TopHarmonicCloseness

fabratu:
This PR introduces the following changes:

- Support for passing a list of nodes, restricting top-k calculation to only this nodes (but based on the truth given by graph G).
- Tests for new parameter
- Minor refactor of documentation, bug fixes for both Python + C++ code
- Refactor TopHarmonicCloseness to support multiple consecutive runs

If the number of nodes in the list << number of nodes in the graph, the running time is improved significantly. Example: 

- 200k nodes in graph
- 100 nodes in input list
- speedup 165x (TopCloseness)
- speedup 55x (TopHarmonicCloseness)

~~Currently the PR is still work in progress.~~

### DIFF
--- a/include/networkit/centrality/TopCloseness.hpp
+++ b/include/networkit/centrality/TopCloseness.hpp
@@ -41,9 +41,11 @@ public:
      * degree.
      * @param sec_heu If true, the BFSbound is re-computed at each iteration. If
      * false, BFScut is used.
-     * @
+     * @param nodeList Restrict closeness computation to a grou of nodes from the graph. In order to
+     * work, these nodes have to be present.
      */
-    TopCloseness(const Graph &G, count k = 1, bool first_heu = true, bool sec_heu = true);
+    TopCloseness(const Graph &G, count k = 1, bool first_heu = true, bool sec_heu = true,
+                 const std::vector<node> &nodeList = std::vector<node>());
 
     /**
      * Computes top-k closeness on the graph passed in the constructor.
@@ -78,6 +80,7 @@ private:
     count k;
     bool first_heu, sec_heu;
     std::vector<node> topk;
+    std::vector<node> nodeList;
     count visEdges;
     count n_op;
     count trail;
--- a/include/networkit/centrality/TopCloseness.hpp
+++ b/include/networkit/centrality/TopCloseness.hpp
@@ -41,8 +41,9 @@ public:
      * degree.
      * @param sec_heu If true, the BFSbound is re-computed at each iteration. If
      * false, BFScut is used.
-     * @param nodeList Restrict closeness computation to a grou of nodes from the graph. In order to
-     * work, these nodes have to be present.
+     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
+     * If the list is empty (default), all nodes from the graph are used for computation.
+     * Note: Actual existence of included nodes in the graph is not checked.
      */
     TopCloseness(const Graph &G, count k = 1, bool first_heu = true, bool sec_heu = true,
                  const std::vector<node> &nodeList = std::vector<node>());
--- a/include/networkit/centrality/TopCloseness.hpp
+++ b/include/networkit/centrality/TopCloseness.hpp
@@ -41,12 +41,8 @@ public:
      * degree.
      * @param sec_heu If true, the BFSbound is re-computed at each iteration. If
      * false, BFScut is used.
-     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
-     * If the list is empty (default), all nodes from the graph are used for computation.
-     * Note: Actual existence of included nodes in the graph is not checked.
      */
-    TopCloseness(const Graph &G, count k = 1, bool first_heu = true, bool sec_heu = true,
-                 const std::vector<node> &nodeList = std::vector<node>());
+    TopCloseness(const Graph &G, count k = 1, bool first_heu = true, bool sec_heu = true);
 
     /**
      * Computes top-k closeness on the graph passed in the constructor.
@@ -75,13 +71,24 @@ public:
      */
     std::vector<edgeweight> topkScoresList(bool includeTrail = false);
 
+    /**
+     * @brief Restricts the top-k closeness computation to a subset of nodes.
+     * If the given list is empty, all nodes in the graph will be considered.
+     * Note: Actual existence of included nodes in the graph is not checked.
+     *
+     * @param nodeList Subset of nodes.
+     */
+    void restrictTopKComputationToNodes(const std::vector<node> &nodeList) {
+        nodeListPtr = &nodeList;
+    };
+
 private:
     const Graph &G;
     count n;
     count k;
     bool first_heu, sec_heu;
     std::vector<node> topk;
-    std::vector<node> nodeList;
+    const std::vector<node> *nodeListPtr;
     count visEdges;
     count n_op;
     count trail;
--- a/include/networkit/centrality/TopHarmonicCloseness.hpp
+++ b/include/networkit/centrality/TopHarmonicCloseness.hpp
@@ -41,12 +41,8 @@ public:
      * @param k Number of nodes with highest harmonic closeness that have to be found.
      * For example, k = 10, the top 10 nodes with highest harmonic closeness will be computed.
      * @param useNBbound If true, the NBbound variation will be used, otherwise NBcut.
-     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
-     * If the list is empty (default), all nodes from the graph are used for computation.
-     * Note: Actual existence of included nodes in the graph is not checked.
      */
-    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false,
-                                  const std::vector<node> &nodeList = std::vector<node>());
+    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false);
 
     ~TopHarmonicCloseness() override;
 
@@ -80,6 +76,17 @@ public:
      */
     std::vector<edgeweight> topkScoresList(bool includeTrail = false);
 
+    /**
+     * @brief Restricts the top-k harmonic closeness computation to a subset of nodes.
+     * If the given list is empty, all nodes in the graph will be considered.
+     * Note: Actual existence of included nodes in the graph is not checked.
+     *
+     * @param nodeList Subset of nodes.
+     */
+    void restrictTopKComputationToNodes(const std::vector<node> &nodeList) {
+        nodeListPtr = &nodeList;
+    };
+
 private:
     const Graph *G;
     const count k;
@@ -92,7 +99,7 @@ private:
 
     std::vector<node> topKNodes;
     std::vector<double> topKScores;
-    std::vector<node> nodeList;
+    const std::vector<node> *nodeListPtr;
 
     // For NBbound
     std::vector<count> reachU, reachL;
@@ -110,8 +117,8 @@ private:
         }
     }
 
-    tlx::d_ary_addressable_int_heap<node, 2, Aux::GreaterInVector<double>> prioQ;
     tlx::d_ary_addressable_int_heap<node, 2, Aux::LessInVector<double>> topKNodesPQ;
+    tlx::d_ary_addressable_int_heap<node, 2, Aux::GreaterInVector<double>> prioQ;
     std::vector<node> trail;
 
     // For weighted graphs
@@ -122,6 +129,7 @@ private:
 
     omp_lock_t lock;
 
+    void init();
     void runNBcut();
     void runNBbound();
     bool bfscutUnweighted(node source, double kthCloseness);
--- a/include/networkit/centrality/TopHarmonicCloseness.hpp
+++ b/include/networkit/centrality/TopHarmonicCloseness.hpp
@@ -41,8 +41,9 @@ public:
      * @param k Number of nodes with highest harmonic closeness that have to be found.
      * For example, k = 10, the top 10 nodes with highest harmonic closeness will be computed.
      * @param useNBbound If true, the NBbound variation will be used, otherwise NBcut.
-     * @param nodeList Restrict closeness computation to a grou of nodes from the graph. In order to
-     * work, these nodes have to be present.
+     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
+     * If the list is empty (default), all nodes from the graph are used for computation.
+     * Note: Actual existence of included nodes in the graph is not checked.
      */
     explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false,
                                   const std::vector<node> &nodeList = std::vector<node>());
--- a/include/networkit/centrality/TopHarmonicCloseness.hpp
+++ b/include/networkit/centrality/TopHarmonicCloseness.hpp
@@ -41,8 +41,11 @@ public:
      * @param k Number of nodes with highest harmonic closeness that have to be found.
      * For example, k = 10, the top 10 nodes with highest harmonic closeness will be computed.
      * @param useNBbound If true, the NBbound variation will be used, otherwise NBcut.
+     * @param nodeList Restrict closeness computation to a grou of nodes from the graph. In order to
+     * work, these nodes have to be present.
      */
-    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false);
+    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false,
+                                  const std::vector<node> &nodeList = std::vector<node>());
 
     ~TopHarmonicCloseness() override;
 
@@ -88,6 +91,7 @@ private:
 
     std::vector<node> topKNodes;
     std::vector<double> topKScores;
+    std::vector<node> nodeList;
 
     // For NBbound
     std::vector<count> reachU, reachL;
--- a/networkit/cpp/centrality/TopCloseness.cpp
+++ b/networkit/cpp/centrality/TopCloseness.cpp
@@ -20,8 +20,9 @@
 
 namespace NetworKit {
 
-TopCloseness::TopCloseness(const Graph &G, count k, bool first_heu, bool sec_heu)
-    : G(G), k(k), first_heu(first_heu), sec_heu(sec_heu) {}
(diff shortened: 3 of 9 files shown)


networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1293,6 +1293,37 @@ TEST_P(CentralityGTest, testTopCloseness) {
...
+        for (auto secHeu : {true, false}) {
+            TopCloseness topC(G, k, firstHeu, secHeu);
+            topC.run();
+            auto topCNodes = topC.topkNodesList();
+            TopCloseness topCR(G, k, firstHeu, secHeu, nodeList);
+            topCR.run();
+            auto topCRNodes = topCR.topkNodesList();
+            EXPECT_EQ(topCNodes[0], topCRNodes[0]);
+            for (node i : {0, 1, 2, 3, 4, 5}) {
+                EXPECT_EQ(std::find(topCNodes.begin(), topCNodes.end(), i), topCNodes.end());
```
angriman, line 1316: This can be simplified a bit using something like `EXPECT_THAT(topCNodes, Contains(i))`. See here for more: https://github.com/google/googletest/blob/main/docs/reference/matchers.md

angriman: You can also avoid the loop by using `EXPECT_THAT(topCNodes, UnorderedElementsAre({0, 1, ...}));`. The link above gives you a wide range of container matchers.

fabratu: Thanks for the review ... wasn't aware of gmock-based tests. Don't know if that makes it really simpler though ;) Had to use `IsSuperSetOf`, since `UnorderedElementsAre` (or similar) require the same number of elements from the argument and the matcher.

networkit/cpp/centrality/TopCloseness.cpp
```diff
@@ -345,10 +346,14 @@ void TopCloseness::run() {
...
     tlx::d_ary_addressable_int_heap<node, 2, Aux::LessInVector<double>> Q{farness};
 
-    std::vector<node> nodes;
-    nodes.reserve(G.numberOfNodes());
-    G.forNodes([&](const node u) { nodes.emplace_back(u); });
-    Q.build_heap(std::move(nodes));
+    if (nodeList.empty()) {
+        std::vector<node> nodes;
+        nodes.reserve(G.numberOfNodes());
+        G.forNodes([&](const node u) { nodes.emplace_back(u); });
```
angriman, line 352:
```suggestion
        std::vector<node> nodes(G.nodeRange().begin(), G.nodeRange().end());
```

networkit/cpp/centrality/TopCloseness.cpp
```diff
@@ -345,10 +346,14 @@ void TopCloseness::run() {
...
+        Q.build_heap(std::move(nodes));
```
angriman, line 353: You can probably use
```suggestion
        Q.build_heap(G.nodeRange().begin(), G.nodeRange().end());
```

networkit/cpp/centrality/test/CentralityGTest.cpp
angriman: Ah true, then `IsSuperSetOf` is what you need. Overall I think that now it's easier to understand what you are testing than having all those `std::find`.

angriman: Btw, you are testing that some nodes are in the top-k ranking. Shouldn't you also test the ranking?

fabratu: Hm, you mean what we expect as ranking vector (when entering a `nodeList`)?

networkit/cpp/centrality/TopCloseness.cpp
fabratu: Good point.

networkit/cpp/centrality/test/CentralityGTest.cpp
angriman: As far as I understand, `topkNodesList` returns the top-k nodes sorted by non-increasing closeness score. If that's the case, then we should test it.

fabratu: Ah, right! Added tests for ensuring proper rank-sorting.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1293,6 +1294,36 @@ TEST_P(CentralityGTest, testTopCloseness) {
...
+            topCR.run();
+            auto topCRNodes = topCR.topkNodesList();
+            auto topCRScores = topCR.topkScoresList();
+            EXPECT_EQ(topCNodes[0], topCRNodes[0]);
+            EXPECT_THAT(topCRNodes, testing::IsSupersetOf({0, 1, 2, 3, 4, 5, 26, 48, 64}));
+            EXPECT_THAT(topCNodes, testing::IsSupersetOf({26, 48, 64}));
+            EXPECT_THAT(topCNodes, testing::Not(testing::IsSupersetOf({0, 1, 2, 3, 4, 5})));
+            for (count i = 0; i < k - 1; i++) {
+                EXPECT_GE(topCRScores[i], topCRScores[i + 1]);
+            }
```
angriman, line 1322: I recommend to use `std::is_sorted` https://en.cppreference.com/w/cpp/algorithm/is_sorted

include/networkit/centrality/TopCloseness.hpp
```diff
@@ -41,9 +41,12 @@ class TopCloseness final : public Algorithm {
      * degree.
      * @param sec_heu If true, the BFSbound is re-computed at each iteration. If
      * false, BFScut is used.
-     * @
+     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
+     * If the list is empty (default), all nodes from the graph are used for computation.
+     * Note: Actual existence of included nodes in the graph is not checked.
```
angriman, line 46: What about
```suggestion
     * @param nodeList Restricts the top-k closeness computation to a subset of nodes.
     * If the list is empty (default), all nodes in the graph will be considered.
     * Note: Actual existence of included nodes in the graph is not checked.
```

include/networkit/centrality/TopCloseness.hpp
```diff
@@ -78,6 +81,7 @@ class TopCloseness final : public Algorithm {
     count k;
     bool first_heu, sec_heu;
     std::vector<node> topk;
+    std::vector<node> nodeList;
```
angriman, line 84: Can we avoid to copy the node list?
```suggestion
    const std::vector<node> &nodeList;
```

include/networkit/centrality/TopHarmonicCloseness.hpp
```diff
@@ -41,8 +41,12 @@ class TopHarmonicCloseness final : public Algorithm {
      * @param k Number of nodes with highest harmonic closeness that have to be found.
      * For example, k = 10, the top 10 nodes with highest harmonic closeness will be computed.
      * @param useNBbound If true, the NBbound variation will be used, otherwise NBcut.
+     * @param nodeList Restrict closeness computation to a group of nodes from the graph.
+     * If the list is empty (default), all nodes from the graph are used for computation.
+     * Note: Actual existence of included nodes in the graph is not checked.
      */
-    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false);
+    explicit TopHarmonicCloseness(const Graph &G, count k = 1, bool useNBbound = false,
+                                  const std::vector<node> &nodeList = std::vector<node>());
```
angriman, line 49: This seems a quite specific feature and I think that adding an additional parameter to the constructor to use it is not a good idea. Could we maybe set the list via a separate function (e.g., `void restrictTopKComputationToNodes(const std::vector<node> &nodeList);`)? One could call that function before `run`.

fabratu: I also thought about whether to integrate it into the constructor or having a calling function, however came up with the opposite conclusion initially. Your argument makes sense though.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1293,6 +1294,34 @@ TEST_P(CentralityGTest, testTopCloseness) {
...
+            auto topCNodes = topC.topkNodesList();
+            TopCloseness topCR(G, k, firstHeu, secHeu, nodeList);
+            topCR.run();
+            auto topCRNodes = topCR.topkNodesList();
+            auto topCRScores = topCR.topkScoresList();
+            EXPECT_EQ(topCNodes[0], topCRNodes[0]);
+            EXPECT_THAT(topCRNodes, testing::IsSupersetOf({0, 1, 2, 3, 4, 5, 26, 48, 64}));
+            EXPECT_THAT(topCNodes, testing::IsSupersetOf({26, 48, 64}));
+            EXPECT_THAT(topCNodes, testing::Not(testing::IsSupersetOf({0, 1, 2, 3, 4, 5})));
+            EXPECT_TRUE(std::is_sorted(topCRScores.begin(), topCRScores.end()));
```
angriman, line 1320: I think you should use a different comparator, probably:
```suggestion
            EXPECT_TRUE(std::is_sorted(topCRScores.begin(), topCRScores.end(), std::greater_equal<double>{}));
```
By default, `std::is_sorted` checks that the given range is sorted by non-increasing order.

include/networkit/centrality/TopCloseness.hpp
fabratu: Changed it to be a pointer due to removing it from the constructor parameter list.

include/networkit/centrality/TopHarmonicCloseness.hpp
```diff
@@ -105,8 +117,8 @@ class TopHarmonicCloseness final : public Algorithm {
         }
     }
 
-    tlx::d_ary_addressable_int_heap<node, 2, Aux::GreaterInVector<double>> prioQ;
-    tlx::d_ary_addressable_int_heap<node, 2, Aux::LessInVector<double>> topKNodesPQ;
+    std::unique_ptr<tlx::d_ary_addressable_int_heap<node, 2, Aux::LessInVector<double>>>
+        topKNodesPQPtr;
```
angriman, line 121: Why did you change this to a `unique_ptr`?

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1313,7 +1343,6 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
                     continue;
                 TopHarmonicCloseness topcc(G, k, useNBbound);
                 topcc.run();
-
```
angriman, line 1316: This was probably introduced by mistake.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1329,6 +1358,30 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
+TEST_F(CentralityGTest, testTopHarmonicClosenessWithNodeList) {
+    METISGraphReader reader;
+    Graph G = reader.read("input/lesmis.graph");
+    constexpr count k = 10;
+    const std::vector<node> nodeList{0, 1, 2, 3, 4, 5, 6, 11, 27, 48};
+
+    // We expect complete TopHarmonicCloseness to not contain nodes 0-6,
+    // while restricted should. {11, 27, 48} should be present in both
+    // results (but position might differ).
+    for (bool useNBbound : {true, false}) {
```
angriman, line 1370: I'd recommend having two tests, one per heuristic. If this test fails, we'll not be able to tell which of the two heuristics was running.

include/networkit/centrality/TopHarmonicCloseness.hpp
fabratu: Good point. I had issues with non-emptied `tlx::d_ary_addressable_int_heap` during refactoring and was unsure, whether the datastructure really gets reset between runs. This is not anymore needed - will revert it.

networkit/cpp/centrality/test/CentralityGTest.cpp
fabratu: Not sure whether we want to also do this for `TopCloseness` (2x4 more tests with a lot of duplication).

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1323,9 +1323,10 @@ TEST_F(CentralityGTest, testTopClosenessWithNodeList) {
     }
 }
 
-TEST_P(CentralityGTest, testTopHarmonicCloseness) {
+TEST_P(CentralityGTest, testTopHarmonicClosenessUseNBbound) {
     const count size = 400;
     const double tol = 1e-6;
+    bool useNBbound = true;
```
angriman, line 1329: I guess this is not needed anymore right?

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1337,28 +1338,62 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
-
-                topkScores = topcc.topkScoresList(true);
-
-                for (count i = 0; i < topkScores.size(); ++i)
-                    EXPECT_NEAR(ranking[i].second, topkScores[i], tol);
-                for (count i = k; i < topkScores.size(); ++i)
-                    EXPECT_NEAR(topkScores[i], topkScores[k - 1], tol);
-            }
+        for (count k : {5, 10, 20}) {
+            if (isWeighted() && useNBbound)
```
angriman, line 1342:
```suggestion
            if (isWeighted())
```

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1337,28 +1338,62 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
+                continue;
+            TopHarmonicCloseness topcc(G, k, useNBbound);
```
angriman, line 1344:
```suggestion
            TopHarmonicCloseness topcc(G, k, true);
```

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1337,28 +1338,62 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
+            for (count i = k; i < topkScores.size(); ++i)
+                EXPECT_NEAR(topkScores[i], topkScores[k - 1], tol);
+        }
+    }
+}
+
+TEST_P(CentralityGTest, testTopHarmonicClosenessUseNBcut) {
+    const count size = 400;
+    const double tol = 1e-6;
+    bool useNBbound = false;
```
angriman, line 1364: Same here.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1337,28 +1338,62 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
+        if (isWeighted()) {
+            G = GraphTools::toWeighted(G);
+            G.forEdges([&G](node u, node v) { G.setWeight(u, v, Aux::Random::probability()); });
+        }
+        HarmonicCloseness cc(G, false);
+        cc.run();
+        const auto ranking = cc.ranking();
+        for (count k : {5, 10, 20}) {
+            if (isWeighted() && useNBbound)
+                continue;
```
angriman, line 1378: This is never executed.

networkit/cpp/centrality/test/CentralityGTest.cpp
```diff
@@ -1337,28 +1338,62 @@ TEST_P(CentralityGTest, testTopHarmonicCloseness) {
...
+            TopHarmonicCloseness topcc(G, k, useNBbound);
```
angriman, line 1379:
```suggestion
            TopHarmonicCloseness topcc(G, k, false);
```

networkit/cpp/centrality/test/CentralityGTest.cpp
angriman: Yes, this would introduce duplication. One option to avoid it is to implement TopCloseness in separate parametric GTests where NBbound as another parameter of the tests.

fabratu: Yes, makes sense. My first idea was to split Centrality and Top/TopHarmonicCloseness tests into different files in order to avoid multiple `TEST_SUITE` definitions in `CentralityGTests.cpp`. I also saw the possibiliy to just add a list of parameters. However for bools this would not really improve the situation if something fails (without taking a dive into the code). Do you have other suggestions about how to implement this?

networkit/cpp/centrality/TopHarmonicCloseness.cpp
```diff
@@ -93,12 +113,18 @@ void TopHarmonicCloseness::run() {
 }
 
 void TopHarmonicCloseness::runNBcut() {
+    tlx::d_ary_addressable_int_heap<node, 2, Aux::GreaterInVector<double>> prioQ{hCloseness};
```
angriman, line 116: I don't think that moving this here is a good idea because doing so you allocate O(n) memory every time you run nbcut. If you just need to ensure that the heap is empty then call `clear`; this won't require to re-allocate the heap every time you run the algorithm.

networkit/cpp/centrality/test/CentralityGTest.cpp
fabratu: Opted now for splitting the tests in different files. Having multiple test suites in one huge file decreases readability too much imho.

networkit/cpp/centrality/TopHarmonicCloseness.cpp
fabratu: Right, fixed.

networkit/cpp/centrality/TopHarmonicCloseness.cpp
```diff
@@ -173,7 +204,7 @@ void TopHarmonicCloseness::runNBbound() {
         if (u == none)
             break;
 
-        bfsbound(u);
+        bfsbound(u, prioQ);
```
angriman, line 207: `prioQ` is a member of the class, you don't need to pass it as a parameter.

angriman (commented):
LGTM. Unfortunately we are still getting the `bad allocation` error that I was getting in my PLP PR. We should probably roll that back.

angriman (approved):

Issue #963, opened 2022-07-01
Core dumped while reading graph with NetworkitBinaryReader if remove nodes before write graph


Jerevia:
For example,
```python
import networkit as nk

nkG = nk.Graph(weighted=True)

nkG.addNodes(100)
nkG.addEdge(0, 1)
nkG.addEdge(2, 3)
nkG.removeNode(10)

nk.writeGraph(nkG, 'graph.bin', nk.Format.NetworkitBinary)
nk.readGraph('graph.bin', nk.Format.NetworkitBinary)
```
it prints
```
Segmentation fault (core dumped)
```

But if I don' t remove nodes, this code could work. 

Another problem is, if I replace the writing format with, for instance, nk.Format.LFR, this code could also work. But all the nodes and edges deleted before saving seems to appear again.

Jerevia:
FYI，community detection algorithms, I tried louvain and leiden, give bad results or core dumped if removing nodes before running this algorithms

Jerevia:
@Relux-the-Relux Hi, do you have any plans for fixing this issue?

axelbrowne:
I just wrote added an issue before seeing that this is the same issue. Whoops. Here's my write-up on the same thing.

### Reading from NetworkitBinary file segfaults for graphs written with uncontinuous node-indices #996

graphio.writeGraph will write a graph with uncontinuous node indices (i.e. nodes are not indexed [0...G.numberOfNodes()) ) to a .networkit file without error, but reading from the same file will result in a segmentation fault.

Ex:
```
# G : a nk.Graph with continuous node indexing
# nodes : a list of uncontinuous/arbitrary nodes indices
# path : some file path string. doesn't matter


g = nk.graphtools.GraphTools.subgraphFromNodes(G, nodes)

# this passes
assert g.numberOfNodes() == len(nodes)

# runs successfully
nk.writeGraph(g, path, nk.Format.NetworkBinary)
```

Later...
```
>>> g = nk.readGraph(path, nk.Format.NetworkBinary)
Segmentation Fault (Core Dumped)
```

fabratu:
If I am not mistaken, this issue is fixed with #1009 (bug in the writer) and therefore available in NetworKit version 10.1. To avoid this, you have to write the graph using 10.1.

Will close this issue for now. Feel free to reopen it if the problem still persists.

PR #1009, opened 2022-11-11
Fix NetworkitBinaryWriter - writing graphs with deleted nodes

Relux-the-Relux:
This addresses  #963 by changing an array with a map to avoid errors when writing.

### DIFF
--- a/networkit/cpp/io/test/IOGTest.cpp
+++ b/networkit/cpp/io/test/IOGTest.cpp
@@ -1227,6 +1227,24 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
     }
 }
 
+TEST_F(IOGTest, testNetworkitWriterNonContinuousNodesIds) {
+    Graph G(20, true);
+    G.removeNode(10);
+    std::string path = "output/test.gt";
+    NetworkitBinaryWriter{}.write(G, path);
+    Graph GRead = NetworkitBinaryReader{}.read(path);
+    EXPECT_EQ(G.numberOfNodes(), GRead.numberOfNodes());
+    EXPECT_EQ(G.numberOfEdges(), GRead.numberOfEdges());
+    EXPECT_EQ(G.isDirected(), GRead.isDirected());
+    EXPECT_EQ(G.isWeighted(), GRead.isWeighted());
+    G.forNodes([&](node u) {
+        G.forEdgesOf(u, [&](node v) {
+            EXPECT_TRUE(GRead.hasEdge(u, v));
+            EXPECT_DOUBLE_EQ(G.weight(u, v), GRead.weight(u, v));
+        });
+    });
+}
+
 TEST_F(IOGTest, testMatrixMarketReaderUnweightedUndirected) {
     CSRMatrix csr = MatrixMarketReader{}.read("input/chesapeake.mtx");
     EXPECT_EQ(csr.numberOfRows(), 39);
--- a/networkit/cpp/io/NetworkitBinaryWriter.cpp
+++ b/networkit/cpp/io/NetworkitBinaryWriter.cpp
@@ -7,6 +7,7 @@
 #include <cstring>
 #include <fstream>
 #include <string>
+#include <unordered_map>
 
 #include <tlx/math/clz.hpp>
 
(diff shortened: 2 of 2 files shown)


fabratu:
Could you also add a test for reading/writing graphs with non-cont. node ids?

angriman:
Thank you, could you please add a test where the previous implementation fails?

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1225,4 +1225,18 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
...
+    G.removeNode(10);
+    NetworkitBinaryReader reader;
+    NetworkitBinaryWriter writer;
+    std::string path = "output/test.gt";
+    writer.write(G, path);
+    Graph Gread = reader.read(path);
+    EXPECT_EQ(G.numberOfNodes(), Gread.numberOfNodes());
+    EXPECT_EQ(G.numberOfEdges(), Gread.numberOfEdges());
+    EXPECT_EQ(G.isDirected(), Gread.isDirected());
+    EXPECT_EQ(G.isWeighted(), Gread.isWeighted());
```
angriman, line 1239: This only checks the **number** of nodes and edges. Shouldn't we also check that the two graphs are the same? My suggestion is to add such a check here for now. Then, in another PR, we could implement it in the `==` operator in `Graph` or as a utility function in `GraphTools`.

angriman (commented):
LGTM. I left a few minor comments.

networkit/cpp/io/test/IOGTest.cpp
fabratu: Maybe for this function it is good to have both. Implment it in `GraphTools`, but also overload the operator and call the `GraphTools` function.

fabratu: For two graphs to be the same (not isomorphism) we also need to check the existence of every node/edge (+ possible weights). For edges also the endpoints should be the same.

fabratu:
@angriman Pushed an update, since @Relux-the-Relux is currently working on smth. else and this PR should be in the next (soonish) release. Ready for review.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1227,6 +1227,24 @@ TEST_F(IOGTest, testNetworkitBinaryZigzag) {
...
+    NetworkitBinaryWriter{}.write(G, path);
+    Graph GRead = NetworkitBinaryReader{}.read(path);
+    EXPECT_EQ(G.numberOfNodes(), GRead.numberOfNodes());
+    EXPECT_EQ(G.numberOfEdges(), GRead.numberOfEdges());
+    EXPECT_EQ(G.isDirected(), GRead.isDirected());
+    EXPECT_EQ(G.isWeighted(), GRead.isWeighted());
+    G.forNodes([&](node u) {
+        G.forEdgesOf(u, [&](node v) {
+            ASSERT_TRUE(GRead.hasEdge(u, v));
+            ASSERT_EQ(G.weight(u, v), GRead.weight(u, v));
```
angriman, line 1243:
```suggestion
            EXPECT_TRUE(GRead.hasEdge(u, v));
            EXPECT_DOUBLE_EQ(G.weight(u, v), GRead.weight(u, v));
```

angriman (commented):
Looks fine to me, just a minor comment.

angriman (approved):

Jerevia:
@Relux-the-Relux Now writing and reading could work without core dumped but I get a different graph after reading from the dumped file

```python
import networkit as nk

nkG = nk.Graph(weighted=True)

nkG.addNodes(100)
nkG.addEdge(0, 1)
nkG.addEdge(2, 3)
nkG.removeNode(10)

nkG.compactEdges()

nk.writeGraph(nkG, 'graph.bin', nk.Format.NetworkitBinary)
nkG_1 = nk.readGraph('graph.bin', nk.Format.NetworkitBinary)

print(set(nkG.iterNodes()) - set(nkG_1.iterNodes()), set(nkG_1.iterNodes()) - set(nkG.iterNodes()))
```
It prints
```
({99}, {10})
```
It seems that the ID of nodes was compacted when saved in the graph, without retaining information about the deleted nodes

PR #1012, opened 2022-11-20
New Algorithm: Pruned Landmark Labeling

angriman:
This PR brings the Pruned Landmark Labeling algorithm by Akiba et al. [1]. The algorithm computes distance labels which are used to answer shortest-path distance queries.

I wrote this code a few years ago and, finally, I found some time to polish it.

[1] https://dl.acm.org/doi/10.1145/2463676.2465315

### DIFF
--- a/include/networkit/distance/PrunedLandmarkLabeling.hpp
+++ b/include/networkit/distance/PrunedLandmarkLabeling.hpp
@@ -1,0 +1,76 @@
+#ifndef NETWORKIT_DISTANCE_PRUNED_LANDMARK_LABELING_HPP_
+#define NETWORKIT_DISTANCE_PRUNED_LANDMARK_LABELING_HPP_
+
+#include <utility>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class PrunedLandmarkLabeling : public Algorithm {
+
+public:
+    /**
+     * Pruned Landmark Labeling algorithm based on the paper "Fast exact shortest-path distance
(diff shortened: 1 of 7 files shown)


include/networkit/distance/PrunedLandmarkLabeling.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+class PrunedLandmarkLabeling : public Algorithm {
+
+public:
+    /**
+     * Pruned Landmark Labeling algorithm based on the paper "Fast exact shortest-path distance
+     * queries on large networks by pruned landmark labeling" from Akiba et al., ACM SIGMOD 2013.
+     * The algorithm computes distance labels by performing pruned breadth-first searches from each
+     * vertex. Labels are used to quickly retrieve shortest-path distances between node pairs.
+     * Note: this algorithm only works for unweighted graphs.
```
fabratu, line 20: Better use @note.

networkit/distance.pyx
```diff
@@ -1791,3 +1791,53 @@ cdef class ReverseBFS(SSSP):
...
+		self._this = new _PrunedLandmarkLabeling(G._this)
+
+	def __dealloc__(self):
+		self._G = None
+
+	def query(self, node u, node v):
+		"""
+		query(u, v)
+
+		Returns the shortest-path distance between the two nodes.
```
fabratu, line 1829: I think we should add some kind of feedback (or test), if `query` is called before `run`.

fabratu (changes requested):
LGTM, only some minor comments. I merged the PLP-revert, you can also rebase to fix CI errors.

networkit/distance.pyx
angriman: I added `assureFinished()` in the `.cpp` file.

PR #1017, opened 2022-11-30
PLP: Revert loop to previous version

fabratu:
This PR reverts the loop to the version before #983. The other changes (variable initialization) remain.

With the changes from #983 PLP the debug build on Windows does time-out due to infinite loops. Sometimes this also leads to bad allocation errors, likely due to remains of loop datastructure. Note that this cannot be reproduced locally (with the same compiler/environment). If we track the output on CI, we see that different nodes exchange their labels in each run. Since the testing instances are small (n=1k), one update is enough to keep the main loop alive.

Log from CI:
```
2022-11-30T11:58:36.6150800Z 1: ITERATION: 435
2022-11-30T11:58:36.6150978Z 1: Label: 788
2022-11-30T11:58:36.6151267Z 1: Label: 981
2022-11-30T11:58:36.6151590Z 1: Label: 788
2022-11-30T11:58:36.6151831Z 1: nUpdated: 3
2022-11-30T11:58:36.6152120Z 1: ITERATION: 436
2022-11-30T11:58:36.6152379Z 1: Label: 981
2022-11-30T11:58:36.6152610Z 1: Label: 788
2022-11-30T11:58:36.6152838Z 1: Label: 981
2022-11-30T11:58:36.6153072Z 1: nUpdated: 3
2022-11-30T11:58:36.6153308Z 1: ITERATION: 437
2022-11-30T11:58:36.6153540Z 1: Label: 788
2022-11-30T11:58:36.6153765Z 1: Label: 981
2022-11-30T11:58:36.6153986Z 1: Label: 788
2022-11-30T11:58:36.6154221Z 1: nUpdated: 3
2022-11-30T11:58:36.6154457Z 1: ITERATION: 438
2022-11-30T11:58:36.6154687Z 1: Label: 981
2022-11-30T11:58:36.6154916Z 1: Label: 788
2022-11-30T11:58:36.6155144Z 1: Label: 981
2022-11-30T11:58:36.6155379Z 1: nUpdated: 3
2022-11-30T11:58:36.6155613Z 1: ITERATION: 439
2022-11-30T11:58:36.6155833Z 1: Label: 788
2022-11-30T11:58:36.6156058Z 1: Label: 981
2022-11-30T11:58:36.6156288Z 1: Label: 788
2022-11-30T11:58:36.6156523Z 1: nUpdated: 3
...
```

Sometimes the update converges though. Error happens roughly 9 out of 10 times.

Things I noticed:
- The error only happens for test `testLabelPropagationOnManySmallClusters` with k=100 clusters. Other tests with a smaller number of cluster seem not be affected.
- Looking into the details of the implementation it seems there are some possible race conditions. Both `nUpdated` (which was fixed in #983) as also `activeNodes` can be written/read by different threads at the same time. Especially concerning `activeNodes` it is unclear why does not create problems. We can construct cases, where all nodes are set to inactive altough their neigbors should set them active. One explanation would be, that neighbors won't change their label in the same iteration, but I don't think we can rule that out.

Things already tried:
- Remove parallelism
- Fix race conditions
- Use different compiler versions and older Windows-images
- Use different mechanism for insert/query `std::unordered_map`

The only remedy is to exchange `std::unordered_map` back to `std::map` (state before #983).

If necessary I think we should fix the race conditions (not in this PR). Also maybe exchanging `std::map` with a hash table doesn't make sense in this case after all. It seems for lower degrees (below ~1k), a hash table with random access and a lot of inserts may not be faster in the end (example using gcc 11, Linux: https://stackoverflow.com/a/69490123). Altough that might need some proper benchmarking.

### DIFF
--- a/networkit/cpp/community/PLP.cpp
+++ b/networkit/cpp/community/PLP.cpp
@@ -6,7 +6,6 @@
  */
 
 #include <omp.h>
-#include <unordered_map>
 
 #include <networkit/Globals.hpp>
 #include <networkit/auxiliary/Log.hpp>
@@ -80,17 +79,12 @@ void PLP::run() {
                 return; // node is isolated
 
             // neighborLabelCounts maps label -> frequency in the neighbors
-            std::unordered_map<label, edgeweight> labelWeights;
+            std::map<label, double> labelWeights;
 
             // weigh the labels in the neighborhood of v
             G->forNeighborsOf(v, [&](node w, edgeweight weight) {
                 label lw = result.subsetOf(w);
-                auto it = labelWeights.find(lw);
-                // add weight of edge {v, w}
-                if (it == labelWeights.end())
-                    labelWeights.emplace(lw, weight);
-                else
-                    it->second += weight;
+                labelWeights[lw] += weight; // add weight of edge {v, w}
             });
 
             // get heaviest label
@@ -104,7 +98,7 @@ void PLP::run() {
             if (result.subsetOf(v) != heaviest) { // UPDATE
                 result.moveToSubset(heaviest, v); // result[v] = heaviest;
 #pragma omp atomic
-                ++nUpdated;
+                nUpdated += 1;
                 G->forNeighborsOf(v, [&](node u) { activeNodes[u] = true; });
             } else {
                 activeNodes[v] = false;


networkit/cpp/community/PLP.cpp
```diff
@@ -76,38 +75,35 @@ void PLP::run() {
...
+                label heaviest = std::max_element(labelWeights.begin(), labelWeights.end(),
+                                                  [](const std::pair<label, edgeweight> &p1,
+                                                     const std::pair<label, edgeweight> &p2) {
+                                                      return p1.second < p2.second;
+                                                  })
+                                     ->first;
+
+                if (result.subsetOf(v) != heaviest) { // UPDATE
+                    result.moveToSubset(heaviest, v); // result[v] = heaviest;
+                    nUpdated += 1;                    // TODO: atomic update?
```
angriman, line 99: Shouldn't we keep the atomic update?

networkit/cpp/community/PLP.cpp
```diff
@@ -76,38 +75,35 @@ void PLP::run() {
         nUpdated = 0;
 
         G->balancedParallelForNodes([&](node v) {
-            if (!activeNodes[v] || G->degree(v) == 0)
-                return; // node is isolated
```
angriman, line 80: We could also keep this. It simplifies the code and does not hurt.

angriman (commented):
LGTM. I think we can save a couple of changes from the previous PR. Details in the comments.

angriman (approved):

fabratu:
Will merge this early since it is needed for fixing CI problems.

PR #1032, opened 2023-02-14
Fix EdmondsKarp max flow for directed and min cut source set for both directed/undirected graphs 

### LINKED ISSUE #1031: Fix getSourceSet of EdmonKarp max flow

CharJon:
See #1031.

### DIFF
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -92,7 +92,7 @@ edgeweight EdmondsKarp::directedBFS(const std::vector<count> &reverseEdges,
             return gain[sink];
         }
 
-        graph->forInNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
+        graph->forInNeighborsOf(u, [&](node, node v, edgeid eid) {
             if (!sinkReached && (reverseEdges[eid] == none) && (flow[eid] > 0)
                 && (pred[v] == none)) {
                 pred[v] = u;
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -118,6 +118,15 @@ std::vector<node> EdmondsKarp::getSourceSet() const {
                 visited[v] = true;
             }
         });
+        if (graph->isDirected()) {
+            // Follow backwards edges in residual network
+            graph->forInNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
+                if (!visited[v] && flow[eid] > 0) {
+                    Q.push(v);
+                    visited[v] = true;
+                }
+            });
+        }
     }
 
     return sourceSet;
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -18,7 +18,8 @@ namespace NetworKit {
 EdmondsKarp::EdmondsKarp(const Graph &graph, node source, node sink)
     : graph(&graph), source(source), sink(sink) {}
 
-edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node> &pred) const {
+edgeweight EdmondsKarp::BFS(const std::vector<edgeweight> &residFlow,
+                            std::vector<node> &pred) const {
     std::fill(pred.begin(), pred.end(), none);
     pred.resize(graph->upperNodeIdBound(), none);
     std::vector<edgeweight> gain(graph->upperNodeIdBound(), 0);
@@ -55,16 +56,121 @@ edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node
     return 0.0;
 }
 
+edgeweight EdmondsKarp::directedBFS(const std::vector<count> &reverseEdges,
+                                    std::vector<node> &pred) const {
+    std::fill(pred.begin(), pred.end(), none);
+    pred.resize(graph->upperNodeIdBound(), none);
+    std::vector<edgeweight> gain(graph->upperNodeIdBound(), std::numeric_limits<edgeweight>::max());
+
+    std::queue<node> Q;
+    Q.push(source);
+    pred[source] = source;
+    gain[source] = std::numeric_limits<edgeweight>::max();;
+
+    while (!Q.empty()) {
+        node u = Q.front();
+        Q.pop();
+        bool sinkReached = false;
+
+        graph->forNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
+            if (!sinkReached && pred[v] == none && flow[eid] < weight) {
+                pred[v] = u;
+                if (v != sink && !sinkReached) {
+                    Q.push(v);
+                } else {
+                    sinkReached = true;
+                }
+
+                auto residCapa =
+                    weight - flow[eid] + (reverseEdges[eid] == none ? 0 : flow[reverseEdges[eid]]);
+                gain[v] = std::min(gain[u], residCapa);
+            }
+        });
+
+        if (sinkReached) {
+            return gain[sink];
+        }
+
+        graph->forInNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
+            if (!sinkReached && (reverseEdges[eid] == none) && (flow[eid] > 0)
+                && (pred[v] == none)) {
+                pred[v] = u;
+                gain[v] = std::min(gain[u], flow[eid]);
+
+                if (v != sink && !sinkReached) {
+                    Q.push(v);
+                } else {
+                    sinkReached = true;
+                }
+            }
+        });
+
+        if (sinkReached) {
+            return gain[sink];
+        }
+    }
+
+    return 0.0;
+}
+
 void EdmondsKarp::run() {
     if (!graph->hasEdgeIds()) {
         throw std::runtime_error("edges have not been indexed - call indexEdges first");
     }
     flow.clear();
     flow.resize(graph->upperEdgeIdBound(), 0.0);
+    flowValue = 0;
 
+    if (graph->isDirected()) {
+        runDirected();
+    } else {
+        runUndirected();
+    }
+
+    hasRun = true;
+}
+
+void EdmondsKarp::runDirected() {
+    std::vector<count> reverseEdges(graph->upperEdgeIdBound(), none);
+    graph->parallelForEdges([&](node u, node v, edgeid eid) {
+        reverseEdges[eid] = graph->hasEdge(v, u) ? graph->edgeId(v, u) : none;
+    });
+    std::vector<node> pred;
+    edgeweight gain = directedBFS(reverseEdges, pred);
+
+    while (gain > 0) {
+        flowValue += gain;
+        node v = sink;
+        while (v != source) {
+            node u = pred[v];
+
+            if (graph->hasEdge(u, v)) {
+                edgeid eid = graph->edgeId(u, v);
+                unsigned long reverseId = reverseEdges[eid];
+                if (reverseId != none) {
+                    auto reverseFlow = flow[reverseId];
+                    assert(!((reverseFlow > 0) && (flow[eid] > 0)));
+                    flow[reverseId] = std::max(reverseFlow - gain, 0.);
+                    flow[eid] += gain - reverseFlow;
+                } else {
+                    flow[eid] += gain;
+                    assert(flow[eid] <= graph->weight(u, v));
+                }
+            } else {
+                edgeid eid = graph->edgeId(v, u);
+                flow[eid] -= gain;
+                assert(flow[eid] >= 0);
+            }
+
+            v = u;
+        }
+        gain = directedBFS(reverseEdges, pred);
+    }
+}
+
+void EdmondsKarp::runUndirected() {
     std::vector<edgeweight> residFlow(graph->upperEdgeIdBound(), 0.0);
 
-    flowValue = 0;
     do {
         std::vector<node> pred;
         edgeweight gain = BFS(residFlow, pred);
@@ -89,8 +195,6 @@ void EdmondsKarp::run() {
 
     graph->parallelForEdges(
         [&](node, node, edgeid eid) { flow[eid] = std::max(flow[eid], residFlow[eid]); });
-
-    hasRun = true;
 }
 
 edgeweight EdmondsKarp::getMaxFlow() const {
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -7,7 +7,6 @@
  */
 
 #include <limits>
-#include <math.h>
 #include <queue>
 #include <stdexcept>
 
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -6,6 +6,7 @@
  *              Michael Hamann <michael.hamann@kit.edu>
  */
 
+#include <algorithm>
 #include <limits>
 #include <queue>
 #include <stdexcept>
@@ -26,7 +27,7 @@ edgeweight EdmondsKarp::BFS(std::vector<node> &pred) const {
     Q.push(source);
     pred[source] = source;
     gain[source] = std::numeric_limits<edgeweight>::max();
-    while (!Q.empty()) {
+    do {
         node u = Q.front();
         Q.pop();
 
@@ -50,7 +51,7 @@ edgeweight EdmondsKarp::BFS(std::vector<node> &pred) const {
         if (sinkReached) {
             return gain[sink];
         }
-    }
+    } while (!Q.empty());
 
     return 0.0;
 }
--- a/networkit/cpp/flow/EdmondsKarp.cpp
+++ b/networkit/cpp/flow/EdmondsKarp.cpp
@@ -172,12 +172,9 @@ void EdmondsKarp::runDirected() {
 void EdmondsKarp::runUndirected() {
     std::vector<edgeweight> residFlow(graph->upperEdgeIdBound(), 0.0);
 
-    do {
-        std::vector<node> pred;
-        edgeweight gain = BFS(residFlow, pred);
-        if (gain == 0)
-            break;
-
+    std::vector<node> pred;
+    edgeweight gain = BFS(residFlow, pred);
+    while (gain > 0) {
         flowValue += gain;
         node v = sink;
         while (v != source) {
@@ -192,10 +189,8 @@ void EdmondsKarp::runUndirected() {
             }
             v = u;
         }
-    } while (true);
-
-    graph->parallelForEdges(
-        [&](node, node, edgeid eid) { flow[eid] = std::max(flow[eid], residFlow[eid]); });
+        gain = BFS(residFlow, pred);
+    }
 }
 
 edgeweight EdmondsKarp::getMaxFlow() const {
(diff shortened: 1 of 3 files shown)


networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -118,6 +118,15 @@ std::vector<node> EdmondsKarp::getSourceSet() const {
                 visited[v] = true;
             }
         });
+        if (graph->isDirected()) {
+            // Follow backwards edges in residual network
+            graph->forInNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
```
fabratu, line 123: I think you can use the signature `graph->forInNeighborsOf(u, [&](node, node v, edgeid eid)` here.

CharJon: Indeed, weight is unused. Done.

CharJon:
I try to add some more testcases over the next days and remove the "Draft:" afterwards.

fabratu:
Hm, for both of the new test instances the previous implementation of `getSourceSet` also works. Can you give an example or reason, why the function has to be modified to work for directed graphs?

CharJon:
> Hm, for both of the new test instances the previous implementation of `getSourceSet` also works (in a sense, that we have a min cut for edges `(s,t)` with `s in Source` and `t not in Source`) . Can you give an example or reason, why the function has to be modified to work for directed graphs?

For the testcase "testEdmondsKarpDirected2" the original implementation reports {0,2,3} as the sourceSet. But the vertex partitioning {0,2,3}, {1,4} does not imply the min-cut of value 1. From my perspective this is not in line with what the docstring states "@return The set of nodes that form the (smallest) source side of the flow/minimum cut.". At least I as a user expected the function to return an actual partitioning corresponding to the min-cut. Is this not the intended functionality?

CharJon:
I completely dismissed the undirected case in this PR, but a simliar problem seems to exist. If I'm not missing something, a more general fix is needed, instead of my !isDirected distinction. I added a directed version of "testEdmondsKarpDirected2" ("testEdmondsKarpP3") where again, the sourceSet does not yield the correct min-cut.

CharJon:
@fabratu there seems to be an additional bug for directed input graphs. "testEdmondsKarpDirected7", which is a direct translation of a wikipedia example, fails. 

Do you prefer fixing the directed and undirected cases in one PR or in should two separated one be opened (or sth else)?

fabratu:
> Do you prefer fixing the directed and undirected cases in one PR or should a separated one be opened (or sth else)?

I would prefer fixing this in one PR. I looked a bit deeper into the code and my previous comment was wrong indeed. Both the old implementation and your addition fail in some cases. For example, an undirected version of "testEdmondsKarpDirected2" would also fail currently.

Since the implementation tries to save memory for bookkeeping of the residual graph, I think we have to modify the tests in `getSourceSet` to also take into account backward edges. 

For undirected:

```
graph->forNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
            if (!visited[v] && ((u < v && flow[eid] < weight) || (u > v && flow[eid] > 0))) {
```

For directed: iterating like undirected + do the same tests for incoming neighbors.

CharJon:
Okay, I try to implement fixes in the following order (if you agree):

1. Fix flow calculation for directed graphs, via case distinction (to make "testEdmondsKarpDirected7" pass
2. Fix / updated getSourceSet for undirected graphs

For 2.:  I'm not sure if your suggested fix works
```
graph->forNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
            if (!visited[v] && ((u < v && flow[eid] < weight) || (u > v && flow[eid] > 0))) {
```
The last line of the current run method
```
graph->parallelForEdges(
        [&](node, node, edgeid eid) { flow[eid] = std::max(flow[eid], residFlow[eid]); });
```
removes information about direction. But is this line really necessary anyways? Is there a clean definition of undirected flow, that makes it reasonable to take the max?
Also, while at it: I personally would like to rename the residFlow for undirected Graphs. Conceptually "flow" and "residFlow" store flow from u->v and v->u respectively for u <= v. Calling the second one "residualFlow" seems a bit missleading to me.

CharJon:
Done. I tried to keep the new code close to the original one, but there might be room for improvement and some details might be worth to discuss?

- As in the undirected case I did not add any edges, but went with an implicit represenation of the residual network. This leads to a bunch of if/else blocks. Not sure if adding reverse edges would be better from a performance perspective
- To fix 2. from above I encoded the direction of flow via signs into the flow vector (this might be considered hacky, but I tried to avoid adding a member variable which would only be useful in the directed case)

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -55,22 +56,125 @@ edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node
...
+    pred.resize(graph->upperNodeIdBound(), none);
+    std::vector<edgeweight> gain(graph->upperNodeIdBound(), std::numeric_limits<edgeweight>::max());
+
+    std::queue<node> Q;
+    Q.push(source);
+    pred[source] = source;
+    gain[source] = std::numeric_limits<edgeweight>::max();
+    ;
+
+    while (!Q.empty()) {
```
angriman, line 71: Nit: I suggest to change this with a `do { ... } while (!Q.empty());`. `Q` is never empty in the first iteration.

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -55,22 +56,125 @@ edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node
...
+        node u = Q.front();
+        Q.pop();
+        bool sinkReached = false;
+
+        graph->forNeighborsOf(u, [&](node, node v, edgeweight weight, edgeid eid) {
```
angriman, line 76: This can be simplified with:
```cpp
for (node v : G->inNeighborRange(u))
  // ...
```
so that you do not have to check `sinkReached`.

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -55,22 +56,125 @@ edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node
...
+                    weight - flow[eid] + (reverseEdges[eid] == none ? 0 : flow[reverseEdges[eid]]);
+                gain[v] = std::min(gain[u], residCapa);
+            }
+        });
+
+        if (sinkReached) {
+            return gain[sink];
+        }
+
+        graph->forInNeighborsOf(u, [&](node, node v, edgeid eid) {
```
angriman, line 95: As above.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -38,6 +38,9 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
     EXPECT_EQ(1, edKa.getFlow(5, 6));
 
     std::vector<node> sourceSet(edKa.getSourceSet());
+    for (auto u : sourceSet) {
+        std::cout << "u: " << u << std::endl;
+    }
```
angriman, line 43: This was used for debugging right? If you don't need it anymore it should be removed.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(2, 4, 5);
+    G.addEdge(3, 5, 15);
+    G.addEdge(4, 5, 5);
+
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 0, 5);
+    edKa.run();
+
+    EXPECT_EQ(15, edKa.getMaxFlow()) << "max flow is not correct";
```
angriman, line 71: `getMaxFlow` returns a double, I suggest to use `EXPECT_DOUBLE_EQ(15.0, ...);`

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    edKa.run();
+    EXPECT_EQ(1, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 1), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 2), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
+
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 4), sourceSet.end());
```
angriman, line 93: These checks can be simplified with appropriate [container matchers](https://github.com/google/googletest/blob/main/docs/reference/matchers.md#container-matchers). For example:
```cpp
EXPECT_THAT(sourceSet, ::testing::Contains(0));
```

Or maybe even better:
```cpp
EXPECT_THAT(sourceSet, UnorderedElementsAre({0, 1, 2, 3, 4}));
```

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(0, 1);
+    G.addEdge(0, 2);
+    G.addEdge(1, 3);
+    G.addEdge(2, 3);
+    G.addEdge(3, 4);
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 0, 4);
+    edKa.run();
+    EXPECT_EQ(1, edKa.getMaxFlow()) << "max flow is not correct";
```
angriman, line 85: As above.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+
+    EdmondsKarp edKa(G, 0, 3);
+    edKa.run();
+    EXPECT_EQ(15, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 1),
+              std::find(sourceSet.begin(), sourceSet.end(), 2));
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
```
angriman, line 113: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    edKa.run();
+    EXPECT_EQ(1, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 1), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 2), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
+
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 4), sourceSet.end());
```
angriman, line 135: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    EXPECT_EQ(4, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 1), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 2), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 4), sourceSet.end());
+
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 8), sourceSet.end());
```
angriman, line 164: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(1, 0, 2);
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 8, 0);
+    edKa.run();
+    EXPECT_EQ(4, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 8), sourceSet.end());
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
```
angriman, line 188: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(1, 0, 4.4);
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 3, 0);
+    edKa.run();
+    EXPECT_EQ(2.4, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
```
angriman, line 204: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(1, 0, 4.4);
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 3, 0);
+    edKa.run();
+    EXPECT_EQ(0, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 3), sourceSet.end());
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
```
angriman, line 220: Ditto.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -49,6 +52,200 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+    G.addEdge(5, 6, 9);
+    G.indexEdges();
+
+    EdmondsKarp edKa(G, 0, 6);
+    edKa.run();
+    EXPECT_EQ(5, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_NE(std::find(sourceSet.begin(), sourceSet.end(), 0), sourceSet.end());
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 6), sourceSet.end());
```
angriman, line 246: Ditto.

networkit/cpp/flow/EdmondsKarp.cpp
angriman: Actually that's not true, unfortunately neighbor iterators do not provide edge ids, so I'm afraid that we have to keep this loop as it is.

networkit/cpp/flow/test/EdmondsKarpGTest.cpp
```diff
@@ -32,21 +33,196 @@ TEST_F(EdmondsKarpGTest, testEdmondsKarpP1) {
...
+
+    EdmondsKarp edKa(G, 0, 3);
+    edKa.run();
+    EXPECT_DOUBLE_EQ(15, edKa.getMaxFlow()) << "max flow is not correct";
+
+    std::vector<node> sourceSet(edKa.getSourceSet());
+    EXPECT_THAT(sourceSet, testing::Contains(0));
+    // either 1 and 2 are in the set or none
+    EXPECT_EQ(std::find(sourceSet.begin(), sourceSet.end(), 1),
+              std::find(sourceSet.begin(), sourceSet.end(), 2));
```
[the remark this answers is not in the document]
CharJon: Was indeed not correct.

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -6,7 +6,6 @@
  *              Michael Hamann <michael.hamann@kit.edu>
  */
 
-#include <algorithm>
```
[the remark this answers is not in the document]
CharJon: I fail to see where and my IDE tells me to remove it, but I added it again.

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -27,17 +26,76 @@ edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node
...
+    } while (!Q.empty());
+
+    return 0.0;
+}
+
+edgeweight EdmondsKarp::directedBFS(const std::vector<count> &reverseEdges,
+                                    std::vector<node> &pred) const {
+    std::fill(pred.begin(), pred.end(), none);
+    pred.resize(graph->upperNodeIdBound(), none);
+    std::vector<edgeweight> gain(graph->upperNodeIdBound(), std::numeric_limits<edgeweight>::max());
```
angriman, line 63: Since `directedBFS` might is called several times, I suggest having `gain` as a member variable of the class, so that we do not need to allocate space for it multiple times.

networkit/cpp/flow/EdmondsKarp.cpp
```diff
@@ -18,7 +17,7 @@ namespace NetworKit {
 EdmondsKarp::EdmondsKarp(const Graph &graph, node source, node sink)
     : graph(&graph), source(source), sink(sink) {}
 
-edgeweight EdmondsKarp::BFS(std::vector<edgeweight> &residFlow, std::vector<node> &pred) const {
+edgeweight EdmondsKarp::BFS(std::vector<node> &pred) const {
     std::fill(pred.begin(), pred.end(), none);
     pred.resize(graph->upperNodeIdBound(), none);
     std::vector<edgeweight> gain(graph->upperNodeIdBound(), 0);
```
angriman, line 23: Same here.

angriman (commented):
Thank you for addressing the comments. I left two last suggestions that should improve the performance. The rest looks good to me.

CharJon:
@angriman @fabratu this PRs main intention was to fix the bugs and I tried to stay close to the original implementation. Regarding performance: There are definitely places that can be optimized further. Feel free to change them directly before merging.

angriman (approved):

PR #1041, opened 2023-03-19
New algorithm: (partially) dynamic pruned landmark labeling

angriman:
This PR brings part of the dynamic 2-hop Cover labeling algorithm described here https://dl.acm.org/doi/10.1145/3299901.  This implementation supports edge insertions only.


networkit/distance.pyx
```diff
@@ -1841,3 +1841,70 @@ cdef class PrunedLandmarkLabeling(Algorithm):
...
 		return (<_PrunedLandmarkLabeling*>(self._this)).query(u, v)
+
+cdef extern from "<networkit/distance/DynPrunedLandmarkLabeling.hpp>":
+
+	cdef cppclass _DynPrunedLandmarkLabeling "NetworKit::DynPrunedLandmarkLabeling"(_Algorithm):
+		_DynPrunedLandmarkLabeling(_Graph G) except +
+		count query(node u, node v) except +
+		void update(_GraphEvent ev) except +
+
+cdef class DynPrunedLandmarkLabeling(Algorithm):
```
fabratu, line 1852: For the Python implementation, we should wait until #1067 is merged. The PR slightly changes how Cython code for subclasses of `DynAlgorithm` should be written.

include/networkit/distance/DynPrunedLandmarkLabeling.hpp
```diff
@@ -0,0 +1,66 @@
...
+ */
+class DynPrunedLandmarkLabeling final : public PrunedLandmarkLabeling, public DynAlgorithm {
+
+public:
+    /**
+     * Dynamic Pruned Landmark Labeling algorithm based on the paper "Fully Dynamic 2-Hop Cover
+     * Labeling " from D'Angelo et al., ACM JEA 2019. The algorithm computes distance labels by
+     * performing pruned breadth-first searches from each vertex. Distance labels can be updated
+     * efficiently after edge insertions.
+     * @note this algorithm only works for unweighted graphs and only supports edge insertions.
```
fabratu, line 26: What happens for weighted graphs? This case is not tested currently, so likely no error. From the code I would assume, that weights are just ignored?

networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -611,4 +612,78 @@ TEST_P(DistanceGTest, testPrunedLandmarkLabeling) {
...
+    EXPECT_EQ(pll.query(1, 2), 1);
+
+    EXPECT_EQ(pll.query(0, 3), 3);
+    EXPECT_EQ(pll.query(1, 3), 2);
+
+    EXPECT_EQ(pll.query(0, 4), 4);
+    EXPECT_EQ(pll.query(1, 4), 3);
+}
+
+TEST_P(DistanceGTest, testDynPrunedLandmarkLabelingBuildClique) {
```
fabratu, line 654: Really minor: If we look at the coverage, both the [header](https://github.com/networkit/networkit/actions/runs/4460282338/jobs/7833317305#step:6:175) and the [implementation](https://github.com/networkit/networkit/actions/runs/4460282338/jobs/7833317305#step:6:554) are not at 100%. I think we are missing tests for illegal `GraphEvents` and `updateBatch`.

fabratu (changes requested):

include/networkit/distance/DynPrunedLandmarkLabeling.hpp
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Updates the distance labels after an edge insertion on the graph.
+     * @note supports only edge insertions.
+     *
+     * @param e The edge insertion.
+     */
+    void update(GraphEvent e) override;
+
+    /**
+     * Not implemented. The algorithm does not support batch updates.
```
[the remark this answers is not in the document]
bernlu: I think it would be nice if `updateBatch` would just call `update` for each event instead of raising an error. Even though there is no special batch update algorithm, users can still use batches for convenience. Most of our other dynamic algorithms do not support both single and batch updates either, but they all forward the call to the other function.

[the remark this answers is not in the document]
angriman: Unfortunately that would not work. The algorithm does not update the distances properly if one adds multiple edges at a time and then calls `update` for each of them.

include/networkit/distance/DynPrunedLandmarkLabeling.hpp
angriman: True, I clarified the comment.

networkit/distance.pyx
angriman: Yes, it's great that now we can inherit from multiple classes in Cython. I rebased and updated the class accordingly.

networkit/cpp/distance/test/DistanceGTest.cpp
angriman: Done.

angriman:
True, I removed the changes in `extlibs`.

fabratu (approved):

PR #1047, opened 2023-03-23
Fix DynBfs

angriman:
`DynBetweennessGTest.testDynApproxBetweenessGeneratedGraph` fails with segfault, see the screenshot below:

Screenshot 2023-03-23 at 23 51 46

I compiled NetworKit on macOS 13.2.1 (M2 chip), llvm Clang 15.0.7 and with these flags:
```
cmake -DNETWORKIT_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -DNETWORKIT_WITH_SANITIZERS=address -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_STANDARD=17 ..
```

This happens for two reasons:
- There seems to be a pretty obvious mistake in `DynBFS.cpp`: the `queues` vector has size `maxDistance + 1`
https://github.com/networkit/networkit/blob/333cb9a7e431dfadf016323491dcfd24227112c5/networkit/cpp/distance/DynBFS.cpp#L77 and the heap-buffer-overflow happens at this line
https://github.com/networkit/networkit/blob/333cb9a7e431dfadf016323491dcfd24227112c5/networkit/cpp/distance/DynBFS.cpp#L88 which could access the vector outside its. Thus, the size of `queues` should be `maxDistance + 2`.
- `maxDistance` is not correctly updated. I observed that, when `updateBatch` is called, `maxDistance` is sometimes smaller than the highest non-infinite value in `distances`.

This PR fixes these issues. For the second one, I simply get the highest non-infinite value in `distances`. This is not the optimal solution, but it should not pessimize considerably the performance of the algorithm. The logic of the algorithm is quite complex, and I did not have time to find the bug in updating `maxDistance`.

Additionally, I included a few improvements in a separate commit.

### DIFF
--- a/networkit/cpp/distance/DynBFS.cpp
+++ b/networkit/cpp/distance/DynBFS.cpp
@@ -78,8 +78,9 @@ void DynBFS::update(GraphEvent e) {
 void DynBFS::updateBatch(const std::vector<GraphEvent> &batch) {
     std::vector<std::queue<node>> queues(maxDistance + 2);
     mod = false;
+
     // insert nodes from the batch whose distance has changed (affected nodes) into the queues
-    for (GraphEvent edge : batch) {
+    for (const GraphEvent &edge : batch) {
         if (edge.type == GraphEvent::EDGE_ADDITION) {
             if (distances[edge.u] == infDist && (G->isDirected() || distances[edge.v] == infDist))
                 continue;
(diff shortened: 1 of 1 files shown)


fabratu:
Hm, I would rather have this fixed than introduce an additional O(n). After looking at the code it seems not really clear, how an erroneous `maxDistance` can be produced. 

Likely, it is related to an edge deletion. Do you have maybe an example for reproducing this behavior?

angriman:
Strangely, I could not reproduce the error, it was probably somewhere else. I added an assertion like in this picture, but it did not fail. I can avoid adding the additional loop.
Screenshot 2023-04-06 at 13 45 17

fabratu (approved):

PR #1050, opened 2023-03-29
Fix unused variable warnings.

angriman:
- Remove unused variables
- Use `tlx::unused` for variables used only in debug mode
- Use pre-increment instead of post-increment where possible

### DIFF
--- a/networkit/cpp/centrality/ApproxSpanningEdge.cpp
+++ b/networkit/cpp/centrality/ApproxSpanningEdge.cpp
@@ -10,6 +10,8 @@
 #include <omp.h>
 #include <queue>
 
+#include <tlx/unused.hpp>
+
 #include <networkit/centrality/ApproxSpanningEdge.hpp>
 #include <networkit/components/BiconnectedComponents.hpp>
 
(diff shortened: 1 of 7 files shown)


fabratu:
In addition to the changes mentioned, you also change `&` to `&&`. The reasoning is, that logic comparison should be done with logic operators?

Also, can you split the changes concerning `unused` and the ones concerning style/opt?

angriman:
> In addition to the changes mentioned, you also change `&` to `&&`. The reasoning is, that logic comparison should be done with logic operators?

Yes, we were getting this warning: https://clang.llvm.org/docs/DiagnosticsReference.html#wbitwise-instead-of-logical

> Also, can you split the changes concerning `unused` and the ones concerning style/opt?

Sure!

fabratu:
Hm, I should have been more attentive concerning your PR. Seems I fixed a lot of problems for the second time. In addition, switching to pre-increment also seems to make some `unused` markings unnecessary.

After a rebase, we can merge this PR right away. Sorry for creating even more work ;)

angriman:
No worries, conflicts were minimal. Now it should be ok!

fabratu (approved):

PR #1054, opened 2023-04-27
CI: Add pyproject.toml to fix editable pip install, update environments + fix styling / build errors.

fabratu:
This PR introduces `pyproject.toml` as a Python packaging file. 

Over time, `setup.py` will be partially or fully migrated to the `toml` file. While for a normal installation the file is not (yet) required, an editable installation using the most recent `setuptools` (like done on the CI) uses `[build-system]` requirements to detect the environment. Without it, `Cython` is not detected (see here for an example run: https://github.com/networkit/networkit/actions/runs/4808153671/jobs/8557789989

Furthermore `libomp` and `clang` are updated to version 16.X for latest macOS and Linux builds. For clang-tidy and clang-format builds version 15.X of clang is still used. Changing it introduces new issues, which will be fixed out of this PR.


fabratu:
Due to new errors for different compilers and tools, more pushes + a bit of trial-and-error is needed until the PR is finished. Will squash all commits in the end.

fabratu:
Ok, PR should be final now. CI is fixed and environments are up-to-date.

include/networkit/Globals.hpp
```diff
@@ -23,16 +23,16 @@ using omp_index = int64_t;
...
 using node = index;        ///< node indices are 0-based
 using edgeweight = double; ///< edge weight type
 using edgeid = index;      ///< edge id
 
-constexpr index none = std::numeric_limits<index>::max(); ///< value for not existing nodes/edges
 constexpr edgeweight defaultEdgeWeight = 1.0;
 constexpr edgeweight nullWeight = 0.0;
-
-using coordinate = double;
+constexpr index none = std::numeric_limits<index>::max(); ///< value for not existing nodes/edges
```
angriman, line 35: Are these changes necessary?

networkit/cpp/centrality/ApproxSpanningEdge.cpp
```diff
@@ -161,7 +161,7 @@ void ApproxSpanningEdge::run() {
 
             queue.push(source);
             toVisit[source] = NodeStatus::NOT_IN_COMPONENT;
-            count nVisited = 0;
+            [[maybe_unused]] count nVisited = 0;
```
angriman, line 164: For better consistency with the rest of the codebase, can't we use `tlx::unused`? We could replace it with something else (e..g, `[[maybe_unused]]`) in the future.

networkit/cpp/distance/test/SSSPGTest.cpp
```diff
@@ -67,8 +67,8 @@ TEST_F(SSSPGTest, testShortestPaths) {
     });
     count dist = bfs.distance(x);
     std::set<std::vector<node>> paths = bfs.getPaths(x, true);
-    count i = 0;
-    for (auto path : paths) {
+    [[maybe_unused]] count i = 0;
+    for (auto const &path : paths) {
```
angriman, line 71: For consistency with the codebse, use west-const:
```suggestion
    for (const auto &path : paths) {
```

angriman (commented):
Thanks for the fixes, I left a few comments.

include/networkit/Globals.hpp
fabratu: Unfortunately yes ... with clang-format 15 and 16 there is a more aggressive alignment of comments. Without rearranging the include guards and clang-format offend each other on line `#endif // _MSC_VER`. The other fix would be either to change the include guards or disable the comment alignment.

networkit/cpp/centrality/ApproxSpanningEdge.cpp
fabratu: Hm, good point. Will exchange the occassions.

angriman (approved):

fabratu:
@angriman Thanks for the review! I will merge this early in order to fix the pipeline on the other PRs.

PR #1057, opened 2023-05-05
Python unit tests: improve structure and order functions lexicographically

PetholzA:
splits test_algorithms.py into the remaining test modules. it also renames / orders various functions and classes to make it more consistent.


angriman:
Thank you, python tests needed a better structure. To simplify reviewing this PR, could you please split it in multiple commits?

fabratu:
Hm, I think that is too much hassle. Over 90% of the commit is moving the tests from `test_algorithms.py` to the appropriate module files. The other ~10% are renaming tests, removing comments and introduce sorting. Since I discussed the process leading to this PR already with @PetholzA, I will also take care of the review. 

Also, reworking/adding tests will be out-of-scope of this PR (we will start working on that shortly).

networkit/test/test_centrality.py
```diff
@@ -4,11 +4,92 @@
 import networkit as nk
 import numpy as np
 import scipy
-
+import random
 
 class Test_Centrality(unittest.TestCase):
```
fabratu, line 9: Sry for the inconvinience - can you please rename the classes to UpperCamelCase?

networkit/test/test_centrality.py
```diff
@@ -4,11 +4,92 @@
...
+		if not c1.numberOfSubsets() == c2. numberOfSubsets(): return False
+		for i in range(0,c1.numberOfElements()):
+			if not c1.subsetsOf(i) == c2.subsetsOf(i): return False
+		return True
+	
+	def setUp(self):
+		self.L = nk.readGraph("input/looptest1.gml", nk.Format.GML) #without self-loops
+		self.LL = nk.readGraph("input/looptest2.gml", nk.Format.GML) #with self-loops sprinkled in
+	
+	def testCentralityApproxBetweenness(self):
```
fabratu, line 22: I think we can delete the module name from the function definition without loosing information. On the command line a single test execution looks like this: 

```
testCentralityCoreDecomposition (test_centrality.Test_Centrality.testCentralityCoreDecomposition) ... ok
```

fabratu (changes requested):
In addition to the change requests - can you please give a proper commit message?

fabratu (approved):

Issue #1007, opened 2022-11-07
Accessing degreeIn of a node in an empty graph segfaults.


jfrog-tamirb:
When I run the following code:

```python
import networkit as nk
nk.Graph().degreeIn(0)
```

With:
- Networkit 10.0
- The Python:3.10 docker image, or Windows

I get a segfault:
```
root@8e2595a8e77c:/# python
Python 3.10.5 (main, Jun 23 2022, 10:42:52) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import networkit as nk
<frozen importlib._bootstrap>:241: ReducedFunctionalityWarning: WARNING: module 'pandas' not found, some functionality will be restricted
>>> nk.Graph().degreeIn(0)
Segmentation fault
```

This seems to be the case for many operations or queries. 
It seems odd when other operations, like `Graph.addEdge`, raise an exception if a node is missing.

When writing in Python, I'd expect all operations to be bounds-checked, and raise exceptions if they fail.

If this is not the case - is there a general rule to follow as to which operations are bounds-checked and which are not?

fabratu:
Thanks for reaching out.

In general we omit checks for performance critical functions (if the check is itself is costly). However this is normally included in the documentation.

Concerning your issue: Functions which include acces to a node, should always include an existience check (since it is cheap). As for `degreeIn` and related functions, this is missing and therefore a bug. Will fix this.

A temp. workaround would be to call `G.hasNode(u)`[1] before calling `degreeIn`.

[1] https://networkit.github.io/dev-docs/python_api/graph.html?highlight=hasnode#networkit.graph.Graph.hasNode

fabratu:
Based on a discussion in #1058 we did not include a check in `degreeIn` since it is not so cheap if used repeatedly (which is the case for many algorithms). Also, from a design perspective - in order to maintain the high throughput and speed and NetworKit - sanity checks are for most parts disabled in the release package.

If you need to access neighbors of a node, which might be present or not, calling `G.hasNode(u)` is likely the best workaround.

Will close this issue for now. Feel free to reopen it if there are any more questions.

PR #1058, opened 2023-05-08
Graph.hpp: fixes segmenation fault for degree functions

PetholzA:
fixes {degree, degreeOut, degreeIn} functions when called with an empty Graph


fabratu:
Thanks for the PR. Please mention the linked issue (using `#` + the number of the issue).

PetholzA:
#1007

include/networkit/graph/Graph.hpp
```diff
@@ -1752,15 +1756,21 @@ class Graph final {
      * @return The number of incoming neighbors.
      * @note If the graph is not directed, the outgoing degree is returned.
      */
-    count degreeIn(node v) const { return directed ? inEdges[v].size() : outEdges[v].size(); }
+    count degreeIn(node v) const { 
+        if (!hasNode(v))
+            throw std::runtime_error("Error, the node does not exist!");
```
angriman, line 1761:
```suggestion
    count degreeIn(node v) const { 
        assert(hasNode(v));
```

angriman:
I still disagree with this change. There are examples of algorithms that often use this function, for example [1], [2], [3], and many more. This change would slow them down. I would rather update the documentation of `degree` and `degreeIn` and mention that bound checks are not performed, as done for example by std::vector [4].
As already mentioned, one can always use `assert` to have an informative message if the issue occurs without pessimizing the performance.

[1] https://github.com/networkit/networkit/blob/master/networkit/cpp/centrality/TopCloseness.cpp#L80
[2] https://github.com/networkit/networkit/blob/master/networkit/cpp/centrality/GroupDegree.cpp#LL43C44-L43C44
[3] https://github.com/networkit/networkit/blob/master/networkit/cpp/centrality/KadabraBetweenness.cpp#L645
[4] https://en.cppreference.com/w/cpp/container/vector/operator_at

hmeyerhenke:
I have not looked at the code, but I agree with Eugenio's reasoning. We should avoid performance degradation if other options are available.

fabratu:
In principle, I agree. As already pointed out: For once, we have decided not to integrate checks, if they hamper the performance. Also, `degree`/`degreeIn` are used a lot in algorithms.

Nevertheless, I asked @PetholzA to create a PR to have a discussion about how to find a better / more consistent sweet spot between performance and safety checks. The reasoning is the following: Especially for `Graph.hpp` we include existence checks for nodes in for several functions (also performance critical ones like `forNodes`). Some functions don't have checks at all. We also occasionally use optional parameters to set call functions safely. So there is inconsistency, which create issues for (new) users.

What do you think would be a good option here?
(i) We could of course leave everything as it is. This would lead to problems, especially for Python users, where it is not common to have zero safety checks. Also, this might lead to confusion due to inconsistency.
(ii) Use default parameters. If I have not missed something in the language standard, the C++ function calls can get more messy, since we often have more than one optional parameter. We could add wrapper functions in this case.
(iii) Use asserts. From a user perspective (especially end-users with release builds), I don't think this helps a lot. For development / testing, that makes sense.

angriman:
> In principle, I agree. As already pointed out: For once, we have decided not to integrate checks, if they hamper the performance. Also, `degree`/`degreeIn` are used a lot in algorithms.
> 
> Nevertheless, I asked @PetholzA to create a PR to have a discussion about how to find a better / more consistent sweet spot between performance and safety checks. The reasoning is the following: Especially for `Graph.hpp` we include existence checks for nodes in for several functions (also performance critical ones like `forNodes`). Some functions don't have checks at all. We also occasionally use optional parameters to set call functions safely. So there is inconsistency, which create issues for (new) users.
> 
> What do you think would be a good option here?
> 
> (i) We could of course leave everything as it is. This would lead to problems, especially for Python users, where it is not common to have zero safety checks. Also, this might lead to confusion due to inconsistency. (ii) Use default parameters. If I have not missed something in the language standard, the C++ function calls can get more messy, since we often have more than one optional parameter. We could add wrapper functions in this case. (iii) Use asserts. From a user perspective (especially end-users with release builds), I don't think this helps a lot. For development / testing, that makes sense.

Because NetworKit is a "tool suite for high-performance network analysis", we should not sacrifice performance just to inform users that they are making a mistake in **their** code (e.g., accessing the degree of a deleted or non-existing node). My suggestion is to go with (iii), plus clarify in the documentation that `Graph` does not implement bound checks for performance reasons.

angriman:
@PetholzA thanks for addressing the comments, this looks good to me. It can be merged once the code style issue is fixed.

fabratu:
After having more discussions about this topic - I agree. Let's add asserts where possible for sanity checks and also use the documentation to inform users about possible unsafe behavior of a function. @PetholzA Please add a hint (in both C++ and Python), that the function does not check for existence of a node (and reference `hasNode` to check for it).

Besides that (and the aforementioned style issues), the PR looks good to me.

angriman (approved):

PetholzA:
note that we also changed the GraphNotebook.ipynb, which included a error

fabratu (approved):

PR #1059, opened 2023-05-08
community.pyx: fixes getValues function

PetholzA:
removes reference from vector, fixes the output function


fabratu:
Due to another PR (#1057) changing the structure of Python tests, this bug fix will not add the missing test for the Python interface. We will try to increase the coverage more thoroughly after the restructuring is merged.

fabratu (approved):

PR #1064, opened 2023-05-16
Fix #1063: BiconnectedComponents bug

### LINKED ISSUE #1063: Biconnected components algorithm breaks if nodes got deleted

CharJon:
Fixes #1063


angriman:
Thank you for fixing this bug! The changes look fine to me, I just recommend writing more informative commit messages.
For example:
- Add `BiconnecteedComponents` testcase with a deleted node.
- Use `emplace` instead of `std::make_pair`.
- Fix segmentation fault (I assume that's what happens) in `BiconnectedComponents`.

angriman (approved):

PR #1067, opened 2023-05-23
Add DynAlgorithm bindings to python

bernlu:
This PR adds python bindings for the `DynAlgorithm` class in a new `dynalgorithm.pxd`/`dynalgorithm.pyx` file and updates all existing dynamic algorithms to use this base class.
Previously, all python bindings of dynamic algorithms implement the same `update` and `updateBatch` forwarding code, which is now moved to the `DynAlgorithm` base class.

The setup is a little weird because of Cython limitations. First, a figure of the new class hierarchy, where `MyAlgorithm` is any dynamic algorithm:
```
      _CythonParentClass
       ^              ^
       |              |
Algorithm          _CythonSubclassDynAlgorithm
       ^                     ^
       |                     |
MyAlgorithm    -->    DynAlgorithm
```

Explanation:
- Cython (or Python in general) does not allow multiple inheritance of `cdef class`es (also called 'extension types'). Since we use multiple inheritance for dynamic algorithms in c++ (inherit from both `Algorithm` and `DynAlgorithm`), we do need this functionality though. The workaround is to define a wrapper class (`DynAlgorithm`) in pure python which wraps the cdef class. Then we can inherit from `Algorithm` and the pure python wrapper.
- Cython also requires that all classes in a inheritance hierarchy have a common base class. Since there is no such class in C++ (and adding one would probably decrease performance because of more vtables), we instead add a dummy base class just for Cython (`CythonParentClass`). We move the `_Algorithm` pointer to this class s.t. both the `Algorithm` and the `DynAlgorithm` subclass trees have access to that pointer.

In the `_CythonSubclassDynAlgorithm.update` and `updateBatch` methods, we have to use a `dynamic_cast` call to cast the `_Algorithm` pointer to a `_DynAlgorithm` pointer. Regarding performance, I would not expect users to call this often enough to matter, but I also did a quick test and the difference seems to be around 100ns for a single call of dynamic_cast in this context.

With this new setup, implementation of a new dynamic algorithm `MyAlgorithm` now requires less boilerplate and repeated code: 
```py
from .base cimport _Algorithm, Algorithm
from .dynbase cimport _DynAlgorithm
from .dynbase import DynAlgorithm

cdef extern from "<MyAlgorithm.hpp>":
	cdef cppclass _MyAlgorithm "NetworKit::MyAlgorithm"(_Algorithm, _DynAlgorithm):
    # ...

cdef class MyAlgorithm(Algorithm, DynAlgorithm):
    # ... methods specific to MyAlgorithm
    # run, update, updateBatch are inherited from Algorithm and DynAlgorithm
```

### DIFF
--- a/networkit/base.pyx
+++ b/networkit/base.pyx
@@ -1,6 +1,23 @@
 # distutils: language=c++
 
-cdef class Algorithm:
+
+cdef class _CythonParentClass:
+	""" Abstract base class for Cython 
+	The purpose of this class is to provide a single combined base cdef class for all classes that we use in this project.
+	Cython cannot handle multiple inheritance without this single common base class."""
+	def __init__(self, *args, **namedargs):
+		if type(self) == _CythonParentClass:
+			raise RuntimeError("Error, you may not use _CythonParentClass directly, use a sub-class instead")
+
+	def __cinit__(self, *args, **namedargs):
+		self._this = NULL
+
+	def __dealloc__(self):
+		if self._this != NULL:
+			del self._this
+		self._this = NULL
+
+cdef class Algorithm(_CythonParentClass):
 	""" Abstract base class for algorithms """
 	def __init__(self, *args, **namedargs):
 		if type(self) == Algorithm:
--- a/networkit/base.pyx
+++ b/networkit/base.pyx
@@ -4,7 +4,8 @@
 cdef class _CythonParentClass:
 	""" Abstract base class for Cython 
 	The purpose of this class is to provide a single combined base cdef class for all classes that we use in this project.
-	Cython cannot handle multiple inheritance without this single common base class."""
+	Cython cannot handle multiple inheritance without this single common base class.
+	"""
 	def __init__(self, *args, **namedargs):
 		if type(self) == _CythonParentClass:
 			raise RuntimeError("Error, you may not use _CythonParentClass directly, use a sub-class instead")
--- a/networkit/dynbase.pyx
+++ b/networkit/dynbase.pyx
@@ -1,0 +1,70 @@
+# distutils: language=c++
+
+from libcpp.cast cimport dynamic_cast
+
+from .dynamics cimport _GraphEvent, GraphEvent
+from .base cimport _CythonParentClass
+from typing import List
+
+ctypedef _DynAlgorithm* _DynAlgorithmPtr
+
+cdef class _CythonSubclassDynAlgorithm(_CythonParentClass):
+	""" Abstract base class for dynamic algorithms
+    Cython does not allow direct inheritance from two extension types (cdef classes).
+    Instead, inherit from the DynAlgorithm class! """
+	def __init__(self, *args, **namedargs):
+		if type(self) == _CythonSubclassDynAlgorithm:
+			raise RuntimeError("Error, you may not use _CythonSubclassDynAlgorithm directly, use a sub-class instead")
+
+	def __cinit__(self, *args, **namedargs):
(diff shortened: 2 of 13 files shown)


fabratu:
As additional information. Importing the necessary `cppclass _DynAlgorithm` and `class DynAlgorithm` now has to be split (as seen on the PR changes):

```
from .dynbase cimport _DynAlgorithm
from .dynbase cimport DynAlgorithm
...
```

We found no way, creating a workaround using only `cimport`.

networkit/dynbase.pyx
```diff
@@ -0,0 +1,80 @@
...
+
+from .dynamics cimport _GraphEvent, GraphEvent
+from .base cimport _CythonParentClass
+from typing import List
+
+ctypedef _DynAlgorithm* _DynAlgorithmPtr
+
+cdef class _CythonSubclassDynAlgorithm(_CythonParentClass):
+	""" Abstract base class for dynamic algorithms
+    Cython does not allow direct inheritance from two extension types (cdef classes).
```
fabratu, line 13: Indentation for the the docstring looks of. Also move the closing `"""` to another line.

networkit/dynbase.pyx
```diff
@@ -0,0 +1,80 @@
...
+		update(ev)
+
+		The generic update method for updating data structure after an update.
+
+		Parameters
+		----------
+		ev : networkit.dynamics.GraphEvent
+			A graph event.
+
+        Returns
```
fabratu, line 38: Indentation error.

networkit/dynbase.pyx
```diff
@@ -0,0 +1,80 @@
...
+		updateBatch(batch)
+
+		The generic update method for updating data structure after a batch of updates.
+
+		Parameters
+		----------
+		batch : list(networkit.dynamics.GraphEvent)
+			List of graph events.
+
+        Returns
```
fabratu, line 61: As above.

networkit/base.pyx
```diff
@@ -1,6 +1,23 @@
 # distutils: language=c++
 
-cdef class Algorithm:
+
+cdef class _CythonParentClass:
+	""" Abstract base class for Cython 
+	The purpose of this class is to provide a single combined base cdef class for all classes that we use in this project.
+	Cython cannot handle multiple inheritance without this single common base class."""
```
fabratu, line 7: Move the closing `"""` to another line.

fabratu (changes requested):

fabratu (approved):
LGTM.

bernlu:
Small change: in our `*.pyx` files where we define `cdef cppclass _MyAlgorithm` we should not re-declare `update` and `updateBatch` (it suffices to declare them once in `dynbase`).
This is how we handle `run` as well (most of the time).
It seems like Cython can deal these multi-declarations, even when sometimes functions are declared with `nogil` and sometimes without, but for consistency they should be declared only once in base/dynbase. 
Fixed with this new commit.

fabratu:
> Small change: in our `*.pyx` files where we define `cdef cppclass _MyAlgorithm` we should not re-declare `update` and `updateBatch` (it suffices to declare them once in `dynbase`). This is how we handle `run` as well (most of the time). It seems like Cython can deal these multi-declarations, even when sometimes functions are declared with `nogil` and sometimes without, but for consistency they should be declared only once in base/dynbase. Fixed with this new commit.

Right, missed that in the previous review. Only checked, that update/updateBatch from cdef class were gone.

PR #1069, opened 2023-05-24
Fix #1068: KruskalMST documentation implementation discrepancy

### LINKED ISSUE #1068: Kruskal MSF calculates maximum spanning tree, but docu states minimum spanning tree

CharJon:
Fixes #1068

### DIFF
--- a/networkit/cpp/graph/test/SpanningGTest.cpp
+++ b/networkit/cpp/graph/test/SpanningGTest.cpp
@@ -34,7 +34,7 @@ TEST_F(SpanningGTest, testKruskalMinSpanningForest) {
     }
 }
 
-TEST_F(SpanningGTest, testKruskalMinSpanningForestSimple) {
+TEST_F(SpanningGTest, testKruskalMinSpanningForestSimple1) {
     Graph g(5, true);
     g.addEdge(0, 1, 1);
     g.addEdge(1, 2, 1);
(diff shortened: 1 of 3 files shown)


CharJon:
The "getTotalWeight" functionality is obviously not  requeired, but I find it useful and this might be worth adding to the spanning forest interface?

networkit/cpp/graph/test/SpanningGTest.cpp
```diff
@@ -32,7 +32,7 @@ TEST_F(SpanningGTest, testKruskalMinSpanningForest) {
     }
 }
 
-TEST_F(SpanningGTest, testKruskalMinSpanningForestSimple) {
+TEST_F(SpanningGTest, testKruskalMinSpanningForestSimple1) {
```
angriman, line 35: Please use a more representative name for these two tests. For example, in this test the MFS is a MST since all edges have the same weight so we could call it `testKruskalMinimumSpanningForestIsMinimumSpanningTree` or `testKruskalMinimumSpanningForestUnweightedGraph`.

For the second test, we could just test `getTotalWeight` since the correctness of the algorithm is already tested above; so I suggest to remove the last part of the test and name it `testKruskalMinimumSpanningForestGetTotalWeight`.

angriman (changes requested):

fabratu:
@CharJon Thanks for the bug fix! If you are still interested in merging this and need assistance, just let us know. Otherwise, besides the comments by @angriman the PR looks good in general.

CharJon:
> @CharJon Thanks for the bug fix! If you are still interested in merging this and need assistance, just let us know. Otherwise, besides the comments by @angriman the PR looks good in general.

I will try to come back to this next week. I do not have strong opinions on test names.

CharJon:
Updated my suggestion. Feel free to change any test-names etc.
I do not see any downsides of testing for the solution to be a forest in every test, but this can obviously be removed.
I still think the getTotalWeight functionality would be nice for the general msf interface, but I think this is fine for now.

fabratu:
Thanks for posting an update. Changes look good to me now.  There are still some failed tests on CI, but the culprit is likely a failed sync with the LLVM-repos (so not related to your code).

If that is not the case, you might need to rebase the PR on top of master, due to an update for the code styling / clang tidy checks.

fabratu (dismissed):

angriman (approved):

fabratu (approved):

fabratu (approved):

PR #1062, opened 2023-05-16
[WIP] MatricesGTest.hpp & test_centrality_py: adds testing functions

PetholzA:
adds testMatrixToGraph() in cpp, and new test functions for: test_centrality test_clique and test_community in python 

this not the final commit! it is meant to to check if the coverage increases

PR #1074, opened 2023-06-02
Unit testing: adds test_algebraic.py and more tests for MatricesGtest.cpp

PetholzA:
this PR splits up #1062 into semantically linked parts

### DIFF
--- a/networkit/cpp/algebraic/test/MatricesGTest.cpp
+++ b/networkit/cpp/algebraic/test/MatricesGTest.cpp
@@ -120,10 +120,23 @@ public:
     template <class Matrix>
     void testParallelForNonZeroElementsInRowOrder();
 
+    template <class Matrix>
+    void testMatrixToGraph();
+
     // TODO: Test other matrix classes
 
     // TODO: Test mmT multiplication, etc.!
 };
+
+template <class Matrix>
+void MatricesGTest::testMatrixToGraph() {
+
+    Matrix mat = Matrix::adjacencyMatrix(METISGraphReader{}.read("input/power.graph"));
+    Graph G = MatrixTools::matrixToGraph(mat);
+
+    EXPECT_EQ(G.numberOfNodes(), 4941);
+    EXPECT_EQ(G.numberOfEdges(), 6594);
+}
 
 template <class Matrix>
 void MatricesGTest::testDimension() {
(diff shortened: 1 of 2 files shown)


networkit/test/test_algebraic.py
```diff
@@ -0,0 +1,32 @@
...
+		self.assertEqual(A[3][3], 0.0)
+		self.assertEqual(A[4][4], 0.0)
+		A2 = nk.algebraic.adjacencyMatrix(G, "sparse")
+		self.assertIsInstance(A2, scipy.sparse.csr_matrix)
+		B = nk.algebraic.laplacianMatrix(G)
+		self.assertIsInstance(B, scipy.sparse.coo_matrix)
+		eigenB = nk.algebraic.laplacianEigenvectors(G)
+		self.assertIsInstance(eigenB[0][0], np.complex128)		
+
+	def testEigenvectorValues(self):
```
angriman, line 26: Here you are just testing that the first eigenvector type is complex. Shouldn't you check that the function produces the expected eigenvectors?

networkit/cpp/algebraic/test/MatricesGTest.cpp
```diff
@@ -1200,6 +1212,11 @@ TEST_F(MatricesGTest, testDimension) {
     testDimension<DenseMatrix>();
 }
 
+TEST_F(MatricesGTest, testMatrixToGraph) {
+    testMatrixToGraph<DynamicMatrix>();
+    testMatrixToGraph<CSRMatrix>();
+}
```
angriman, line 1218: There is a more elegant way to instantiate a type-parametrized test suites: http://google.github.io/googletest/reference/testing.html#TYPED_TEST_SUITE_P

Anyway, this is not something introduced by this PR. This is ok for now, but we should change it later.

networkit/cpp/algebraic/test/MatricesGTest.cpp
```diff
@@ -120,11 +120,23 @@ class MatricesGTest : public testing::Test {
...
     // TODO: Test mmT multiplication, etc.!
 };
 
+template <class Matrix>
+void MatricesGTest::testMatrixToGraph() {
+
+    Matrix mat = Matrix::adjacencyMatrix(METISGraphReader{}.read("input/power.graph"));
+    Graph G = MatrixTools::matrixToGraph(mat);
+
+    EXPECT_EQ(G.numberOfEdges(), 6594);
```
angriman, line 137: What about testing also the number of nodes and that the produced edges are correct? Further, which kind of graph do we expect? Should be weighted/unweighted, directed/undirected? I think these properties should be tested.

angriman (changes requested):
Thank you very much for the new tests. I left some comments.

fabratu:
Currently, the coverage is at 81%. Since the module is small, try to reach 100%.

angriman (approved):

PR #1075, opened 2023-06-02
python unit tests: adds test_clique and adds functions in test_community

PetholzA:
this PR improves #1062

### DIFF
--- a/networkit/test/test_clique.py
+++ b/networkit/test/test_clique.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+import unittest
+import os
+import networkit as nk
+import numpy as np
+import scipy
+
+class TestCliques(unittest.TestCase):
+
+	
+	def testMaxClique(self):
+		G = nk.readGraph("input/jazz2_directed.gml",nk.Format.GML)
+		G.removeSelfLoops()
+		#without callback, testing maxOnly, expecting single Clique
+		A1 = nk.clique.MaximalCliques(G, maximumOnly=True).run().getCliques()
+		self.assertListEqual(A1, [[2]])
+		
+	def testMaxCliqueWithCallback(self):
+		G = nk.readGraph("input/karate.graph",nk.Format.METIS)
+		G.removeSelfLoops()
+		
+		def checkCliqueNodes(maxClique):
+			for u in maxClique:
+				self.assertTrue(G.hasNode(u))
+				
+		#with callback, testing allMaxCliques
+		nk.clique.MaximalCliques(G, maximumOnly=False, callback=checkCliqueNodes).run()
+
+if __name__ == "__main__":
+	unittest.main()
--- a/networkit/test/test_community.py
+++ b/networkit/test/test_community.py
@@ -30,6 +30,55 @@ class TestCommunity(unittest.TestCase):
 			for j in CLLP.getMembers(i):
 				reconstructedSet.append(j)
 		self.assertEqual(set(self.LL.iterNodes()), set(reconstructedSet))
+	
+	def testCoverHubDominance(self):
+		G = nk.Graph(20)
+		G.addEdge(1,2)
+		G.addEdge(2,4)
+		cover = nk.Cover(20) 
+		cover.setUpperBound(3)
+		cover.addToSubset(0,0)
+		cover.addToSubset(1,1)
+		cover.addToSubset(2,2)
+		sim=nk.community.CoverHubDominance(G, cover)
+		sim.run()
+		self.assertEqual(sim.getWeightedAverage(), 1.0)
+		self.assertEqual(sim.getUnweightedAverage(), 1.0)
+		self.assertEqual(sim.getValue(0), 1.0)
+		self.assertListEqual(sim.getValues(),[1.0, 1.0, 1.0])
+		self.assertEqual(sim.getMaximumValue(), 1.0)
+		self.assertEqual(sim.getMinimumValue(), 1.0)
+
+	def testStablePartitionNodes(self):
+		PLM = nk.community.PLM(self.L)
+		PLM.run()	
+		PLMP = PLM.getPartition()		
+		SPN = nk.community.StablePartitionNodes(self.L, PLMP)
+		SPN.run()
+		self.assertEqual(SPN.getWeightedAverage(), 1.0)
+		self.assertEqual(SPN.getUnweightedAverage(), 1.0)
+		self.assertTrue(SPN.isStable(0))	
+
+	def testIntrapartitionDensity(self):
+		G = nk.community.PLM(self.L).run()
+		Gpart = G.getPartition()
+		dense=nk.community.IntrapartitionDensity(self.L, Gpart)
+		dense.run()
+		self.assertAlmostEqual(dense.getWeightedAverage(), 0.685, 3)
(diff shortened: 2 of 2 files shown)


networkit/test/test_clique.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+import unittest
+import os
+import networkit as nk
+import numpy as np
+import scipy
+
+class Test_Cliques(unittest.TestCase):
```
angriman, line 8: To be consistent with other Python tests:
```suggestion
class TestCliques(unittest.TestCase):
```

networkit/test/test_clique.py
```diff
@@ -0,0 +1,17 @@
...
+
+	def test_maxClique(self):
```
angriman, line 10:
```suggestion
	def testmaxClique(self):
```

networkit/test/test_community.py
```diff
@@ -87,7 +125,7 @@ def testJaccardMeasure(self):
 		PLPLLP = PLPLL.getPartition()
 		JM = nk.community.JaccardMeasure()
 		self.assertIsInstance(JM.getDissimilarity(self.LL, PLMLLP, PLPLLP),float)
-
+
```
angriman, line 128: Drop this change.

networkit/test/test_community.py
```diff
@@ -198,8 +245,16 @@ def testPLP(self):
...
+		self.assertIsInstance(PLPL.getTiming(), list)
 	
-	def testCutClustering(self):
+	def testSpectralPartitioner(self):
+		G = nk.Graph(10)
+		spec = nk.community.SpectralPartitioner(G, 1)
+		spec.run()
+		self.assertEqual(spec.getPartition().numberOfSubsets(), 0)
+		
+	def testCutClustering2(self):
```
angriman, line 257:
```suggestion
	def testCutClustering(self):
```

networkit/test/test_community.py
```diff
@@ -77,7 +116,6 @@ def testHubdominance(self):
 		HD = nk.community.HubDominance()
 		self.assertIsInstance(HD.getQuality(PLMLLP, self.LL),float )
 
-
```
angriman, line 80: Drop this change.

networkit/test/test_community.py
```diff
@@ -218,15 +273,6 @@ def testLFM(self):
...
 			self.assertGreaterEqual(len(cover.subsetsOf(u)), 1)
-	
-	def testPLM(self):
-		G = nk.readGraph("input/PGPgiantcompo.graph",nk.Format.METIS)
-		comms = nk.community.detectCommunities(G, nk.community.PLM(G))
-
-		# each node must be assigned
-		for v in G.iterNodes():
-			self.assertTrue(comms.contains(v))			
-
```
angriman, line 276: Why is this deleted?

networkit/test/test_community.py
```diff
@@ -31,6 +31,45 @@ def testCutClustering(self):
...
+
+	def testIntrapartitionDensity(self):
+		G = nk.community.PLP(self.L).run()
+		Gpart = G.getPartition()
+		dense=nk.community.IntrapartitionDensity(self.L, Gpart)
+		dense.run()
+		self.assertIsInstance(dense.getWeightedAverage(), float)
+		self.assertIsInstance(dense.getUnweightedAverage(), float)
+		self.assertIsInstance(dense.getValue(0), float)
+		self.assertIsInstance(dense.getGlobal(), float)
```
angriman, line 60: We should check that these values are correct, not that they are floats.

networkit/test/test_clique.py
```diff
@@ -0,0 +1,17 @@
...
+		G = nk.readGraph("input/jazz2_directed.gml",nk.Format.GML)
+		G.removeSelfLoops()
+		A = nk.clique.MaximalCliques(G, callback=[1,2]).run().getCliques()
```
angriman, line 13: According to the [documentation](https://networkit.github.io/dev-docs/python_api/clique.html), `callback` should be a callable, but here you are passing a list. Please clarify what are you trying to test here.

angriman (changes requested):

networkit/test/test_clique.py
```diff
@@ -0,0 +1,24 @@
...
+	
+	def testMaxClique(self):
+		G = nk.readGraph("input/jazz2_directed.gml",nk.Format.GML)
+		G.removeSelfLoops()
+		#without callback, testing maxOnly, expecting single Clique
+		A1 = nk.clique.MaximalCliques(G, maximumOnly=True).run().getCliques()
+		self.assertEqual(len(A1), 1)
+		#with callback
+		def testCallback():
+			return [1,2,3]
```
angriman, line 19: I don't think this is how the callback is supposed to be used. Please check the `MaximalClique` C++ tests for an example -- I am not suggesting to test the algorithm twice, just make sure that the callback works correctly also in Python.

networkit/test/test_clique.py
```diff
@@ -0,0 +1,24 @@
...
+		A2 = nk.clique.MaximalCliques(G, testCallback()).run().getCliques()
+		self.assertListEqual(A2, [[2]])
```
angriman, line 21: Move this to a separate `testMaxCliqueWithCallback` test.

networkit/test/test_clique.py
```diff
@@ -0,0 +1,26 @@
...
+		G.removeSelfLoops()
+		#without callback, testing maxOnly, expecting single Clique
+		A1 = nk.clique.MaximalCliques(G, maximumOnly=True).run().getCliques()
+		self.assertEqual(len(A1), 1)
+		
+	def testMaxCliqueWithCallback(self):
+		G = nk.readGraph("input/jazz2_directed.gml",nk.Format.GML)
+		G.removeSelfLoops()
+		callback = lambda x,y : 1,2
+		A2 = nk.clique.MaximalCliques(G, callback).run().getCliques()
```
fabratu, line 22: This is not the correct way to call `MaximalCliques` with a callback function. The second parameter is a bool.

fabratu:
Coverage of the `community` module is at 56%. However, based on an offline discussion, I rather want to finish this PR and add the remaining (or at least get above 80%) tests in a future PR.

networkit/test/test_community.py
```diff
@@ -198,7 +295,18 @@ def testPLP(self):
...
+		self.assertIsInstance(PLPL.getTiming(), list)
 	
+	def testSpectralPartitioner(self):
+		#small instance
+		specSm = nk.community.SpectralPartitioner(self.L, 3)
+		specSm.run()
+		#medium instance
+		jazz = nk.readGraph("input/jazz.graph",nk.Format.METIS)
+		spec = nk.community.SpectralPartitioner(jazz, 4, balanced=False)
+		spec.run()
```
fabratu, line 308: I think we should be more precise when testing this code, since it is mostly Python-code. I would suggest testing for `getPartition()` after `run()`(also inspect the result concerning expectations). If you do this, better als split small and medium instances into two tests.

fabratu (changes requested):

fabratu (approved):

fabratu:
PR has been stable for some time, the only part missing was the fix in #1108. Merging this early as a result.

PR #1079, opened 2023-06-07
python tests: improves test_centrality

PetholzA:
this is not the final commit

### DIFF
--- a/networkit/test/test_centrality.py
+++ b/networkit/test/test_centrality.py
@@ -29,6 +29,11 @@ class TestCentrality(unittest.TestCase):
 		for i in range(len(CL.ranking())):
 			self.assertAlmostEqual(CL.ranking()[i][1], CLL.ranking()[i][1], delta=0.2*CL.ranking()[i][1])
 
+	def testApproxBetweennessNumberOfSamples(self):
+		CL = nk.centrality.ApproxBetweenness(self.L, epsilon=0.01, delta=0.1)
+		CL.run()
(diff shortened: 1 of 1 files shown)


angriman (changes requested):
Thank you for the tests, I left some comments.

I see that, while adding new tests, you followed what was already done for other algorithms (e.g., validating an approximation algorithm against itself or just checking the length of the scores list without testing the actual values). It is ok to replicate existing tests on other algorithms, I just don't think that the Python tests we have are appropriate.
In particular, I don't see the point of just checking that Python algorithms yield output of a given size/type. For example, instead of:
```
self.assertEqual(len(algo.ranking()), 4)
```
we should have
```
self.assertListEqual(algo.ranking(), [1,2,3,4])
```
While this might seem redundant because we already have C++ tests, it ensures that we haven't introduced bugs in the Cython layer. So my suggestion is to avoid replicating existing tests. New tests should:
- Check one behavior, do not test multiple functionalities of an algorithm in the same test.
- Validate the algorithm's output.

PetholzA:
@angriman Thank you very much for your comments, i try to include your suggestions from now on and retrospectivly. At multiple occasions it seemed very redundant to write a new testFunction for every single algorithm function, therefore i lumped them together, i can stop that according to your suggestions. Best regards

fabratu:
Coverage is at 84%, which is fine for `centrality`. I think we can finish this PR in order not to overload it. Removing the draft state.

fabratu (approved):

PR #1083, opened 2023-06-13
KadabraBetweenness: Fix heap corruption bug in Cython 

fabratu:
… by adding missing parameter in constructor.

This error was detected during runs of the new tests in #1079. A unit test for Cython/Python will also be integrated in that PR.


angriman (approved):
Good catch, thanks!

fabratu:
Merging this early, since the number of changes is small, and they are needed for other PRs.

PR #1085, opened 2023-06-13
python tests: adds more tests into test_component.py

PetholzA:
mainly to increase testing coverage

### DIFF
--- a/networkit/test/test_component.py
+++ b/networkit/test/test_component.py
@@ -25,6 +25,34 @@ class TestComponent(unittest.TestCase):
 				cc.run()
 				self.assertEqual(cc.numberOfComponents(), 1)
 			G1.forNodes(testNode)
+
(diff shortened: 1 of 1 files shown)


networkit/test/test_component.py
```diff
@@ -25,6 +25,36 @@ def testNode(v):
 				cc.run()
 				self.assertEqual(cc.numberOfComponents(), 1)
 			G1.forNodes(testNode)
+
+	def testBiconnectedComponentsSize(self):
+		G = nk.Graph(5)
+		for i in range(2):
+			G.addEdge(i, i+1)
```
angriman, line 32: I suggest to keep things simple :) same for the others instances below.
```suggestion
		G.addEdge(0, 1)
		G.addEdge(1, 2)
```

fabratu (approved):

PR #1088, opened 2023-06-19
python test: adds tests in test_matching.py

PetholzA:


fabratu:
In principle, we do not have to reach 100% coverage. For this small module however, I would suggest doing so.

Two things are missing: `toPartition` and the `PathGrowingMatcher` constructor with given `edgeScores`.

fabratu (approved):

fabratu (approved):

PR #1089, opened 2023-06-20
python test: adds test in test_reachability.py

PetholzA:


fabratu:
I think the only thing missing is `forAllSimplePaths`. Best add a new test `testAllSimplePathsForIterator` for this function.

fabratu (approved):

PR #1091, opened 2023-06-20
python test: adds tests in test_scd.py

PetholzA:


networkit/test/test_scd.py
```diff
@@ -46,5 +46,41 @@ def testSCD(self):
...
+		self.assertEqual(gt.getAverageJaccard(), 0.5)
+
+	def testExpandOneCommunity(self):
+		nk.setSeed(42, False)
+		ground = nk.Cover(self.G.upperNodeIdBound())
+		ground.setUpperBound(1)
+		for i in range(10):
+			ground.addToSubset(0, i)
+		
+		bfs = nk.scd.RandomBFS(self.G, ground)
```
fabratu, line 80: Better test `RandomBFS` itself as part of `testSCD` and test `expandOneCommunity` (especially the TypeError exception) on its own

PetholzA: the type error exception is added, the problem with including randomBFS into the existing testSCD is that the getConductance() value does not fit in the current assertions

fabratu (approved):

PR #1095, opened 2023-06-23
python test: adds test in test_independentset.py

PetholzA:
this PR should also fix the occuring build fail, caused by testLubyAlgorithm by removing a wrong assertion


fabratu (approved):

fabratu:
Merging this early, since the faulty assertion creates pipeline errors.

Issue #1070, opened 2023-05-24
nk.community.kCoreCommunityDetection error


yushangdi:
I'm trying to run nk.community.kCoreCommunityDetection, and got an error AttributeError: 'networkit.centrality.CoreDecomposition' object has no attribute 'cores'. It seems that this line https://github.com/networkit/networkit/blob/master/networkit/community.pyx#L1847 is not valid anymore?

clstaudt:
Hi @fabratu, I am the developer behind the [trIAge project](https://github.com/trIAgelab/trIAge) which is developing an AI assistant to make life easier for open source communities. 

I am currently exploring how the bot performs on real-world issues and how much context it needs to be supplied with to solve them. I hope you don't mind the direct approach, but NetworKit is a great test case because I am a former maintainer who is not actively working on the project anymore. Generating this response still involved some manual steps that we are currently trying to validate and automate. 

I'd appreciate your feedback, especially whether the bot response solves the issue, and you would consider accepting a pull request from the trIAge bot in the future.

fabratu:
@clstaudt Nice tool, this might indeed help both users in using the software as also developers in improving the API. Will try to give feedback wherever possible. What kind of PRs can we expect from trIAge?

Answer to the issue:

The problem here is a mere typo. It should be `scores()` not `cores()`. While retrieving the cover is also a solution, the canonical way is to use `scores()` for centrality measures. You can see an example of how to use it here: https://networkit.github.io/dev-docs/notebooks/User-Guide.html#Core-Decomposition

We currently test the code, however the test is misleading and therefore does not detect erroneous behavior. The doc string is also not complete - all in all time to refresh the code. If possible, maybe you can post a PR fixing this issue? If not, we will schedule a fix. In the time being, as a workaround, you can call the included calls on your own.

clstaudt:
@fabratu 

> The problem here is a mere typo. It should be scores() not cores()

Interesting, the model came up with a correct solution but not the preferred solution.

> What kind of PRs can we expect from trIAge?

How can an LLM like GPT4 assist, given it has access to the code base and the documentation? In the first steps we want to provide a GitHub bot with assistance capabilities, e.g. suggest how to improve the way the issue was reported, detect when the question has already been answered in the documentation, generate test cases for the reported issue, suggest bugfixes and other solutions... The goal is to take workload from maintainers and provide faster support to users. Automated creation of Pull Requests that fix issues may not be too far in the future.

clstaudt:
[...]

PR #1096, opened 2023-06-23
Fix kCoreCommunityDetection

yushangdi:
Fix kCoreCommunityDetection. Previous implementation has typo and subgraphFromNodes method needs to be updated.

Issue: https://github.com/networkit/networkit/issues/1070#issuecomment-1604676495


fabratu:
Thanks for posting the fix. Based on the discussion in #1070, better use `core.getPartition()` and fill `kCore` until `k` is reached.

yushangdi:
@fabratu Thanks for the comment. I've updated the implementation to use `getPartition()` now.

fabratu (approved):

PR #1097, opened 2023-06-27
New algorithm: dynamic approx electrical closeness

bernlu:
This algorithm is based on the `ApproxElectricalCloseness` algorithm in networkit.
The dynamic algorithm supports:
- edge addition. Algorithm details can be found in our recent paper (which I think is not published yet (?))
- edge deletion. Based on the same idea as edge addition, with small changes to the math / reordering some terms. 

This implementation currently does not support arbitrary edge deletions: edges on the BFS tree from the pivot element `u` (which is used to approximate r(u,v) for all `v`) may not be deleted, because this would require a larger update computation. In principle, this is possible, but for my use case this is not a problem right now. The code can easily be extended when the need arises.

Additional changes in this PR:
- updated the description of `ApproxElectricalCloseness` to a more accurate description of the results properties
- created a new GTest file which now contains the tests for the static and dynamic algorithm

### DIFF
--- a/include/networkit/centrality/ApproxElectricalCloseness.hpp
+++ b/include/networkit/centrality/ApproxElectricalCloseness.hpp
@@ -30,13 +30,13 @@ class ApproxElectricalCloseness final : public Centrality {
 public:
     /**
      * Approximates the electrical closeness of all the vertices of the graph by approximating the
-     * diagonal of the laplacian's pseudoinverse of @a G. Every element of the diagonal is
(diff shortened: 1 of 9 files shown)


angriman (commented):
Thank you for the PR. I see that this algorithm shares a lot of data structures and code with `ApproxElectricalCloseness`. I'd suggest declaring `DynApproxElectricalCloseness` as a subclass of `ApproxElectricalCloseness`, so that you can re-use existing code and avoid duplication.

PR #1101, opened 2023-07-05
docu: fix wikipedia link

PetholzA:
this PR adresses a link that did not work properly


fabratu:
Merging this early, since it is a trivial fix.

PR #1102, opened 2023-07-09
Refactor Matrices and MatricesGTest

angriman:
The first commit adds missing matrix functions and tests:
- For CSRMatrix: functions for iterating over all elements of the matrix, including zeros.
- For DenseMatrix: functions to compute the adjacency, Laplacian, and incidence matrix of a graph and a diagonal matrix.
- For DynamcMatrix: functions to iterate over the (non-zero) elements.
In this way, all matrix classes offer the same set of APIs.

The second commit refactors the unit tests for matrices. It defines a typed test suite (see https://google.github.io/googletest/advanced.html#typed-tests) that performs the unit tests on all the matrix types. This is much more convenient than defining a templated unit test (as we were doing before) because:
- Failures much are easier to track: if a test fails, the type is reported in the logs.
- Tests are more concise (the new version has 220 fewer lines of code).

Other minor changes:
- Avoid Yoda conditions in tests https://en.wikipedia.org/wiki/Yoda_conditions
- Replace `ASSERT_` with `EXPECT_` where it is reasonable to do so (https://google.github.io/googletest/primer.html#assertions)

### DIFF
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -987,6 +987,27 @@ public:
     void forElementsInRow(index i, L handle) const;
 
     /**
+     * Iterate in parallel over all elements (including zeros) of row @a row in the matrix and call
+     * handler(index column, double value)
+     */
+    template <typename L>
+    void parallelForElementsInRow(index row, L handle) const;
+
+    /**
+     * Iterate over all elements of the matrix in row order and call handler (lambda
+     * closure).
+     */
+    template <typename L>
+    void forElementsInRowOrder(L handle) const;
+
+    /**
+     * Iterate in parallel over all rows and call handler (lambda closure) on elements of
+     * the matrix.
+     */
+    template <typename L>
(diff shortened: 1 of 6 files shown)


fabratu:
I merged #1074 - the parts for C++ are likely obsolete with this PR and can be overwritten.

angriman:
I rebased on top of master and included the new test.

fabratu:
Hm, somehow the coverage does decrease significantly for the involved classes:

- `DenseMatrix.cpp`: 81% to 17%
- `DynamicMatrix.cpp`: 73% to 8%
- `CSRGeneralMatrix.hpp`: 94% to 60%
- `DenseMatrix.hpp`: 97% to 8%
...

Not sure, what causes this.

angriman:
The reason is that `MatricesGTest` is never executed. The test pattern is, for example, `MatricesGTest/0.testMatrixToGraph` while our test filter is `*Test.test*`, which filters out the new tests. I changed it so that `MatricesGTest` are not filtered out.

angriman:
Coverage seems ok now:
Screenshot 2023-07-16 at 13 16 23

fabratu (approved):

Issue #1106, opened 2023-07-20
Building The Python Module


larsgottesbueren:
I'm having trouble building the python module.  The following is the output for the command python3 setup.py build_ext

```
[315/408] Building CXX object CMakeFiles/helpers.dir/networkit/helpers.cpp.o
FAILED: CMakeFiles/helpers.dir/networkit/helpers.cpp.o 
/usr/bin/g++ -Dhelpers_EXPORTS -I/home/gottesbueren/libs/networkit/include -I/home/gottesbueren/libs/networkit/extlibs/ttmath -I/usr/include/python3.11 -I/home/gottesbueren/libs/networkit/extlibs/tlx -fuse-ld=lld -O3 -DNDEBUG -std=gnu++17 -fPIC  -DNETWORKIT_RELEASE_LOGGING -fopenmp -MD -MT CMakeFiles/helpers.dir/networkit/helpers.cpp.o -MF CMakeFiles/helpers.dir/networkit/helpers.cpp.o.d -o CMakeFiles/helpers.dir/networkit/helpers.cpp.o -c /home/gottesbueren/libs/networkit/networkit/helpers.cpp
/home/gottesbueren/libs/networkit/networkit/helpers.cpp: In function ‘int __pyx_pymod_exec_helpers(PyObject*)’:
/home/gottesbueren/libs/networkit/networkit/helpers.cpp:24330:17: error: invalid conversion from ‘void*’ to ‘__pyx_t_9networkit_10structures_edgeweight*’ {aka ‘double*’} [-fpermissive]
24330 |   __pyx_k__30 = __pyx_t_10;
      |                 ^~~~~~~~~~
      |                 |
      |                 void*
/home/gottesbueren/libs/networkit/networkit/helpers.cpp:24331:17: error: invalid conversion from ‘void*’ to ‘__pyx_t_9networkit_10structures_node*’ {aka ‘long unsigned int*’} [-fpermissive]
24331 |   __pyx_k__31 = __pyx_t_10;
      |                 ^~~~~~~~~~
      |                 |
      |                 void*
/home/gottesbueren/libs/networkit/networkit/helpers.cpp:24332:17: error: invalid conversion from ‘void*’ to ‘double*’ [-fpermissive]
24332 |   __pyx_k__32 = __pyx_t_10;
      |                 ^~~~~~~~~~
      |                 |
      |                 void*
[318/408] Building CXX object CMakeFiles/graphio.dir/networkit/graphio.cpp.o
[...]

PR #1107, opened 2023-07-21
pyproject.toml: Add wheel as requirement and pin stable versions of cython, setuptools.

fabratu:
This PR fixes invalid C++ code, which the recently released version of Cython (3.0.0) produces (see: #1106). Until this is fixed, the previously stable 0.29.X version is used.

Furthermore, wheel is added as build requirement in order to create wheels when doing non-editable install via pip.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.12.1
         env:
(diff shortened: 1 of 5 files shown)


bernlu (approved):

fabratu:
Merging this early, since the fix is trivial and it fixes problems on CI. In a separate PR, the code base will be adopted to be compatible with Cython 3.X (likely after Patch/Bug-releases are out).

PR #1112, opened 2023-08-11
graphio: add function to guess file format

bernlu:
This PR adds a new `guessFileFormat` function to the `graphio` module (python). This function is used to guess the file format when `graphio.readGraph(path, fileformat)` is called with `fileformat=None` - which is now the default value for the `fileformat` parameter. Guessing may fail and raise an exception (in which case one needs to provide the format manually as before). 

The guessing strategies are quite simple and will probably fail for some cases where the format is actually rather easy to guess, but for now it works for the most common file structures and types (see tests). We can improve this function when new ideas/problems arise.

Direct use:
```py
>>> import networkit as nk
>>> nk.graphio.guessFileFormat("input/airfoil1.graph")
<Format.METIS: 6>
```
With `readGraph`:
```py
>>> import networkit as nk
>>> nk.readGraph("input/airfoil1.graph")
<networkit.graph.Graph object at 0x7fa2a591c570>
```

One problem which is unfixable is that a file can be both in METIS and EdgeList format, e.g.
```
3 3
1 2
0 2
0 1
```
Depending on the format, this is either a 3 or 4 node graph. When this case is detected, an error is raised and the format has to be provided manually.

### DIFF
--- a/networkit/graphio.pyx
+++ b/networkit/graphio.pyx
@@ -9,6 +9,7 @@ from libcpp.map cimport map as cmap
 from libcpp.unordered_map cimport unordered_map
 from .helpers import stdstring
 
+import re
 import os
 import logging
 import numpy
@@ -1021,7 +1022,7 @@ class Format(__AutoNumber):
 	- networkit.graphio.Format.GraphViz
 	- networkit.graphio.Format.GEXF
 	- networkit.graphio.Format.GML
-	- networkit.graphio.Format.KONEC
+	- networkit.graphio.Format.KONECT
 	- networkit.graphio.Format.LFR
 	- networkit.graphio.Format.METIS
 	- networkit.graphio.Format.NetworkitBinary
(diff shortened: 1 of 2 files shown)


bernlu:
Maybe it would be nice to add support for MatrixMarket files as well, but I think currently the matrix market file reader is c++ only.. which means this would be a larger change and probably its own PR.

bernlu:
Added python support for reading matrix market files to graphs. this does not use the cpp matrix market reader because that reader creates a CSRMatrix (which is not what we need for readGraph). I decided to instead just read the matrix market file as edge list.

fabratu (commented):
Only two minor comments concerning documentation - otherwise, I think we can merge this. As discussed offline, the `MatrixMarket` reader for Cython/Python (returning a graph, not a `CSRMatrix`) is added in another PR.

networkit/graphio.pyx
```diff
@@ -1098,17 +1101,161 @@ def getReader(fileformat, *kargs, **kwargs):
...
+	with open(filepath, 'r') as f:
+		n = None
+		m = None
+		noncommentlines = 0
+		edges = 0
+		for line in f.readlines():
+			# snap check: keep verifying line format while the file could be a snap file
+			# check that all lines have format \d+ \d+ (or are empty lines)
+			if snapFound:
+					match = re.match(r'(^\d+\s\d+$)|(^\s*$)', line)
```
bernlu, line 1187: snap format: lines may end with whitespaces

fabratu (approved):

PR #1114, opened 2023-08-11
Complex paths

fidus58:
complex paths as new centrality algorithm

tests are still missing

### DIFF
--- a/include/networkit/centrality/ComplexPaths.hpp
+++ b/include/networkit/centrality/ComplexPaths.hpp
@@ -6,42 +6,42 @@
  *      Author: Klaus Ahrens
  *              <ahrens@informatik.hu-berlin.de>
  *
- *  adapted and reimplemented from 
+ *  adapted and reimplemented from
  *
  *            https://github.com/drguilbe/complexpaths.git
  *
- *  see [ Guilbeault, D., Centola, D. Topological measures for 
- *        identifying and predicting the spread of complex contagions. 
- *        Nat Commun 12, 4430 (2021). 
+ *  see [ Guilbeault, D., Centola, D. Topological measures for
+ *        identifying and predicting the spread of complex contagions.
+ *        Nat Commun 12, 4430 (2021).
  *        https://doi.org/10.1038/s41467-021-24704-6 ]
  *
  */
 
-#ifndef NETWORKIT_COMPLEX_PATHS_HPP_
-#define NETWORKIT_COMPLEX_PATHS_HPP_
+#ifndef NETWORKIT_CENTRALITY_COMPLEX_PATHS_HPP_
+#define NETWORKIT_CENTRALITY_COMPLEX_PATHS_HPP_
 
 #include <random>
 #include <vector>
 
-#include <networkit/graph/Graph.hpp>
 #include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
 
-using std::vector;
 using std::sample;
+using std::vector;
 
 namespace NetworKit {
 
 /**
  * @ingroup centrality
  * computes
- *      in Mode::singleNode : 
+ *      in Mode::singleNode :
  *	complex path graphs (with inner connection degree above threshold)
  *	from a start node in Mode::singleNode
  *	OR
- *      in Mode::allNodes : 
- *	complex path lengths (maximal distance in complex path graphs for 
+ *      in Mode::allNodes :
+ *	complex path lengths (maximal distance in complex path graphs for
  *	starting nodes u) for all nodes u
- * 
+ *
  */
 
 class ComplexPathAlgorithm : public Algorithm {
@@ -49,13 +49,13 @@ public:
     /**
      * operation modes
      */
-    enum Mode {singleNode, allNodes};
+    enum Mode { singleNode, allNodes };
 
     /**
      * Constructs the ComplexPathAlgorithm class for the given Graph @a G.
      * depending on the @a mode the algorithm the algorithm
      *   in @a mode <code>Mode::singleNode</code> starting from @a start
-     *      constructs a subgraph of @a G in which all inner nodes have 
+     *      constructs a subgraph of @a G in which all inner nodes have
      *      more than @a threshold neighbors
      *      call <code>getComplexGraph</code> after run to get it
      *   in @a mode <code>Mode::allNodes</code> constructs these subgraphs for all nodes
@@ -67,38 +67,36 @@ public:
      * @param G The graph.
      * @param threshold number of neighbors needed
      * @param mode as explained above
-     * @param start start node for Mode::singleNode 
+     * @param start start node for Mode::singleNode
      */
-    ComplexPathAlgorithm(const Graph& G, 
-                         count threshold = 3, 
-                         Mode mode = Mode::allNodes, 
+    ComplexPathAlgorithm(const Graph &G, count threshold = 3, Mode mode = Mode::allNodes,
                          node start = none);
 
     /**
      * normalize path lengths
      */
     void normalize();
-		 
+
     void run() override;
 
-    /** 
+    /**
      * [normalized] results after running in <code>Mode::allNodes</code>
      */
     vector<double> getPLci();
 
-    /** 
+    /**
      * resulting graph after running in <code>Mode::singleNode</code>
      */
     Graph getComplexGraph();
 
-    /** 
-     * nodes in the resulting graph after running in <code>Mode::singleNode</code> 
+    /**
+     * nodes in the resulting graph after running in <code>Mode::singleNode</code>
      * which are connected to start by at least threshold paths
      */
     vector<node> getAdopters();
 
 private:
-    const Graph& inputGraph;
+    const Graph &inputGraph;
     Graph complexPathGraph;
     vector<double> complexPathsLengths;
     vector<node> adopters;
@@ -107,10 +105,10 @@ private:
     const count threshold;
     bool normPaths;
 
-    Graph complexPathsGraph(node seed, count threshold, vector<node>* adopters);
+    Graph complexPathsGraph(node seed, count threshold, vector<node> *adopters);
     vector<double> complexPathLength(count t);
-}; 
+};
 
 } /* namespace NetworKit */
 
-#endif // NETWORKIT_COMPLEX_PATHS_HPP_
+#endif // NETWORKIT_CENTRALITY_COMPLEX_PATHS_HPP_
--- a/include/networkit/centrality/ComplexPaths.hpp
+++ b/include/networkit/centrality/ComplexPaths.hpp
@@ -26,9 +26,6 @@
 #include <networkit/base/Algorithm.hpp>
 #include <networkit/graph/Graph.hpp>
 
-using std::sample;
-using std::vector;
-
 namespace NetworKit {
 
 /**
@@ -82,7 +79,7 @@ public:
     /**
      * [normalized] results after running in <code>Mode::allNodes</code>
      */
-    vector<double> getPLci();
+    std::vector<double> getPLci();
 
     /**
      * resulting graph after running in <code>Mode::singleNode</code>
@@ -93,20 +90,21 @@ public:
      * nodes in the resulting graph after running in <code>Mode::singleNode</code>
      * which are connected to start by at least threshold paths
      */
-    vector<node> getAdopters();
+    std::vector<node> getAdopters();
 
 private:
-    const Graph &inputGraph;
+    const Graph *inputGraph;
     Graph complexPathGraph;
-    vector<double> complexPathsLengths;
-    vector<node> adopters;
+    std::vector<double> complexPathsLengths;
+    std::vector<node> adopters;
     const Mode mode;
     const node start;
     const count threshold;
     bool normPaths;
 
-    Graph complexPathsGraph(node seed, count threshold, vector<node> *adopters);
-    vector<double> complexPathLength(count t);
+    Graph complexPathsGraph(node seed, count threshold, std::vector<node> *adopters);
+    std::vector<double> complexPathLength(count t);
+    std::vector<node> generateSeeds(node seed, const Graph &g, count threshold);
 };
 
 } /* namespace NetworKit */
--- a/include/networkit/centrality/ComplexPaths.hpp
+++ b/include/networkit/centrality/ComplexPaths.hpp
@@ -1,0 +1,116 @@
+/*
+ *  ComplexPaths.hpp
+ *
+ *
+ *  Created on:16.06.2023
+ *      Author: Klaus Ahrens
+ *              <ahrens@informatik.hu-berlin.de>
+ *
+ *  adapted and reimplemented from 
+ *
+ *            https://github.com/drguilbe/complexpaths.git
+ *
+ *  see [ Guilbeault, D., Centola, D. Topological measures for 
+ *        identifying and predicting the spread of complex contagions. 
+ *        Nat Commun 12, 4430 (2021). 
+ *        https://doi.org/10.1038/s41467-021-24704-6 ]
+ *
+ */
+
+#ifndef NETWORKIT_COMPLEX_PATHS_HPP_
+#define NETWORKIT_COMPLEX_PATHS_HPP_
+
+#include <random>
+#include <vector>
+
+#include <networkit/graph/Graph.hpp>
+#include <networkit/base/Algorithm.hpp>
+
+using std::vector;
+using std::sample;
+
+namespace NetworKit {
+
+/**
+ * @ingroup centrality
+ * computes
+ *      in Mode::singleNode : 
+ *	complex path graphs (with inner connection degree above threshold)
+ *	from a start node in Mode::singleNode
+ *	OR
+ *      in Mode::allNodes : 
+ *	complex path lengths (maximal distance in complex path graphs for 
+ *	starting nodes u) for all nodes u
+ * 
+ */
+
+class ComplexPathAlgorithm : public Algorithm {
+public:
+    /**
+     * operation modes
+     */
+    enum Mode {singleNode, allNodes};
+
+    /**
+     * Constructs the ComplexPathAlgorithm class for the given Graph @a G.
+     * depending on the @a mode the algorithm the algorithm
+     *   in @a mode <code>Mode::singleNode</code> starting from @a start
+     *      constructs a subgraph of @a G in which all inner nodes have 
+     *      more than @a threshold neighbors
+     *      call <code>getComplexGraph</code> after run to get it
+     *   in @a mode <code>Mode::allNodes</code> constructs these subgraphs for all nodes
+     *      and calculates their longest paths from their start nodes
+     *      these length can be absolute or normalized to [0..1] by calling
+     *      normalize before or after <code>run</code>
+     *      call <code>getPLci</code> after run to get these lengths
+     *
+     * @param G The graph.
+     * @param threshold number of neighbors needed
+     * @param mode as explained above
+     * @param start start node for Mode::singleNode 
+     */
+    ComplexPathAlgorithm(const Graph& G, 
+                         count threshold = 3, 
+                         Mode mode = Mode::allNodes, 
+                         node start = none);
+
+    /**
+     * normalize path lengths
+     */
+    void normalize();
+		 
+    void run() override;
+
+    /** 
+     * [normalized] results after running in <code>Mode::allNodes</code>
+     */
+    vector<double> getPLci();
+
+    /** 
+     * resulting graph after running in <code>Mode::singleNode</code>
+     */
+    Graph getComplexGraph();
+
+    /** 
+     * nodes in the resulting graph after running in <code>Mode::singleNode</code> 
+     * which are connected to start by at least threshold paths
+     */
+    vector<node> getAdopters();
+
+private:
+    const Graph& inputGraph;
+    Graph complexPathGraph;
+    vector<double> complexPathsLengths;
+    vector<node> adopters;
+    const Mode mode;
+    const node start;
+    const count threshold;
+    bool normPaths;
+
+    Graph complexPathsGraph(node seed, count threshold, vector<node>* adopters);
+    vector<double> complexPathLength(count t);
+}; 
+
+} /* namespace NetworKit */
+
+#endif // NETWORKIT_COMPLEX_PATHS_HPP_
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -2933,8 +2933,6 @@ class SciPyPageRank(SpectralCentrality):
 		self.eigenvector = spectrum[1][0]
 		self.eigenvalue = spectrum[0][0]
 
-##### from branch ComplexPaths (K. Ahrens)
-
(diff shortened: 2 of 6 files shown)


networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+		vector[double] getPLci() except +
+		void normalize() except +
+		vector[node] getAdopters() except +
+		_Graph &getComplexGraph() except +
+
+cdef class ComplexPaths(Algorithm):
+	""" 
+	ComplexPaths(G, threshold, mode, start)
+
+	according to:
```
fabratu, line 2960: Please elaborate more what complex paths does. Take a look at the documentation of other recently implemented algorithms for reference.

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+ 
+	Parameters
+	----------
+	G : networkit.Graph
+		The graph.
+
+	threshold : int
+		The complex path width: minimal number of neighbors
+
+	mode : ComplexPathMode
```
fabratu, line 2974: Writing the parameter this way will likely break the rendering of the doc. Take a look at other examples in NetworKit at how this is written - for example `Diameter`:  https://github.com/networkit/networkit/blob/master/networkit/distance.pyx#L487

Also, the overview over included enum-parameters should be added above the parameter section.

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+			mode=ComplexPathMode.singleNode
+		self._G = G
+		self._this = new _ComplexPaths(G._this, threshold, mode, start)
+
+	def getPLci(self):
+		""" 
+		getPLci()
+
+		Returns complex path lengths for every node in G
+		absolute or scaled to [0,1] when normalize() was called before run
```
fabratu, line 2998:
```
... before run.
This function is available only if networkit.centrality.ComplexPathMode.ALL_NODES was choosen.
```

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+
+cdef extern from "<networkit/centrality/ComplexPaths.hpp>" namespace "NetworKit::ComplexPathAlgorithm":
+
+	cpdef enum _Mode  "NetworKit::ComplexPathAlgorithm::Mode":
+		singleNode,
+		allNodes
+
+cdef class ComplexPathMode(object):
+	singleNode = _Mode.singleNode
+	allNodes   = _Mode.allNodes
```
fabratu, line 2946: Enums in Python are written in capital letters with underscore (SREAMING_SNAKE_CASE): `SINGLE_NODE` and `ALL_NODES`

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+		""" 
+		getPLci()
+
+		Returns complex path lengths for every node in G
+		absolute or scaled to [0,1] when normalize() was called before run
+		in mode ComplexPathMode.allNodes only
+
+		Returns
+		-------
+		list(double)
```
fabratu, line 3003: `float`

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+			A vector containing complex path lengths for all nodes
+		"""
+		return (<_ComplexPaths*>(self._this)).getPLci()
+
+	def getComplexGraph(self):
+		""" 
+		getComplexGraph()
+
+		Returns complex path (sub)graph of G starting in start
```
fabratu, line 3012: `in node start` (also use `:code` for `start`)

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+		in mode ComplexPathMode.singleNode only
```
fabratu, line 3013: New sentence

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+		"""
+		return Graph().setThis((<_ComplexPaths*>(self._this)).getComplexGraph())
+
+	def getAdopters(self):
+		"""
+		getAdopters()
+
+		Returns all nodes in the complex subgraph with at least
+		threshold neighbors (those who are adopted/infected when 
+		starting in start)
```
fabratu, line 3027: MIssing punctuation

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+
+		Returns
+		-------
+		list(node)
```
fabratu, line 3031: `int`

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+	def normalize(self):
+		"""
+		normalize()
+
+		when called before run() all complex path lengths 
+		returned by getPLci are scaled to [0,1]
+
+		in mode ComplexPathMode.allNodes only
+
+		Returns
```
fabratu, line 3045: Remove the `Returns` section from the docstring.

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+		list(node) 
+			A vector all adopted/infected nodes
+		"""
+		return (<_ComplexPaths*>(self._this)).getAdopters()
+
+	def normalize(self):
+		"""
+		normalize()
+
+		when called before run() all complex path lengths
```
fabratu, line 3040: `When called before run(), all ...`

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+			A vector all adopted/infected nodes
+		"""
+		return (<_ComplexPaths*>(self._this)).getAdopters()
+
+	def normalize(self):
+		"""
+		normalize()
+
+		when called before run() all complex path lengths 
+		returned by getPLci are scaled to [0,1]
```
fabratu, line 3041: `are scaled to [0,1].`

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
...
+
+		in mode ComplexPathMode.allNodes only
```
fabratu, line 3043: No new line and convert it to a proper sentence.

networkit/centrality.pyx
```diff
@@ -2928,3 +2932,120 @@ class SciPyPageRank(SpectralCentrality):
 
 		self.eigenvector = spectrum[1][0]
 		self.eigenvalue = spectrum[0][0]
+
+##### from branch ComplexPaths (K. Ahrens)
```
fabratu, line 2936: Remove this line (and unnecessary empty lines).

fabratu (changes requested):

include/networkit/centrality/ComplexPaths.hpp
```diff
@@ -0,0 +1,114 @@
...
+#define NETWORKIT_CENTRALITY_COMPLEX_PATHS_HPP_
+
+#include <random>
+#include <vector>
+
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+using std::sample;
+using std::vector;
```
angriman, line 30: I think we should drop this, it does not make the code easier to read.

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+vector<double> ComplexPathAlgorithm::complexPathLength(count t) {
+    const auto &g = inputGraph;
+    auto n = g.numberOfNodes();
+    vector<double> PLci(n);
+    constexpr auto infDist = std::numeric_limits<edgeweight>::max();
+
+    g.parallelForNodes([&](node u) {
+        auto c_g = complexPathsGraph(u, t, nullptr);
```
angriman, line 185:
```suggestion
        Graph c_g = complexPathsGraph(u, t, nullptr);
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        auto noPaths = false;
```
angriman, line 186:
```suggestion
        bool noPaths = false;
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        NetworKit::BFS bfs(c_g, u, noPaths);
```
angriman, line 187: We are already in the `NetworKit` namespace.
```suggestion
        BFS bfs(c_g, u, noPaths);
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        bfs.run();
+        auto distances = bfs.getDistances();
+        for (auto &d : distances)
+            if (d == infDist)
+                d = 0.0;
+        auto distanceSum = std::accumulate(distances.begin(), distances.end(), 0.0);
+        PLci[u] = distanceSum / n;
```
angriman, line 194: Division between `double` and `uint64_t`
```suggestion
        PLci[u] = distanceSum / static_cast<double>(n);
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+    g.parallelForNodes([&](node u) {
+        auto c_g = complexPathsGraph(u, t, nullptr);
+        auto noPaths = false;
+        NetworKit::BFS bfs(c_g, u, noPaths);
+        bfs.run();
+        auto distances = bfs.getDistances();
+        for (auto &d : distances)
+            if (d == infDist)
+                d = 0.0;
+        auto distanceSum = std::accumulate(distances.begin(), distances.end(), 0.0);
```
angriman, line 193: I think that we can do something like described in https://en.cppreference.com/w/cpp/algorithm/transform_reduce:
```suggestion
        const double distanceSum = std::transform_reduce(distances.cbegin(), distances.cend(), 0.0, std::plus{}, [](double val) { return val == infDist ? 0.0 : val; });
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+static void addNewEdge(Graph &g, node u, node v) {
+    if (g.hasEdge(u, v))
+        return;
+    g.addEdge(u, v);
+}
+
+Graph ComplexPathAlgorithm::complexPathsGraph(node seed, count threshold, vector<node> *adopters) {
+    const auto &g = inputGraph;
```
angriman, line 90:
```suggestion
    const Graph &g = inputGraph;
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+            auto indirectNeighbors = Graph::NeighborRange<>(g, u);
+            for (auto in : indirectNeighbors) {
+                if (in != seed
+                    && std::find(moreSeeds.begin(), moreSeeds.end(), in) == moreSeeds.end())
+                    moreSeeds.push_back(in);
+            }
+            sample(moreSeeds.begin(), moreSeeds.end(), std::back_inserter(seeds), needMore,
+                   randomGen);
+            if (seeds.size() < needSeeds)
+                WARN("complexPathsGraph: ", seed, ": not enough indirect neighbors.");
```
angriman, line 120: This is inefficient and overly complicated. We basically need to sample a fixed number of seeds from the nodes up to distance 2 from the seed. Why don't we:
- Sample from the neighbors of `seed` only if they are enough, otherwise
- Run a BFS that stops at level 2 and use two queues to keep all the nodes at distance 1 and 2 separate. Then insert all the nodes at level 1 and sample the remaining number of seeds from the nodes at level 2.

In this way the code would be simpler and we won't need to run all these `std::find` (which are linear searches).

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+
+    seeds.insert(seeds.begin(), seed); // put in front, so that direct neighbors are activated first
+
+    for (auto u : seeds) {
```
angriman, line 126:
```suggestion
    for (node u : seeds) {
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        activated[u] = true;
+        if (adopters)
+            adopters->push_back(u);
+        auto neighbors = Graph::NeighborRange<>(g, u);
+
+        for (auto v : neighbors) {
```
angriman, line 132:
```suggestion
        for (node v : neighbors) {
```

angriman (changes requested):
Thanks for the new algorithm, I left some comments. My comments about the usage of `auto` are probably arguable, but as a reviewer I found some parts of the code hard to follow because of that. If using `auto` confuses the reader, then it was probably misused.

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    activated[u] = true;
+                    if (adopters)
+                        adopters->push_back(u);
+                    spread = true;
+                }
+                auto uNeighbors = Graph::NeighborRange<>(g, u);
+                for (auto v : uNeighbors) {
+                    addNewEdge(complex_g, u, v);
+                }
+            }
```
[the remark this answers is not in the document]
PetholzA: This results in the compiler warning: "control reaches the end of non-void function".

include/networkit/centrality/ComplexPaths.hpp
```diff
@@ -0,0 +1,114 @@
...
+    Graph getComplexGraph();
+
+    /**
+     * nodes in the resulting graph after running in <code>Mode::singleNode</code>
+     * which are connected to start by at least threshold paths
+     */
+    vector<node> getAdopters();
+
+private:
+    const Graph &inputGraph;
```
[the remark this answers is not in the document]
fabratu: Can you elaborate more on what "these cases" means? 

I remember some discussion in the past about const ref vs. `const` pointer. However, I only see an usecase for pointer, if we don't initialize the member variable directly in the constructor.

fabratu (dismissed):

fabratu:
@angriman Thanks for the review - I want to merge this new algorithm into the next release. Any objections?

networkit/cpp/centrality/ComplexPaths.cpp
angriman: Not resolved.

networkit/cpp/centrality/ComplexPaths.cpp
angriman: Not resolved.

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,203 @@
...
+        });
+        if (spread) { // new influences in growing complex_g
+            complex_g.parallelForNodes(computeInfluence);
+        }
+    } while (spread);
+
+    return complex_g;
+}
+
+static auto min_max_norm(vector<double> data) {
```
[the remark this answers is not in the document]
angriman: Not resolged.

angriman:
It seems to me that some comments were marked as resolved without resolving them nor a justification of why they were not resolved. I marked them as unresolved. If you disagree with some of them, please let me know.

networkit/cpp/centrality/ComplexPaths.cpp
PetholzA: I did forgot to mention, that i tried using std::transform_reduce, this resulted in some older build on the CI failing (iirc: Linux - cpp8), because it did not knew std::transform_reduce yet.

networkit/cpp/centrality/ComplexPaths.cpp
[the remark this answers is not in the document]
angriman: Maybe in my suggestion I forgot a parenthesis. Anyway, the idea is to avoid an unnecessary indentation level. Instead of wrapping everything around `if (influence[u] >= threshold)`, you can use `if (influence[u] < threshold) return;`, the rest of the lambda remains the same. That lambda returns `void` so that compiler error is unrelated the suggestion.

networkit/cpp/centrality/ComplexPaths.cpp
angriman: Oh, pity.

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+        });
+        if (spread) { // new influences in growing complex_g
+            complex_g.parallelForNodes(computeInfluence);
+        }
+    } while (spread);
+
+    return complex_g;
+}
+
+static auto min_max_norm(std::vector<double> data) {
```
angriman, line 168:
```suggestion
static double min_max_norm(const std::vector<double> &data) {
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+    auto [min, max] = std::minmax_element(data.begin(), data.end());
+    double diff = *max - *min;
+    std::vector<double> norm(data);
+    // Hint: Use structured bind "min" as init-capture in order to avoid compiler errors
+    // for C++17 and Clang (at least up until C++20).
+    std::transform(norm.cbegin(), norm.cend(), norm.begin(),
+                   [diff, &min = min](auto e) { return (e - *min) / diff; });
```
angriman, line 175:
```suggestion
                   [diff, &min = min](double e) { return (e - *min) / diff; });
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    std::vector<node> seeds;
+    seeds.reserve(4);
+
+    auto neighbors = Graph::NeighborRange<>(g, seed);
+
+    auto randomGen = Aux::Random::getURNG();
+    std::sample(neighbors.begin(), neighbors.end(), std::back_inserter(seeds), seedsNeeded,
+                randomGen);
+    auto needMore = seedsNeeded - seeds.size();
```
angriman, line 99: If `seeds.size()` is bigger than `seedsNeeded` this will overflow.
```suggestion
    if (seeds.size() >= seedsNeeded) return seeds;
    const count additionalSeeds = seedsNeeded - seeds.size();
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/centrality/ComplexPaths.hpp>
+#include <networkit/distance/BFS.hpp>
+#include <networkit/distance/PrunedLandmarkLabeling.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+ComplexPathAlgorithm::ComplexPathAlgorithm(const Graph &G, count threshold, Mode mode, node start)
+    : inputGraph{&G}, mode{mode}, start{start}, threshold{threshold}, normPaths{false} {
```
angriman, line 36: Add a check for `threshold > 0`. Below in the code you have `auto seedsNeeded = threshold - 1` which will overflow if `threshold` is 0.

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+static void addNewEdge(Graph &g, node u, node v) {
+    if (g.hasEdge(u, v))
+        return;
+    g.addEdge(u, v);
+}
+
+std::vector<node> ComplexPathAlgorithm::generateSeeds(node seed, const Graph &g, count threshold) {
+    // activate threshold-1 random neighbors of seed:
+    auto seedsNeeded = threshold - 1;
```
angriman, line 89: Can we be 100% sure
```suggestion
    const count seedsNeeded = threshold - 1;
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+    auto needMore = seedsNeeded - seeds.size();
+    if (needMore > 0) {
+        WARN("complexPathsGraph: ", seed, ": not enough direct neighbors.");
+
+        // collect all nodes at distance 2 before sampling to establish equal distribution
+        std::vector<node> indirNeighbors;
+
+        for (node u : neighbors) {
+            auto indirectNeighbors = Graph::NeighborRange<>(g, u);
+            for (node in : indirectNeighbors) {
```
angriman, line 108:
```suggestion
            for (node in : g.neighborRange(u)) {
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+Graph ComplexPathAlgorithm::complexPathsGraph(node seed, count threshold,
+                                              std::vector<node> *adopters) {
+    const Graph *g = inputGraph;
+    const auto n = g->numberOfNodes();
+    Graph complex_g = Graph(n); // all nodes, no edges so far
+    std::vector<bool> activated(n);
+    std::vector<count> influence(n);
+
+    auto seeds = generateSeeds(seed, *g, threshold);
```
angriman, line 128:
```suggestion
    std::vector<node> seeds = generateSeeds(seed, *g, threshold);
```

networkit/cpp/centrality/ComplexPaths.cpp
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+        std::sample(indirNeighbors.begin(), indirNeighbors.end(), std::back_inserter(seeds),
+                    needMore, randomGen);
+    }
+    return seeds;
+}
+
+Graph ComplexPathAlgorithm::complexPathsGraph(node seed, count threshold,
+                                              std::vector<node> *adopters) {
```
angriman, line 121: Why don't we use a reference?
```suggestion
                                              std::vector<node> &adopters) {
```

angriman (changes requested):
Thanks, I left a few more comments.

networkit/cpp/centrality/ComplexPaths.cpp
PetholzA: As far as i can tell, because of line 191, where we call complexPathGraphs using the nullpointer. (is there a way around this?)

angriman: I see. Then let's keep the pointer.

angriman (approved):
Thanks, LGTM.

PR #1117, opened 2023-08-15
gtests: adds randomSpanningForest test

PetholzA:
- also adjusts the algorithm, so that the spanning forest gets properly returned


networkit/cpp/components/test/ConnectedComponentsGTest.cpp
```diff
@@ -585,4 +586,18 @@ TEST_F(ConnectedComponentsGTest, testExtractLargestStronglyConnectedComponent) {
     }
 }
 
+TEST_F(ConnectedComponentsGTest, testRandomSpanningForest) {
+    Graph G(5);
```
fabratu, line 590: Better create a test graph in such a way, that a subset of edges with constant size is needed for creating a spanning tree. You can then also test for the number of edges in the spanning tree.

fabratu (changes requested):

fabratu (approved):

PR #1122, opened 2023-08-25
gtests: adds distance tests

PetholzA:


networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -147,6 +152,91 @@ TEST_F(DistanceGTest, testAStar) {
...
+
+TEST_P(DistanceGTest, testAlgebraicDistance) {
+    Aux::Random::setSeed(42, false);
+    auto G = generateERGraph(500, 0.03);
+    G.indexEdges();
+
+    AlgebraicDistance AGD(G, 10UL, 30UL, 0.5, 0UL, true);
+    AGD.preprocess();
+    node source = GraphTools::randomNode(G);
+    node target = GraphTools::randomNode(G);
```
angriman, line 176: I suggest picking a fixed pair of nodes. Sampling two random nodes here just makes code harder to debug if the test fails and does not bring benefits.

networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -147,6 +152,91 @@ TEST_F(DistanceGTest, testAStar) {
...
+    EXPECT_GT(AGD.distance(source, target), 0.2);
+    auto res = AGD.getEdgeScores();
+    for (auto dist : res) {
+        EXPECT_LE(dist, 1);
+    }
+}
+
+TEST_P(DistanceGTest, testEccentricity) {
+    // only undirected and unweighted implemented
+    if (!isDirected() && !isWeighted()) {
```
angriman, line 186: Here we should use `TEST_F` since only one combination of the parameters is supported.

networkit/cpp/distance/test/DistanceGTest.cpp
```diff
@@ -147,6 +152,91 @@ TEST_F(DistanceGTest, testAStar) {
...
 }
 
+TEST_P(DistanceGTest, testAdamicAdar) {
+    Aux::Random::setSeed(43, false);
+    auto G = generateERGraph(500, 0.03);
+    G.indexEdges();
+
+    AdamicAdarDistance AAD(G);
+    AAD.preprocess();
+    auto rndEdge = GraphTools::randomEdge(G);
```
angriman, line 162: As in the comment in `testAlgebraicDistance`, just pick a fixed edge for the test.

fabratu (approved):

PR #1125, opened 2023-08-31
gtests: adds coarsening test

PetholzA:


networkit/cpp/coarsening/test/CoarseningGTest.cpp
```diff
@@ -330,4 +331,33 @@ TEST_F(CoarseningGTest, testMatchingContractorWithSelfLoop) {
     EXPECT_EQ(G.totalEdgeWeight(), coarseG.totalEdgeWeight());
 }
 
+TEST_F(CoarseningGTest, testClusteringProjectorWithNClusterings) {
+    ErdosRenyiGenerator gen(100, 0.5);
+    Graph Gfine = gen.generate();
+
+    for (int n = 0; n < 10; ++n) {
```
fabratu, line 338: I don't think 0 makes any sense in this case. Let alone, `zetafine` should be empty (the maximum id is `-1`).

fabratu (changes requested):

fabratu (approved):

PR #1126, opened 2023-08-31
adds independentset test

PetholzA:
also changes luby algorithm to be deterministic


networkit/cpp/independentset/Luby.cpp
```diff
@@ -42,15 +42,15 @@ std::vector<bool> Luby::run(const Graph &G) {
     while (!empty()) {
         // choose set S - weighted choice of active nodes with probability $1 / 2 \omega(v)$
         std::vector<bool> S(G.numberOfNodes(), false);
-        G.parallelForNodes([&](node u) {
+        G.forNodes([&](node u) {
```
fabratu, line 45: Why is this change (and the following concerning removal of parallelism) done?

fabratu (changes requested):

networkit/cpp/independentset/Luby.cpp
```diff
@@ -75,7 +78,7 @@ std::vector<bool> Luby::run(const Graph &G) {
         });
 
         // remove S' and all neighboring nodes from V'
-        G.parallelForNodes([&](node u) {
+        G.forNodes([&](node u) {
```
fabratu, line 81: Same strategy as above should apply here (likely better performance).

fabratu (changes requested):

networkit/cpp/independentset/Luby.cpp
```diff
@@ -41,30 +47,30 @@ std::vector<bool> Luby::run(const Graph &G) {
...
-                    // remove node with smaller degree
-                    edgeweight wu = weightedDegree(u);
-                    edgeweight wv = weightedDegree(v);
-                    if (wu > wv) {
-                        S[v] = false;
-                    } else if (wv > wu) {
-                        S[u] = false;
-                    } else {          // tie
-                        S[v] = false; // arbitrary decision
+        G.parallelForNodes([&](node u) {
```
fabratu, line 61: While the code works on CI, it is not guaranteed to work due to race conditions. The atomicity of bool and `std::atomic` helps when doing atomic operations like read/write. So it might happen, that after checking for `(S[u] && S[v])`, another thread is already setting `S[v]` or `S[u]` to `false` - leading to wrong results.

I think there is no way around using a critical section:

```
        std::vector<bool> S(G.numberOfNodes());
        ...
        // remove non-independent nodes from S to get S'
        G.parallelForNodes([&](node u) {
            if (S[u]) {
                G.forNeighborsOf(u, [&](node v) {
                    if (u < v) { // exclude self-loops and enforce strict order
#pragma omp critical
                        {
                            // remove node with smaller degree from S
                            if (S[u] && S[v]) {
                                if (weightedDegree(u) > weightedDegree(v)) {
                                    S[v] = false;
                                } else {
                                    S[u] = false;
                                }
                            }
                        }
                    }
                });
            }
        });
```

While the code is slower, it is not worsened too much (benchmarked with 8 cores):

```
# Before (w. race conditions): 
❯ python benchmark_luby.py
Time:  0.018196725845336915 +/- 0.001036126555402532

# After (w. critical section):
❯ python benchmark_luby.py
Time:  0.05442938804626465 +/- 0.013922199304217321

**Review by fabratu — CHANGES_REQUESTED** [review]

File: networkit/cpp/independentset/Luby.cpp [hunk 4]
```diff
@@ -5,17 +5,23 @@
...
 
 std::vector<bool> Luby::run(const Graph &G) {
-
+    if (G.isDirected()) {
+        throw std::runtime_error("current implementation can only deal with undirected graphs");
+    }
     std::vector<bool> I(G.numberOfNodes(), false); // independent set $I = \emptyset$
     // instead of pruning the graph, store here whether a node in G is still in G'
-    std::vector<bool> V(G.numberOfNodes(), true);
+    std::vector<std::atomic<bool>> V(G.numberOfNodes());
```
fabratu, line 21: Since the new version uses `critical` for sections with potential data races, you should revert the changes with `std::atomic` (e.g. use `std::vector<bool>` instead of `std::vector<std::atomic<bool>>`. (Same for all occurrences).

fabratu (changes requested):

networkit/cpp/independentset/Luby.cpp
```diff
@@ -42,28 +44,33 @@ std::vector<bool> Luby::run(const Graph &G) {
...
-                    edgeweight wu = weightedDegree(u);
-                    edgeweight wv = weightedDegree(v);
-                    if (wu > wv) {
-                        S[v] = false;
-                    } else if (wv > wu) {
-                        S[u] = false;
-                    } else {          // tie
-                        S[v] = false; // arbitrary decision
+        G.parallelForNodes([&](node u) {
+            if (S[u]) {
```
angriman, line 58: You can avoid an indentation level:
```suggestion
            if (!S[u]) return;
```

networkit/cpp/independentset/Luby.cpp
```diff
@@ -42,28 +44,33 @@ std::vector<bool> Luby::run(const Graph &G) {
...
+                G.forNeighborsOf(u, [&](node v) {
+                    if (u < v) { // exclude self-loops and enforce strict order
+#pragma omp critical
+                        {
+                            // remove node with smaller degree from S
+                            if (S[u] && S[v]) {
```
angriman, line 64: You can avoid two indentation levels:
```suggestion
                    // exclude self-loops and enforce strict order
                    if (u >= v ) return; 
#pragma omp critical
                        {
                            // remove node with smaller degree from S
                            if (!(S[u] && S[v])) return;
```

PetholzA: This suggestion is not cleanly implementable: Returns are not allowed in a 'pragma omp critical' section, and using any 'omp cancellation' turns into error: ‘#pragma omp cancel’ may only be used in compound statements.

PR #1134, opened 2023-10-03
Use string_view instead of const string&.

angriman:
`std::string_view` is a non-owning read-only view of a string, and it helps to avoid unnecessary memory allocations [1]. I think we should start using it instead of `const std::string&`. 

[1] https://www.modernescpp.com/index.php/c-17-avoid-copying-with-std-string-view/

### DIFF
--- a/include/networkit/auxiliary/StringTools.hpp
+++ b/include/networkit/auxiliary/StringTools.hpp
@@ -10,6 +10,7 @@
 
 #include <algorithm>
 #include <string>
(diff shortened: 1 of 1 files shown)


fabratu:
That makes sense (especially for the substring methods you refactored) - I think we can merge this PR as is. Will insert the refactoring of the remaining code base to the project plan.

fabratu (approved):

PR #1137, opened 2023-10-07
Clang-format creates a new line at the end of the file.

angriman:
There are files that have an empty line at the end of a file and others that don't. Editors format files differently and these changes could end up in PR diffs. I think that we should be consistent and adopt one of the two. My recommendation is to always have a new line at the end of a file.


fabratu (approved):
I agree, good idea.

PR #1140, opened 2023-10-15
Use string_view instead of const string&.

angriman:
As done in #1134, I found other places in `Log` where we can replace `const std::string&` with `std::string_view`.


fabratu:
With the refactoring, Is it possible to remove the include of `<string>`?

angriman:
We still need it because other functions use `std::string`, like `getLogLevel()` in `Log` and `toStringF()` in `StringBuilder`.

fabratu (approved):

PR #1142, opened 2023-11-02
Feature/girgs

chistopher:
I tried to add a  generator for [GIRGs](https://doi.org/10.1017/nws.2022.32).
In contrast to #320, this PR is just a wrapper around our [implementation](https://github.com/chistopher/girgs) and I tried to keep the code complexity as low as possible. 
The GIRGs repo is added as a git submodule and builds a static library.
Networkit or its components link against it but do not expose it further.

There still remain some issues that I don't like and I would greatly appreciate your input.
- in a non-monolith build, all components link against GIRGs although only the generators module needs it
- GIRGs enables `-Wswitch-default` warnings that trigger a lot here. I can remove this from the public interface of GIRGs if desired.
- the algorithm does not catch SIGINT. Since the implementation is very fast, we could catch SIGINT in the wrapper and just ignore it to prevent an interactive python shell from exiting.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "extlibs/ttmath"]
 	path = extlibs/ttmath
 	url = https://github.com/networkit/ttmath
+[submodule "extlibs/girgs"]
+	path = extlibs/girgs
+	url = git@github.com:chistopher/girgs.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,6 +231,18 @@ else()
 endif()
 
 ################################################################################
+# Use girgs as a CMake submodule
+if(EXISTS "${PROJECT_SOURCE_DIR}/extlibs/girgs/CMakeLists.txt")
+	option(BUILD_SHARED_LIBS	""	OFF)
+	option(OPTION_BUILD_CLI		""	OFF)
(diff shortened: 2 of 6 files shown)


fabratu:
If I have not missed something, `-Wswitch-default` only triggers for the edge iterators. I would rather not add a default case, since these are exhaustive switch cases. If that's not a problem, I would recommend to disable it as already suggested.

Given the time, I will read into #320 in the next days and then give further implications on how we proceed with the integration.

fabratu:
While we can fix the first and third of your points, I am still a bit indecisive on what to do with this feature integration in general. After looking at #320, the decision was to add the generator as an external library and I can see pros and cons about it.

Can you maybe elaborate more on the following topics?

- Are there any plans to continue development on the GIRGS/HyperGIRGS tool?
- Given the integration into NetworKit was started, how much effort would it be to still do it?
- Is there a possibility to pass a function as a parameter to `girgs::generateEdges`, so that edges are getting passed to that function/handle? In that way, we might be able to remove the doubling of edges in memory.

Example without sanity checks (given `u`and `v` represent the generated edge ids):

```
template <typename L, typename T>
generateEdges(..., L handle) {
...
L(static_cast<T>(u), static_cast<T>(v));
...
}
```

PR #1127, opened 2023-08-31
fix: compactEdges

PetholzA:

PR #1146, opened 2023-11-12
Restore original removeEdge API.

angriman:
The two boolean parameters introduced to `removeEdge` in #1127 should be removed because we are assuming that the users are familiar with the internal implementation of Graph and know how `sortEdges` and `indexEdges` work. `removeEdge` should be kept as simple as it was and the additional complexity of dealing with sorted edges or indexed edges should be moved elsewhere.

I suggest that we introduce two new public member variables in Graph so that:
- advanced users that want to keep the edges/edge ids sorted can do so,
- we don't complicate NetworKit APIs.

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1979,22 +1979,12 @@ public:
      */
     bool addPartialOutEdge(Unsafe, node u, node v, edgeweight ew = defaultEdgeWeight,
                            uint64_t index = 0, bool checkForMultiEdges = false);
-
     /**
      * Removes the undirected edge {@a u,@a v}.
      * @param u Endpoint of edge.
      * @param v Endpoint of edge.
-     * @param maintainSortedEdges If set to true, the ingoing and outgoing adjacency vectors will
-     * automatically be updated to maintain a sorting (if it existed before) by performing up to n-1
-     * swaps. If the user plans to remove multiple edges, better set it to false and call
-     * sortEdges() afterwards to avoid redundant swaps. Default = false.
-     * @param maintainCompactEdgeIDs If set to true, the EdgeIDs will automatically be adjusted,
-     * so that no gaps in between IDs exist. If the user plans to remove multiple edges, better set
-     * it to false and call indexEdges(force=true) afterwards to avoid redundant re-indexing.
-     * Default = false.
      */
-    void removeEdge(node u, node v, bool maintainSortedEdges = false,
-                    bool maintainCompactEdgeIDs = false);
+    void removeEdge(node u, node v);
 
     /**
      * Removes all the edges in the graph.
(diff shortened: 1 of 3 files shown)


fabratu:
Not that I am really against handling this problem in another way - just to give some rationale why the parameters were introduced:

Currently, sorting edges and indexing edges is linked at several places with the algebraic module. When dealing with dynamic graphs (e.g. removing edges), matrices and operations are giving wrong results due to inconsistency in the datastructure. What makes it more complicated, is that the errors are not necessarily obvious for the user. 

As a result, I think the affected user has to deal with the consequences. I agree, that dealing with dynamic graphs + the algebraic module is rather advanced. Therefore, we went with adding parameters, which normal users don't touch and which does not break existing code. Also, this technique is used a lot in our code so far. Either use algorithms/functions the normal way or change default parameter for advanced users. We also tried to combine both variables to one (something like `maintainConsistency`) - however both variables have different impact on running time and solve different things. We tried to cover/explain the new parameters in the documentation, so that confusion is hopefully minimized. Now that I am re-reading it, the doc can be further improved. For example, we could add a hint, that one only needs to change these parameters when dealing with dynamic graphs.

One argument against public variables would be, that we are currently not using this broadly in NetworKit (at least I am not aware of). We would have to teach this concept to users.

angriman:
Thanks Fabian for the explanation. Still, I think that either having public constants or having private ones with corresponding setters and getters is preferable than the current solution. What we should avoid is exposing additional parameters that cover very specific use cases of the underlying data structure in such a simple API such as `removeEdge`.

If you take a look at the remove edge function in other tools like [networkx](https://networkx.org/documentation/stable/_modules/networkx/classes/graph.html#Graph.remove_edge) or [graph_tool](https://graph-tool.skewed.de/static/doc/autosummary/graph_tool.Graph.html#graph_tool.Graph.remove_edge) they take as input the edge only - as the user will expect.

The [graph_tool](https://graph-tool.skewed.de/static/doc/autosummary/graph_tool.Graph.html#graph_tool.Graph.remove_edge) example in particular seems to be doing what I'm suggesting, they have a [`set_fast_edge_removal()`](https://graph-tool.skewed.de/static/doc/autosummary/graph_tool.Graph.html#graph_tool.Graph.set_fast_edge_removal) function which changes the behavior of `remove_edge`. What about using the same idea in NetworKit? We could have private variables and introduce new `setKeepEdgesSorted()` and `setMaintainCompactEdges()` functions.

fabratu:
While I see the point about a simple `removeEdge` function, I don't think optional parameters distract developers/users too much. At least this concept is used a lot in e.g. scipy or scikit-learn. Nevertheless, I agree with the argument of doing it differently for such a fundamental function (in terms of graph manipulation).

It's maybe the best approach to not reinvent the wheel here and go with what you suggested about graph_tool. @PetholzA Can you add the two functions `setKeepEdgesSorted()`/`setMaintainCompactEdges()` and move the parameters from `removeEdge()` to a private member variable?

angriman:
Closing this because implemented in #1158

PR #1149, opened 2023-11-22
fix ErdosRenyiEnumeratorGTest

PetholzA:
adjusted the tests, because when called with many threads the assertions failed

### DIFF
--- a/networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
+++ b/networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
@@ -12,6 +12,7 @@
 #include <tuple>
 
 #include <networkit/Globals.hpp>
+#include <networkit/auxiliary/Parallelism.hpp>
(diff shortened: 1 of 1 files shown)


networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
```diff
@@ -82,7 +81,7 @@ static void testEre(const bool directed, const node n, const double prob) {
         } else {
             // count that we produced a number of edges, near the expected number
             // we could apply much sharper bounds here if we allow for rare errors
-            ASSERT_NEAR(num_edges, ere.expectedNumberOfEdges(), 0.6 * ere.expectedNumberOfEdges());
+            ASSERT_NEAR(num_edges, ere.expectedNumberOfEdges(), 0.3 * ere.expectedNumberOfEdges());
```
fabratu, line 84: Even with scaling the number of nodes, is it still necessary to worsen the bounds here?

fabratu (approved):

PR #1153, opened 2023-12-01
add const access to graph attributes

bernlu:
currently one cannot access node or edge attributes from const graphs. This PR fixes that issue by adding a few const variants of exisiting functions.


bernlu:
my clang-format has some problems currently. I will fix the code style next week.

fabratu (approved):

PR #1155, opened 2023-12-14
Remove previous deprecations for release 11.0.

fabratu:
Preparation PR for 11.0. Removes all deprecated functions/classes, which were present for 10.0 and 10.1.


bernlu (changes requested):
The changes.md lists `time()` and `timeStep()` (2x) as removed functions, but the changes do not match that claim (`time()` is not modified at all, `timeStep()` is not actually removed)

fabratu:
Thanks for the review! Right, missed removing this from `Changes.md`. `time()` and `timeStep()` were undeprecated in a previous PR.

bernlu (approved):

PR #1157, opened 2023-12-15
Fix: networkit binary format explanation and update guess format

bernlu:
The binary format description (in networkit/cpp/io/NetworKitBinaryGraph.md) is outdated. The current version of our binary format is `nkbg003` while the file says `nkbg002`. You can see here that our writer sets the version to 3:
https://github.com/networkit/networkit/blob/cb48d134b09bc9c4be6e121a01c0efee4e0b245f/networkit/cpp/io/NetworkitBinaryWriter.cpp#L272

Generally, it seems that commit 1bf051d596f18ae1a1be9cb2fc5168b47f009ab5 is just kind of broken (reverts the version information, adds duplicate text to the readme), so  I fixed that.

Since the binary reader supports both version 2 and 3, I have modifed the nkb files in `input` to include both versions. The guess format now recognizes both versions.


fabratu:
Is there also a version `nkbg001`?

bernlu:
I have never seen a `nkbg001` file and there seems to be no mention of that version in the code. Our current reader only supports 2 and 3.

fabratu:
Alright - it would be good to add the version information in the documentation more clearly. So that the user knows that there is only version 2 and 3, what the difference between those versions is, and that a re-write of a 2-version network creates a 3-version file.

fabratu (approved):

PR #1164, opened 2024-01-19
cpp: improves RmatGenerator and tests

PetholzA:
This code originated as bachelors thesis at MACSy, it has been copied without extensive sanity checks, feel free to comment on possible improvements.

### DIFF
--- a/include/networkit/generators/RmatGenerator.hpp
+++ b/include/networkit/generators/RmatGenerator.hpp
@@ -24,10 +24,19 @@ namespace NetworKit {
 class RmatGenerator final : public StaticGraphGenerator {
     count scale; ///< n = 2^scale
     count edgeFactor;
-    double a, b, c; ///< probabilities
     double defaultEdgeWeight;
     bool weighted;
     count reduceNodes;
+    bool discardSelfLoops;
+
+    // Data for the alias table:
+    std::vector<std::pair<uint32_t, uint32_t>> bits;
+    std::vector<uint8_t> numberOfBits;
+    std::vector<uint32_t> coinFlipProbability;
+    std::vector<uint32_t> coinFlipReplacement;
+    uint32_t mask;
+    std::pair<uint32_t, uint32_t> curBits{0, 0};
+    uint32_t remainingBits = 0;
 
 public:
     /**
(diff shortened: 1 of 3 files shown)


networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -12,17 +15,155 @@
...
+                curUnderfullIndex = lastOverfullIndex;
+            } else
+                break;
+        }
+    }
+}
+
+void RmatGenerator::sample(std::mt19937_64 &urng) {
+    std::uniform_int_distribution<uint64_t> dist;
+    uint64_t randomNumber = dist(urng);
```
fabratu, line 140: This number can be directly sampled using `Aux` functions.

include/networkit/generators/RmatGenerator.hpp
```diff
@@ -39,14 +48,20 @@ class RmatGenerator final : public StaticGraphGenerator {
...
-                  bool weighted = false, count reduceNodes = 0);
+                  bool weighted = false, count reduceNodes = 0, bool discardSelfLoops = true);
 
     /**
      * @return Graph to be generated according to parameters specified in constructor.
      */
     Graph generate() override;
+
+private:
+    void sample(std::mt19937_64 &urng);
```
fabratu, line 62: `sample` is a generic function name. Also, it is only used in `sampleEdge` and using internal states based on that. Better implement it directly as a lambda function inside `sampleEdge`.

networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -12,17 +15,155 @@
...
+    numberOfBits.resize(size);
+    coinFlipProbability.resize(size);
+    coinFlipReplacement.resize(size);
+    for (count i = 0; i < size; i++) {
+        bits[i] = std::make_pair(entryList[i].i, entryList[i].j);
+        numberOfBits[i] = entryList[i].numberOfBits;
+        coinFlipProbability[i] = 0;
+        coinFlipReplacement[i] = 0;
+    }
+    double baseProbability = 1.0 / (double)size;
```
fabratu, line 107: As above (casting).

networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -12,17 +15,155 @@
...
+    count lastOverfullIndex = 0;
+    count lastUnderfullIndex = 0;
+    while (true) {
```
fabratu, line 110: Avoid using `while(true)` for both readability and maintainability. Use `do {} while(...)` instead.

networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -34,36 +175,11 @@ Graph RmatGenerator::generate() {
...
-            u = u | (q >> 1);
-            v = v | (q & 1);
-        }
+    // Reset the internal state of the alias table:
+    curBits = {0, 0};
+    remainingBits = 0;
 
-        return std::make_pair(u, v);
-    });
+    auto drawEdge([&]() { return sampleEdge(Aux::Random::getURNG(), (uint8_t)scale); });
```
fabratu, line 182: As above (casting). Also use `sampleEdge`, since `drawEdge` simply now calls this.

include/networkit/generators/RmatGenerator.hpp
```diff
@@ -39,14 +48,20 @@ class RmatGenerator final : public StaticGraphGenerator {
...
+
+    std::pair<node, node> sampleEdge(std::mt19937_64 &urng, uint8_t bits);
```
fabratu, line 64: Check whether `urng` can be moved to a member variable. This is what is normally done in other algorithm classes. This would then also remove the need for passing it as a parameter.

fabratu (changes requested):
Thanks for taking care of the integration. In addition, please update the Python bindings and add tests (if needed) for the new parameter.

PR #1165, opened 2024-01-19
cpp: improves BA generator, introduces ParallelGraphBuilder.cpp

PetholzA:
This code originated as bachelors thesis at MACSy, it has been copied without extensive sanity checks, feel free to comment on possible improvements

PR #1175, opened 2024-01-31
introduces parallel BA generator

PetholzA:
new version of #1165 , because of previously incorrect fork

### DIFF
--- a/include/networkit/generators/BarabasiAlbertGenerator.hpp
+++ b/include/networkit/generators/BarabasiAlbertGenerator.hpp
@@ -71,6 +71,8 @@ public:
     BarabasiAlbertGenerator(count k, count nMax, const Graph &initGraph, bool batagelj = true);
 
     Graph generate() override;
+    Graph generateParallel();
+    Graph generateBatagelj();
 };
 
 } /* namespace NetworKit */
--- a/networkit/cpp/generators/BarabasiAlbertGenerator.cpp
+++ b/networkit/cpp/generators/BarabasiAlbertGenerator.cpp
@@ -5,13 +5,18 @@
  *      Author: forigem, Manuel Penschuck <networkit@manuel.jetzt>
  */
 
+#include <atomic>
+#include <chrono>
+#include <iostream>
 #include <random>
 #include <unordered_set>
+#include <tlx/math/clz.hpp>
 
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/Random.hpp>
 #include <networkit/auxiliary/SignalHandling.hpp>
 #include <networkit/generators/BarabasiAlbertGenerator.hpp>
+#include <networkit/graph/ParallelGraphBuilder.hpp>
 
 namespace NetworKit {
 BarabasiAlbertGenerator::BarabasiAlbertGenerator(count k, count nMax, count n0, bool batagelj)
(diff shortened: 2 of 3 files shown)


fabratu:
Since the code from this PR uses #1178, a review will be added once the corresponding feature is merged.

networkit/cpp/generators/BarabasiAlbertGenerator.cpp
```diff
@@ -46,14 +51,119 @@ BarabasiAlbertGenerator::BarabasiAlbertGenerator(count k, count nMax, const Grap
...
+                return val;
+        }
+    }
+};
+
 Graph BarabasiAlbertGenerator::generate() {
     if (!nMax)
         return Graph();
 
-    if (!batagelj)
```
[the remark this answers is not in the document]
PetholzA: This has been addressed by the parameter `parallel`. However this reverts the logic, if this function is not called with the default parameter. (When the user previously used batagelj=true this is now interpreted as parallel=true, which results in potentially different behaviour)

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -43,16 +46,16 @@ class BarabasiAlbertGenerator final : public StaticGraphGenerator {
      * @param k     Number of attachments per node
      * @param nMax  Number of nodes in the graph
      * @param n0    Number of connected nodes to begin with
-     * @param batagelj Specifies whether to use Batagelj and Brandes's method (much faster)
-     *              rather than the naive one; default: true
+     * @param parallel Specifies whether to use Batagelj and Brandes's method or a parallel variant.
```
fabratu, line 49: I suggest using the following doc-string: 

```
Specifies whether to use parallelism for the generator. If set to false, the generator follows the logic of Batagelj and Brandes's algorithm.
```

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -21,7 +21,10 @@ class BarabasiAlbertGenerator final : public StaticGraphGenerator {
     count k{0};    //!< Attachments made per node
     count nMax{0}; //!< The maximal number of nodes attached
     count n0{0};   //!< The number of initial connected nodes
-    bool batagelj; //!< Specifies whether to use batagelj's method or the original one
+    bool parallel; //!< Specifies whether to use batagelj's method or a parallel one.
```
fabratu, line 24: I suggest reversing order of description (since this follows more the logic of the conditional operator):

```
Specifies whether to use a parallel algorithm or instead the algorithm of Batagelj/Brandes.
```

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -30,7 +33,7 @@ class BarabasiAlbertGenerator final : public StaticGraphGenerator {
      * This generator implements the preferential attachment model as introduced by Barabasi and
```
fabratu, line 33: I think it is time to also change the doc string in general. Only name/reference Brandes + Batagelj and the new parallel implementation.

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -43,16 +46,16 @@ class BarabasiAlbertGenerator final : public StaticGraphGenerator {
...
-     * @param batagelj Specifies whether to use Batagelj and Brandes's method (much faster)
-     *              rather than the naive one; default: true
+     * @param parallel Specifies whether to use Batagelj and Brandes's method or a parallel variant.
+     * Default: false.
      */
-    BarabasiAlbertGenerator(count k, count nMax, count n0 = 0, bool batagelj = true);
+    BarabasiAlbertGenerator(count k, count nMax, count n0 = 0, bool parallel = false);
 
     /**
      * This generator implements the preferential attachment model as introduced by Barabasi and
```
fabratu, line 55: Same here.

networkit/cpp/generators/BarabasiAlbertGenerator.cpp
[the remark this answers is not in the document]
fabratu: Makes sense, this changes the default behavior which we should avoid.

Also not perfect, but then I would suggest using something like this for the constructor:

```
BarabasiAlbertGenerator::BarabasiAlbertGenerator(count k, count nMax, count n0, bool disableParallelism) 
... parallel(!disableParallelism) {
...
}
```

fabratu (changes requested):

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -18,57 +18,67 @@ namespace NetworKit {
...
 
     /**
-     * This generator implements the preferential attachment model as introduced by Barabasi and
-     * Albert[1]. The original algorithm is very slow and thus, the much faster method from Batagelj
-     * and Brandes[2] is implemented and the current default. The original method is no longer
-     * supported and the \p batagelj parameter will be removed in future releases.
+     * This generator uses the preferential attachment model as introduced by Barabasi and
+     * Albert[1], implemented in the much faster method from Batagelj and Brandes[2] per default
+     * where the running time is O(n+m). Furthermore there is a parallel version from Sanders and
+     * Schulz[3] implemented which can be selected by setting sequential=false which empirically
```
fabratu, line 37:
```
Schulz[3] implemented. This implementation can be selected by setting sequential=false. Empirically, the parallel version shows better running times ...
```

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -18,57 +18,67 @@ namespace NetworKit {
...
```
fabratu, line 62: Same here.

include/networkit/generators/BarabasiAlbertGenerator.hpp
```diff
@@ -18,57 +18,67 @@ namespace NetworKit {
...
+     * showed better runtimes (when executed multi-threaded).
+     *
      * [1] Barabasi, Albert: Emergence of Scaling in Random Networks
      *   http://arxiv.org/pdf/cond-mat/9910332.pdf
      * [2] ALG 5 of Batagelj, Brandes: Efficient Generation of Large Random Networks
      *   https://kops.uni-konstanz.de/bitstream/handle/123456789/5799/random.pdf?sequence=1
-     * Running time is O(n+m)
+     * [3] Peter Sanders, Christian Schulz: Scalable generation of scale-free graphs
```
fabratu, line 69: As a follow-up from the review of the Chung-Lu Model PR. We should be consistent with how we cite references in the comments. For URLs, use markdown style. Using reference numbers certainly makes the comments more readable. I suggest transferring the style from this PR also to the Chung-Lu PR.

networkit/generators.pyx
```diff
@@ -77,20 +77,27 @@ cdef extern from "<networkit/generators/BarabasiAlbertGenerator.hpp>":
...
-	This generator implements the preferential attachment model as introduced by Barabasi and Albert[1].
-	The original algorithm is very slow and thus, the much faster method from Batagelj and Brandes[2] is
-	implemented and the current default.
-	The original method can be chosen by setting \p batagelj to false.
-	[1] Barabasi, Albert: Emergence of Scaling in Random Networks http://arxiv.org/pdf/cond-mat/9910332.pdf
-	[2] ALG 5 of Batagelj, Brandes: Efficient Generation of Large Random Networks https://kops.uni-konstanz.de/bitstream/handle/123456789/5799/random.pdf?sequence=1
-
+	BarabasiAlbertGenerator(k, nMax, n0=0, sequential=True)
+
+	This generator uses the preferential attachment model as introduced by Barabasi and
```
fabratu, line 87: Indentation looks off here (but this might be due to github website).

fabratu (changes requested):

fabratu (approved):

PR #1178, opened 2024-01-31
introduces parallelGraphBuilder

PetholzA:
part of re-upload from #1165


networkit/cpp/graph/ParallelGraphBuilder.cpp
```diff
@@ -0,0 +1,107 @@
...
+            for (auto edge : edges) {
+                G.addPartialOutEdge(Unsafe{}, edge.source, edge.destination);
+            }
+#pragma omp atomic
+            numberOfHalfEdges += edges.size();
+        }
+        if (directed) {
+            for (count &edgeCount : edgeCounts) {
+                edgeCount = 0;
+            }
```
angriman, line 75:
```suggestion
            std::fill(edgeCount.begin(), edgeCount.end(), 0);
```

networkit/cpp/graph/ParallelGraphBuilder.cpp
```diff
@@ -0,0 +1,107 @@
...
+            for (count i = 0;; i++) {
+                node v = i * omp_get_max_threads() + omp_get_thread_num();
+                if (v >= n)
+                    break;
+                G.preallocateDirectedInEdges(v, edgeCounts[i] + G.degreeIn(v));
+            }
+            for (auto &edgesfromThread : inEdgesPerThread) {
+                auto &edges = edgesfromThread[omp_get_thread_num()];
+                for (auto edge : edges) {
```
angriman, line 90:
```suggestion
                for (HalfEdge edge : edges) {
```

include/networkit/graph/ParallelGraphBuilder.hpp
```diff
@@ -0,0 +1,38 @@
...
+    bool directed;
+    std::vector<std::vector<std::vector<HalfEdge>>> outEdgesPerThread;
+    std::vector<std::vector<std::vector<HalfEdge>>> inEdgesPerThread;
+
+public:
+    ParallelGraphBuilder(count n, bool directed = false);
+
+    void addEdgeParallel(index a, index b);
+
+    void addHalfEdgeParallel(Unsafe, index source, index destination);
```
angriman, line 25: Why not passing `HalfEdge`?
```suggestion
    void addHalfEdgeParallel(Unsafe, HalfEdge halfEdge);
```

include/networkit/graph/ParallelGraphBuilder.hpp
```diff
@@ -0,0 +1,38 @@
+
+#ifndef NETWORKIT_GRAPH_PARALLEL_GRAPH_BUILDER_HPP_
+#define NETWORKIT_GRAPH_PARALLEL_GRAPH_BUILDER_HPP_
+
+#include <networkit/Globals.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class ParallelGraphBuilder {
```
angriman, line 10: Add documentation.

networkit/cpp/graph/ParallelGraphBuilder.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+        for (count i = 0;; i++) {
+            node v = i * omp_get_max_threads() + omp_get_thread_num();
+            if (v >= n)
+                break;
+            G.preallocateUndirected(v, edgeCounts[i] + G.degreeOut(v));
+        }
+        for (auto &edgesfromThread : outEdgesPerThread) {
+            auto &edges = edgesfromThread[omp_get_thread_num()];
+            for (auto edge : edges) {
```
angriman, line 66:
```suggestion
            for (HalfEdge edge : edges) {
```

networkit/cpp/graph/ParallelGraphBuilder.cpp
```diff
@@ -0,0 +1,107 @@
...
+        addHalfInEdgeParallel(Unsafe{}, b, a);
+    } else {
+        addHalfEdgeParallel(Unsafe{}, a, b);
+        addHalfEdgeParallel(Unsafe{}, b, a);
+    }
+}
+
+void ParallelGraphBuilder::addHalfEdgeParallel(Unsafe, index a, index b) {
+    assert(a != b);
+    outEdgesPerThread[omp_get_thread_num()][a % omp_get_max_threads()].emplace_back(HalfEdge{a, b});
```
angriman, line 34: If you use emplace this should be enough, same for the other functions below:
```suggestion
    outEdgesPerThread[omp_get_thread_num()][a % omp_get_max_threads()].emplace_back(a, b);
```

angriman (changes requested):
Thanks for the PR, I left some comments. Since we already have a `GraphBuilder` class which also supports edge weights, why don't we refactor that class to run in parallel instead of creating a new one?

include/networkit/graph/ParallelGraphBuilder.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    void addHalfEdgeParallel(Unsafe, index source, index destination);
+
+    void addHalfInEdgeParallel(Unsafe, index source, index destination);
+
+    void addHalfOutEdgeParallel(Unsafe, index source, index destination);
+
+    Graph toGraphParallel();
+
+    void addEdgesToGraphParallel(Unsafe, Graph &G);
```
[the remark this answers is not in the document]
PetholzA: Thanks for the comments, i've refactored the parallel graph building into the existing graph builder. Therefore, the suffix might be useful to distinguish between sequential and parallel functions, what do you think?

networkit/cpp/graph/ParallelGraphBuilder.cpp
PetholzA: the current implementation of ´HalfEdge` does not have such a constuctor

fabratu: Good idea to add it then. Improves readability.

include/networkit/graph/GraphBuilder.hpp
```diff
@@ -76,8 +87,9 @@ class GraphBuilder {
      * @param n Number of nodes.
      * @param weighted If set to <code>true</code>, the graph has edge weights.
      * @param directed If set to @c true, the graph will be directed.
+     * @param parallel If set to @c true, the graph will be build in parallel.
      */
-    GraphBuilder(count n = 0, bool weighted = false, bool directed = false);
+    GraphBuilder(count n = 0, bool weighted = false, bool directed = false, bool parallel = false);
```
fabratu, line 92: Given the performance is increasing (as well if executed on one thread) and If the result is the same (and it should be maybe except for the order of nodes in the adjacency list) for both variants, then `parallel` can be set to true.

fabratu:
Please also add tests for the new functions. Thx!

include/networkit/graph/ParallelGraphBuilder.hpp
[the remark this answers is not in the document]
fabratu: After taking a more in-depth look at the code, I would also recommend refactoring the existing code. There is already a parallel builder, which, according to the contributor of the code, is a lot slower than the new concept.

As already indicated, the old implementation supports weights, which adds a layer of complexity due to additional conditionals (even though branch prediction will likely cause some remedy here). Also, the old implementation, when calling `completeGraph` (this corresponds to `toGraphParallel` in the new implementation) does `shrinkToFit()` and `reset()` as a final step. Would something like this also be necessary with the new refactored implementation? Does this have performance implications? What happens if the new parallel build is completed twice?

I would suggest doing the following:

1. Implement support for weighted edges for the new parallel builder
2. Benchmark the old vs. the new implementation
3. If the new implementation is still faster, refactor old `toGraphParallel()` to use the new `toGraphParallel`. 
 
For the refactoring step, I also left some other comments.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -366,4 +378,97 @@ count GraphBuilder::numberOfEdges(const Graph &G) {
...
+                }
+#pragma omp atomic
+                numberOfHalfEdges += edges.size();
+            }
+        }
+    }
+    G.setEdgeCount(unsafe, numberOfHalfEdges / 2);
+}
+
+Graph GraphBuilder::toGraphParallel() {
```
fabratu, line 469: This corresponds to `toCompleteGraph` in the former implementation.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -366,4 +378,97 @@ count GraphBuilder::numberOfEdges(const Graph &G) {
...
+}
+
+void GraphBuilder::addHalfOutEdgeParallel(Unsafe, HalfEdge halfEdge) {
+    node a = halfEdge.source;
+    node b = halfEdge.destination;
+    assert(a != b);
+    outEdgesPerThread[omp_get_thread_num()][a % omp_get_max_threads()].emplace_back(HalfEdge{a, b});
+}
+
+void GraphBuilder::addEdgesToGraphParallel(Unsafe, Graph &G) {
```
fabratu, line 414: This function only gives valid results, if `addHalf/Out/In/EdgeParallel` was used for adding edges. Either make sure to check this accordingly or better change the public API during refactoring.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -366,4 +378,97 @@ count GraphBuilder::numberOfEdges(const Graph &G) {
     }
 }
 
+// Parallel functions
+
+void GraphBuilder::addEdgeParallel(node a, node b) {
```
fabratu, line 383: See comment at function `addEdgesInParallel`: If the new concept is faster (also for one thread), then drop the previous `addHalfEdge` function (and others) in favor of the new one(s).  If for the sequential case, nothing is gained, a wrapper/template is needed to distinguish between sequential and parallel insertion.

include/networkit/graph/ParallelGraphBuilder.hpp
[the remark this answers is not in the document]
angriman: @fabratu I agree with your points. About 3., assuming that the parallel implementation is faster than the sequential, why don't we just implement a single `toGraph` function which runs in parallel, and drop `toGraphSequential` and `toGraphParallel`? I don't see the point of maintaining two different functions that achieve the same goal.

PR #1179, opened 2024-01-31
improves rmat generator + adds tests

PetholzA:
re-open #1164 from different fork

### DIFF
--- a/networkit/cpp/generators/RmatGenerator.cpp
+++ b/networkit/cpp/generators/RmatGenerator.cpp
@@ -3,19 +3,79 @@
  *
  *  Created on: 18.03.2014
  *      Author: Henning, cls
+ *
+ * Uses the algorithm described by Hübschle-Schneider and Sanders in
+ * "Linear Work Generation of R-MAT Graphs" https://arxiv.org/abs/1905.03525
  */
 
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/NumericTools.hpp>
+#include <networkit/auxiliary/PrioQueue.hpp>
 #include <networkit/auxiliary/Random.hpp>
 #include <networkit/generators/RmatGenerator.hpp>
 
 namespace NetworKit {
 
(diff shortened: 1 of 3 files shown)


networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -3,26 +3,143 @@
...
+            curBits.second &= (1 << remainingBits) - 1;
+            return result;
+        }
+        result.first = result.first << remainingBits | curBits.first;
+        result.second = result.second << remainingBits | curBits.second;
+        input_bits -= remainingBits;
+
+        // sample step
+        [this]() {
+            uint64_t randomNumber = Aux::Random::integer();
```
fabratu, line 132: Avoid `Aux::Random::integer();`, since this function causes a slowdown due to every time creating a new distribution.

networkit/cpp/generators/RmatGenerator.cpp
```diff
@@ -34,36 +151,11 @@ Graph RmatGenerator::generate() {
...
-            u = u | (q >> 1);
-            v = v | (q & 1);
-        }
+    // Reset the internal state of the alias table:
+    curBits = {0, 0};
+    remainingBits = 0;
 
-        return std::make_pair(u, v);
-    });
+    urng = Aux::Random::getURNG();
```
fabratu, line 158: Use: `auto &urng = Aux::Random::getURNG();` (make it a member variable of the function) and move it directly to `sampleEdge`. Instead, introduce a member variable of the class for the integer-distribution and initialize it here (in general: one distribution per thread is needed). This avoids creating a new distribution each time (see: `Aux::Random::integer()`).

fabratu (changes requested):

fabratu:
The remaining issue with the code style is due to a clang-format minor update, changing the styling rules. This is oos of this PR.

fabratu (approved):

PR #1181, opened 2024-02-01
Pin clang-format version to 17 and mitigate ValueErrors

fabratu:
Reasons for this PR:

- Other versions of clang-format in general lead to style errors
- For some distributions, "clang-format" also includes binaries/tools with a non-numeric suffix, like `clang-format-diff`. These are now checked.

### DIFF
--- a/extrafiles/tooling/CppClangFormat.py
+++ b/extrafiles/tooling/CppClangFormat.py
@@ -31,15 +31,18 @@ def isSupported(cmd):
 	# Read major revision number from "clang-format version XX.XX.XX ... "
 	version = str(subprocess.check_output([cmd, "--version"],
 		universal_newlines=True, env=getEnv())).strip().split()[-1].split('.')[0]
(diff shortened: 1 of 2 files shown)


extrafiles/tooling/CppClangFormat.py
```diff
@@ -51,12 +54,11 @@ def findClangFormat():
 
 	raise FileNotFoundError("clang-format binary not found. We searched for:\n " + "\n ".join(allowed))
 
-def runClangFormat(inputFilename, outputFilename, clangFormatBinary = 'clang-format-16'):
+def runClangFormat(inputFilename, outputFilename, clangFormatBinary = 'clang-format-17'):
```
angriman, line 57: Would this work?

```suggestion
def runClangFormat(inputFilename, outputFilename, clangFormatBinary = f"clang-format-{nkt.CLANG_FORMAT_VERSION"):
```

fabratu:
Will squash on merge in the end.

angriman (approved):

PR #1186, opened 2024-02-07
generators: adds ConfigurationModel

PetholzA:
adds new static degree sequence generator (code originated as bachelors thesis at MACSy)

### DIFF
--- a/networkit/cpp/generators/ConfigurationModel.cpp
+++ b/networkit/cpp/generators/ConfigurationModel.cpp
@@ -1,0 +1,85 @@
+
+#include <networkit/generators/ConfigurationModel.hpp>
+
+#include <networkit/auxiliary/Random.hpp>
+#include <networkit/generators/HavelHakimiGenerator.hpp>
+#include <networkit/graph/Graph.hpp>
+
+#include <iostream>
+
+namespace NetworKit {
+
+ConfigurationModel::ConfigurationModel(const std::vector<count> &sequence)
+    : StaticDegreeSequenceGenerator(sequence) {
+
+    // Throw an error if the sequence isn't realizable:
+    std::sort(seq.begin(), seq.end(), std::greater<count>());
+    HavelHakimiGenerator(seq).generate();
+}
+
+Graph ConfigurationModel::generate() {
+    do {
+        Graph result{seq.size()};
+
+        std::vector<index> urn{};
+        std::vector<count> remaining(seq.size());
+        std::vector<bool> alreadyTaken(seq.size(), false);
+        index remainingNodes = seq.size();
+
+        // Fill the urn:
+        for (index i = 0; i < seq.size(); ++i) {
+            for (count j = 0; j < seq[i]; ++j) {
+                urn.push_back(i);
+            }
+            remaining[i] = seq[i];
+            if (seq[i] == 0) {
+                --remainingNodes;
+            }
+        }
+
+        count remainingSize = urn.size();
+        index curIndex = 0;
+
+        auto sample = [&remainingSize, &urn, &alreadyTaken, &curIndex]() -> index {
+            index result;
+            do {
+                --remainingSize;
+                index i;
+                do {
+                    i = Aux::Random::integer(remainingSize);
+                    result = urn[i];
+                } while (alreadyTaken[result]);
+                urn[i] = urn[remainingSize];
+            } while (result <= curIndex);
+            return result;
+        };
+
+        std::vector<index> neighbors;
+        for (curIndex = 0; curIndex < seq.size(); ++curIndex) {
+            if (remaining[curIndex] == 0)
+                continue;
+            remainingNodes--;
+            if (remaining[curIndex] > remainingNodes)
+                break; // Impossible
+            neighbors.clear();
+            for (node j = 0; j < remaining[curIndex]; ++j) {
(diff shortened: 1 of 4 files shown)


networkit/cpp/generators/ConfigurationModel.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        auto sample = [&remainingSize, &urn, &alreadyTaken, &curIndex]() -> index {
+            index result;
+            do {
+                --remainingSize;
+                index i;
+                do {
+                    i = Aux::Random::integer(remainingSize);
+                    result = urn[i];
+                } while (alreadyTaken[result]);
```
angriman, line 51: This becomes very inefficient if many elements are already taken. Why don't we keep a `std::vector<index>` of available indices, sample one element from it using `std::sample`, and remove the sampled element from the sequence?

networkit/cpp/generators/ConfigurationModel.cpp
```diff
@@ -0,0 +1,85 @@
...
+            return result;
+        };
+
+        std::vector<index> neighbors;
+        for (curIndex = 0; curIndex < seq.size(); ++curIndex) {
+            if (remaining[curIndex] == 0)
+                continue;
+            remainingNodes--;
+            if (remaining[curIndex] > remainingNodes)
+                break; // Impossible
```
angriman, line 63: If this should never happen, then consider adding an `assert`.

networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -228,6 +229,65 @@ TEST_F(GeneratorsGTest, testClusteredRandomGraphGeneratorSparseCommunities) {
     }
 }
 
+TEST_F(GeneratorsGTest, testConfigurationModelGeneratorWithEdgeSwitchingOnRealSequence) {
+    METISGraphReader reader;
+    auto graphs = {"input/jazz.graph", "input/lesmis.graph"};
+
+    for (auto path : graphs) {
+        Graph G = reader.read(path);
+        count n = G.numberOfNodes();
```
angriman, line 238:
```suggestion
        count n = G.upperNodeIdBound();
```

networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -228,6 +229,65 @@ TEST_F(GeneratorsGTest, testClusteredRandomGraphGeneratorSparseCommunities) {
...
+        if (volume < 50000) {
+            std::vector<count> testSequence(n);
+            G2.forNodes([&](node u) { testSequence[u] = G2.degree(u); });
+            Aux::Parallel::sort(testSequence.begin(), testSequence.end(), std::greater<count>());
+            Aux::Parallel::sort(sequence.begin(), sequence.end(), std::greater<count>());
+
+            for (index i = 0; i < n; ++i) {
+                EXPECT_EQ(sequence[i], testSequence[i]);
+            }
+        }
```
angriman, line 258: If this test fails, it's not clear for which instance it failed. I'd suggest splitting this test into two tests, one where the volume is expected to be < 50000 and another one where the volume is expected to be >= 50000.

networkit/cpp/generators/test/GeneratorsGTest.cpp
```diff
@@ -228,6 +229,65 @@ TEST_F(GeneratorsGTest, testClusteredRandomGraphGeneratorSparseCommunities) {
...
+    }
+}
+
+TEST_F(GeneratorsGTest, testConfigurationModelGeneratorWithRejectionSamplingOnRealSequence) {
```
angriman, line 262: Same for this test.

angriman (changes requested):

networkit/cpp/generators/ConfigurationModel.cpp
PetholzA: As far as i could test, this way is more efficient, here's why: We are sampling from the urn (which represents the degree seq in the following way: If the largest degree is 36, there are 36 `0` entries, second largest degree is 22, there are 22 `1` entries and so on ..) 

When sampling from the urn, the probability of being sampled is proportional to the degree of the nodes. Therefore, when we want to sample in a similar fashion, we need a practical copy of the urn, which only contains "available indices". Updating this rather large vector for every sample takes much longer than re-sampling (according to local testing). 

If you have any other ideas for a more efficient implementation, feel free to elaborate.

networkit/cpp/generators/ConfigurationModel.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+namespace NetworKit {
+
+ConfigurationModel::ConfigurationModel(const std::vector<count> &sequence)
+    : StaticDegreeSequenceGenerator(sequence) {
+
+    std::sort(seq.begin(), seq.end(), std::greater<count>());
+
+    // Throw an error if the sequence isn't realizable:
+    HavelHakimiGenerator(seq).generate();
```
fabratu, line 18: This creates `O(n+m)` temp. memory overhead. Is the Havel Hakimi algorithm faster than calling `isRealizable()` from the base class?

fabratu (changes requested):

fabratu (approved):

Issue #1104, opened 2023-07-11
CSRMatrix::setValue does not update isSorted causing wrong results when accessing values


bernlu:
The internal flag `isSorted` (tracks whether each row in the `columnIdx` vector is sorted) is not consistent with the actual state of the `columnIdx` vector. Inserting new values with `setValue` does not always keep the vector sorted, but will never update the state of `isSorted`. This causes incorrect results when accessing values with `CSRMatrix(i,j)`.

Small example:
```c++
CSRMatrix A(2);
A.setValue(0, 0, 1);
A.setValue(0, 1, -1);
A.setValue(1, 0, -1);
A.setValue(1, 1, 1);
// A =
// 1 -1
// -1 1
for (index i = 0; i < A.numberOfRows(); i++) {
    for (index j = 0; j < A.numberOfColumns(); j++)
        std::cout << A(i, j) << ", ";
    std::cout << std::endl;
}
```
Output:
```
0, -1,
0, 1,
```

PR #1158, opened 2023-12-22
restore_removeEdge_api

PetholzA:
this commit restores the original removeEdge api ( as referenced in #1146 ) and adjusts the tests

PR #1190, opened 2024-02-28
Bug fix/csr matrix set value binary search bug

FreestyleBuild:
A bug in the binary search of GeneralCSRMatrix caused the collumn index array to be not sorted, if setValue operations are made in the wrong order. An example can be seen in #1104. The binary search function has been replaced at the appropriate places, now allowing setValue operations in any order, while keeping the values sorted. A test has been added to check this behavior for the example from #1104 and two other matrices.

### DIFF
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -64,8 +64,8 @@ public:
      * @param isSorted If the matrix representation should uses sorted vectors.
      */
     CSRGeneralMatrix(count nRows, count nCols, ValueType zero = 0)
-        : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols),
-          isSorted(true), zero(zero) {}
+        : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols), isSorted(true),
+          zero(zero) {}
 
     /**
      * Constructs the @a dimension x @a dimension Matrix from the elements at
@@ -286,7 +286,7 @@ public:
 
         index colIdx = none;
 
-        if (! isSorted) {
+        if (!isSorted) {
             for (index k = rowIdx[i]; k < rowIdx[i + 1]; ++k) {
                 if (columnIdx[k] == j) {
                     colIdx = k;
@@ -298,7 +298,7 @@ public:
             colIdx = static_cast<index>(it - columnIdx.begin());
         }
 
-        if (! isSorted) {
+        if (!isSorted) {
             if (colIdx != none) {
                 if (value != zero) { // update existing value
                     nonZeros[colIdx] = value;
@@ -390,9 +390,7 @@ public:
     /**
      * Make the matrix use an unsorted representation.
      */
-    void makeUnsorted() {
-        isSorted = false;
-    }
+    void makeUnsorted() { isSorted = false; }
 
     /**
      * @return Row @a i of this matrix as vector.
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -50,7 +50,6 @@ public:
      * Constructs the CSRGeneralMatrix with size @a dimension x @a dimension.
      * @param dimension Defines how many rows and columns this matrix has.
      * @param zero The zero element (default = 0).
-     * @param isSorted If the matrix representation should uses sorted vectors.
      */
     CSRGeneralMatrix(count dimension, ValueType zero = 0)
         : rowIdx(dimension + 1), columnIdx(0), nonZeros(0), nRows(dimension), nCols(dimension),
@@ -61,7 +60,6 @@ public:
      * @param nRows Number of rows.
      * @param nCols Number of columns.
      * @param zero The zero element (default = 0).
-     * @param isSorted If the matrix representation should uses sorted vectors.
      */
     CSRGeneralMatrix(count nRows, count nCols, ValueType zero = 0)
         : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols), isSorted(true),
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -286,7 +286,7 @@ public:
 
         index colIdx = none;
 
-        if (!sorted()) {
+        if (! isSorted) {
             for (index k = rowIdx[i]; k < rowIdx[i + 1]; ++k) {
                 if (columnIdx[k] == j) {
                     colIdx = k;
@@ -298,7 +298,7 @@ public:
             colIdx = static_cast<index>(it - columnIdx.begin());
         }
 
-        if (!sorted()) {
+        if (! isSorted) {
             if (colIdx != none) {
                 if (value != zero) { // update existing value
                     nonZeros[colIdx] = value;
@@ -316,7 +316,7 @@ public:
 
                 // update rowIdx
                 for (index k = i + 1; k < rowIdx.size(); ++k) {
-                    rowIdx[k]++;
+                    ++rowIdx[k];
                 }
             }
         } else {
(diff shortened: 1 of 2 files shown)


networkit/cpp/algebraic/test/MatricesGTest.cpp
```diff
@@ -1367,5 +1367,165 @@ TEST_F(CSRMatrixGTest, testCSRMatrixSort) {
...
+    A.setValue(0, 0, 1);
+    A.setValue(0, 1, -1);
+    A.setValue(1, 0, -1);
+    A.setValue(1, 1, 1);
+    A.setValue(0, 2, 1);
+    A.setValue(1, 2, -1);
+
+    ASSERT_FALSE(A.sorted());
+
+    ASSERT_EQ(A(0, 0), 1);
```
angriman, line 1465: Same here, please use `EXPECT_EQ`.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -69,20 +50,22 @@ class CSRGeneralMatrix {
      * Constructs the CSRGeneralMatrix with size @a dimension x @a dimension.
      * @param dimension Defines how many rows and columns this matrix has.
      * @param zero The zero element (default = 0).
+     * @param isSorted If the matrix representation should uses sorted vectors.
      */
-    CSRGeneralMatrix(count dimension, ValueType zero = 0)
+    CSRGeneralMatrix(count dimension, ValueType zero = 0, bool isSorted = true)
```
angriman, line 55: As discussed in #1146 and #1158, I'd suggest not complicating the constructor with additional parameters. Instead, as done in #1158, we could add some getter and setter functions for the `isSorted` variable.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -69,20 +50,22 @@ class CSRGeneralMatrix {
...
      * Constructs the CSRGeneralMatrix with size @a nRows x @a nCols.
      * @param nRows Number of rows.
      * @param nCols Number of columns.
      * @param zero The zero element (default = 0).
+     * @param isSorted If the matrix representation should uses sorted vectors.
      */
-    CSRGeneralMatrix(count nRows, count nCols, ValueType zero = 0)
-        : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols), isSorted(true),
-          zero(zero) {}
+    CSRGeneralMatrix(count nRows, count nCols, ValueType zero = 0, bool isSorted = true)
```
angriman, line 66: As above.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -299,50 +285,60 @@ class CSRGeneralMatrix {
...
+                                       columnIdx.begin() + rowIdx[i + 1], j);
+            colIdx = static_cast<index>(it - columnIdx.begin());
         }
 
-        if (colIdx != none && colIdx >= rowIdx[i]
-            && columnIdx[colIdx] == j) { // the matrix already has an entry at (i,j) => replace it
-            if (value == getZero()) {    // remove the nonZero value
-                columnIdx.erase(columnIdx.begin() + colIdx);
-                nonZeros.erase(nonZeros.begin() + colIdx);
+        if (!sorted()) {
```
angriman, line 301:
```suggestion
        if (isSorted) {
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -299,50 +285,60 @@ class CSRGeneralMatrix {
         assert(j < nCols);
 
         index colIdx = none;
-        if (nnzInRow(i) == 0) {
-            colIdx = none;
-        } else if (!sorted()) {
+
+        if (!sorted()) {
```
angriman, line 289:
```suggestion
        if (isSorted) {
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -299,50 +285,60 @@ class CSRGeneralMatrix {
...
+                    }
+                }
+            } else if (value != zero) { // don't add zero values
+                columnIdx.emplace(std::next(columnIdx.begin(), rowIdx[i + 1]), j);
+                nonZeros.emplace(std::next(nonZeros.begin(), rowIdx[i + 1]), value);
 
                 // update rowIdx
                 for (index k = i + 1; k < rowIdx.size(); ++k) {
-                    --rowIdx[k];
+                    rowIdx[k]++;
```
angriman, line 319:
```suggestion
                    ++rowIdx[k];
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -299,50 +285,60 @@ class CSRGeneralMatrix {
...
+            } else if (value != zero) { // don't add zero values
+                columnIdx.emplace(std::next(columnIdx.begin(), colIdx), j);
+                nonZeros.emplace(std::next(nonZeros.begin(), colIdx), value);
 
-            // update rowIdx
-            for (index k = i + 1; k < rowIdx.size(); ++k) {
-                rowIdx[k]++;
+                // update rowIdx
+                for (index k = i + 1; k < rowIdx.size(); ++k) {
+                    rowIdx[k]++;
```
angriman, line 340:
```suggestion
                    ++rowIdx[k];
```

angriman (changes requested):
Thanks, I left a few more comments.

include/networkit/algebraic/CSRGeneralMatrix.hpp
FreestyleBuild: Would setting the sorted value not require unnecessary effort, if it is set from false to true? Or should the matrix be sorted by default and the setter only allows to create an unsorted version. In this case I would suggest a makeUnsorted() function or similar, instead of a setter-function.

angriman: We could have the matrix sorted by default. If some user does not require that functionality, we could allow them to turn it off with some `makeUnosrted()` function as you suggest. What we should avoid is polluting the constructors with variables that control an internal behavior of the data structure.

FreestyleBuild: Latest commit includes the makeUnsorted function and the old constructors

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -69,6 +50,7 @@ class CSRGeneralMatrix {
      * Constructs the CSRGeneralMatrix with size @a dimension x @a dimension.
      * @param dimension Defines how many rows and columns this matrix has.
      * @param zero The zero element (default = 0).
+     * @param isSorted If the matrix representation should uses sorted vectors.
```
angriman, line 53: This should be removed.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -79,6 +61,7 @@ class CSRGeneralMatrix {
      * @param nRows Number of rows.
      * @param nCols Number of columns.
      * @param zero The zero element (default = 0).
+     * @param isSorted If the matrix representation should uses sorted vectors.
```
angriman, line 64: Same here.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -79,6 +61,7 @@ class CSRGeneralMatrix {
...
      */
     CSRGeneralMatrix(count nRows, count nCols, ValueType zero = 0)
         : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols), isSorted(true),
```
angriman, line 67:
```suggestion
        : rowIdx(nRows + 1), columnIdx(0), nonZeros(0), nRows(nRows), nCols(nCols),
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
FreestyleBuild: Changed

include/networkit/algebraic/CSRGeneralMatrix.hpp
FreestyleBuild: Done

PetholzA:
Thanks again for your contribution @FreestyleBuild, the bug has been adressed in a more general overhaul of the algebraic module #1228, which has been merged already. Thereby, this can be closed.

PR #1191, opened 2024-02-28
New feature: convert networkx attributes to networkit attributes

bernlu:
Adds support for node and edge attributes to the networkx -> networkit graph conversion function.
Additionally, adds python bindings for the attribute getter functions.

This PR is not complete yet. The code should be finished (imo), but there is a weird error when accessing edge attributes of directed graphs. I think this problem is not related to the new code in this PR, but more investigation is necessary.

### DIFF
--- a/networkit/nxadapter.py
+++ b/networkit/nxadapter.py
@@ -1,6 +1,7 @@
 """
 This module handles compatibility between NetworKit and NetworkX
 """
+from typing import Union
 
 # local imports
 from . import graph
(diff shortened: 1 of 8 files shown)


networkit/nxadapter.py
```diff
@@ -1,60 +1,166 @@
...
+                    if typeMap
+                    else _inferType(value)
+                )
+                try:
+                    attribute = nkG.getEdgeAttribute(str(key), valueType or str)
+                except RuntimeError:  # attribute does not exist or is of different type
+                    try:
+                        attribute = nkG.attachEdgeAttribute(str(key), valueType or str)
+                        if valueType is None:
+                            print(
```
fabratu, line 145: Better use `warn` instead of `print`

fabratu (dismissed):

fabratu (approved):

fabratu:
Merging this early, since the feature is needed for ND 2024 and the recent changes on the PR were only minor CI fixes.

PR #1194, opened 2024-03-21
introduces new parallel graph builder and deprecates old one

PetholzA:
re-upload of #1178

### DIFF
--- a/include/networkit/graph/GraphBuilder.hpp
+++ b/include/networkit/graph/GraphBuilder.hpp
@@ -44,23 +44,36 @@ class GraphBuilder {
     count n;         //!< current number of nodes
     count selfloops; //!< currently encountered number of self loops
 
-    bool weighted; //!< true if the graph will be weighted, false otherwise
-    bool directed; //!< true if the graph will be directed, false otherwise
+    bool weighted;                  //!< true if the graph will be weighted, false otherwise
+    bool directed;                  //!< true if the graph will be directed, false otherwise
+    bool autoCompleteEdges = false; //!< if set to true, all function calls of addHalfEdge() will
+                                    //!< add both half edges, reducing the time to build
 
-    std::vector<std::vector<node>>
-        outEdges; //!< (outgoing) edges, for each edge (u, v) v is saved in
-                  //!< outEdges[u] and for undirected also u in outEdges[v]
-    std::vector<std::vector<edgeweight>>
-        outEdgeWeights; //!< same schema (and same order!) as outEdges
+    index indexInOutEdgeArrayPerThread(node u, node v) const;
 
-    std::vector<std::vector<node>> inEdges; //!< only used for directed graphs, inEdges[v] contains
-                                            //!< all nodes u that have an edge (u, v)
-    std::vector<std::vector<edgeweight>>
-        inEdgeWeights; //!< only used for directed graphs, same schema as inEdges
+    index indexInInEdgeArrayPerThread(node u, node v) const;
 
-    index indexInOutEdgeArray(node u, node v) const;
+    //!< only used in parallel graph building
+    struct HalfEdge {
+        node source;
+        node destination;
+        // HalfEdge(node source, node destination);
+        HalfEdge(){};
+        HalfEdge(node source, node destination) : source(source), destination(destination){};
+    };
 
-    index indexInInEdgeArray(node u, node v) const;
+    std::vector<std::vector<std::vector<HalfEdge>>>
+        outEdgesPerThread; //!< outgoing adjacency list for each thread, where
+                           //!< outEdges[cur_thread][u%max_threads] contains v if the edge (u,v)
+                           //!< exists
+    std::vector<std::vector<std::vector<edgeweight>>>
+        outEdgeWeightsPerThread; //!< corresponding weights (if weighted)
+    std::vector<std::vector<std::vector<HalfEdge>>>
+        inEdgesPerThread; //!< ingoing adjacency list (if directed) for each thread, where
+                          //!< inEdges[cur_thread][u%max_threads] contains v if the edge (u,v)
+                          //!< exists
+    std::vector<std::vector<std::vector<edgeweight>>>
+        inEdgeWeightsPerThread; //!< corresponding weights (if weighted)
 
 public:
     /**
@@ -76,8 +89,11 @@ public:
      * @param n Number of nodes.
      * @param weighted If set to <code>true</code>, the graph has edge weights.
      * @param directed If set to @c true, the graph will be directed.
+     * @param autoCompleteEdges If set to @c true, the edges will automatically be added to the
+     * adjacency lists of both nodes (decreases the time to build).
      */
-    GraphBuilder(count n = 0, bool weighted = false, bool directed = false);
+    GraphBuilder(count n = 0, bool weighted = false, bool directed = false,
+                 bool autoCompleteEdges = false);
 
     void reset(count n = 0);
 
@@ -122,13 +138,13 @@ public:
     /**
      * Insert an edge between the nodes @a u and @a v. If the graph is weighted
      * you can optionally set a weight for this edge. The default weight is 1.0.
+     * If setUseWholeEdges(true) has been called prior, this sets both half edges, which improves
+     * the performance of the graph builder.
      * @param u Endpoint of edge.
      * @param v Endpoint of edge.
      * @param weight Optional edge weight.
      */
-    void addHalfEdge(node u, node v, edgeweight ew = defaultEdgeWeight) {
-        addHalfOutEdge(u, v, ew);
-    }
+    void addHalfEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
     void addHalfOutEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
     void addHalfInEdge(node u, node v, edgeweight ew = defaultEdgeWeight);
 
@@ -160,9 +176,19 @@ public:
     void increaseInWeight(node u, node v, edgeweight ew);
 
     /**
-     * Generates a Graph instance. The graph builder will be reseted at the end.
+     * Generates a Graph instance. The graph builder will be resetted at the end.
      */
-    Graph completeGraph(bool parallel);
+    Graph completeGraph();
+
+    /**
+     * DEPRECATED: use completeGraph() instead which uses the parallel mode by default (if
+     * possible).
+     */
+    Graph TLX_DEPRECATED(completeGraph([[maybe_unused]] bool parallel)) {
+        WARN("GraphBuilder::completeGraph(bool parallel) is deprecated, use "
+             "GraphBuilder::completeGraph() instead");
+        return completeGraph();
+    }
 
     /**
      * Iterate over all nodes of the graph and call @a handle (lambda closure).
@@ -199,9 +225,17 @@ public:
     template <typename L>
     void parallelForNodePairs(L handle) const;
 
+    /*
+     * If set to true, the graph builder automatically adds the second part of the halfEdge when
+     * calling <code>addHalfEdge(u,v)</code>. This enables the graph building process to be more
+     * efficient, while adding the halfEdges takes slightly longer.
+     */
+    void setAutoCompleteEdges(bool completeEdges = false) { autoCompleteEdges = completeEdges; };
+
 private:
     void toGraphSequential(Graph &G);
     void toGraphParallel(Graph &G);
+    void addHalfEdgesToGraph(Graph &G);
 
     template <typename T>
     static void copyAndClear(std::vector<T> &source, std::vector<T> &target);
--- a/networkit/cpp/generators/ErdosRenyiGenerator.cpp
+++ b/networkit/cpp/generators/ErdosRenyiGenerator.cpp
@@ -29,7 +29,7 @@ Graph ErdosRenyiGenerator::generate() {
         });
     }
 
-    return builder.completeGraph(false);
+    return builder.completeGraph();
 }
 
 } /* namespace NetworKit */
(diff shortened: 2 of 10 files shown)


networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
     return n++;
 }
 
-void GraphBuilder::addHalfOutEdge(node u, node v, edgeweight ew) {
-    assert(indexInOutEdgeArray(u, v) == none);
-    outEdges[u].push_back(v);
-    if (weighted) {
-        outEdgeWeights[u].push_back(ew);
+void GraphBuilder::addHalfEdge(index a, index b, edgeweight ew) {
+    if (!useWholeEdges) {
```
fabratu, line 94: Reverse logic -> start wiht positive case `if(useWholeEdges)`

networkit/cpp/generators/ErdosRenyiGenerator.cpp
```diff
@@ -22,14 +22,14 @@ Graph ErdosRenyiGenerator::generate() {
 
     {
         ErdosRenyiEnumeratorDefault impl(nNodes, prob, directed);
-        impl.forEdgesParallel([&](int /*tid*/, node u, node v) {
+        impl.forEdges([&](int /*tid*/, node u, node v) {
```
fabratu, line 25: This actually makes the ER-Generator slower (because most of the time is spent in generating edges). If it's necessary for the ordering of the resulting adjacency lists, better call `completeGraph` sequentially (with one thread).

fabratu (changes requested):

include/networkit/graph/GraphBuilder.hpp
```diff
@@ -76,8 +86,11 @@ class GraphBuilder {
      * @param n Number of nodes.
      * @param weighted If set to <code>true</code>, the graph has edge weights.
      * @param directed If set to @c true, the graph will be directed.
+     * @param useWholeEdges If set to @c true, the edges will automatically be added to the
+     * adjacency lists of both nodes (decreases the time to build).
      */
-    GraphBuilder(count n = 0, bool weighted = false, bool directed = false);
+    GraphBuilder(count n = 0, bool weighted = false, bool directed = false,
+                 bool useWholeEdges = false);
```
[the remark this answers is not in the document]
angriman: I think that this is another example of a parameter that we should not add to a constructor. Also, you already implemented a `setUseWholeEdges`. I suggest not adding this parameter to the constructor, and better document the behavior (i.e., what is the default behavior, and what changes if a user calls `setUseWholeEdges(true)`) .

include/networkit/graph/GraphBuilder.hpp
```diff
@@ -199,9 +214,15 @@ class GraphBuilder {
     template <typename L>
     void parallelForNodePairs(L handle) const;
 
+    /*
+     * If set to true, this enables the graph builder to
```
angriman, line 218: Documentation missing?

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
 namespace NetworKit {
 
-GraphBuilder::GraphBuilder(count n, bool weighted, bool directed)
-    : n(n), selfloops(0), weighted(weighted), directed(directed), outEdges(n),
-      outEdgeWeights(weighted ? n : 0), inEdges(directed ? n : 0),
-      inEdgeWeights((directed && weighted) ? n : 0) {}
+GraphBuilder::GraphBuilder(count n, bool weighted, bool directed, bool useWholeEdges)
+    : n(n), selfloops(0), weighted(weighted), directed(directed), useWholeEdges(useWholeEdges) {
+    // pre-allocating the adjacency vectors
+    int max_threads = omp_get_max_threads();
```
angriman, line 18:
```suggestion
    const int max_threads = omp_get_max_threads();
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+    outEdgesPerThread.resize(max_threads);
+    inEdgesPerThread.resize(max_threads);
+    for (int i = 0; i < max_threads; i++) {
```
angriman, line 21:
```suggestion
    for (int i = 0; i < max_threads; ++i) {
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+        outEdgesPerThread[i].resize(max_threads);
+        inEdgesPerThread[i].resize(max_threads);
+    }
+    if (weighted) {
+        outEdgeWeightsPerThread.resize(max_threads);
+        inEdgeWeightsPerThread.resize(max_threads);
+        for (int i = 0; i < max_threads; i++) {
```
angriman, line 28:
```suggestion
        for (int i = 0; i < max_threads; ++i) {
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
 
 void GraphBuilder::reset(count n) {
     this->n = n;
     selfloops = 0;
-    outEdges.assign(n, std::vector<node>{});
-    outEdgeWeights.assign(isWeighted() ? n : 0, std::vector<edgeweight>{}),
-        inEdges.assign(isDirected() ? n : 0, std::vector<node>{}),
-        inEdgeWeights.assign((isDirected() && isWeighted()) ? n : 0, std::vector<edgeweight>{});
+
+    int max_threads = omp_get_max_threads();
```
angriman, line 39:
```suggestion
    const int max_threads = omp_get_max_threads();
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+    outEdgesPerThread.assign(max_threads, std::vector<std::vector<HalfEdge>>{});
+    outEdgeWeightsPerThread.assign(isWeighted() ? max_threads : 0,
+                                   std::vector<std::vector<edgeweight>>{}),
```
angriman, line 42: Why not
```suggestion
    if (isWeighted()) {
        outEdgeWeightsPerThread.resize(max_threads);
    }
```

Also for the other two below.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+        inEdgesPerThread.assign(isDirected() ? max_threads : 0,
+                                std::vector<std::vector<HalfEdge>>{}),
+        inEdgeWeightsPerThread.assign((isDirected() && isWeighted()) ? max_threads : 0,
+                                      std::vector<std::vector<edgeweight>>{});
+    outEdgesPerThread.resize(max_threads);
+    for (int i = 0; i < max_threads; i++) {
```
angriman, line 48:
```suggestion
    for (int i = 0; i < max_threads; ++i) {
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+        outEdgesPerThread[i].resize(max_threads);
+    }
+    if (weighted) {
+        outEdgeWeightsPerThread.resize(max_threads);
+        for (int i = 0; i < max_threads; i++) {
```
angriman, line 53:
```suggestion
        for (int i = 0; i < max_threads; ++i) {
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
+            outEdgeWeightsPerThread[i].resize(max_threads);
+        }
+    }
+    inEdgesPerThread.resize(max_threads);
+    if (weighted)
+        inEdgeWeightsPerThread.resize(max_threads);
+    for (int i = 0; i < max_threads; i++) {
```
angriman, line 60:
```suggestion
    for (int i = 0; i < max_threads; ++i) {
```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
-        if (x == v) {
+index GraphBuilder::indexInOutEdgeArrayPerThread(node u, node v) const {
+    int max_threads = omp_get_max_threads();
+    int thread_num = omp_get_thread_num();
+    auto &edges = outEdgesPerThread[thread_num][u % max_threads];
+    for (index i = 0; i < edges.size(); ++i) {
+        if (edges[i].source == u && edges[i].destination == v)
             return i;
-        }
     }
```
angriman, line 74: With https://en.cppreference.com/w/cpp/algorithm/all_any_none_of you can use something like:
```cpp
if (std::any_of(edges.begin(), edges.end(), [u, v](const HalfEdge& halfEdge) {
    return halfEdge.source == u && halfEdge.destination == v;
  }) {
    return i; 
} else {
  return none;
}

```

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
-        if (x == v) {
+index GraphBuilder::indexInInEdgeArrayPerThread(node u, node v) const {
+    int max_threads = omp_get_max_threads();
+    int thread_num = omp_get_thread_num();
+    auto &edges = inEdgesPerThread[thread_num][u % max_threads];
+    for (index i = 0; i < edges.size(); ++i) {
+        if (edges[i].source == u && edges[i].destination == v)
             return i;
-        }
     }
```
angriman, line 85: Same here, use `std::any_of`.

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -12,86 +12,163 @@
...
-        inEdgeWeights[u].push_back(ew);
+        outEdgeWeightsPerThread[omp_get_thread_num()][u % omp_get_max_threads()].emplace_back(ew);
     }
-    if (u == v) {
-#pragma omp atomic
-        selfloops++;
+}
+
+void GraphBuilder::addHalfInEdge(node u, node v, edgeweight ew) {
+    inEdgesPerThread[omp_get_thread_num()][u % omp_get_max_threads()].emplace_back(HalfEdge{u, v});
```
angriman, line 120: Add a proper constructor to `HalfEdge` if needed and use
```suggestion
    inEdgesPerThread[omp_get_thread_num()][u % omp_get_max_threads()].emplace_back(u, v);
```

networkit/cpp/graph/GraphBuilder.cpp
PetholzA: This does not return the index which we are looking for. Besides that, where is the advantage ?

networkit/cpp/graph/GraphBuilder.cpp
PetholzA: Only resizing does not clear the data structure (as the size is the same as before), therefore, we clear the data structure by assigning new empty vectors to ensure proper resetting.

networkit/cpp/graph/GraphBuilder.cpp
fabratu: I also can't see how we extract `i` with `std::any_of`. However, better define the result before the loop (init with `none`). If the index is found, set the result variable and break. This makes the code more readable.

include/networkit/graph/GraphBuilder.hpp
```diff
@@ -160,9 +176,11 @@ class GraphBuilder {
     void increaseInWeight(node u, node v, edgeweight ew);
 
     /**
-     * Generates a Graph instance. The graph builder will be reseted at the end.
+     * Generates a Graph instance. The graph builder will be resetted at the end.
      */
-    Graph completeGraph(bool parallel);
+    Graph completeGraph();
+
+    Graph TLX_DEPRECATED(completeGraph(bool parallel));
```
fabratu, line 183: Add a documentation hint for deprecation (and what function to use instead).

networkit/cpp/graph/GraphBuilder.cpp
```diff
@@ -101,60 +180,64 @@ void GraphBuilder::swapNeighborhood(node u, std::vector<node> &neighbours,
...
     if (vi != none) {
-        inEdgeWeights[u][vi] += ew;
+        inEdgeWeightsPerThread[omp_get_thread_num()][u % omp_get_max_threads()][vi] += ew;
     } else {
-        addHalfInEdge(u, v, ew);
+        addHalfEdge(u, v, ew);
     }
 }
 
-Graph GraphBuilder::completeGraph(bool parallel) {
```
fabratu, line 144: The deprecated implementation is missing. Look at other deprecated functions. The implementation includes a warning and usually a subsequent call to the new implementation.

Example:
```
void MyClass:myDeprecatedFunction(...) {
    WARN("DEPRECATED: Use function ... instead.");
    newFunction(...);
}
```

fabratu (changes requested):

include/networkit/graph/GraphBuilder.hpp
```diff
@@ -160,9 +176,20 @@ class GraphBuilder {
...
+    Graph completeGraph();
+
+    /**
+     * DEPRECATED: use completeGraph() instead which uses the parallel mode by default (if
+     * possible).
+     */
+    Graph TLX_DEPRECATED(completeGraph(bool parallel)) {
+        WARN("GraphBuilder::completeGraph(bool parallel) is deprecated, use "
+             "GraphBuilder::completeGraph() instead");
+        if (parallel || !parallel) // otherwise throws "-WERROR-unusedParameter"
```
fabratu, line 190: Better use `maybe_unused` for the `parallel` variable: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

PR #1198, opened 2024-03-26
Update CMakeLists.txt

### LINKED ISSUE #1197: ttmath header files missing
Quite some time ago, #491 mentioned that make install did not copy the ttmath header files.
#498 added code to fix this, by copying all files with ".hpp" ending.
At least the current ttmath library uses ".h" as file suffix for header files. As a result, the install step currently is not copying any header files for ttmath.

CharJon:
Fix #1197


fabratu (approved):

fabratu:
The pipeline errors are not due to the change, but a general CI problem and are fixed in #1199.

PR #1199, opened 2024-03-26
CI: fix Windows build and update CI + build settings

fabratu:
Notes: 

- GCC is not updated to 13.X for macOS, since the compiler is currently broken when working with `libomp` on CI (not reproducible locally).
- clang-tidy is not updated to 18.X. This is out of scope for this PR, due to non-minor refactoring.
- The `-permissive` flag is added for MSVC builds, since when building with Visual Studio, the flag is enabled by default. This marks nonconforming code (related to MSVC + a CPP-standard).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR
 	endif()
 
 elseif (MSVC)
-	set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /DNETWORKIT_OMP2 /MP /DNETWORKIT_WINDOWS")
+	set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /DNETWORKIT_OMP2 /MP /DNETWORKIT_WINDOWS /permissive-")
 
 	# TTMath requires running an ASM for MSVC which we disable here at the cost of worse performance
 	set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} /DTTMATH_NOASM=1")
--- a/networkit/cpp/distance/NeighborhoodFunctionHeuristic.cpp
+++ b/networkit/cpp/distance/NeighborhoodFunctionHeuristic.cpp
@@ -21,8 +21,8 @@ namespace NetworKit {
 NeighborhoodFunctionHeuristic::NeighborhoodFunctionHeuristic(const Graph &G, count nSamples,
                                                              SelectionStrategy strategy)
     : Algorithm(), G(&G),
-      nSamples(!nSamples ? (count)std::ceil(
-                   std::max((double)0.15f * G.numberOfNodes(), std::sqrt(G.numberOfEdges())))
+      nSamples(!nSamples ? (count)std::ceil(std::max((double)0.15f * G.numberOfNodes(),
+                                                     std::sqrt(G.numberOfEdges())))
                          : nSamples),
       strategy(strategy) {
 
--- a/networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
+++ b/networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
@@ -41,8 +41,7 @@ static void testEre(const bool directed, const node n, const double prob) {
         // initialize data
         edge_exists.assign(n * n, 0);
 
-        constexpr size_t cacheline_scale = 17; // avoid false sharing
-        std::vector<count> num_edges_thread(omp_get_max_threads() * cacheline_scale);
+        std::vector<count> num_edges_thread(omp_get_max_threads());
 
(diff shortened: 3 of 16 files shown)


CMakeLists.txt
```diff
@@ -110,32 +113,32 @@ endif()
...
+if(NOT "${NETWORKIT_CXX_SANITIZERS}" STREQUAL "")
+	set(NETWORKIT_CXX_FLAGS "-fsanitize=${NETWORKIT_CXX_SANITIZERS} ${NETWORKIT_CXX_FLAGS}")
+	set(NETWORKIT_LINK_FLAGS "-fsanitize=${NETWORKIT_CXX_SANITIZERS} ${NETWORKIT_LINK_FLAGS}")
 endif()
 
 # Check if coverage support is enabled
-if (NETWORKIT_COVERAGE)
-  set(NETWORKIT_CXX_FLAGS "${NETWORKIT_CXX_FLAGS} --coverage")
-  set(NETWORKIT_LINK_FLAGS "${NETWORKIT_LINK_FLAGS} --coverage")
+if(NETWORKIT_COVERAGE)
```
c-bebop, line 131: Would be better to make a special format PR for changes as such :smiley_cat:

networkit/cpp/distance/NeighborhoodFunctionHeuristic.cpp
```diff
@@ -21,8 +21,8 @@ namespace NetworKit {
 NeighborhoodFunctionHeuristic::NeighborhoodFunctionHeuristic(const Graph &G, count nSamples,
                                                              SelectionStrategy strategy)
     : Algorithm(), G(&G),
-      nSamples(!nSamples ? (count)std::ceil(
-                   std::max((double)0.15f * G.numberOfNodes(), std::sqrt(G.numberOfEdges())))
+      nSamples(!nSamples ? (count)std::ceil(std::max((double)0.15f * G.numberOfNodes(),
```
c-bebop, line 24: I am confused as to why all these formatting changes are part of this PR? Are these changes intentional?

networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
```diff
@@ -54,9 +53,9 @@ static void testEre(const bool directed, const node n, const double prob) {
             ASSERT_FALSE(edge_exists[edgeindex(u, v, n)]);
             edge_exists[edgeindex(u, v, n)] = 1;
 
-            // collect statistics for probabilisitic checks
+            // collect statistics for probabilistic checks
             edge_counts[edgeindex(u, v, n)]++;
-            num_edges_thread[cacheline_scale * tid]++;
+            num_edges_thread[tid]++;
```
c-bebop, line 58: Did this test fail before? How does this relate to the issue addressed by this PR?

c-bebop (changes requested):
I am not sure if I understand the scope of this PR correctly. Everything in `ci.yml` makes sense. The format changes in C++ files as well as logic changes in [ErdosRenyiEnumeratorGTest.cpp](https://github.com/networkit/networkit/pull/1199/files#diff-a9ef78e174eeffa57f33d99b6e3b889c678f4577570d5c2091f0caa4856a652c) however seem to me out of scope of this PR.

CMakeLists.txt
fabratu: You are right, I will revert the style changes for `CMakeLists.txt` (since it was not formatted before).

networkit/cpp/distance/NeighborhoodFunctionHeuristic.cpp
fabratu: The PR is about updating the CI + fixing current build issues on the CI. Maybe I should have split it. Nevertheless, all style changes except for `CMakeLists.txt` are due to raising the version of clang-format from 17 to 18.

networkit/cpp/generators/test/ErdosRenyiEnumeratorGTest.cpp
fabratu: Yes, for example, see the current build on master. For MSVC builds, the previous calculation of std::count_if together with the `cacheline_scale` led to wrong results. The refactoring doesn't change the logic of the test, however the build now works.

fabratu:
> I am not sure if I understand the scope of this PR correctly. Everything in `ci.yml` makes sense. The format changes in C++ files as well as logic changes in [ErdosRenyiEnumeratorGTest.cpp](https://github.com/networkit/networkit/pull/1199/files#diff-a9ef78e174eeffa57f33d99b6e3b889c678f4577570d5c2091f0caa4856a652c) however seem to me out of scope of this PR.

Thx for the review - this is partially answered in the comments :) Scope is: fixing CI issues (see master builds) + updating versions. Maybe you are right though, in future I will split those PRs. Also makes it easier to review them.

c-bebop (approved):
Thank you for the clarification!

fabratu:
Will merge this early, since fix is needed for CI.

PR #1210, opened 2024-04-26
Improve LAMG helper classes

bernlu:
Various small changes to the LAMG helper classes to improve readability, performance, and allow use in `const` contexts. Also fixes a few problems such that the template `Matrix` can now be any of our matrix classes - and now uses explicit template instantiation to verfiy this.

### DIFF
--- a/include/networkit/numerics/LAMG/LevelHierarchy.hpp
+++ b/include/networkit/numerics/LAMG/LevelHierarchy.hpp
@@ -40,7 +40,7 @@ public:
                              const std::vector<EliminationStage<Matrix>> &coarseningStages);
     void addAggregationLevel(const Matrix &A, const Matrix &P, const Matrix &R);
     void setLastAsCoarsest();
-    inline DenseMatrix &getCoarseMatrix() { return coarseLUMatrix; }
+    inline const DenseMatrix &getCoarseMatrix() const { return coarseLUMatrix; }
 
     inline count size() const {
         return levelType.size() + 1; // elimination + aggregation levels + finestLevel
@@ -48,7 +48,8 @@ public:
 
     LevelType getType(index levelIdx) const;
     Level<Matrix> &at(index levelIdx);
-    double cycleIndex(index levelIdx);
+    const Level<Matrix> &at(index levelIdx) const;
+    double cycleIndex(index levelIdx) const;
 };
 
 template <class Matrix>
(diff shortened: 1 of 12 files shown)


networkit/cpp/numerics/SolverLamg.cpp
```diff
@@ -0,0 +1,140 @@
...
+        for (omp_index i = 0; i < static_cast<omp_index>(r.getDimension()); ++i) {
+            for (index k = 0, aIdx = ARowIdx[i], eIdx = ERowIdx[i]; k < n; ++k) {
+                double AEvalue = r[i] - rHistory[level][k][i];
+                if (std::fabs(AEvalue) > 1e-25) {
+                    AColumnIdx[aIdx] = k;
+                    ANonZeros[aIdx] = AEvalue;
+                    ++aIdx;
+                }
+
+                double Eval = history[level][k][i] - x[i];
```
fabratu, line 55: Same as above.

networkit/cpp/numerics/SolverLamg.cpp
```diff
@@ -0,0 +1,140 @@
...
+                if (std::fabs(Eval) > 1e-25) {
+                    EColumnIdx[eIdx] = k;
+                    ENonZeros[eIdx] = Eval;
+                    ++eIdx;
+                }
+            }
+        }
+
+        CSRMatrix AE(r.getDimension(), n, ARowIdx, AColumnIdx, ANonZeros, 0.0, true);
```
fabratu, line 64: `A`, `E` and `AE` are suboptimal variable names. If possible, find better names.

networkit/cpp/numerics/SolverLamg.cpp
```diff
@@ -0,0 +1,140 @@
...
+        CSRMatrix E(r.getDimension(), n, ERowIdx, EColumnIdx, ENonZeros, 0.0, true);
+
+        Vector alpha = smoother.relax(CSRMatrix::mTmMultiply(AE, AE), CSRMatrix::mTvMultiply(AE, r),
+                                      Vector(n, 0.0), 10);
+        x += E * alpha;
+    }
+}
+
+template <>
+void SolverLamg<DynamicMatrix>::minRes(const index level, Vector &x, const Vector &r) const {
```
fabratu, line 74: Comments from the `CSRMatrix` instance also apply here.

networkit/cpp/numerics/SolverLamg.cpp
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        Vector alpha = smoother.relax(DynamicMatrix::mTmMultiply(AE, AE),
+                                      DynamicMatrix::mTvMultiply(AE, r), Vector(n, 0.0), 10);
+        x += E * alpha;
+    }
+}
+
+template <>
+void SolverLamg<DenseMatrix>::minRes(const index level, Vector &x, const Vector &r) const {
```
fabratu, line 105: Comments from the `CSRMatrix` instance also apply here.

fabratu (changes requested):

include/networkit/numerics/LAMG/LevelHierarchy.hpp
```diff
@@ -40,15 +42,16 @@ class LevelHierarchy {
...
+    inline const DenseMatrix &getCoarseMatrix() const { return coarseLUMatrix; }
 
     inline count size() const {
         return levelType.size() + 1; // elimination + aggregation levels + finestLevel
     }
 
     LevelType getType(index levelIdx) const;
     Level<Matrix> &at(index levelIdx);
-    double cycleIndex(index levelIdx);
+    const Level<Matrix> &at(index levelIdx) const;
```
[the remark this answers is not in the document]
bernlu: I added a const version because the new LAMG (#1212) uses `const LevelHierarchy` objects in the `solve` functions, where access to the `at` version is required.

bernlu:
To address the comments on variable naming: I agree that the names are not chosen well, but I do not have enough knowledge of the inner workings of this algorithm to understand what the `minRes` function is supposed to do to give better names to the variables and function signature.
The code is basically the same as before, just with more `const` identifiers, and we have tests that confirm that everything still works. I think this PR is finished as is.

fabratu (approved):

Issue #1167, opened 2024-01-19
Lamg solve vs. parallelSolve inconsistencies when solving matrices with zero-rows


bernlu:
The `solve` and `parallelSolve` functions return different results when solving linear systems of matrices that contain zero rows. `parallelSolve` seems to be broken and always returns the zero vector in my tests.

Example code:
```cpp
Graph G(3);
G.addEdge(0, 1);
CSRMatrix A = CSRMatrix::laplacianMatrix(G);

std::cout << A;
// 1, -1, 0
// -1, 1, 0
// 0, 0, 0

Vector rhs({1, 1, 0});

Lamg<CSRMatrix> lamg;
lamg.setup(A);

Vector solution(3);
lamg.solve(rhs, solution);

std::cout << solution; // (-0.5, 0.5, 0)

std::vector<Vector> rhss{rhs};
std::vector<Vector> solutions{Vector(3)};

lamg.parallelSolve(rhss, solutions);
std::cout << solutions[0]; // (0, 0, 0)
```

PR #1212, opened 2024-04-26
Lamg: parallel solve for multiple components

bernlu:
This PR adds support to the Lamg solver to deal with multi-component systems in parallel.
Related issue: #1167

Todos:
- [x] add tests
- [x] make code more const correct
- [x] rework setup functions
- [x] add support for parallelSolve
- [x] benchmarks

Note that this PR depends on #1210 and #1214 which should be merged first. (currently all commits from #1210 are listed here as well, but this PR is only for changes in LAMG.hpp and tests)

### DIFF
--- a/include/networkit/numerics/LinearSolver.hpp
+++ b/include/networkit/numerics/LinearSolver.hpp
@@ -139,11 +139,11 @@ template <class Matrix>
 std::vector<SolverStatus>
 LinearSolver<Matrix>::parallelSolve(const std::vector<Vector> &rhs, std::vector<Vector> &results,
                                     count maxConvergenceTime, count maxIterations) const {
-    std::vector<SolverStatus> stati(rhs.size());
+    std::vector<SolverStatus> stats(rhs.size());
     for (index i = 0; i < rhs.size(); ++i) {
-        stati[i] = solve(rhs[i], results[i], maxConvergenceTime, maxIterations);
+        stats[i] = solve(rhs[i], results[i], maxConvergenceTime, maxIterations);
     }
-    return stati;
+    return stats;
 }
 
 } /* namespace NetworKit */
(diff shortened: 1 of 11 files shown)


include/networkit/numerics/LinearSolver.hpp
```diff
@@ -134,12 +136,14 @@ void LinearSolver<Matrix>::setupConnected(const Graph &graph) {
 }
 
 template <class Matrix>
-void LinearSolver<Matrix>::parallelSolve(const std::vector<Vector> &rhs,
-                                         std::vector<Vector> &results, count maxConvergenceTime,
-                                         count maxIterations) {
+std::vector<SolverStatus>
+LinearSolver<Matrix>::parallelSolve(const std::vector<Vector> &rhs, std::vector<Vector> &results,
+                                    count maxConvergenceTime, count maxIterations) const {
+    std::vector<SolverStatus> stati(rhs.size());
```
fabratu, line 142: Is this referring to the plural of `status`? Maybe then better use `stats` as a name.

networkit/cpp/numerics/test/LAMGGTest.cpp
```diff
@@ -1,121 +1,229 @@
 /*
  * LAMGGTest.cpp
  *
- *  Created on: 20.11.2014
- *      Author: Michael
+ *  Created on: 24.04.2024
```
fabratu, line 4: Why are there now two files for testing? I think at least the file names are not self-explanatory in this case.

networkit/cpp/numerics/test/LAMGSolverGTest.cpp
```diff
@@ -0,0 +1,122 @@
+/*
+ * LAMGGTest.cpp
```
fabratu, line 2: Wrong file name.

fabratu (changes requested):

networkit/cpp/numerics/test/LAMGGTest.cpp
bernlu: There used to be a `LAMGGtest` which did not actually test the `LAMG` class at all - only `SolverLamg`. I moved these tests to a new test file. I will rename it to `SolverLamgGTest` now.
For new `LAMG` tests I use the now empty `LAMGGTest` file.

fabratu (approved):

PR #1214, opened 2024-05-02
add matrix and vector print functions

bernlu:


fabratu (approved):

include/networkit/algebraic/DynamicMatrix.hpp
```diff
@@ -452,4 +454,20 @@ inline void NetworKit::DynamicMatrix::parallelForNonZeroElementsInRowOrder(L han
...
+            os << std::endl;
+        for (index j = 0; j < M.numberOfColumns(); j++) {
+            if (j != 0)
+                os << ", ";
+            os << M(i, j);
+        }
+    }
+    return os;
+}
+} // namespace NetworKit
```
angriman, line 471: Since all matrix classes support a `(row, col)` operator, we should implement this function only once in a `MatrixUtil.hpp`, e.g.:
```cpp
// MatrixUtil.hpp

template <class MatrixType>
std::ostream &printMatrix(std::ostream &ostream, const MatrixType& matrix) {
  // ...
  return os;
}
```

angriman: Also, I suggest using `row` and `col` instead of `i` and `j`, and use pre-increment instead of post-increment.

angriman (changes requested):

include/networkit/algebraic/DynamicMatrix.hpp
bernlu: renamed the variables.

Considering that there are many other duplicate functions in our three matrix classes, I think a more useful change would be to create a `Matrix` superclass for all of these functions, and define the print function in this superclass. That is a project for another time though and for now I think it is fine to implement the print function in all three classes.

angriman: Yes having a `Matrix` superclass would be much better.

angriman (approved):

Issue #1121, opened 2023-08-23
CSRMatrix::forNonZeroElementsInRowOrder inconsistencies


bernlu:
Creating a 2x2 laplacian CSRMatrix from a graph with self-loops is for some reason not equal to a matrix created with the same values by hand (note that the laplacian matrix is actually not corrent, see #1120 , and this issue *may* be related to that). Additionally, the resulting laplacian matrix is not symmetric when checked with `MatrixTools::isSymmetric`.

Example code:

```c++
    Graph G(2);
    G.addEdge(0, 1);

    std::cout << CSRMatrix::laplacianMatrix(G) << std::endl;
    // 1, -1,
    // -1, 1,

    G.addEdge(1, 1);
    auto lp = CSRMatrix::laplacianMatrix(G);

    std::cout << lp << std::endl;
    // 1, -1,
    // -1, -1,

    std::cout << MatrixTools::isSymmetric(lp) << std::endl;
    // 0   -- expect 1

    CSRMatrix mat(2);
    mat.setValue(0, 0, 1);
    mat.setValue(1, 0, -1);
    mat.setValue(0, 1, -1);
    mat.setValue(1, 1, -1);

    std::cout << mat << std::endl;
    // 1, -1,
    // -1, -1,

    std::cout << MatrixTools::isSymmetric(mat) << std::endl;
    // 1

    std::cout << (lp == mat) << std::endl;
    // 0   -- expect 1
```

Both `isSymmetric` and `operator==` use `CSRMatrix::forNonZeroElementsInOrder` which seems to sometimes produce wrong values for diagonal entries. For the laplacian, the lambda is called with `(1,1,1)` even though the value should be -1 instead of 1.

Issue #1120, opened 2023-08-23
CSRMatrix::laplacianMatrix wrong results for self-loops


bernlu:
The laplacian matrix computed by `CSRMatrix::laplacianMatrix` is wrong when the graph contains self-loops.

Example:
```c++
    Graph G(2);
    G.addEdge(0, 1);
    G.addEdge(1, 1);

    std::cout << CSRMatrix::laplacianMatrix(G) << std::endl;
    // 1, -1,
    // -1, -1,
```
I am not 100% sure what the expected result is, but in any case it is not -1 on the diagonal.

Issue #1105, opened 2023-07-17
Module algebraic: laplacianMatrix(G) includes deleted nodes


bernlu:
Given a graph (of size `n`), adding and then removing a node causes the `laplacianMatrix` function to return wrong results: The resulting matrix keeps the larger size of `n+1` even though the node has been removed. This issue appears in at least two (maybe distinct?) places. 

#### Python
```py
import networkit as nk
G = nk.Graph(4)
G.addEdge(0,1)
G.addEdge(0,2)
G.addEdge(0,3)
G.addEdge(1,2)
nk.algebraic.laplacianMatrix(G).toarray()
```
> [ 3., -1., -1., -1.],
       [-1.,  2., -1.,  0.],
       [-1., -1.,  2.,  0.],
       [-1.,  0.,  0.,  1.]

(correct)
```py
newnode = G.addNode()
nk.algebraic.laplacianMatrix(G).toarray()
```
> [ 3., -1., -1., -1.,  0.],
       [-1.,  2., -1.,  0.,  0.],
       [-1., -1.,  2.,  0.,  0.],
       [-1.,  0.,  0.,  1.,  0.],
       [ 0.,  0.,  0.,  0.,  0.]

(correct)
```py
G.removeNode(newnode)
nk.algebraic.laplacianMatrix(G).toarray()
```
> [ 3., -1., -1., -1.,  0.],
       [-1.,  2., -1.,  0.,  0.],
       [-1., -1.,  2.,  0.,  0.],
       [-1.,  0.,  0.,  1.,  0.],
       [ 0.,  0.,  0.,  0.,  0.]

(wrong) this matrix should be equal to the first.

#### C++
```
Graph G(4);
G.addEdge(0, 1);
G.addEdge(0, 2);
G.addEdge(0, 3);
G.addEdge(1, 2);

std::cout << CSRMatrix::laplacianMatrix(G) << std::endl;
node newnode = G.addNode();
std::cout << CSRMatrix::laplacianMatrix(G) << std::endl;
G.removeNode(newnode);
std::cout << CSRMatrix::laplacianMatrix(G) << std::endl;
```
> 3, -1, -1, -1
-1, 2, -1, 0
-1, -1, 2, 0
-1, 0, 0, 1 

(correct)
> 3, -1, -1, -1, 0
-1, 2, -1, 0, 0
-1, -1, 2, 0, 0
-1, 0, 0, 1, 0
0, 0, 0, 0, 0

(correct)
[...]

PR #1228, opened 2024-05-27
c++: adds usability related matrix operators and tests

PetholzA:
adresses usability improvements adressed in https://github.com/orgs/networkit/projects/1/views/1?pane=issue&itemId=62136231 for the algebraic module and adds corresponding tests

### DIFF
--- a/include/networkit/algebraic/CSRGeneralMatrix.hpp
+++ b/include/networkit/algebraic/CSRGeneralMatrix.hpp
@@ -42,24 +42,21 @@ class CSRGeneralMatrix {
     bool isSorted;
     ValueType zero;
 
-    /**
-     * Binary search the sorted columnIdx vector between [@a left, @a right]
-     * for column @a j.
-     * If @a j is not present, the index that is immediately left of the place
-     * where @a j would be is returned.
-     * @param left
-     * @param right
-     * @param j
-     * @return The position of column @a j in columnIdx or the element immediately
-     * to the left of the place where @a j would be.
-     */
-    index binarySearchColumns(index left, index right, index j) const {
-        assert(sorted());
-        const auto it = std::lower_bound(columnIdx.begin() + left, columnIdx.begin() + right, j);
-        if (it == columnIdx.end() || *it != j)
-            return none;
-        return it - columnIdx.begin();
-    }
+    class IndexProxy {
+    public:
+        IndexProxy(CSRGeneralMatrix &mat, index i, index j) : Matrix{mat}, i{i}, j{j} {}
+
+        operator const ValueType &() const {
+            return const_cast<const CSRGeneralMatrix &>(Matrix)(i, j);
+        }
+
+        void operator=(double rhs) { Matrix.setValue(i, j, rhs); }
+
+    private:
+        CSRGeneralMatrix &Matrix;
+        index i;
+        index j;
+    };
 
 public:
     /** Default constructor */
@@ -206,6 +203,8 @@ public:
     /** Default copy assignment operator */
     CSRGeneralMatrix &operator=(const CSRGeneralMatrix &other) = default;
 
+    IndexProxy operator()(index i, index j) { return IndexProxy(*this, i, j); }
+
     /**
      * Compares this matrix to @a other and returns true if the shape and zero
      * element are the same as well as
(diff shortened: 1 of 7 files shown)


include/networkit/algebraic/MatrixTools.hpp
```diff
@@ -129,6 +133,22 @@ NetworKit::Graph matrixToGraph(const Matrix &matrix) {
...
+/**
+ * Converts the @a other matrix into a equivalent matrix of the template parametrized type.
+ * @param other The input Matrix.
+ * @return Matrix of type DenseMatrix, CSRMatrix or DynamicMatrix.
+ */
+template <class ToMatrix, class FromMatrix>
+ToMatrix matrixTo(const FromMatrix &other) {
+    int nRows = other.numberOfRows();
+    int nCols = other.numberOfColumns();
+    ToMatrix M(nRows, nCols, 0.0);
```
bernlu, line 145: the default/zero value of `M` should be the default/zero value of `other` ?
Also, avoid single letter variables

networkit/cpp/algebraic/test/MatricesGTest.cpp
```diff
@@ -681,6 +681,122 @@ TYPED_TEST(MatricesGTest, testMatrixSubtractionEqual) {
...
+
+    EXPECT_TRUE(mat1.isApprox(mat2, eps)); // now epsilon 0.01 > 0.001
+}
+
+TYPED_TEST(MatricesGTest, testMatrixConversion) {
+    std::vector<Triplet> triplets = {{2, 1, 1.0}, {3, 1, 2.0}, {0, 2, 1.5}, {0, 1, 1.0}};
+    // ground truth
+    DenseMatrix den(4, 4, triplets);
+    CSRMatrix csr(4, 4, triplets);
+    csr.sort(); // to ease the correctness check
```
bernlu, line 720: why do we need to sort here? does `==` not work for sorted and unsorted CSR?

bernlu:
List of issues related to this PR:
- #1120 
- #1121 
- #1105 
- #1104 

There is also another older PR (#1190) to fix some of these issues which we forgot about. We should make sure the tests added there are included in this new PR.

PetholzA:
@bernlu 
> There is also another older PR (#1190) to fix some of these issues which we forgot about. We should make sure the tests added there are included in this new PR.

All these tests work, however, considering that the tests in this PR are equivalent while being more general, i would prefer not to additionally add the other ones.

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -56,11 +56,30 @@ class CSRGeneralMatrix {
...
         const auto it = std::lower_bound(columnIdx.begin() + left, columnIdx.begin() + right, j);
-        if (it == columnIdx.end() || *it != j)
-            return none;
-        return it - columnIdx.begin();
+        if (*it > j) {
+            if (it == columnIdx.begin())
+                return none;
+            return std::prev(it) - columnIdx.begin();
+        } else
+            return it - columnIdx.begin();
```
bernlu, line 64: I think this code could be simplified to:
```cpp
if (it == columnIdx.begin())
    return none;
if (*it > j)
    it =  std::prev(it);
return it - columnIdx.begin();
```

include/networkit/algebraic/CSRGeneralMatrix.hpp
```diff
@@ -56,11 +56,30 @@ class CSRGeneralMatrix {
     index binarySearchColumns(index left, index right, index j) const {
```
bernlu, line 56: The code and documentation here do not match. Please check if  we have any code in our project that expects / deals with `binarySearchColumns` returning `none`?


Also, there is an undefined special case in the documentation: what is the return value if we search in the first row for a column that does not exist and there is no "index immediately left of the place where j would be"? (this would be index `-1`?). Currently our code returns `none`, but that is not documented.

bernlu (changes requested):

include/networkit/algebraic/CSRGeneralMatrix.hpp
PetholzA: This function is only used in the 'CSRGeneralMatrix.hpp', the doc is now adjusted

include/networkit/algebraic/CSRGeneralMatrix.hpp
PetholzA: this has been adressed by removing that function entirely

PR #1234, opened 2024-06-25
windows: fix numpy2.0 compatibility

PetholzA:
this PR extends #1226, such that, the compatibility with numpy2.0 works properly with windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,7 +251,7 @@ jobs:
           python3 -c "import networkit"
 
   linux-build-min-support:
-    name: "Linux (${{ matrix.compiler }}, CPython 3.8): full"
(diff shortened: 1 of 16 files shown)


fabratu:
Ah, now we are getting other errors on the CI (see Linux gcc-8 builds). It is likely related to this: https://stackoverflow.com/questions/76389395/attributeerror-module-numpy-has-no-attribute-long

Can you also add a patch here?

fabratu:
Ok, it only seems to work if the old version of numpy (and therefore Python 3.8) is removed from the compatibility matrix.

Therefore, please split the commits into fixing Windows and raising the Python version. Also, please change the README, requirements.txt, pyproject.toml to reflect the new version requirements for both numpy and Python.

After merging, I will try to fix this incompatibility with old versions. There appears to be a compat-header for numpy. It is not clear, however, if this works well with Cython.

PR #1132, opened 2023-09-24
Add MTXGraphReader

friedagerharz:
Adding a reader for the [Matrix Market Format](https://networkrepository.com/mtx-matrix-market-format.html).

- [ ] add tests

PR #1236, opened 2024-07-04
cpp: adds MTXGraphReader and corresponding tests

PetholzA:
This PR finalizes #1132 with the additional tests. Note that the original code has been slightly adjusted, namely the `MTXGraphReader.cpp` now uses `addEdge()` instead of `addPartialEdge()` which resulted in a incomplete Graph data structure in some cases.

### DIFF
--- a/include/networkit/io/MTXGraphReader.hpp
+++ b/include/networkit/io/MTXGraphReader.hpp
@@ -1,0 +1,31 @@
+#ifndef NETWORKIT_IO_MTX_GRAPH_READER_HPP_
+#define NETWORKIT_IO_MTX_GRAPH_READER_HPP_
+
+#include <networkit/io/GraphReader.hpp>
+
+namespace NetworKit {
+
+/**
+ * @ingroup io
+ *
+ * Reader for the matrix market file format documented in
+ * https://networkrepository.com/mtx-matrix-market-format.html
+ *
+ * Does not allow complex fields.
+ *
+ */
+class MTXGraphReader final : public GraphReader {
+public:
+    MTXGraphReader() = default;
+
+    /**
+     * Takes a file path as parameter and returns a graph.
+     *
+     * @param path
+     * @return Graph
+     */
+    Graph read(const std::string &path) override;
+};
+
+} /* namespace NetworKit */
+#endif // NETWORKIT_IO_MTX_GRAPH_READER_HPP_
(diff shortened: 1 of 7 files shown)


networkit/cpp/io/MTXGraphReader.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+    Graph G(std::max(size.rows, size.columns), weighted, !symmetric);
+    std::string graphName =
+        Aux::StringTools::split(Aux::StringTools::split(path, '/').back(), '.').front();
+
+    std::optional<MTXParser::Edge> current_edge = parser.getNext(weighted);
+
+    while (current_edge.has_value()) {
+        const auto [from, to, weight] = current_edge.value();
+        std::cout<<"from= "<<from<<" to= "<<to<<" weight= "<<*weight<<"\n";
```
fabratu, line 30: Based on debugging on MSVC 14.40: The problem is the deref on optional `weight` (given the graph is not weighted). The optional is not default initialized. Instead of fixing it, I suggest use the existing `Graph::Edge` and `Graph::WeightedEdge` together with a template parameter instead of redefining these structs.

The std::cout lines should be removed in the final code anyways.

include/networkit/io/MTXGraphReader.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @ingroup io
+ *
+ * Reader for the matrix market file format documented in
+ * https://networkrepository.com/mtx-matrix-market-format.html
+ *
+ * Does not allow complex fields.
+ *
+ * TODO: check possible combinations of field values e.g.:
```
fabratu, line 16: Is this TODO still valid?

PetholzA: The code does not explicitly check for the examples given. I personally think, here the `garbage in, garbage out` rule applies: Checking for the sanity of certain graph conditions (possibly after parsing each line, or once after parsing the entire matrix) is very costly. Typically, the user knows which graphs they are using, and thereby, which conditions should be fulfilled. So making sure that the resulting graph satisfies those conditions, in my opinion, should be up to the user.

fabratu: I agree 👍 Then, in order to avoid confusion, please remove the `TODO` from the comments.

fabratu (approved):

Issue #1224, opened 2024-05-22
GraphTools.topologicalSort() assumes input graph is compacted 


jurepustos:
I've been pulling my hair out over `GraphTools.topologicalSort()` for about two days, before looking into the source code and noticing that the algorithm actually assumes that node id's go from 0 to number of nodes - 1. The documentation does not mention this and I feel that it should explicitly mention such a restriction. The library could also give a way to pass a node id map to the function if the graph is not compacted.
I plan on making a pull request with a solution in the relative near future, but I want this issue/restriction to be known.

fabratu:
Hi,

yes , thank you - that's correct. We should name the restriction in the documentation. If you need any help in creating a PR for adding (or/and adding a node map), feel free to ask.

jurepustos:
closing since #1241 fixed this issue a while ago

PR #1241, opened 2024-07-26
Fixes to TopologicalSort

jurepustos:
Fixed issue #1224 by adding an optional argument to for a continuous node id mapping, i.e. mapping of current node ids to [0,n). 
Replaced raw accesses into vectors by the index operator with calls to `.at()` which performs bounds checks. This fixes an unreported bug which causes undefined behaviour and specifically segmentation faults when node ids are not in the [0,n) interval. The new behaviour is to throw an exception with a descriptive message instead.
If a node id mapping is passed to the algorithm, it maps all node ids through the mapping (and performs bounds checks as in above paragraph). If node ids are not mapped to [0,n), the bounds checks will cause the algorithm to throw an exception with an appropriate message.
Also refactored tests for TopologicalSort and added new tests for bounds checks and node id mappings.

### DIFF
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -422,7 +422,7 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  * @return                  A vector of node-ids sorted according to their topology.
  */
 std::vector<node> topologicalSort(const Graph &G,
-                                  std::unordered_map<node, node> &nodeIdMapping = defaultNodeIdMap,
+                                  const std::unordered_map<node, node> &nodeIdMapping = {},
                                   bool checkMapping = false);
 
 /**
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -412,13 +412,18 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  * Given a directed graph G, the topology sort algorithm creates one valid topology order of nodes.
  * Undirected graphs are not accepted as input, since a topology sort is a linear ordering of
  * vertices such that for every edge u -> v, node u comes before v in the ordering.
+ * Node ids must either be continuous or you must provide a continuous node id mapping.
  *
  * This is a helper function. Instead of calling it via GraphTools, it is also possible to create a
  * TopologicalSort-object from the base-class.
- * @param   G           Directed input graph
- * @return              A vector of node-ids sorted according to their topology.
+ * @param   G               Directed input graph
+ * @param   nodeIdMapping   Optional continuous node id mapping
+ * @return                  A vector of node-ids sorted according to their topology.
  */
 std::vector<node> topologicalSort(const Graph &G);
+
+std::vector<node> topologicalSort(const Graph &G,
+                                  const std::unordered_map<node, node> &nodeIdMapping);
 
 /**
  * Randomizes the weights of the given graph. The weights are uniformly distributed in
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -5,6 +5,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
+#include "networkit/graph/TopologicalSort.hpp"
 
 #include <tlx/unused.hpp>
 #include <networkit/GlobalState.hpp>
@@ -420,9 +421,8 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  * @param   nodeIdMapping   Optional continuous node id mapping
  * @return                  A vector of node-ids sorted according to their topology.
  */
-std::vector<node> topologicalSort(const Graph &G);
-
-std::vector<node> topologicalSort(const Graph &G, std::unordered_map<node, node> &nodeIdMapping,
+std::vector<node> topologicalSort(const Graph &G,
+                                  std::unordered_map<node, node> &nodeIdMapping = defaultNodeIdMap,
                                   bool checkMapping = false);
 
 /**
--- a/include/networkit/graph/GraphTools.hpp
+++ b/include/networkit/graph/GraphTools.hpp
@@ -422,8 +422,8 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
  */
 std::vector<node> topologicalSort(const Graph &G);
 
-std::vector<node> topologicalSort(const Graph &G,
-                                  const std::unordered_map<node, node> &nodeIdMapping);
+std::vector<node> topologicalSort(const Graph &G, std::unordered_map<node, node> &nodeIdMapping,
+                                  bool checkMapping = false);
 
 /**
  * Randomizes the weights of the given graph. The weights are uniformly distributed in
--- a/include/networkit/graph/TopologicalSort.hpp
+++ b/include/networkit/graph/TopologicalSort.hpp
@@ -7,6 +7,7 @@
 #ifndef NETWORKIT_GRAPH_TOPOLOGICAL_SORT_HPP_
 #define NETWORKIT_GRAPH_TOPOLOGICAL_SORT_HPP_
 
+#include <unordered_map>
 #include <networkit/base/Algorithm.hpp>
 #include <networkit/graph/Graph.hpp>
 
@@ -21,11 +22,19 @@ class TopologicalSort final : public Algorithm {
 public:
     /**
      * Initialize the topological sort algorithm by passing an input graph. Note that topological
-     * sort is defined for directed graphs only.
+     * sort is defined for directed graphs only. Node ids must be continuous in the interval [0, n).
      *
      * @param G The input graph.
      */
     TopologicalSort(const Graph &G);
+
+    /**
+     * Initialize the topological sort algorithm by passing an input graph. Note that topological
+     * sort is defined for directed graphs only. The node id mapping must be a continuous.
+     *
+     * @param G The input graph.
+     */
+    TopologicalSort(const Graph &G, const std::unordered_map<node, node> &nodeIdMapping);
 
     /**
      * Execute the algorithm. The algorithm is not parallel.
@@ -47,6 +56,8 @@ private:
 
     const Graph *G;
 
+    const std::unordered_map<node, node> *nodeIdMap;
+
     // Used to mark the status of each node, one vector per thread
     std::vector<NodeMark> topSortMark;
 
@@ -55,6 +66,8 @@ private:
 
     // Helper structures
     count current;
+
+    void checkDirected();
 
     // Reset algorithm data structure
     void reset();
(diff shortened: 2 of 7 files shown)


fabratu:
Hi,

thank you for your first contribution to NetworKit!

In general the PR looks good. However due to the scope of the project (supporting large scale analysis) avoid using `.at()` when it is called frequently.

jurepustos:
If there's a better solution that avoids out of bounds accesses, I'm all for it. But I think the tradeoff of a small performance impact due to bounds checks vs. accidental memory corruption should be cleary in the favour of bounds checks. With memory corruption, the best scenario is a segmentation fault, while the worst is subtle breakage and wrong calculations in completely unrelated code. 

I did look at `Graph.upperNodeIdBound()` before. Is it equal to the highest node id? If it is, I could eliminate a significant portion of `.at()` calls by checking this number. Or, if that's not the case, I could make the continuous property a part of `Graph` itself. As for the node id mapping, I could make a new structure that encapsulates the mapping and makes sure it stays continuous. I'll greatly appreciate feedback on this.

jurepustos:
>The problem in this case is, that the overhead is not small in general for `.at()`. I did some random access on a mid-sized vector on a M2 macOS and it shows a 50% overhead.

Noted, I was under the impression that the impact wouldn't be this big. I should have benchmarked both versions of the algorithm to see the difference.

>Yes, your impression is correct. It gives the highest node id. There also exists the function `getContinuousNodeIds`, which retrieves the continuous mapping of a graph.

Nice, `upperNodeIdBound` makes things much easier. I'm also aware of `getContinuousNodeIds` - in hindsight, I should have asked if calling this function when a mapping is not passed would be a good decision. I'll implement it like that instead.

Considering that, I'm thinking of the following. Since the caller might have calculated a mapping already, it would still make sense to allow the caller to pass their own mapping, hence my original decision to make it a parameter. However, by having this parameter, I believe that we have the responsibility of checking the caller's mapping for correctness, or at least that mapped node ids fall in [0,n). `GraphTools.getCompactedGraph()`, which similarly takes a continuous node ids mapping, doesn't perform any checks on the mapping, so I'm guessing we're okay with not checking it? I personally dislike this and would love a better solution, perhaps a wrapper class around `std::unordered_map` that maintains continuity. Would that be wanted, or a complete waste of time?

fabratu:
Yeah, I understand the part about not liking it. We occasionally implement an additional parameter, which enables safety checks (like `addEdge` in Python) and document the costs of these check. We could also go this route here.

In general your impression is again correct, most functions per default don't check for sanity (if it is too costly). As a result, the maintenance and correct preparation of data is relayed to the user.

include/networkit/graph/TopologicalSort.hpp
```diff
@@ -47,6 +58,10 @@ class TopologicalSort final : public Algorithm {
 
     const Graph *G;
 
+    std::optional<std::unordered_map<node, node>> computedNodeIdMap;
+
+    std::unordered_map<node, node> *nodeIdMap;
```
fabratu, line 63: I suggest using `const std::unordered_map<node, node> &nodeIdMap;` and change access in rest of the code accordingly.

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -4,40 +4,94 @@
...
+#include <stdexcept>
+#include "networkit/graph/GraphTools.hpp"
 #include <networkit/graph/TopologicalSort.hpp>
 
 namespace NetworKit {
 
-TopologicalSort::TopologicalSort(const Graph &G) : G(&G) {
-    if (!G.isDirected())
+TopologicalSort::TopologicalSort(const Graph &G)
+    : G(&G), computedNodeIdMap(GraphTools::getContinuousNodeIds(G)) {
```
fabratu, line 14: `computedNodeIdMap` get computed twice - once in the intializer list of the constructor and once in line below in the if-case. As discussed, we should avoid the default computation of the list.

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -4,40 +4,94 @@
...
+
+void TopologicalSort::checkDirected() {
+    if (!G->isDirected())
         throw std::runtime_error("Topological sort is defined for directed graphs only.");
 }
 
+void TopologicalSort::checkNodeIdMap() {
+    size_t numberOfNodes = G->numberOfNodes();
+    std::vector<bool> checkTable(numberOfNodes);
+    for (auto it = nodeIdMap->begin(); it != nodeIdMap->end(); it++) {
```
fabratu, line 43:
```
for (auto& [origNode, mappedNode]: nodeIdMap) { 
    ...
```

fabratu (changes requested):

include/networkit/graph/TopologicalSort.hpp
jurepustos: Can't do, for multiple reasons. First, C++ doesn't allow references as class members. Second, I can't have `const` and also use indexing by `[]`. Third, it is perfectly valid for `nodeIdMap` to be `null`.

fabratu: The part about C++ is not true. Take a look for example at `centrality/Centrality.hpp`. There the graph `const Graph& G` is a const reference member of the class.

jurepustos: My bad, I mixed this up with a limitation in Microsoft's C++/CLI. Then I can indeed make this happen.

include/networkit/graph/GraphTools.hpp
```diff
@@ -412,13 +413,17 @@ void sortEdgesByWeight(Graph &G, bool decreasing = false);
...
  * TopologicalSort-object from the base-class.
- * @param   G           Directed input graph
- * @return              A vector of node-ids sorted according to their topology.
+ * @param   G               Directed input graph
+ * @param   nodeIdMapping   Optional continuous node id mapping
+ * @return                  A vector of node-ids sorted according to their topology.
  */
-std::vector<node> topologicalSort(const Graph &G);
+std::vector<node> topologicalSort(const Graph &G,
+                                  std::unordered_map<node, node> &nodeIdMapping = defaultNodeIdMap,
```
fabratu, line 425:
```suggestion
                                  const std::unordered_map<node, node> &nodeIdMapping = {},
```

include/networkit/graph/TopologicalSort.hpp
```diff
@@ -7,11 +7,15 @@
...
 #define NETWORKIT_GRAPH_TOPOLOGICAL_SORT_HPP_
 
+#include <optional>
+#include <unordered_map>
 #include <networkit/base/Algorithm.hpp>
 #include <networkit/graph/Graph.hpp>
 
 namespace NetworKit {
 
+static std::unordered_map<node, node> defaultNodeIdMap = {};
```
fabratu, line 17:
```suggestion

```

include/networkit/graph/TopologicalSort.hpp
```diff
@@ -21,11 +25,13 @@ class TopologicalSort final : public Algorithm {
...
     /**
      * Initialize the topological sort algorithm by passing an input graph. Note that topological
-     * sort is defined for directed graphs only.
+     * sort is defined for directed graphs only. The node id mapping must be a continuous.
      *
      * @param G The input graph.
      */
-    TopologicalSort(const Graph &G);
+    TopologicalSort(const Graph &G,
+                    std::unordered_map<node, node> &nodeIdMapping = defaultNodeIdMap,
```
fabratu, line 33:
```suggestion
                    const std::unordered_map<node, node> &nodeIdMapping = {},
```

include/networkit/graph/TopologicalSort.hpp
```diff
@@ -45,7 +51,11 @@ class TopologicalSort final : public Algorithm {
 private:
     enum class NodeMark : unsigned char { NONE, TEMP, PERM };
 
-    const Graph *G;
+    const Graph &G;
+
+    std::optional<std::unordered_map<node, node>> computedNodeIdMap;
+
+    std::unordered_map<node, node> &nodeIdMap;
```
fabratu, line 58:
```suggestion
    const std::unordered_map<node, node> &nodeIdMap;
```

networkit/cpp/graph/GraphTools.cpp
```diff
@@ -513,8 +513,10 @@ void sortEdgesByWeight(Graph &G, bool decreasing) {
         });
 }
 
-std::vector<node> topologicalSort(const Graph &G) {
-    TopologicalSort topSort(G);
+std::vector<node> topologicalSort(const Graph &G,
+                                  std::unordered_map<node, node> &nodeIdMapping,
```
fabratu, line 517:
```suggestion
                                  const std::unordered_map<node, node> &nodeIdMapping,
```

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -4,40 +4,77 @@
...
  *      Author: Fabian Brandt-Tumescheit
  */
+#include <stdexcept>
+#include "networkit/graph/GraphTools.hpp"
 #include <networkit/graph/TopologicalSort.hpp>
 
 namespace NetworKit {
 
-TopologicalSort::TopologicalSort(const Graph &G) : G(&G) {
+TopologicalSort::TopologicalSort(const Graph &G, std::unordered_map<node, node> &nodeIdMap,
```
fabratu, line 13:
```suggestion
TopologicalSort::TopologicalSort(const Graph &G, const std::unordered_map<node, node> &nodeIdMap,
```

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -47,9 +84,18 @@ void TopologicalSort::run() {
     hasRun = true;
 }
 
+node TopologicalSort::mapNode(node u) {
+    if (!nodeIdMap.empty())
```
fabratu, line 88:
```suggestion
    if (!nodeIdMap.empty()) {
        auto retIt = nodeIdMap.find(u);
        return retIt->second;
    } else if (computedNodeIdMap.has_value())
```

fabratu (changes requested):
I suggest to remove the `static` class member and use `const` for the node map. Added some code suggestion to make this work. The only significant change is in `mapNode`, since `const` does not support `operator[]` for `unordered_map`.

networkit/cpp/graph/TopologicalSort.cpp
jurepustos: For some reason, I was under the impression that `at()` and `find()` have unwanted overhead like in vectors. Of course, there's no overhead, thanks for the suggestion!

jurepustos: For some reason, I had thought `at()` or `find()` would cause a performance overhead, but that's clearly not the case for `unordered_map`. Now it all falls back into place.

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -4,40 +4,76 @@
  *  Created on: 22.11.2021
  *      Author: Fabian Brandt-Tumescheit
  */
+#include <stdexcept>
+#include "networkit/graph/GraphTools.hpp"
```
fabratu, line 8: Sorry for the tedious comment. Please change the include to: `#include <networkit/graph/GraphTools.hpp>`

fabratu (approved):

PR #1245, opened 2024-08-06
cpp: fixDynBetweennessOneNode for weighted graphs

PetholzA:
This PR addresses the issue mentioned [in this discussion](https://github.com/networkit/networkit/discussions/990). Furthermore, it adds the corresponding GTest to ensure proper integration.

### DIFF
--- a/networkit/cpp/centrality/DynBetweennessOneNode.cpp
+++ b/networkit/cpp/centrality/DynBetweennessOneNode.cpp
@@ -13,7 +13,6 @@
 #include <networkit/auxiliary/NumericTools.hpp>
 #include <networkit/auxiliary/PrioQueue.hpp>
 #include <networkit/centrality/DynBetweennessOneNode.hpp>
-
 namespace NetworKit {
 
 DynBetweennessOneNode::DynBetweennessOneNode(Graph &G, node x) : G(G), x(x) {}
@@ -22,6 +21,14 @@ DynBetweennessOneNode::DynBetweennessOneNode(Graph &G, node x) : G(G), x(x) {}
  * Run method that stores a One shortest path for each node pair and stores shortest distances
  */
 void DynBetweennessOneNode::run() {
+    if (G.isWeighted()) {
+        runWeighted();
+    } else {
+        runUnweighted();
+    }
+}
+
+void DynBetweennessOneNode::runUnweighted() {
     bcx = 0;
     distances.resize(G.upperNodeIdBound());
     sigma.resize(G.upperNodeIdBound());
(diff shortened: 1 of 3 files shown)


networkit/cpp/centrality/DynBetweennessOneNode.cpp
```diff
@@ -138,58 +144,66 @@ void DynBetweennessOneNode::update(GraphEvent event) {
...
-                if (distances[u][w] >= weightuv + distances[v][y] + weightyw
-                    && distances[v][w] == distances[v][y] + weightyw) {
-                    Pred[w] = y;
-                    updateQueue(w);
-                }
-            });
         }
+        // computing bc(x)
+        bcx = 0;
+        G.forNodePairs([&](node s, node y) {
```
fabratu, line 202: This has complexity `O(n^2)`, which we should avoid. Before the `bcx` score was computed during the loop. Seems that the logic is different now (`forNodePairs` iterates over the upper triangle of the graph instead of incrementing/decrementing twice as before).

networkit/cpp/centrality/DynBetweennessOneNode.cpp
```diff
@@ -27,51 +30,55 @@ void DynBetweennessOneNode::run() {
...
+            sssp->numberOfPaths(t).ToDouble(sigma[s][t]);
+        });
+
+        // initialize sigmax (paths that pass through x) for source s
         G.forNodes([&](node t) {
-            if (t != x && s != x && sigma[s][t] != 0) {
-                bcx += double(sigmax[s][t]) / sigma[s][t];
+            auto paths = sssp->getPaths(t);
+            for (auto path : paths) {
+                if (std::find(path.begin(), path.end(), x) != path.end()) {
```
fabratu, line 60: Have you measured the time complexity of this change? There can be a lot of paths for dense networks and `find` takes linear time.

networkit/cpp/centrality/DynBetweennessOneNode.cpp
PetholzA: The `run()` method now produces correct betweenness scores for weighted graphs, but indeed, takes longer than previously. We could practically have both variants implemented and choose the one suited for the input graph (not the prettiest solution, but for the unweighted case runtimes should remain similar). 

On the other hand, our current `SSSP` variants do not store the shortest paths explicitly (rather they only store the predecessor(s) for each node). This makes computing the `bc(x)` complicated without designing a new `BFS-like` algorithm that computes `bc(x)` on-the-fly. Here, any recommendations on your behalf are welcome.

networkit/cpp/centrality/DynBetweennessOneNode.cpp
PetholzA: The `run()` method now produces correct betweenness scores for weighted graphs, but indeed, takes longer than previously. We could practically have both variants implemented and choose the one suited for the input graph (not the prettiest solution, but for the unweighted case runtimes should remain similar). 

On the other hand, our current `SSSP` variants do not store the shortest paths explicitly (rather they only store the predecessor(s) for each node). This makes computing the `bc(x)` complicated without designing a new `BFS-like` algorithm that computes `bc(x)` on-the-fly. Here, any recommendations on your behalf are welcome.

PetholzA: The `run()` method has been adjusted to use a priority queue now instead of a regular queue. The runtime takes a slight hit, but the additional support for weighted graphs should outweigh that cost.

fabratu (approved):

PR #1246, opened 2024-08-20
DHB Integration

HazelCC:
This PR integrate the data structure Dynamic Hashed Blocks (DHB) into Networkit. This improves NetworKit for dynamic graph operations.

### **TLTR:**

For a detailed description of DHB, refer to the paper: [*A Fast Data Structure for Dynamic Graphs Based on Hash-Indexed Adjacency Blocks* (2022)](https://drops.dagstuhl.de/storage/00lipics/lipics-vol233-sea2022/LIPIcs.SEA.2022.11/LIPIcs.SEA.2022.11.pdf)

**Why?**

Experiments have shown that DHB outperforms competing dynamic graph structures in edge insertions, updates, deletions, and traversal operations. This integration allows users to leverage DHB to gain performance for dynamic graph operations compared to the (vanilla) Graph class.

**How?**

This integration extends the functionalities of the default `Graph` class by supporting (almost) all methods implemented in `Graph`. When NetworKit is build using DHB, the necessary resources will be pulled in as a sub-module, and the associated code branches will be compiled. If the respective NetworKit modules are not compiled using the DHB integration, `DHBGraph` will fall back using the implementation of the default `Graph` class. 

**Measurement:**

Experiments run on commodity hardware (MacOS, M1, 8 GB RAM) have been run to get some baseline measurements on the performance of default `Graph` class vs. the `DHBGraph` class. For edge insertions, sequential weight update, and sequential edge removal, `DHBGraph` outperforms `Graph` class in an unweighted graph setting. 

image(2)
image(1)
image

TODOs:

- [x]  ~~Review the passing by value semantics for all lambdas passed to a method such as `f(L handle)`. Instead, we might want to pass all lambdas by rvalue-reference as in `f(L&& handle)`.~~
- [x]  Review which type of `omp parallel for` scheduler (guided?) we use.
- [x]  ~~Review possibilities to use a concurrent/parallel version of functions that iterate over edges/vertices. For example, we might make `removeNode` to use the concurrent version of `forNeighbors`.~~
- [x]  Write more tests for `edgeIterator` and `nodeIterator`. (@fabratu)
- [x]  Write tests to check the consistency of ~~unweighted~~ undirected `DHBGraph` objects. An ~~unweighted~~ undirected graph should store one edge in both direction. (@HazelCC)
- [x]  Improve documentation. (@fabratu)
- [x] Fix GCC build issues. @c-bebop 
- [x] Build DHB on Clang and fix issues @c-bebop


.clang-format
```diff
@@ -88,7 +88,7 @@ PenaltyBreakFirstLessLess: 120
 PenaltyBreakString: 1000
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
-PointerAlignment: Right
+PointerAlignment: Left
```
fabratu, line 91: What is the rationale behind the clang-format changes?

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(NETWORKIT_CLANG_TIDY "Check code with clang-tidy" OFF)
 option(NETWORKIT_FLATINSTALL "Install into a flat directory structure (useful when building a Python package)" OFF)
 option(NETWORKIT_COVERAGE "Build with support for coverage" OFF)
 option(NETWORKIT_SANITY_CHECKS "Algorithms will perform further (expensive) checks" OFF)
+option(NETWORKIT_USE_DHB "Enable DHB)" ON)
```
fabratu, line 21: `DHB` is set to enabled by default here. Does this also mean, that it is used instead of `Graph` as data structure by default?

CMakeLists.txt
```diff
@@ -265,6 +285,7 @@ if(NETWORKIT_BUILD_CORE AND NETWORKIT_MONOLITH)
 	endif()
 
 	target_link_libraries(networkit PRIVATE OpenMP::OpenMP_CXX PUBLIC tlx)
+	target_link_libraries(networkit PRIVATE dhb)
```
fabratu, line 288: This again makes `dhb` mandatory. What is the point of using an option then?

fabratu (dismissed):
In order to keep the PR manageable, I suggest to fix all remaining issues concerning the integration in this PR and move optimization and additional thoughts to another PR.

Concerning building / testing: It seems that `DHB` is not correctly added as a submodule or there is some ill-defined logic in CMakeLists.txt. CI does fetch all submodules, however `DHB` is not found.

CMakeLists.txt
c-bebop: No it is not used by default. Instead there is a `DHBGraph` class now that can be instantiated. We did not change the default `Graph` class.

.clang-format
c-bebop: We can also revert those changes, if we want to make this another PR. The rationale is easily explained. To align the pointer symbol on right will clearly show that a type if of pointer or reference type as in:
```c++
int a = 10;
int* b = &a;
```  
Now `b` is of type _pointer to integer_ thus the symbol that makes it a pointer is part of the type not part of the variable name - as it would be with right alignment. The same applies to references. 

The `QualifierAlignment` in line `123` makes sure we are using _east const_ which simply follows the binding of a qualifier to it's left as in:
```c++
int const * const c = &a
```
Where the _const_ qualifier always binds to it's left. If one would want to use _west const_ one has to write:
```c++
const int * const c = &a
```
Basically mixing both alignments (west & east) since it is not possible to write:
```c++
const int const *  c = &a
```

CMakeLists.txt
c-bebop: So the question here is rather: do we want to use `if` branches for example to remove the mandatory linkage to `dhb` or is there another way?

CMakeLists.txt
fabratu: I see. When first reading into the PR I misunderstood how `DHB` is integrated.

CMakeLists.txt
fabratu: Ok, I revoke my wrapper assumption. Maybe I again misunderstood the concept of integrating the library. `DHBGraph.hpp` seems to include two cases: one implementation where `dhb` is available and one where it is not? 

If so, what is the functionality of `DHBGraph.hpp`/`DHBGraph.cpp` if the library is not present? Seems very similar to graph then, right?

c-bebop:
@HazelCC Could you clarify and elaborate what exactly you mean with the point: _Write tests to check the consistency of unweighted DHBGraph objects._?

c-bebop:
> > @HazelCC Could you clarify and elaborate what exactly you mean with the point: _Write tests to check the consistency of unweighted DHBGraph objects._?
> 
> An undirected graph object should store directed edges in both directions. Previously there were issues where logical bugs causing inconsistency for undirected graph, therefore it's good to have more tests addressing this.

For now, we have referenced you as taking over for this task. Is that fine?

c-bebop:
This PR (NetworKit DHB Integration) is now ready for review. All CI defined **builds are running successfully**. For further development please **address failing builds immediately**.

There are a couple of TODOs in code as well as in the PR (comments/description). @fabratu  and @HazelCC please make sure you will either address these as soon as possible, or leave comments in code for further development / improvements.

CMakeLists.txt
```diff
@@ -390,7 +410,8 @@ function(networkit_add_module modname)
 			COMPILE_FLAGS "${NETWORKIT_CXX_FLAGS}"
 			LINK_FLAGS "${NETWORKIT_LINK_FLAGS}")
 
-		target_link_libraries(${MODULE_TARGET} PRIVATE OpenMP::OpenMP_CXX PUBLIC tlx)
```
c-bebop, line 393: For the `networkit` target we link `PRIVATE` for the module target we link `PUBLIC`. In `DHBGraph.hpp` we include DHB. So we must link publicly. We're not getting any errors here in our CI.. Perhaps, we never use `DHBGraph` in combination with a target linking against module `graph`?

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,2882 @@
...
+// forward declaration to randomization/CurveballImpl.hpp
+namespace CurveballDetails {
+class CurveballMaterialization;
+}
+
+/**
+ * @ingroup graph
+ * A graph (with optional weights) and parallel iterator methods.
+ *
+ * TODO:
```
c-bebop, line 43: TODO!

CMakeLists.txt
fabratu: `DHBGraph.hpp` is part of the `graph` module. Outside of the `graph` module we don't use the `dhb`-api directly, therefore it should not be necessary to link against `dhb` in the first place. Have you tried making `dhb` private for the module_target?

include/networkit/graph/DHBGraph.hpp
fabratu: What is meant with "review" here? Is it about correctness? If not and this is about style/performance, I suggest remove any improvements (there are several TODOs in the code) to future PRs.

CMakeLists.txt
```diff
@@ -551,17 +575,23 @@ function(networkit_add_extra IS_TEST MOD NAME)
...
+			# add networkit modules to be linked against your target.
 			target_link_libraries(${TARGET_NAME}
 				PRIVATE
 					gtest networkit_gtest_main
 				PRIVATE
-					networkit_${MOD}
-					tlx
 					OpenMP::OpenMP_CXX
+					tlx
+					dhb
```
c-bebop, line 593: When linking publicly against `dhb` we may not need to link against `dhb` again here for the tests.

include/networkit/graph/DHBGraph.hpp
c-bebop: Fine for me!

networkit/cpp/graph/test/DHBGraphGTest.cpp
```diff
@@ -0,0 +1,2543 @@
...
+}
+
+TEST_P(DHBGraphGTest, testWeight) {
+    this->Ghouse.forNodes([&](node u) {
+        this->Ghouse.forNodes(
+            [&](node v) { ASSERT_EQ(this->Ahouse[u][v], this->Ghouse.weight(u, v)); });
+    });
+}
+
+TEST_P(DHBGraphGTest, testSetWeight) {
```
c-bebop, line 1070: This test could also be separated into more tests .. as the comments suggest.

c-bebop (dismissed):
Looks very good, now just some small fixes.

networkit/cpp/graph/DHBGraph.cpp
```diff
@@ -0,0 +1,509 @@
...
+                ++omega;
+            }
+        });
+    }
+    edgesIndexed = true;
+}
+
+edgeid DHBGraph::edgeId(node u, node v) const {
+    if (!edgesIndexed) {
+        throw std::runtime_error(
```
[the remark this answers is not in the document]
fabratu: I would rather stay with the current logic in order to mimic the behavior from `Graph`.

networkit/cpp/graph/test/DHBGraphGTest.cpp
```diff
@@ -0,0 +1,2543 @@
...
+    // if orig is not directed, but copy to a directed graph, edge number should double.
+    ASSERT_EQ(this->Ghouse.numberOfEdges(), D.numberOfEdges());
+
+    ASSERT_TRUE(DW.isWeighted());
+    ASSERT_TRUE(DW.isDirected());
+    ASSERT_EQ(this->Ghouse.numberOfNodes(), DW.numberOfNodes());
+    ASSERT_EQ(this->Ghouse.numberOfEdges(), DW.numberOfEdges());
+    this->Ghouse.forNodes([&](node v) {
+        count d = this->Ghouse.degree(v);
+        // count dUndirected = isDirected() ? d + this->Ghouse.degreeIn(v) : d;
```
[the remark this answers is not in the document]
fabratu: Remnants of previous versions, deleted them.

networkit/cpp/graph/test/DHBGraphGTest.cpp
```diff
@@ -0,0 +1,2543 @@
...
+    };
+
+    for (node u = 0; u < this->Ghouse.upperNodeIdBound(); u++) {
+        for (node v = 0; v < this->Ghouse.upperNodeIdBound(); v++) {
+            auto edge = std::make_pair(u, v);
+            auto edgeReverse = std::make_pair(v, u);
+            bool hasEdge = containsEdge(edge);
+            bool hasEdgeReverse = containsEdge(edgeReverse);
+            if (this->Ghouse.isDirected()) {
+                ASSERT_EQ(hasEdge, this->Ghouse.hasEdge(u, v));
```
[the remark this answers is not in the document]
fabratu: I agree, that using a tracking variable might reduce the number of asserts and make the test run faster. However, in the current state I don't see a running time problem. In addition I disagree with the argument concerning the improvement of error output. Currently the test is written in a way, that we know exactly for which edge the `assert` fails and therefore it makes it easier to search for bugs/problems. With a tracking variable, you would lose this information or would have to go through an array of booleans.

networkit/cpp/graph/DHBGraph.cpp
```diff
@@ -0,0 +1,509 @@
...
+    }
+    return false;
+}
+
+void DHBGraph::removeAllEdges() {
+    m_dhb_graph = dhb::Matrix<EdgeData>(m_dhb_graph.vertices_count());
+    m = 0;
+    storedNumberOfSelfLoops = 0;
+    omega = 0;
+}
```
bernlu, line 320: this function should invalidate all edgeAttributes as well.

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+     * Get an upper bound for the node ids in the graph.
+     * @return An upper bound for the node ids.
+     */
+    index upperNodeIdBound() const noexcept;
+
+    /**
+     * Return edge weight of edge {@a u,@a v}. Returns 0 if edge does not
+     * exist.
+     * BEWARE: Running time is \Theta(1) for DHB implementation but \Theta(deg(u)) for fallback
+     * NetworKit graph DS.
```
bernlu, line 1131: i think this fallback nk graph is outdated information?

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+     * neighbor exists.
+     */
+    node getIthNeighbor(node u, index i) const;
+
+    /**
+     * Return the weight to the i-th (outgoing) neighbor of @a u.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return @a edge weight to the i-th (outgoing) neighbor of @a u, or @c +inf if no such
```
bernlu, line 1247: The function actually returns the default edge weight and not inf when the neighbor does not exist.

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+     */
+    edgeweight getIthNeighborWeight(node u, index i) const;
+
+    /**
+     * Get i-th (outgoing) neighbor of @a u and the corresponding edge weight.
+     *
+     * @param u Node.
+     * @param i index; should be in [0, degreeOut(u))
+     * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
+     * edge weight, or @c defaultEdgeWeight if unweighted.
```
bernlu, line 1258: we should add that `none` is returned when the neighbor does not exist.

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+     *
+     * This function is not supported for DHB. Please use getIthNeighborWithWeight(node u, index i).
+     */
+    std::pair<node, edgeweight> getIthNeighborWithWeight(Unsafe, node u, index i) const;
```
bernlu, line 1272: why is this function declared twice, but once we claim that it is not supported while the other one is implemented?

bernlu (dismissed):
We should probably discuss some of these comments, some are quite detailed and maybe out of scope for this PR.

An additional general comment: I think it would be nice to add time complexity to the documentation for (almost) all functions.

c-bebop:
There are still some review comments but.. are we merging this one?

fabratu:
@c-bebop This is just a gentle reminder :)

include/networkit/graph/Edge.hpp
```diff
@@ -0,0 +1,62 @@
...
+    Edge(node _u, node _v, bool sorted = false) {
+        u = sorted ? std::min(_u, _v) : _u;
+        v = sorted ? std::max(_u, _v) : _v;
+    }
+};
+
+/**
+ * A weighted edge used for the graph constructor with
+ * initializer list syntax.
+ */
```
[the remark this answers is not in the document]
c-bebop: Agreed. I removed the doc string.

include/networkit/graph/Edge.hpp
```diff
@@ -0,0 +1,62 @@
...
+struct WeightedEdge : Edge {
+    edgeweight weight;
+
+    // Needed by cython
```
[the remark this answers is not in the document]
c-bebop: What about `std::vector`?

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+namespace NetworKit {
+
+// forward declaration to randomization/CurveballImpl.hpp
+namespace CurveballDetails {
+class CurveballMaterialization;
+}
```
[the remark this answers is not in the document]
c-bebop: Seems so. Removed.

networkit/cpp/graph/DHBGraph.cpp
```diff
@@ -0,0 +1,509 @@
...
+    bool is_inserted_out = false;
+    bool is_inserted = false;
+
+    // Insert edges depending on whether the graph is directed or undirected
+    if (!directed) {
+        is_inserted_out = m_dhb_graph.insert(u, v, edge_data, do_update);
+        is_inserted_in = m_dhb_graph.insert(v, u, edge_data, do_update);
+    } else {
+        is_inserted = m_dhb_graph.insert(u, v, edge_data, do_update);
+    }
```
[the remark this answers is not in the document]
c-bebop: Agree with that. Done.

networkit/cpp/graph/DHBGraph.cpp
```diff
@@ -0,0 +1,509 @@
...
+namespace NetworKit {
+
+/** CONSTRUCTORS **/
+
+DHBGraph::DHBGraph(count n, bool weighted, bool directed, bool edgesIndexed)
+    : weighted(weighted), directed(directed), m(0), storedNumberOfSelfLoops(0), omega(0),
+      edgesIndexed(edgesIndexed), // edges are not indexed by default
+      nodeAttributeMap(this), edgeAttributeMap(this) {
+
+    m_dhb_graph = dhb::Matrix<EdgeData>(n);
```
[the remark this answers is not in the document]
c-bebop: :pray: Done.

CMakeLists.txt
```diff
@@ -264,7 +280,11 @@ if(NETWORKIT_BUILD_CORE AND NETWORKIT_MONOLITH)
 		target_include_directories(networkit_state BEFORE PUBLIC "${PROJECT_SOURCE_DIR}/include")
 	endif()
 
-	target_link_libraries(networkit PRIVATE OpenMP::OpenMP_CXX PUBLIC tlx)
+	target_link_libraries(networkit PRIVATE 
+									dhb 
+									OpenMP::OpenMP_CXX 
+									PUBLIC 
+									tlx)
```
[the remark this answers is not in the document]
c-bebop: Done.

include/networkit/graph/DHBGraph.hpp
```diff
@@ -0,0 +1,1991 @@
...
+
+/**
+ * @ingroup graph
+ * A graph (with optional weights) and parallel iterator methods.
+ *
+ * TODO:
+ * - [ ] Review the passing by value semantics for all lambdas passed
+ *       to a method such as f(L handle). Instead, we might want to pass
+ *       all lambdas by rvalue-reference as in f(L&& handle).
```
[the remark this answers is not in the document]
c-bebop: Done. Revised doc of DHB graph class as well.

include/networkit/graph/DHBGraph.hpp
c-bebop: It's actually just this one which is not implemented. Marked this one as `deleted`.

include/networkit/graph/DHBGraph.hpp
c-bebop: True. :pray: Done.

networkit/cpp/graph/DHBGraph.cpp
c-bebop: Done.

include/networkit/graph/DHBGraph.hpp
c-bebop: I return `0.0` as in `nullWeight` now for all functions like `getIthNeighbor...()` which aligns to what you would expect from an edge that does not exist? :cowboy_hat_face:

include/networkit/graph/DHBGraph.hpp
c-bebop: True. Fixed.

c-bebop (dismissed):
The current state looks good to me. 🍏

fabratu (approved):

fabratu:
Maybe there are still some sub-optimal parts in the code + the class is not templated. Since this PR has been open for a very long time, we will merge it now and deal with open ends later on.

c-bebop:
🥳 🎉

c-bebop:
Thank you @HazelCC @fabratu and @bernlu for your contributions to integrate DHB into Networkit.

PR #1249, opened 2024-08-26
replace const std::string& by std::string_view

PetholzA:
This PR replaces each occasion of `const std::string&` by `std::string_view`. [This forum post](https://stackoverflow.com/questions/40127965/how-exactly-is-stdstring-view-faster-than-const-stdstring) leads to the assumption that using `std::string_view` provides runtime advantages, while having almost no downsides.


fabratu (approved):

moritzbeier:
A downside is that there is no automatic conversion from std::filesystem::path anymore.
I had to change all instances of `reader.read(path)` to `reader.read(std::string(path))` because of this change.

PR #1252, opened 2024-09-04
fix DynamicMatrix * DynamicMatrix

bernlu:
For two matrices A (n\*m) and B(m\*p), the size of the resulting matrix C=AB is n\*p. The sparse accumulator should have size p - which is `B.numberOfColumns()`


PetholzA (approved):
LGTM

PR #1264, opened 2024-10-18
Add RB Matrix and Graph reader

bernlu:


networkit/cpp/io/RBGraphReader.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+Graph RBGraphReader::read(std::string_view path) {
+    std::ifstream in(path.data());
+    if (!in.is_open()) {
+        throw std::runtime_error("could not open: " + std::string(path));
+    }
+
+    RBMatrixReader reader;
+
+    reader.readToVectors(in);
```
fabratu, line 17: Hm, we effectively allocate memory for edges twice. How about implementing the logic from `readToVectors` in a parser class and either then use it to fill the vectors or the graph directly?

networkit/cpp/io/RBMatrixReader.cpp
```diff
@@ -0,0 +1,148 @@
...
+    // start reading data
+    while (getline(in, line)) {
+        row++;
+        std::istringstream in_line(line);
+
+        if (state == POINTERS) {
+            int p;
+            while (in_line >> p) {
+                pointers.push_back(
+                    p - 1); // data in file is 1-indexed. shift to 0-index for networkit
```
fabratu, line 100: Is it safe to assume that the file is 1-indexed?

fabratu (changes requested):

networkit/cpp/io/RBMatrixReader.cpp
bernlu: yes

networkit/cpp/io/RBGraphReader.cpp
bernlu: There is no way to avoid storing the IC and JR vectors of the csc matrix, because we have to read the JR and Val vectors at the same time to know which values map to which rows/columns. Hence it is not worth duplicating/changing the code.

fabratu (approved):

PR #1265, opened 2024-11-01
Raise cmake minimum version to 3.14.

fabratu:


bernlu:
Are there specific reasons for version 3.14 here?

fabratu:
> Are there specific reasons for version 3.14 here?

Yes, this version is the oldest, which has been released <5 years ago.

PR #1276, opened 2025-01-08
Add Left-Right Planarity Test Algorithm to NetworKit

Schwarf:
### Overview
This PR introduces an implementation of the Left-Right Planarity Test algorithm, which is used to determine whether a given graph is planar. The implementation was guided by the paper ["The Left-Right Planarity Test"](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=7963e9feffe1c9362eb1a69010a5139d1da3661e) as well as the corresponding implementation in NetworkX.
### Unit Tests 
To ensure correctness and robustness, I have included 18 unit tests for the algorithm. However, given the complexity of the problem, I would be happy to add more test cases with known outcomes if provided or suggested. 
### Graph API Extension
I added a new method, `sortNeighbors`, to the `Graph` class. This method sorts the neighbors of a graph node based on a binary predicate, and I have included a corresponding unit test for this functionality. However, it’s possible that this could be achieved with existing APIs, and I would appreciate guidance if such an alternative already exists.
### Notes
This PR represents my first feature contribution to NetworKit. There is likely room for optimization, and I welcome any feedback or suggestions for improvement.

I look forward to your review and guidance to help refine this contribution.


fabratu (changes requested):
Thanks for your contribution!

In general, the PR is in a good shape. I added some comments, where things need to be addressed. In addition to the comments, please change the boilerplate code in the new class/header files to be more aligned with the current style (this is not automatically checked):

```
/* <FILENAME>
 *
 *	Created on: <DD.MM.YYYY>
 *    Authors: <NAME> <FAMILY NAME> (<MAIL>)
 *             
 */
```

You can leave out any personal data (name, mail, ...) if you want or also use a github username.

If there are any questions, let me know.

include/networkit/planarity/LeftRightPlanarityTest.hpp
```diff
@@ -0,0 +1,75 @@
...
+//
+
+#ifndef NETWORKIT_PLANARITY_LEFT_RIGHT_PLANARITY_TEST_HPP_
+#define NETWORKIT_PLANARITY_LEFT_RIGHT_PLANARITY_TEST_HPP_
+#include <networkit/base/Algorithm.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class LeftRightPlanarityTest final : public Algorithm {
```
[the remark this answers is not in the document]
Schwarf: I see your point and propose `LeftRightPlanarityCheck`?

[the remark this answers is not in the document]
fabratu: Yes, that name is appropriate.

networkit/cpp/planarity/LeftRightPlanarityTest.cpp
```diff
@@ -0,0 +1,261 @@
...
+    graph_->forNodes([&](const node currentNode) {
+        if (heights[currentNode] == noneHeight) {
+            heights[currentNode] = 0;
+            roots.push_back(currentNode);
+            this->dfsOrientation(currentNode);
+        }
+    });
+
+    sortAdjacencyListByNestingDepth();
+    isPlanar_ = true;
```
angriman, line 25: Here you can use `std::ranges::all_of` https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2374,4 +2374,33 @@ TEST_P(GraphGTest, testEdgeIdsAfterRemoveWithoutSortingOrIDs) {
...
+        G.sortNeighbors(currentNode, [&](const node neighbor1, const node neighbor2) {
+            return neighbor1 < neighbor2;
+        });
+    });
+
+    G.forNodes([&](const node currentNode) {
+        const auto &sortedNeighbors = G.neighborRange(currentNode);
+        std::vector<node> sortedNeighborVector(sortedNeighbors.begin(), sortedNeighbors.end());
+
+        EXPECT_TRUE(std::is_sorted(sortedNeighborVector.begin(), sortedNeighborVector.end()));
```
angriman, line 2398: Here you can use `std::ranges::is_sorted`, same below https://en.cppreference.com/w/cpp/algorithm/ranges/is_sorted

networkit/cpp/planarity/test/LeftRightPlanarityTestGtest.cpp
```diff
@@ -0,0 +1,637 @@
+//
+// Created by andreas on 05.01.25.
+//
+#include <gtest/gtest.h>
+#include <networkit/planarity/LeftRightPlanarityTest.hpp>
+
+namespace NetworKit {
+class LeftRightPlanarityTestGTest : public testing::Test {
```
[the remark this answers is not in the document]
Schwarf: This already tested via the test: testCompleteGraph

include/networkit/graph/Graph.hpp
```diff
@@ -941,6 +941,22 @@ class Graph final {
...
 
+    /**
+     * Sorts the neighbors of a given node according to a user-defined comparison function.
+     *
+     * @param u The node whose neighbors will be sorted.
+     * @param lambda A binary predicate used to compare two neighbors. The predicate should
+     *               take two nodes as arguments and return true if the first node should
+     *               precede the second in the sorted order.
+     */
+    void sortNeighbors(const node u, const std::function<bool(node, node)> &lambda) {
```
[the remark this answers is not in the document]
Schwarf: Thank you for the feedback! 
To confirm my understanding of the requirements for `sortNeighbors``:
1. Weighted Graphs: The method should sort the adjacency list (outEdges) and also permute the associated   outEdgeWeights vector, ensuring weights remain aligned with their respective edges.
2. Directed Graphs: In addition to sorting outEdges, the method should handle inEdges (and their associated weights and edge IDs) for directed graphs.
3.Indexed Graphs:The outEdgeIds (and inEdgeIds for directed graphs) should also be permuted during sorting to maintain alignment with the adjacency lists.

Is my understanding  correct?

include/networkit/graph/Graph.hpp
```diff
@@ -45,7 +45,7 @@ namespace NetworKit {
 struct Edge {
     node u, v;
 
-    Edge() : u(none), v(none) {}
+    constexpr Edge() : u(none), v(none) {}
```
angriman, line 48: As far as I understand, you don't need this to be `constexpr` since there is no `constexpr` function using it.

include/networkit/graph/Graph.hpp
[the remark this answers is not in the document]
angriman: IIUC this function can replace the `sortAdjacencyArrays` lambda in `sortEdges` for sorting the neighbors for a specific node. If that's the case, then please change `sortEdges` to use this function as a black-box.
Additional changes that I'd suggest:
- Move the implementation outside the class (e.g., above the implementation of `sortEdges`
- Avoid `std::function` and use a template parameter.
- Do not use `const` for primitive type function parameters.

include/networkit/planarity/LeftRightPlanarityCheck.hpp
```diff
@@ -0,0 +1,112 @@
...
+     * https://en.wikipedia.org/wiki/Planar_graph
+     *
+     * The algorithm achieves (almost) linear runtime complexity. The only non-linear component
+     * arises from sorting the nodes of the depth-first search tree.
+     *
+     * @param graph The input graph to test for planarity. The graph should be undirected.
+     */
+    LeftRightPlanarityCheck(const Graph &graph) : graph_(&graph) {
+        if (graph.isDirected()) {
+            throw std::logic_error("The graph is not an undirected graph.");
```
angriman, line 32: In case of invalid inputs, we usually throw a `std::runtime_error`.

include/networkit/planarity/LeftRightPlanarityCheck.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+        void swap() { std::swap(left, right); }
+
+        friend bool operator==(const ConflictPair &lhs, const ConflictPair &rhs) {
+            return lhs.left == rhs.left && lhs.right == rhs.right;
+        }
+    };
+    const ConflictPair NoneConflictPair{Interval(), Interval()};
+
+    const Graph *graph_;
```
angriman, line 90: Please be consistent with the rest of the codebase and avoid trailing underscores for private member variables.

include/networkit/planarity/LeftRightPlanarityCheck.hpp
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Returns whether the input graph is planar.
+     * The result is only valid after the `run()` method has been called.
+     *
+     * @return True if the graph is planar, false otherwise.
+     * @throws std::logic_error if called before `run()` has been executed.
+     */
+    bool isPlanar() const {
+        if (!hasRun)
+            throw std::logic_error("isPlanar() called before the run() method was called.");
```
angriman, line 54: Use `Algorithm::assureFinished()` instead of checking for `hasRun`.

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2374,4 +2374,133 @@ TEST_P(GraphGTest, testEdgeIdsAfterRemoveWithoutSortingOrIDs) {
     });
 }
 
+TEST_P(GraphGTest, testSortNeighbors) {
```
angriman, line 2377: Please break down this test in multiple small tests that evaluate whether the function produced the expected result on different graph types (e.g., directed/undirected, weighted/unweighted, with/without edge ids, etc).

Also, please avoid generating random graphs. Instead, define a small test graph yourself and test that the neighbors are sorted as expected. I know that generating random instances is widely done in NetworKit, but this can lead to flaky tests, and I think that we should avoid doing this in the future.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+void LeftRightPlanarityCheck::run() {
+    // Euler-criterion
+    if (graph_->numberOfNodes() > 2 && graph_->numberOfEdges() > 3 * graph_->numberOfNodes() - 6) {
+        hasRun = true;
+        isPlanar_ = false;
+        return;
+    }
+
+    heights.assign(graph_->upperNodeIdBound(), noneHeight);
+    graph_->forNodes([&](const node currentNode) {
```
angriman, line 21: Avoid using `const` for primitive types in functions input parameters, also below.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    sortAdjacencyListByNestingDepth();
+    isPlanar_ =
+        std::ranges::all_of(roots, [this](const auto rootNode) { return dfsTesting(rootNode); });
+    hasRun = true;
+}
+
+void LeftRightPlanarityCheck::sortAdjacencyListByNestingDepth() {
+
```
angriman, line 36: Nit: remove this empty line.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+                return it1->second < it2->second;
+            }
+            return false;
+        });
+    });
+}
+
+bool LeftRightPlanarityCheck::conflicting(const Interval &interval, const Edge &edge) {
+    return !interval.isEmpty() && lowestPoint.contains(interval.high) && lowestPoint.contains(edge)
+           && lowestPoint[interval.high] > lowestPoint[edge];
```
angriman, line 51: Use `lowestPoint.find` instead of using `.contains` and `[]`, so that you need to do only one lookup instead of two.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+bool LeftRightPlanarityCheck::applyConstraints(const Edge edge, const Edge parentEdge) {
+    auto tmpConflictPair = ConflictPair{};
```
angriman, line 55: What about `ConflictPair tmpConflictPair{};` ?

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        auto currentConflictPair = stack.top();
+        stack.pop();
+        if (!currentConflictPair.left.isEmpty()) {
+            currentConflictPair.swap();
+        }
+        if (!currentConflictPair.left.isEmpty()) {
+            return false;
+        }
+        if (lowestPoint.contains(currentConflictPair.right.low) && lowestPoint.contains(parentEdge)
+            && lowestPoint[currentConflictPair.right.low] > lowestPoint[parentEdge]) {
```
angriman, line 66: Same here, use `find` instead of `contains` and `[]`.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+count LeftRightPlanarityCheck::getLowestLowPoint(const ConflictPair &conflictPair) {
+    if (conflictPair.left.isEmpty())
+        return lowestPoint[conflictPair.right.low];
+    if (conflictPair.right.isEmpty())
+        return lowestPoint[conflictPair.left.low];
+    return std::min(lowestPoint[conflictPair.right.low], lowestPoint[conflictPair.left.low]);
+}
+
+void LeftRightPlanarityCheck::removeBackEdges(Edge edge) {
+    auto parentNode = edge.u;
```
angriman, line 115: Nit: avoid `auto` here.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+count LeftRightPlanarityCheck::getLowestLowPoint(const ConflictPair &conflictPair) {
+    if (conflictPair.left.isEmpty())
+        return lowestPoint[conflictPair.right.low];
+    if (conflictPair.right.isEmpty())
+        return lowestPoint[conflictPair.left.low];
+    return std::min(lowestPoint[conflictPair.right.low], lowestPoint[conflictPair.left.low]);
+}
+
+void LeftRightPlanarityCheck::removeBackEdges(Edge edge) {
```
angriman, line 114: IIUC `edge` never changes, maybe use `const Edge& edge`?

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        if (conflictPair.right.high == noneEdge && conflictPair.right.low != noneEdge) {
+            ref[conflictPair.right.low] = conflictPair.left.low;
+            conflictPair.right.low = noneEdge;
+        }
+        stack.push(conflictPair);
+    }
+
+    if (!stack.empty() && lowestPoint[edge] < heights[parentNode]) {
+        auto highestReturnEdgeLeft = stack.top().left.high;
+        auto highestReturnEdgeRight = stack.top().right.high;
```
angriman, line 145: Nit: avoid `auto` here.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+            ref[edge] = highestReturnEdgeRight;
+        }
+    }
+}
+
+bool LeftRightPlanarityCheck::dfsTesting(node startNode) {
+    std::stack<node> dfsStack;
+    dfsStack.emplace(startNode);
+    auto neighborIterators =
+        std::unordered_map<node, decltype(dfsGraph.neighborRange(startNode).begin())>{};
```
angriman, line 160: `std::unordered_map<node, decltype(dfsGraph.neighborRange(startNode).begin())> neighborIterators;` is simpler.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    auto preprocessedEdges = std::unordered_set<Edge>{};
```
angriman, line 162: Same here: `std::unordered_set<Edge>{} preprocessedEdges;`

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+    while (!dfsStack.empty()) {
```
angriman, line 163: Nit: `dfsStack` is never empty at the beginning, so you can replace this with a `do {} while (!dfsStack.empty());` loop.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        const auto currentNode = dfsStack.top();
```
angriman, line 164: Please avoid `auto` here and below in this function, it makes is really hard to follow the code. One exception where you can keep it is for iterator types.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        dfsStack.pop();
+        const auto parentEdge = parentEdges[currentNode];
+        bool callRemoveBackEdges{true};
+        if (auto it = neighborIterators.find(currentNode); it == neighborIterators.end()) {
+            neighborIterators[currentNode] = dfsGraph.neighborRange(currentNode).begin();
+        }
+        auto &neighborIterator = neighborIterators[currentNode];
+        while (neighborIterator != dfsGraph.neighborRange(currentNode).end()) {
+            const auto neighbor = *neighborIterator;
+            auto currentEdge = Edge(currentNode, neighbor);
```
angriman, line 174: `Edge currentEdge(currentNode, neighbor);`

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+    while (!dfsStack.empty()) {
+        const auto currentNode = dfsStack.top();
+        dfsStack.pop();
+        const auto parentEdge = parentEdges[currentNode];
+        bool callRemoveBackEdges{true};
+        if (auto it = neighborIterators.find(currentNode); it == neighborIterators.end()) {
+            neighborIterators[currentNode] = dfsGraph.neighborRange(currentNode).begin();
+        }
+        auto &neighborIterator = neighborIterators[currentNode];
+        while (neighborIterator != dfsGraph.neighborRange(currentNode).end()) {
```
angriman, line 172: Consider moving this loop to a separate function. There are multiple nested blocks and it is not easy to follow what the code is doing.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+                    preprocessedEdges.insert(currentEdge);
+                    callRemoveBackEdges = false;
+                    break;
+                }
+                lowestPointEdge[currentEdge] = currentEdge;
+                stack.emplace(Interval{}, Interval(currentEdge, currentEdge));
+            }
+
+            if (lowestPoint.contains(currentEdge)
+                && lowestPoint[currentEdge] < heights[currentNode]) {
```
angriman, line 189: As above, avoid `contains` + `[]` and use `.find`.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        if (callRemoveBackEdges) {
+            if (parentEdge != noneEdge)
+                removeBackEdges(parentEdge);
+        }
+    }
+    return true;
+}
+
+void LeftRightPlanarityCheck::dfsOrientation(const node startNode) {
+
```
angriman, line 209: Nit: remove empty line.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+        if (callRemoveBackEdges) {
+            if (parentEdge != noneEdge)
+                removeBackEdges(parentEdge);
+        }
+    }
+    return true;
+}
+
+void LeftRightPlanarityCheck::dfsOrientation(const node startNode) {
```
angriman, line 208: `node` is a primitive type, so drop `const`.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    std::stack<node> dfsStack;
+    dfsStack.emplace(startNode);
+    auto preprocessedEdges = std::unordered_set<Edge>{};
```
angriman, line 212: `std::unordered_set<Edge>{} preprocessedEdges;`

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,262 @@
...
+    while (!dfsStack.empty()) {
```
angriman, line 213: As above, replace with `do {} while` and avoid using `auto` for non-iterator types.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+
+namespace NetworKit {
+class LeftRightPlanarityCheckGTest : public testing::Test {
+public:
+    LeftRightPlanarityCheckGTest() = default;
+    static constexpr int maxNumberOfNodes{100};
+    Graph pathGraph(const count numNodes) {
+        Graph graph;
+        graph.addNodes(numNodes);
+        for (count i{}; i < numNodes - 1; ++i) {
```
angriman, line 18: `count i = 0;`, same below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+ */
+#include <gtest/gtest.h>
+#include <networkit/planarity/LeftRightPlanarityCheck.hpp>
+
+namespace NetworKit {
+class LeftRightPlanarityCheckGTest : public testing::Test {
+public:
+    LeftRightPlanarityCheckGTest() = default;
+    static constexpr int maxNumberOfNodes{100};
+    Graph pathGraph(const count numNodes) {
```
angriman, line 15: Drop `const`, same for the functions below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+        for (count i = 0; i < n; ++i) {
+            graph.addEdge(i, n + i);
+        }
+
+        return graph;
+    }
+};
+
+TEST_F(LeftRightPlanarityCheckGTest, testDirectedGraphThrows) {
+    auto graph = Graph(0, false, true, false);
```
angriman, line 130: Just use `Graph graph(...);` same for the other tests below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+        FAIL() << "Expected std::logic_error";
+    } catch (const std::logic_error &e) {
+        EXPECT_STREQ(e.what(), "isPlanar() called before the run() method was called.");
+    } catch (...) {
+        FAIL() << "Expected std::logic_error but got a different exception.";
+    }
+}
+
+TEST_F(LeftRightPlanarityCheckGTest, testEmptyGraph) {
+
```
angriman, line 155: Drop this empty line, same for the tests below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+TEST_F(LeftRightPlanarityCheckGTest, testCompleteGraph) {
+    for (count numberOfNodes{2}; numberOfNodes <= maxNumberOfNodes; ++numberOfNodes) {
+        auto graph = completeGraph(numberOfNodes);
+        LeftRightPlanarityCheck test(graph);
+        test.run();
+        if (numberOfNodes < 5)
+            EXPECT_TRUE(test.isPlanar());
+        else
+            EXPECT_FALSE(test.isPlanar());
```
angriman, line 228: A unit test should evaluate a specific behavior of your code. Here you are testing two different things: that the graph is planar if the number of nodes is less than 5, and not planar otherwise.

I'd suggest splitting this test in two tests.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+    graph.addEdge(35, 37);
+    graph.addEdge(35, 46);
+    graph.addEdge(36, 49);
+
+    LeftRightPlanarityCheck test(graph);
+    test.run();
+    EXPECT_TRUE(test.isPlanar());
+}
+
+TEST_F(LeftRightPlanarityCheckGTest, testRandomGeneratedPlanarGraph100Nodes) {
```
angriman, line 402: This graph is too large for a unit test, please use a smaller one.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+    graph.addEdge(4, 8);
+    graph.addEdge(5, 6);
+    graph.addEdge(5, 7);
+    graph.addEdge(5, 8);
+    LeftRightPlanarityCheck test(graph);
+    test.run();
+    EXPECT_FALSE(test.isPlanar());
+}
+
+TEST_F(LeftRightPlanarityCheckGTest, testRandomGeneratedPlanarGraph50Nodes) {
```
angriman, line 342: This graph is too large, please use a smaller one or drop the test.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+    graph.addEdge(80, 89);
+    graph.addEdge(84, 86);
+    graph.addEdge(89, 95);
+    LeftRightPlanarityCheck test(graph);
+    test.run();
+    EXPECT_TRUE(test.isPlanar());
+}
+
+TEST_F(LeftRightPlanarityCheckGTest, testRandomGeneratedPlanarConnectedGraph51Nodes) {
+    Graph graph(51);
```
angriman, line 514: Same here, the graph is too large.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -0,0 +1,666 @@
...
+    graph.addEdge(10, 13);
+    graph.addEdge(11, 18);
+    graph.addEdge(16, 17);
+    graph.addEdge(16, 18);
+    graph.addEdge(17, 18);
+    LeftRightPlanarityCheck test(graph);
+    test.run();
+    EXPECT_TRUE(test.isPlanar());
+}
+TEST_F(LeftRightPlanarityCheckGTest, testPetersenGraphs) {
```
angriman, line 631: Aren't Petersen graphs already tested above? I'd suggest running this test on a single representative example, not on hundreds of graphs.

angriman (changes requested):
Thank you for your contribution! I left some comments, most of them are just minor style changes to be consistent with the rest of the codebase.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: > `count i = 0;`, same below.

Thank you for pointing this out! I’m curious about the reason for this change request. My understanding is that `{}` is generally considered the better choice in modern C++ because it ensures value initialization and prevents unintended narrowing conversions.

angriman: That's true. In this case we have a simple for-loop, the `= 0` initialization does not have unintended consequences and I find it more expressive of what you are doing.

networkit/cpp/graph/test/GraphGTest.cpp
Schwarf: Thank you for your explanation. I followed the "test style" of the existing tests in this file, but I agree that using random graphs may not be ideal. I also didn’t feel entirely comfortable with relying on random graph generation, as it can make debugging more challenging.

Schwarf:
Hi @angriman,

Thank you for your feedback! I’d like to provide some context and arguments for the choice of graph sizes in the tests and the use of many test cases:

1. Multiple and Larger Graphs 
During the initial implementation, I relied on smaller, single-graph tests, which passed successfully. However, when I extended the tests to include many graphs of the same class (e.g., multiple grid graphs instead of a single 3x3 grid graph), some cases emerged, revealing bugs in the algorithm. Additionally, larger and random graphs helped me uncover issues that smaller graphs missed. These are the main reasons I included these tests.

2. Algorithm extension
I plan to extend my contribution by modifying the algorithm to provide planar embeddings for planar graphs. This may require minor refactoring of the current implementation. Similarly, potential performance improvements might also lead to refactoring. Having a variety of test cases (including larger and multiple graphs) will make it easier for me and others to ensure correctness during these changes.

3. Negligible Runtime Cost
On my machine, the runtime difference between testing a single graph versus hundreds of graphs is of the order of a few milliseconds, which is negligible compared to tests that take up to 10 seconds. The same applies to the included large random graphs. While I understand the importance of keeping tests efficient ("Kleinvieh macht auch Mist"), I currently believe the benefits of including larger and multiple graphs, outweigh the costs.

Given these reasons, I’d appreciate it if we could retain the current test cases in the file `LeftRightPlanarityCheckGTest.cpp`. Of course, I’m happy to hear your thoughts.

What do you think?

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: You are right there has been some redundancy. I would like to keep this test and remove the tests:
`testGeneralizedPetersenGraph5_1` and `testGeneralizedPetersenGraph5_2`. See my arguments below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: I will split the test.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: See my comment below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: See my comment below.

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: See my comment below.

fabratu:
> Hi @angriman,
> 
> Thank you for your feedback! I’d like to provide some context and arguments for the choice of graph sizes in the tests and the use of many test cases:
> 
>     1. Multiple and Larger Graphs
>        During the initial implementation, I relied on smaller, single-graph tests, which passed successfully. However, when I extended the tests to include many graphs of the same class (e.g., multiple grid graphs instead of a single 3x3 grid graph), some cases emerged, revealing bugs in the algorithm. Additionally, larger and random graphs helped me uncover issues that smaller graphs missed. These are the main reasons I included these tests.
> 
>     2. Algorithm extension
>        I plan to extend my contribution by modifying the algorithm to provide planar embeddings for planar graphs. This may require minor refactoring of the current implementation. Similarly, potential performance improvements might also lead to refactoring. Having a variety of test cases (including larger and multiple graphs) will make it easier for me and others to ensure correctness during these changes.
> 
>     3. Negligible Runtime Cost
>        On my machine, the runtime difference between testing a single graph versus hundreds of graphs is of the order of a few milliseconds, which is negligible compared to tests that take up to 10 seconds. The same applies to the included large random graphs. While I understand the importance of keeping tests efficient ("Kleinvieh macht auch Mist"), I currently believe the benefits of including larger and multiple graphs, outweigh the costs.
> 
> 
> Given these reasons, I’d appreciate it if we could retain the current test cases in the file `LeftRightPlanarityCheckGTest.cpp`. Of course, I’m happy to hear your thoughts.
> 
> What do you think?

There are positives and negatives about having random generated graphs for testing. In the past it helped us discover some quite subtle bugs. However, sometimes these can come from the generators itself, leading to the case that a test implicitly checks the functionalities of different code units. Also you can not predict when these errors turn up. Overall there are better systems to generate these tests, but these would need a complete pipeline overhaul. 

We had more bugs in algorithms / core functionalities which only use a pre-defined set of graphs, since these are often too narrow to cover all corner-cases and are tailored around the original use-cases of the algorithm. Therefore I would not drop the strategy of testing random graphs.

It is true though that this one algorithm does incorporate a lot of generated graphs. While fast for now, running times can explode when new features are encoded or data structures change. I would therefore suggest to drop some redundancy / cases. As a rule of thumb, most algorithms don't test for more than 10 different graphs.

Schwarf:
> > Hi @angriman,
> > Thank you for your feedback! I’d like to provide some context and arguments for the choice of graph sizes in the tests and the use of many test cases:
> > ```
> > 1. Multiple and Larger Graphs
> >    During the initial implementation, I relied on smaller, single-graph tests, which passed successfully. However, when I extended the tests to include many graphs of the same class (e.g., multiple grid graphs instead of a single 3x3 grid graph), some cases emerged, revealing bugs in the algorithm. Additionally, larger and random graphs helped me uncover issues that smaller graphs missed. These are the main reasons I included these tests.
> > 
> > 2. Algorithm extension
> >    I plan to extend my contribution by modifying the algorithm to provide planar embeddings for planar graphs. This may require minor refactoring of the current implementation. Similarly, potential performance improvements might also lead to refactoring. Having a variety of test cases (including larger and multiple graphs) will make it easier for me and others to ensure correctness during these changes.
> > 
> > 3. Negligible Runtime Cost
> >    On my machine, the runtime difference between testing a single graph versus hundreds of graphs is of the order of a few milliseconds, which is negligible compared to tests that take up to 10 seconds. The same applies to the included large random graphs. While I understand the importance of keeping tests efficient ("Kleinvieh macht auch Mist"), I currently believe the benefits of including larger and multiple graphs, outweigh the costs.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Given these reasons, I’d appreciate it if we could retain the current test cases in the file `LeftRightPlanarityCheckGTest.cpp`. Of course, I’m happy to hear your thoughts.
> > What do you think?
> 
> There are positives and negatives about having random generated graphs for testing. In the past it helped us discover some quite subtle bugs. However, sometimes these can come from the generators itself, leading to the case that a test implicitly checks the functionalities of different code units. Also you can not predict when these errors turn up. Overall there are better systems to generate these tests, but these would need a complete pipeline overhaul.
> 
> We had more bugs in algorithms / core functionalities which only use a pre-defined set of graphs, since these are often too narrow to cover all corner-cases and are tailored around the original use-cases of the algorithm. Therefore I would not drop the strategy of testing random graphs.
> 
> It is true though that this one algorithm does incorporate a lot of generated graphs. While fast for now, running times can explode when new features are encoded or data structures change. I would therefore suggest to drop some redundancy / cases. As a rule of thumb, most algorithms don't test for more than 10 different graphs.

Hi @farbratu,

Thank you for your detailed feedback! I have to admit that I might not have fully grasped all the concerns raised by @angriman. Thank you for providing additional context and suggestions.

1. I agree with your suggestion to reduce redundancy and limit the number of graphs tested for each type. I’ll adjust the tests to include around 10 graphs per type and in some cases, perhaps up to 20.
2. Regarding the use of random graphs: these are programmatically generated. I will reduce their size from 50–100 nodes to a more manageable 10–20 nodes to ensure the tests remain efficient while still being diverse.

Before I proceed, is the above proposal acceptable for your guys?

angriman:
> > > Hi @angriman,
> > > Thank you for your feedback! I’d like to provide some context and arguments for the choice of graph sizes in the tests and the use of many test cases:
> > > ```
> > > 1. Multiple and Larger Graphs
> > >    During the initial implementation, I relied on smaller, single-graph tests, which passed successfully. However, when I extended the tests to include many graphs of the same class (e.g., multiple grid graphs instead of a single 3x3 grid graph), some cases emerged, revealing bugs in the algorithm. Additionally, larger and random graphs helped me uncover issues that smaller graphs missed. These are the main reasons I included these tests.
> > > 
> > > 2. Algorithm extension
> > >    I plan to extend my contribution by modifying the algorithm to provide planar embeddings for planar graphs. This may require minor refactoring of the current implementation. Similarly, potential performance improvements might also lead to refactoring. Having a variety of test cases (including larger and multiple graphs) will make it easier for me and others to ensure correctness during these changes.
> > > 
> > > 3. Negligible Runtime Cost
> > >    On my machine, the runtime difference between testing a single graph versus hundreds of graphs is of the order of a few milliseconds, which is negligible compared to tests that take up to 10 seconds. The same applies to the included large random graphs. While I understand the importance of keeping tests efficient ("Kleinvieh macht auch Mist"), I currently believe the benefits of including larger and multiple graphs, outweigh the costs.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Given these reasons, I’d appreciate it if we could retain the current test cases in the file `LeftRightPlanarityCheckGTest.cpp`. Of course, I’m happy to hear your thoughts.
> > > What do you think?
> > 
> > 
> > There are positives and negatives about having random generated graphs for testing. In the past it helped us discover some quite subtle bugs. However, sometimes these can come from the generators itself, leading to the case that a test implicitly checks the functionalities of different code units. Also you can not predict when these errors turn up. Overall there are better systems to generate these tests, but these would need a complete pipeline overhaul.
> > We had more bugs in algorithms / core functionalities which only use a pre-defined set of graphs, since these are often too narrow to cover all corner-cases and are tailored around the original use-cases of the algorithm. Therefore I would not drop the strategy of testing random graphs.
> > It is true though that this one algorithm does incorporate a lot of generated graphs. While fast for now, running times can explode when new features are encoded or data structures change. I would therefore suggest to drop some redundancy / cases. As a rule of thumb, most algorithms don't test for more than 10 different graphs.
> 
> Hi @farbratu,
> 
> Thank you for your detailed feedback! I have to admit that I might not have fully grasped all the concerns raised by @angriman. Thank you for providing additional context and suggestions.
> 
> 1. I agree with your suggestion to reduce redundancy and limit the number of graphs tested for each type. I’ll adjust the tests to include around 10 graphs per type and in some cases, perhaps up to 20.
> 2. Regarding the use of random graphs: these are programmatically generated. I will reduce their size from 50–100 nodes to a more manageable 10–20 nodes to ensure the tests remain efficient while still being diverse.
> 
> Before I proceed, is the above proposal acceptable for you guys?

I agree that testing on random graphs is very helpful for finding bugs in the algorithm you are developing (I've used them a lot in the past). However, I think that, when possible, they should not be part of unit tests because of the following reasons (more about testing best practices can be found in [1]):
- Reproducibility: random tests can be flaky when tested on different architectures. It is important to have reproducible tests when possible.
- If they fail, they don't help you understand what in the algorithm is not behaving as intended.

My recommendation is to test on a fixed and representative set of instances to evaluate that the algorithm behaves as intended. I understand that sometimes this is easier said than done because the algorithm may be complex. If that's the case, we can keep tests on random graphs.

Regarding large graphs: it's ok to have individual tests on larger instances (you can also use those in the input/ directory in NetworKit). What I'm against is writing a unit test with 50+ lines used for graph edges, which makes it impossible for a future reader to understand what's going on. Why do you need do define such a large graph? What kind of behavior are you trying to test that cannot be tested with a smaller instance?

If you just need to test your algorithm on a larger graph, then my recommendation is to replace the custom graphs with 50+ nodes (`testRandomGeneratedPlanarGraph50/51/100Nodes`) with graphs in the `input/` directory.

Let me know if you have further questions!

[1] https://abseil.io/resources/swe-book/html/ch12.html

Schwarf:
> > > > Hi @angriman,
> > > > Thank you for your feedback! I’d like to provide some context and arguments for the choice of graph sizes in the tests and the use of many test cases:
> > > > ```
> > > > 1. Multiple and Larger Graphs
> > > >    During the initial implementation, I relied on smaller, single-graph tests, which passed successfully. However, when I extended the tests to include many graphs of the same class (e.g., multiple grid graphs instead of a single 3x3 grid graph), some cases emerged, revealing bugs in the algorithm. Additionally, larger and random graphs helped me uncover issues that smaller graphs missed. These are the main reasons I included these tests.
> > > > 
> > > > 2. Algorithm extension
> > > >    I plan to extend my contribution by modifying the algorithm to provide planar embeddings for planar graphs. This may require minor refactoring of the current implementation. Similarly, potential performance improvements might also lead to refactoring. Having a variety of test cases (including larger and multiple graphs) will make it easier for me and others to ensure correctness during these changes.
> > > > 
> > > > 3. Negligible Runtime Cost
> > > >    On my machine, the runtime difference between testing a single graph versus hundreds of graphs is of the order of a few milliseconds, which is negligible compared to tests that take up to 10 seconds. The same applies to the included large random graphs. While I understand the importance of keeping tests efficient ("Kleinvieh macht auch Mist"), I currently believe the benefits of including larger and multiple graphs, outweigh the costs.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Given these reasons, I’d appreciate it if we could retain the current test cases in the file `LeftRightPlanarityCheckGTest.cpp`. Of course, I’m happy to hear your thoughts.
> > > > What do you think?
> > > 
> > > 
> > > There are positives and negatives about having random generated graphs for testing. In the past it helped us discover some quite subtle bugs. However, sometimes these can come from the generators itself, leading to the case that a test implicitly checks the functionalities of different code units. Also you can not predict when these errors turn up. Overall there are better systems to generate these tests, but these would need a complete pipeline overhaul.
> > > We had more bugs in algorithms / core functionalities which only use a pre-defined set of graphs, since these are often too narrow to cover all corner-cases and are tailored around the original use-cases of the algorithm. Therefore I would not drop the strategy of testing random graphs.
> > > It is true though that this one algorithm does incorporate a lot of generated graphs. While fast for now, running times can explode when new features are encoded or data structures change. I would therefore suggest to drop some redundancy / cases. As a rule of thumb, most algorithms don't test for more than 10 different graphs.
> > 
> > 
> > Hi @farbratu,
> > Thank you for your detailed feedback! I have to admit that I might not have fully grasped all the concerns raised by @angriman. Thank you for providing additional context and suggestions.
> > 
> > 1. I agree with your suggestion to reduce redundancy and limit the number of graphs tested for each type. I’ll adjust the tests to include around 10 graphs per type and in some cases, perhaps up to 20.
> > 2. Regarding the use of random graphs: these are programmatically generated. I will reduce their size from 50–100 nodes to a more manageable 10–20 nodes to ensure the tests remain efficient while still being diverse.
> > 
> > Before I proceed, is the above proposal acceptable for you guys?
> 
> I agree that testing on random graphs is very helpful for finding bugs in the algorithm you are developing (I've used them a lot in the past). However, I think that, when possible, they should not be part of unit tests because of the following reasons (more about testing best practices can be found in [1]):
> 
>     * Reproducibility: random tests can be flaky when tested on different architectures. It is important to have reproducible tests when possible.
> 
>     * If they fail, they don't help you understand what in the algorithm is not behaving as intended.
> 
> 
> My recommendation is to test on a fixed and representative set of instances to evaluate that the algorithm behaves as intended. I understand that sometimes this is easier said than done because the algorithm may be complex. If that's the case, we can keep tests on random graphs.
> 
> Regarding large graphs: it's ok to have individual tests on larger instances (you can also use those in the input/ directory in NetworKit). What I'm against is writing a unit test with 50+ lines used for graph edges, which makes it impossible for a future reader to understand what's going on. Why do you need do define such a large graph? What kind of behavior are you trying to test that cannot be tested with a smaller instance?
> 
> If you just need to test your algorithm on a larger graph, then my recommendation is to replace the custom graphs with 50+ nodes (`testRandomGeneratedPlanarGraph50/51/100Nodes`) with graphs in the `input/` directory.
> 
> Let me know if you have further questions!
> 
> [1] https://abseil.io/resources/swe-book/html/ch12.html

Thank you for the detailed explanation of your concerns. I considered the large graphs of the input folder before but shied away from the effort to independently determine if they are planar or not. I will go ahead and try while removing the "random graphs" from the unit tests. For the "typed graphs" I will reduce the number of generated graphs as proposed above and reorganize the tests to focus on specific behaviors. Thank your for the link to the article. It was a great read and a helpful reminder of best practices.

include/networkit/graph/Graph.hpp
Schwarf: I added this to define
``` static constexpr Edge noneEdge{}; ```
in ```LeftRightPlanarityCheck.hpp```.
But I will change it back, since I don't really need the definition at compile-time.

File: networkit/cpp/planarity/LeftRightPlanarityCheck.cpp (hunk 24 above)
Comment (Schwarf) [reply networkit/cpp/planarity/LeftRightPlanarityCheck.cpp:172]: I've moved the inner-loop into a lambda and hope this change improves readability by reducing the nesting.

File: include/networkit/graph/Graph.hpp (hunk 6 above)
Comment (angriman) [reply include/networkit/graph/Graph.hpp:48]: This was marked as resolved but the comment was not addressed, same for other comments. Please either address the comment or add some details about why it was not resolved.

Comment (Schwarf) [reply include/networkit/graph/Graph.hpp]: Sorry. I did resolve the conversations, where I fixed the review findings... but I plan to make one push with all commits.

File: include/networkit/graph/Graph.hpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (Schwarf) [reply include/networkit/graph/Graph.hpp:952]: If this function can indeed be reused, I am willing to address that in a subsequent PR. This PR is already significant in scope, and I prefer not to mix the concerns of implementing a new feature with refactoring existing code.

Regarding the preference for a template parameter over std::function: Could you please elaborate on the reasoning? For sorting neighbors, the API requires a binary predicate that returns a bool. In my opinion, using std::function clearly communicates this requirement at the API level, whereas a template parameter does not.

File: networkit/cpp/planarity/test/LeftRightPlanarityTestGtest.cpp [hunk 39]
```diff
@@ -0,0 +1,637 @@
...
+    graph.addEdge(4, 8);
+    graph.addEdge(5, 6);
+    graph.addEdge(5, 7);
+    graph.addEdge(5, 8);
+    LeftRightPlanarityTest test(graph);
+    test.run();
+    EXPECT_FALSE(test.isPlanar());
+}
+
+TEST_F(LeftRightPlanarityTestGTest, RandomGeneratedPlanarGraph50Nodes) {
```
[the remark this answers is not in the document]
Schwarf: Tests are removed

networkit/cpp/graph/test/GraphGTest.cpp
```diff
@@ -2374,4 +2374,330 @@ TEST_P(GraphGTest, testEdgeIdsAfterRemoveWithoutSortingOrIDs) {
...
+
+    // Validate that neighbors are sorted according to weights
+    G.forNodes([&](const node currentNode) {
+        const auto sortedNeighbors = G.neighborRange(currentNode);
+        std::vector<edgeweight> sortedWeights;
+        for (const node neighbor : sortedNeighbors) {
+            sortedWeights.push_back(G.weight(currentNode, neighbor));
+        }
+        // Ensure weights are sorted in ascending order
+        EXPECT_TRUE(std::is_sorted(sortedWeights.begin(), sortedWeights.end()));
```
angriman, line 2576: Use `std::ranges::is_sorted`, also below.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,282 @@
...
+                stack.emplace(Interval{}, Interval(currentEdge, currentEdge));
+            }
+
+            if (auto currentEdgeIterator = lowestPoint.find(currentEdge);
+                currentEdgeIterator != lowestPoint.end()
+                && currentEdgeIterator->second < heights[currentNode]) {
+
+                if (neighbor == *dfsGraph.neighborRange(currentNode).begin()) {
+                    lowestPointEdge[parentEdges[currentNode]] = lowestPointEdge[currentEdge];
+                } else {
```
angriman, line 193: This should be equivalent and you don't need an additional indentation:
```cpp
else if (!applyConstraints(...)) {
   return false;
}
```

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,282 @@
...
+        if (auto it = neighborIterators.find(currentNode); it == neighborIterators.end()) {
+            neighborIterators[currentNode] = dfsGraph.neighborRange(currentNode).begin();
+        }
+
+        if (!processNeighborEdges(currentNode, callRemoveBackEdges)) {
+            return false;
+        }
+
+        if (callRemoveBackEdges) {
+            if (parentEdge != noneEdge) {
```
angriman, line 220:
```cpp
if (callRemoveBackEdges && parentEdge != noneEdge) { ... }
```

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+    } while (!dfsStack.empty());
+
+    return true;
+}
+
+void LeftRightPlanarityCheck::dfsOrientation(node startNode) {
+    std::stack<node> dfsStack;
+    dfsStack.emplace(startNode);
```
angriman, line 231: `emplace` is useful for constructing objects inplace, but using it for primitive type is equivalent to `push`. So for this stack I'd suggest to use `push`, see more here https://abseil.io/tips/112.

networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
```diff
@@ -0,0 +1,282 @@
...
+                    continue;
+                dfsGraph.addEdge(currentNode, neighbor);
+                lowestPoint[currentEdge] = heights[currentNode];
+                secondLowestPoint[currentEdge] = heights[currentNode];
+                if (heights[neighbor] == noneHeight) // Tree edge
+                {
+                    parentEdges[neighbor] = currentEdge;
+                    heights[neighbor] = heights[currentNode] + 1;
+                    dfsStack.emplace(currentNode);
+                    dfsStack.emplace(neighbor);
```
angriman, line 252: Same comment as above.

angriman (approved):
Thank you for addressing the comments, I left a few minor ones, the rest LGTM.

include/networkit/graph/Graph.hpp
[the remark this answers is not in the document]
angriman: Fair point, then please do so in a subsequent PR.

fabratu:
A side-comment: The error in the pipeline for the aarch64 wheel build is fixed in PR #1283 and can be ignored concerning the finish state of this PR.

include/networkit/planarity/LeftRightPlanarityCheck.hpp
```diff
@@ -0,0 +1,111 @@
...
+     * without any edges crossing. For an overview of planar graphs, refer to:
+     * https://en.wikipedia.org/wiki/Planar_graph
+     *
+     * The algorithm achieves (almost) linear runtime complexity. The only non-linear component
+     * arises from sorting the nodes of the depth-first search tree.
+     *
+     * @param G The input graph to test for planarity. The graph should be undirected.
+     */
+    LeftRightPlanarityCheck(const Graph &G) : graph(&G) {
+        if (G.isDirected()) {
```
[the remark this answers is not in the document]
Schwarf: I believe applying assureFinished() here would reduce the clarity of the error message. The current message ("The graph is not an undirected graph.") provides specific feedback. Using assureFinished() would replace this with a misleading message ("Error, run must be called first"), which might confuse the user.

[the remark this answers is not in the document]
angriman: Ah sorry, your code is fine.

angriman:
There are still unresolved comments (e.g., `std::ranges::is_sorted` vs `std::is_sorted`), did you forget to push some commits?

Schwarf:
> There are still unresolved comments (e.g., `std::ranges::is_sorted` vs `std::is_sorted`), did you forget to push some commits?
  
They should all be marked as Outdated. I will go through and resolve them.

angriman (approved):

fabratu (approved):

PR #1283, opened 2025-01-16
CI: Split coverage and leak build, fix leaks, remove aarch64 wheel build

fabratu:
This PR effectively fixes the broken coverage upload. Due to also testing for leaks and continuing on errors, the coverage for the C++ code way empty if a potential address leak is found. Both tests are now split into different runs. This should also speedup the coverage run due to having both leak checks and coverage at the same time.

In addition several potential leaks are fixed. One fix in `random.cc` from libstd is not fixed, since it is intentional. The fix instead is to remove the out-of-bound sanitizer checks for this file by adding a filter.

### DIFF
--- a/networkit/cpp/community/LouvainMapEquation.cpp
+++ b/networkit/cpp/community/LouvainMapEquation.cpp
@@ -36,8 +36,8 @@ convertStringToParallelizationType(std::string_view parallelizationStrategy) {
 
 LouvainMapEquation::LouvainMapEquation(const Graph &graph, bool hierarchical, count maxIterations,
                                        ParallelizationType parallelizationType)
-    : CommunityDetectionAlgorithm(graph), parallel(parallelizationType > ParallelizationType::NONE),
-      parallelizationType(parallelizationType), hierarchical(hierarchical),
+    : CommunityDetectionAlgorithm(graph), parallelizationType(parallelizationType),
+      parallel(parallelizationType > ParallelizationType::NONE), hierarchical(hierarchical),
       maxIterations(maxIterations), clusterCut(graph.upperNodeIdBound()),
       clusterVolume(graph.upperNodeIdBound()),
       locks(parallelizationType == ParallelizationType::RELAX_MAP ? graph.upperNodeIdBound() : 0),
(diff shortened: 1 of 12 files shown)


networkit/cpp/community/LouvainMapEquation.cpp
```diff
@@ -34,6 +34,21 @@ LouvainMapEquation::LouvainMapEquation(const Graph &graph, bool hierarchical, co
     result = Partition(graph.upperNodeIdBound());
 }
 
+LouvainMapEquation::LouvainMapEquation(const Graph &graph, bool hierarchical, count maxIterations,
+                                       std::string_view parallelizationStrategy)
+    : CommunityDetectionAlgorithm(graph), hierarchical(hierarchical), maxIterations(maxIterations),
+      clusterCut(graph.upperNodeIdBound()), clusterVolume(graph.upperNodeIdBound()) {
+    parallelizationType = convertStringToParallelizationType(parallelizationStrategy);
```
bernlu, line 41: this line should be part of the initializer list and not in the function body

bernlu: The other variables in this function could be moved into the initializer list as well

networkit/cpp/generators/ChungLuGeneratorAlamEtAl.cpp
```diff
@@ -121,10 +123,13 @@ Graph ChungLuGeneratorAlamEtAl::generateSequential() {
             const double p_j = static_cast<double>(groups[i].degrees)
                                * static_cast<double>(groups[j].degrees)
                                / static_cast<double>(sum_deg);
-            edgeSkipping(rng, addEdge, i, j, p_j, groups[i].vertexCount * groups[j].vertexCount);
+            if (p_j > 0.0)
```
bernlu, line 126: maybe move this check into `edgeSkipping` instead

networkit/cpp/graph/TopologicalSort.cpp
```diff
@@ -36,7 +34,7 @@ void TopologicalSort::checkDirected() {
 void TopologicalSort::checkNodeIdMap() {
     size_t numberOfNodes = G.numberOfNodes();
     std::vector<bool> checkTable(numberOfNodes);
-    for (auto &[origNode, mappedNode] : nodeIdMap) {
+    for (auto &[origNode, mappedNode] : *nodeIdMap) {
```
bernlu, line 37: `nodeIdMap` could be a `nullptr` - check?

bernlu (changes requested):

bernlu (dismissed):

bernlu (dismissed):

fabratu:
Sry for the drive-by fix, but since the PR is finished and concerned with CI anyways, I added a missing package for the coverage build.

bernlu (approved):

Issue #938, opened 2022-05-05
Suppress `ReducedFunctionalityWarning`


Thijss:
Hi there,

Would it be possible to suppress the `ReducedFunctionalityWarning`?
https://github.com/networkit/networkit/blob/2aaa0eecc82560c0d6b212e494706a23497dc8ea/networkit/simulation.pyx#L11-L14

If I understand correctly, I think it is only relevant when using `EpidemicSimulationSEIR.getData()`
https://github.com/networkit/networkit/blob/2aaa0eecc82560c0d6b212e494706a23497dc8ea/networkit/simulation.pyx#L65

The warning is now raised on import.

Perhaps the same logic can be applied as applied here?
https://github.com/networkit/networkit/blob/6205384e771c7e3ba1a1e2f46731b5ea01e1f80f/networkit/plot.py#L12-L17

The warning could then be logged upon running `.getData()` if `not have_pandas`

or an error could be thrown like the `plot` module:
https://github.com/networkit/networkit/blob/6205384e771c7e3ba1a1e2f46731b5ea01e1f80f/networkit/plot.py#L123-L124

I am not too familiair with `.pyx` files, so not sure this is possible

Thijss:
I am not too familiair with `.pyx` files, so not sure this is possible

angriman:
This seems quite simple to do; we could use an internal variable that suppresses the warning.

PR #1296, opened 2025-02-06
Issue #938: suppress warning for pandas in networkit/simulation.pyx

domischweisgut:
In networkit/simulation.pyx: We suppress now the warning after import of the pandas package and instead throw an error only when we actually want to use this package. This looks like this: 

Output without pandas when: 
'
>>> from networkit import *
>>> G = readGraph("./input/airfoil1.graph", Format.METIS)
>>> sim = simulation.EpidemicSimulationSEIR(G, 1, 0.1, 1, 0, 0)
>>> sim.getData()
' 

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "networkit/simulation.pyx", line 70, in networkit.simulation.EpidemicSimulationSEIR.getData
    raise MissingDependencyError("pandas")
networkit.support.MissingDependencyError: Optional dependency pandas is not installed

##############################################

Output with pandas installed:

Empty DataFrame
Columns: [zero, time, state, count]
Index: []

### DIFF
--- a/networkit/simulation.pyx
+++ b/networkit/simulation.pyx
@@ -58,7 +58,7 @@ cdef class EpidemicSimulationSEIR(Algorithm):
 		"""
 		try:
 			import pandas
-		return pandas.DataFrame((<_EpidemicSimulationSEIR*>(self._this)).getData(), columns=["zero", "time", "state", "count"])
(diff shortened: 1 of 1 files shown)


bernlu:
#938

bernlu (approved):

c-bebop (commented):
Looks fine.

Issue #909, opened 2022-03-24
Segfault writing empty graph in NetworkitBinary format


opatut:
Writing an empty graph to file with `writeGraph` causes a segmentation fault.

Here's a minimal example for this bug:

```python
>>> import networkit
>>> g = networkit.graph.Graph()
>>> networkit.graphio.writeGraph(g, 'x', networkit.Format.NetworkitBinary)
[1]    4538 segmentation fault (core dumped)  python
```

My machine:

```
Linux garlic 5.16.15-arch1-1 #1 SMP PREEMPT Thu, 17 Mar 2022 00:30:09 +0000 x86_64 GNU/Linux
Python 3.9.7
Networkit 9.1.1 from pypi
```

The segfault does not occur if you add a `g.addNode()` before writing the graph.

angriman:
The segfault occurs because the NetworKit binary writer does not support empty graphs. We'll try to fix this asap.

Thijss:
I just encountered this too. Great to hear it will be fixed soon :D

Thijss:
My current workaround for anyone interested:

```python
def cache_nk_graph(nk_graph: Graph, cache_path: PosixPath) -> PosixPath:
    if nk_graph.numberOfNodes() > 0:
        nk.writeGraph(nk_graph, str(cache_path.resolve()), nk.Format.NetworkitBinary)
    else:
        nk.writeGraph(nk_graph, str(cache_path.resolve()), nk.Format.EdgeList)
    return cache_path


def load_nk_graph(nk_cache_path: PosixPath) -> Graph:
    try:
        return nk.readGraph(str(nk_cache_path), nk.Format.NetworkitBinary)
    except TypeError:
        return nk.readGraph(str(nk_cache_path), nk.Format.EdgeList)
```

Thijss:
My current workaround for anyone interested:
```python
def cache_nk_graph(nk_graph: Graph, cache_path: PosixPath) -> PosixPath:
    if nk_graph.numberOfNodes() > 0:
        nk.writeGraph(nk_graph, str(cache_path.resolve()), nk.Format.NetworkitBinary)
    else:
        nk.writeGraph(nk_graph, str(cache_path.resolve()), nk.Format.LFR)
    return cache_path


def load_nk_graph(nk_cache_path: PosixPath) -> Graph:
    try:
        return nk.readGraph(str(nk_cache_path), nk.Format.LFR)
    except TypeError:
[...]

PR #1298, opened 2025-02-06
Bugfix: Writing Networkit Binary (Empty) Graph, Issue #909

c-bebop:
Issue #909 concerns a bug where writing an empty graph in the NetworKit binary format leads to a segmentation fault.

This PR fixes the issue by:
- check first if there was any data to be written
- if not we do not de-reference the vectors containing data but simply write 0

### DIFF
--- a/networkit/test/test_graphio.py
+++ b/networkit/test/test_graphio.py
@@ -131,6 +131,16 @@ class TestGraphIO(unittest.TestCase):
                 G1 = nk.graphio.readGraph(filename, format, *kargs)
             self.checkStatic(G, G1)
 
+    def testWriteEmptyGraph(self):
+        empty_graph_write = nk.graph.Graph()
+        # empty_graph_write.addNode()
+        output_path = 'output/empty_graph.bin'
+        nk.graphio.writeGraph(empty_graph_write, output_path, nk.Format.NetworkitBinary)
+        empty_graph_read = nk.readGraph(output_path, nk.Format.NetworkitBinary)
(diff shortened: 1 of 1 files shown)


networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -826,6 +839,19 @@ TEST_F(IOGTest, testKONECTGraphReader) {
     ASSERT_EQ(G.weight(0, 1), 1.261404);
     ASSERT_EQ(G.weight(127, 48), 0.03050447);
 }
+
+TEST_F(IOGTest, testNetworkitBinaryWriteEmpty) {
+    Graph graph(0, false, true);
+    NetworkitBinaryWriter writer;
+    std::filesystem::path const path = "empty_graph.bin";
```
fabratu, line 846: The path/filename should be "output/empty_graph.nkb".

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -826,6 +839,19 @@ TEST_F(IOGTest, testKONECTGraphReader) {
     ASSERT_EQ(G.weight(0, 1), 1.261404);
     ASSERT_EQ(G.weight(127, 48), 0.03050447);
 }
+
+TEST_F(IOGTest, testNetworkitBinaryWriteEmpty) {
```
fabratu, line 843: Since the test for GraphTool and NetworKit binary writer both do the same, I suggest to use the same naming scheme.

fabratu (changes requested):

PR #1303, opened 2025-02-08
Refactor sortNeighbors with OpenMP Parallelization & Evaluate Reuse in sortEdges

Schwarf:
# Refactor `sortNeighbors` with OpenMP Parallelization & Evaluate Reuse in `sortEdges`

## Summary  
This PR refactors `sortNeighbors` to leverage OpenMP for parallel sorting. The goal was to explore whether `sortNeighbors` could replace the `sortAdjacencyArrays` lambda inside `sortEdges`, as suggested in a [previous discussion](https://github.com/networkit/networkit/pull/1276#discussion_r1921507461). 

## Key Changes  
- Parallelized `sortNeighbors` similar to `sortEdges`
- Resolved race conditions that arose due to parallel execution.  

## Challenges & Findings  
- While the parallelization improves `sortNeighbors`, I did not see an obvious way how to reuse it in `sortEdges`.  
- Both `sortNeighbors` and `sortEdges` now utilize OpenMP, but their structures differ, raising the question of whether they should remain separate or be unified.  

## Next Steps & Request for Feedback  
I would appreciate input on:  
- Whether the refactored `sortNeighbors` should be merged as is, even if it doesn't fully replace the existing logic in `sortEdges`.  
- Any suggestions for unifying the sorting logic while maintaining performance and correctness are welcome since I have only limited experience in multi-threaded environments.

Tagging @angriman and other reviewers for their thoughts. Any guidance, best practices regarding multi-threaded applications would be very welcome! 🚀

### DIFF
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -2289,17 +2289,10 @@ void Graph::sortNeighbors(Lambda lambda) {
         // Sort the outEdge-Attributes
         auto &outIndices = threadData[omp_get_thread_num()];
         outIndices.clear();
-        if (outEdges[w].empty())
-            return;
         outIndices.resize(outEdges[w].size());
 
         std::iota(outIndices.begin(), outIndices.end(), 0);
         std::ranges::sort(outIndices, [&](index a, index b) {
-            if (a >= outEdges[w].size() || b >= outEdges[w].size()) {
-                std::cerr << "Invalid index in outEdges: " << a << ", " << b
-                          << " (max: " << outEdges[w].size() << ")\n";
-                std::terminate();
-            }
             return extLambda(w, outEdges[w][a], outEdges[w][b]);
         });
 
@@ -2317,10 +2310,12 @@ void Graph::sortNeighbors(Lambda lambda) {
         }
         // For directed graphs we need to sort the inEdge-Attributes separately
         if (directed) {
+            if (inEdges[w].empty())
+                return;
             auto &inIndices = threadData[omp_get_thread_num()];
             inIndices.clear();
-            if (inEdges[w].size() > inIndices.size())
-                inIndices.resize(inEdges[w].size());
+
+            inIndices.resize(inEdges[w].size());
             std::iota(inIndices.begin(), inIndices.end(), 0);
 
             std::ranges::sort(inIndices, [&](index a, index b) {
--- a/include/networkit/graph/Graph.hpp
+++ b/include/networkit/graph/Graph.hpp
@@ -1002,7 +1002,7 @@ public:
      *               precede the second in the sorted order.
      */
     template <typename Lambda>
-    void sortNeighbors(node u, Lambda lambda);
+    void sortNeighbors(Lambda lambda);
 
     /**
      * Sorts the adjacency arrays by node id. While the running time is linear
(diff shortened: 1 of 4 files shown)


fabratu:
> Another approach could be to extract the common sorting logic (as shown in this PR) into a new reusable function.

Maybe I am overlooking something here, but shouldn't it be possible to leave the API of `sortNeighbors` as is (with as is I mean before the changes introduced in this PR). Inside `sortNeighbors` then the parameters for the lambdas are formed as they are created in `sortAdjacencyArrays`.

```
template <typename Lambda>
void Graph::sortNeighbors(node u, Lambda lambda) {
...
        if (isWeighted()) {
            if (hasEdgeIds())
                std::sort(indices.begin(), indicesEnd, [&](auto a, auto b) -> bool {
                    return lambda(WeightedEdgeWithId{u, outEdges[a], outWeights[a], outEdgeIds[a]},
                                  WeightedEdgeWithId{u, outEdges[b], outWeights[b], outEdgeIds[b]});
                });
...
```

So the call would be like this in `sortEdges`:

```
template <class Lambda>
void Graph::sortEdges(Lambda lambda) {
    ...
    balancedParallelForNodes([&](const node u) {
        if (degree(u) < 2)
            return;
        
        sortNeighbors(u, lambda);
    });
```

The case for directed graphs is missing by directly calling the lambda, so maybe it makes sense to move `sortAdjacencyArrays` inside sortNeighbors and call it from there for out/inEdges.

Schwarf:
> > Another approach could be to extract the common sorting logic (as shown in this PR) into a new reusable function.
> 
> Maybe I am overlooking something here, but shouldn't it be possible to leave the API of `sortNeighbors` as is (with as is I mean before the changes introduced in this PR). Inside `sortNeighbors` then the parameters for the lambdas are formed as they are created in `sortAdjacencyArrays`.
> 
> ```
> template <typename Lambda>
> void Graph::sortNeighbors(node u, Lambda lambda) {
> ...
>         if (isWeighted()) {
>             if (hasEdgeIds())
>                 std::sort(indices.begin(), indicesEnd, [&](auto a, auto b) -> bool {
>                     return lambda(WeightedEdgeWithId{u, outEdges[a], outWeights[a], outEdgeIds[a]},
>                                   WeightedEdgeWithId{u, outEdges[b], outWeights[b], outEdgeIds[b]});
>                 });
> ...
> ```
> 
> So the call would be like this in `sortEdges`:
> 
> ```
> template <class Lambda>
> void Graph::sortEdges(Lambda lambda) {
>     ...
>     balancedParallelForNodes([&](const node u) {
>         if (degree(u) < 2)
>             return;
>         
>         sortNeighbors(u, lambda);
>     });
> ```
> 
> The case for directed graphs is missing by directly calling the lambda. Maybe it makes sense to move `sortAdjacencyArrays` inside sortNeighbors and call it from there for out/inEdges.

This approach might work when looking at sortNeighbors and sortEdges in isolation, but the sortNeighbors API was specifically introduced to support sorting in [this part of the code](https://github.com/networkit/networkit/blob/2e5e7222759d25ae6cf2a64b7d33a2e47087803f/networkit/cpp/planarity/LeftRightPlanarityCheck.cpp#L39).
The Left-Right-Planarity-Check algorithm (and the nestingDepth data structure) does not require the additional indirection via WeightedEdgeWithId, and I would prefer to avoid introducing it, as it is unnecessary for the algorithm to function.

Schwarf:
Thanks for the suggestion! I’ve been working on implementing this approach, but I’ve run into some issues. Specifically, I’m encountering segmentation faults and infinite loops in tests related to the Left-Right-Planarity-Check when using the modified sortNeighbors.

I suspect there might be a race condition or an unintended change in memory access patterns due to OpenMP, but I haven’t fully pinpointed the issue yet. I started debugging it by testing sortNeighbors in isolation and checking for differences in lambda behavior, but I’m still stuck.

This might take me some time to resolve. Just wanted to give a quick update—I'll keep you posted!

Schwarf:
Given the current changes in this PR, I have a question about NetworKit’s policy on breaking changes.
If I modify the `sortEdgesByWeight` method to:
```
void sortEdgesByWeight(Graph &G, bool decreasing) {
    if (decreasing)
        G.sortNeighbors([&](node currenNode, node neighbor1, node neighbor2) {
            if (G.weight(currenNode, neighbor1) == G.weight(currenNode, neighbor2))
                return neighbor1 < neighbor2;
            return G.weight(currenNode, neighbor1) > G.weight(currenNode, neighbor2);
        });
    else
        G.sortNeighbors([&](node currenNode, node neighbor1, node neighbor2) {
            if (G.weight(currenNode, neighbor1) == G.weight(currenNode, neighbor2))
                return neighbor1 < neighbor2;
            return G.weight(currenNode, neighbor1) < G.weight(currenNode, neighbor2);
        });
}
```
I could simplify `sortEdges` to:
```
template <class Lambda>
void Graph::sortEdges(Lambda lambda) {
    sortNeighbors(lambda);
}
```
However, this change would be breaking for users relying on Graph::sortEdges(Lambda lambda).

fabratu:
General API-changes (both C++ and Python) are possible, but need deprecation warningss. We usually include the new signature in the warning itself. Deprecations are usually in place for two releases and an API-changing release is always major.

We are more restrictive however concerning the core, since it is expected that changes affect most users. This includes (amongst others) the `Graph` object. If there are no good reasons for doing so, changing the behavior or API of core functions is normally kept to a minimum.

Concerning your code example ... I am not sure, if I can follow completely. Is your intent to change the signature or functionality of a function?

Based on my suggestions - maybe it helps if you share the problematic code (preferably in a another PR or a gist), then we might be able to help with insights / debugs.

Schwarf:
Hi @fabratu,
Thanks for the feedback!
I’ve pushed my latest suggestion to restart the discussion.

The changes affect the API of `sortEdges`, which now forwards to a refactored version of `sortNeighbors`. The new `sortNeighbors` accepts a lambda of the form `lambda(node u, node v1, node v2)`, which allows sorting based on the source node as well. This unifies the sorting logic and removes duplication between the two methods.

That said, this does **break the previously introduced `sortEdges` and `sortNeighbors` API**.
Let me know your thoughts.

PR #1318, opened 2025-04-14
Add Prim's algorithm

Schwarf:
## Summary
This PR introduces the `PrimMSF` class, which implements Prim's algorithm to compute a Minimum Spanning Forest (MSF) for an undirected graph.

- For connected graphs, it produces a Minimum Spanning Tree (MST).
- For disconnected graphs, it computes a forest of MSTs.

It extends the base `SpanningForest` interface, requires undirected graphs as input, and reuses the existing Kruskal tests to verify correctness.


bernlu (dismissed):

include/networkit/graph/PrimMSF.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @brief Returns the total weight of the minimum spanning forest.
+     * @return The total weight of the minimum spanning forest.
+     */
+    edgeweight getTotalWeight() const {
+        assureFinished();
+        if (G->isWeighted())
+            return totalWeight;
+        return static_cast<edgeweight>(forest.numberOfEdges());
```
fabratu, line 58: Thanks for your contribution!

Taking a look at the coverage, this line is not executed. Can you add a test using an unweighted graph? In principle, we should aim at 100% coverage for newly added headers/implementations.

Schwarf: Sure. I will add the test case.

networkit/cpp/graph/PrimMSF.cpp
```diff
@@ -0,0 +1,66 @@
...
+    }
+    // Weighted graph
+    forest = GraphTools::copyNodes(*G);
+    const index numberOfNodes = G->numberOfNodes();
+    std::vector<edgeweight> weights(numberOfNodes, infiniteWeight);
+    std::vector<node> parents(numberOfNodes, none);
+    std::vector<bool> visited(numberOfNodes, false);
+    std::priority_queue<std::pair<edgeweight, node>, std::vector<std::pair<edgeweight, node>>,
+                        std::greater<>>
+        minHeap;
```
angriman, line 32: A slightly more efficient priority queue that we use in NetworKit is the tlx::d_ary_heap, see for example https://github.com/networkit/networkit/blob/master/networkit/cpp/centrality/TopCloseness.cpp#L324

networkit/cpp/graph/PrimMSF.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    G->forNodes([&](node startNode) {
+        if (visited[startNode]) {
+            return;
+        }
+        minHeap.emplace(nullWeight, startNode);
+        weights[startNode] = nullWeight;
+        parents[startNode] = startNode;
+        while (!minHeap.empty()) {
+            const auto pair = minHeap.top();
```
angriman, line 42: You can use [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding) to simplify unpacking the pair:
```cpp
const auto [currentWeight, currentNode] = minHeap.top();
```

Schwarf: I used structured bindings before, but some builds were failing — see this commit:
https://github.com/networkit/networkit/pull/1318/commits/235430dacffe1606b7bcc896c89ede25a7a7da52
But I'll let myself be surprised with the d_ary_heap.

Schwarf: No, seems not to work with structured bindings:
image

networkit/cpp/graph/PrimMSF.cpp
Schwarf: Done

networkit/cpp/graph/PrimMSF.cpp
angriman: Ah it looks like an OpenMP limitation, let's not use it then.

angriman (approved):

Schwarf:
Hi @bernlu and @fabratu,

Do you think this PR can be merged, or is something missing?

PR #1324, opened 2025-06-02
[Do not merge] Proposal for adding node and edgeweight template parameters to Graph

angriman:
## Problem

The `NetworKit::Graph` class always uses 64-bit integers for node IDs, but [NetworKit applications](https://scholar.google.com/scholar?cites=2526066274193235127&as_sdt=2005&sciodt=0,5&hl=en) rarely use (if ever) graphs with more than 2^32 nodes. Using smaller node IDs (32-bit or less) significantly reduces the size of the graph and improves scalability.

Edge weights have a similar issue: `NetworKit::Graph` uses 64-bit floats for edge weights, while applications might only require 32-bit floats or integers.

## Proposed solution

Adding a `NodeType` and `EdgeWeightType` template parameters to NetworKit's graph class enables users to select the most appropriate type for node IDs and edge weights. I considered two options for implementing this idea:

### 1. [not recommended] Add template parameters to `NetworKit::Graph` and to all algorithms using `Graph`.
With this option, we'll keep the original `NetworKit::Graph` class, but we'll have to add the same template parameters to all NetworKit algorithms using it:

```cpp
// Graph.hpp

template <class NodeType, class EdgeWeightType>
class Graph { ... };

// MyAlgo.hpp

template <class NodeType, class EdgeWeightType>
class MyAlgo {
public:
  MyAlgo(const Graph<NodeType, EdgeWeightType>& G);
};
```

Alternatively, we could add a `GraphType` template parameter to all algorithms and access the `NodeType` and `EdgeWeightType` in `Graph`:
```cpp
// Graph.hpp

template <class NodeType, class EdgeWeightType>
class Graph {
public:
  typedef NodeType node_type;
  typedef EdgeWeightType edge_weight_type;
};

// MyAlgo.hpp

template <class GraphType>
class MyAlgo {
  using NodeType = typename GraphType::node_type;
  using EdgeWeightType = typename GraphType::edge_weight_type;

public:
  MyAlgo(const GraphType& G);
};
```

I think that none of those solutions are viable because:
1. they require significant changes to the code;
2. they break all external C++ code using `Graph` (after the change, `Graph` will require template parameters).

### 2. [proposal] Rename `Graph` to `DynamicGraph` and define `Graph` as an alias of `DynamicGraph<uint64_t, double>`.
This is the option implemented in this PR. I also included an example of how we can migrate an existing algorithm (the `NetworKit::FloydWarshall`) to the template graph, demonstrating feasibility.
```cpp
// Graph.hpp
template <class NodeType, class EdgeWeightType>
class DynamicGraph { ... };

using Graph = DynamicGraph<uint64_t, double>;

// MyAlgo.hpp

template <class GraphType>
class MyAlgo {
  using NodeType = typename GraphType::node_type;
  using EdgeWeightType = typename GraphType::edge_weight_type;

public:
  MyAlgo(const GraphType& G);
};
```
If we accept renaming `Graph`, this option has two main advantages compared to the previous one:
1. `Graph` and all algorithms will maintain the same behavior as before, no breakage of external code using `Graph`.
2. Incremental migration: since `Graph` maintains the same behavior as before, we can migrate algorithms gradually instead of migrating all of them in one change (see the `FloydWarshall` example).

Further, the idea of adding a `GraphType` template parameter to algorithms will allow keeping a single implementation and generalize it to different graph types. This should be useful to efforts trying to introduce new graph types like #1231 or #1246.

Please let me know if you have any suggestions or if you have other ideas I haven't considered.


fabratu:
We had also offline discussions about this topic. The rationale was basically the same as what is stated here, being more flexible with node ids and edge weights.

Also your proposal makes sense. A few (maybe not thorough thoughts):

- The change shifts NetworKit substantially towards a header only library and large header files. This reduces readability, but I am not aware of a technique for avoidance.
- From a design perspective we now have two namespaces/owners of node ids and edge weights. Code independent from the `Graph`-class currently uses fixed `node` and `edgeweight` via `Globals.hpp`. One way to deal with this would be to also make the datatypes in Globals also templates and provide alias names in order to not break existing code + convenience: `node`, `node32`, `node16`, `edgeweight`, `edgeweight32`, ... This would also provide a fallback, whenever a `GraphType` is not available. 
- Is there a way to avoid to add `using NodeType = ...` and `using EdgeWeightType = ...` at the beginning of the algorithms? It is just convenience, however this will likely lead to different parameter names in different classes, reducing readability further.

Likely, other modules/classes like `GraphTools`, `generators` or `Aux` need also be converted to templates in the end - however with the proposed solution this can be done in reasonable chunks.

angriman:
Thank you both for your feedback.

> The change shifts NetworKit substantially towards a header only library and large header files. This reduces readability, but I am not aware of a technique for avoidance.

That's right. For readability, we could adopt the suggestions from @Schwarf, but we can't avoid other issues like longer compilation time.

> Code independent from the `Graph`-class currently uses fixed `node` and `edgeweight` via `Globals.hpp`. One way to deal with this would be to also make the datatypes in Globals also templates and provide alias names in order to not break existing code.

I think that an easier solution is to template that code and add default template parameters to minimize changes in non-template algorithms using it.  For example, `BucketPQ.hpp` defines an `<int64_t, uint64_t>` priority queue. We could replace it with:
```cpp
// BucketPQ.hpp
template <class KeyType = int64_t, class ValueType = uint64_t>
class BucketPQ : public PrioQueue<KeyType, ValueType> {
    static_assert(std::is_signed<KeyType>() && std::is_integral<KeyType>());
   // ...
};

// SomeAlgo.hpp
// Instead of `BucketPQ prioQ;`
BucketPQ<> prioQ;
```
And then we gradually stop using the types declared in `Globals.hpp` as we template the classes using it.

Let me know if what you are proposing is simper than this, I might have misinterpreted it.

> Is there a way to avoid to add `using NodeType = ...` and `using EdgeWeightType = ...` at the beginning of the algorithms?

Yes, we can do something similar as done with `NodeIterators.hpp`. The disadvantage is that doing so we have to specify the template arguments each time for `GraphType` as well:

```cpp
template <template <class, class> class GraphType, class NodeType, class EdgeWeightType>
class FloydWarshall : public Algorithm {
public:
  // Before:
  // FloydWarshall(const GraphType &G);
  // Now it becomes:
  FloydWarshall(const GraphType<NodeType, EdgeWeightType> &G);
};

// Before:
// template <class GraphType>
// void FloydWarshall<GraphType>::run() {}
// Now it becomes:
template <template <class, class> class GraphType, class NodeType, class EdgeWeightType>
void FloydWarshall<GraphType<NodeType, EdgeWeightType>>::run() {}
```

To me this looks too cluttered. Adding the `using NodeType = ...` and `using EdgeWeightType = ...` at the beginning is redundant, but it significantly simplifies the code. I'm not aware of other solutions.

My question is how do we move from here? If we agree to proceed with this change, I think that we should clarify:
- How we migrate the C++ code: how we name the template parameters (e.g., we could shorten them as `NodeT` and `EdgeWeightT`), whether we prefer having the `using NodeType = ...` and `using EdgeWeightType = ...` or the more verbose `template <template <class, class> class GraphType, class NodeType, class EdgeWeightType>` or others if available, etc.
- Prepare a migration plan: after `Graph`, which algorithms/functions do we want to migrate first? Do we also want to migrate their Cython counterparts? For example, instead of a 64-bit graph, we could expose to the Cython code only a 32-bit graph.

Let me know what you think.

angriman:
@fabratu any updates on this? Are you still interested in a template graph class, or do you prefer keeping the original implementation? If it helps decide, we could run some benchmark to measure the speedup of a 32-bit graph compared to the current 64-bit implementation.

Schwarf:
I would be interested in helping move this proposal forward if there is still interest.

From May 14 to May 17, 2026, I expect to have a good amount of free time and could support a time-boxed feasibility spike around this. For example, I could help with experiments, refactoring, or migrating one or two representative algorithms/utilities to better understand the practical impact.

To make such a spike effective, it would be helpful to align beforehand on a coordinated agenda and a few concrete goals.

fabratu:
Thanks for pinging @angriman !

Certainly a benchmark or representative migration example would be helpful if the time slot of @Schwarf is still not occupied. I will try to create a more lengthy answer later this evening to define a meaningful demo for this.

What still worries me (but maybe this is too constructed) is a potential mismatch between datatypes from helper datastructures (relying on `Globals.hpp`) and the given graph (relying on templates).

angriman:
> What still worries me (but maybe this is too constructed) is a potential mismatch between datatypes from helper datastructures (relying on `Globals.hpp`) and the given graph (relying on templates).

Helper data structures such as `BucketPQ` will have to be migrated to templates before migrating the algorithms relying on them. We can avoid discrepancies in the transition phase by introducing a templated implementation and keeping the original helper data structure name as an alias to the old types from `Globals.hpp`
For example, `BucketPQ` can be migrated as follows:

```cpp
// BucketPQ.hpp

#include <networkit/Globals.hpp>

namespace Aux {

// New template bucket priority queue.
template <class KeyType, class ValueType>
class BucketPriorityQueue : public PrioQueue<KeyType, ValueType> {
    static_assert(std::is_signed<KeyType>() && std::is_integral<KeyType>());
   // ...
};

// Alias for the original BucketPQ. This is necessary to maintain compatibility with non-template algorithms.
using BucketPQ = BucketPriorityQueue <int64_t, NetworKit::index>;

} // namespace Aux
```

Then, non-template algorithms can keep using the original `BucketPQ` with no changes, while template algorithms can start using `BucketPriorityQueue`:
```cpp
// SomeTemplateAlgo.hpp
// Instead of `BucketPQ prioQ;`, use the new template bucket priority queue.
BucketPriorityQueue<int, float> pq;
```

fabratu:
Right, makes sense. 

After thinking again about this PR and other discussions we had in the past, more flexible node id and edge weight types (lets stay with those for now) sound reasonable and give one more tool for the scalability of algorithms.
 
Based on the input so far and given the amount of work this design change would create, I propose the following approach:

Produce a minimal benchmark with an appropriate algorithm to back the intuition and show practical feasibility. To this end, the implementation of Floyd-Warshall is a reasonable choice, since it combines both regular and irregular access patterns and contains parallel regions (and the work is already done). The goal of the benchmark is to measure running time; an advantage in memory consumption is obvious. To get reasonable differences between 64bit and 32bit widths, likely the examined networks need to be of a certain size; likely different sizes are needed.

The benchmark makes it at least appealing to modify one reader/generator to support variable width. Another option would be a width converter function. I would opt for the reader/generator, since it should give us flexible benchmarking results more easily.

The benchmark can also give a preview of how much the compilation time increases.

A potential migration plan needs more consideration; however, a general order can look like this:

1. Graph Datastructure + all dependent files/classes
2. Core Helper Datastructures like `Partition`, `BucketPQ`, ...
3. Generators and IO (to enable testing of algorithms)
4. Algorithms

Concerning Python: While the gains are also desirable, the interface in its current state would benefit the most from other time investments (like fewer copies). In addition, the migration is extensive for only C++ already. Therefore, concerning Cython, my suggestion is to focus at first (as done in the example conversion) on exposing a wrapper with 64bit/64bit.

Something we have not yet covered: Should all migrated algorithms be tested with a given combination of datasizes? If so, what would be an efficient way to modify the unit tests?

fabratu:
Concerning coding style:

- The proposed variant with `DynamicGraph` is likely the best option. At least I don't know another one, and it gives the most flexibility concerning a potential migration.
- In general, the approach should be that code, which is not aware of the changes, should remain unchanged until migrated itself. As a result, we should rename the templated classes whenever a class is used downstream (like `BucketPriorityQueue`). Coming up with reasonable names might be a challenge whenever descriptive class names are already used. Using `Dynamic` as a prefix is a good starting point until we have better suggestions. At some point NetworKit might introduce a clear cut, removing the aliases.
- I'm in favor of splitting declaration and implementation; however, `.inl` or `.tpl` might be misleading since these files normally include both templated and untemplated functions.
- I'm also in favor of shortening the template parameter names due to resemblance to literal datatypes like `int64_t`. So using `NodeT` or `Node_T` instead of `NodeType`, while the first variant is more in line with the current coding style in NetworKit.
- I don't see a real advantage in using `template <template <class, class> class GraphType, class NodeType, class EdgeWeightType>` instead of the `using ...` pattern.

Schwarf:
Given the scope of this refactoring, would there be interest in a short direct synchronization?

The PR comments should of course remain the source of truth, but I think a brief synchronous discussion could help align on the intended direction, priorities, and concrete next steps. It could also help avoid duplicated work if multiple people are interested in contributing to this effort.

For example, we could use a temporary Discord server/thread or any other lightweight channel preferred by the maintainers.

Schwarf:
I rebased the prototype on current `master` and added a small Floyd-Warshall benchmark to get an initial data point for the proposed `DynamicGraph<NodeT, EdgeWeightT>` direction.

This benchmark compares different node-ID widths within the rebased templated implementation. It does not yet compare the rebased `DynamicGraph<uint64_t, double>` path against the original non-templated `Graph` implementation on `master`.
The benchmark compares:

- `DynamicGraph<uint64_t, double>` as the current/default-style baseline
- `DynamicGraph<uint32_t, double>` as the smaller node-ID variant

I used two deterministic graph families:

- complete graphs (weighted)
- path graphs (weighted)

Graph construction happens outside the measured benchmark loop. The measured part is the normal end-to-end `FloydWarshall<GraphType>` execution path, i.e. construction of the algorithm object, `run()`, and one consumed distance query.

I ran repeated Google Benchmark measurements with 10 repetitions and random interleaving. The plots below normalize the CPU-time mean against the corresponding `uint64_t` baseline for each graph family and input size. Values below `1.0` are faster than the `uint64_t` baseline. The lower panel shows the coefficient of variation.

### `OMP_NUM_THREADS=1`

Floyd-Warshall OMP_NUM_THREADS=1 benchmark

### Default threading

Floyd-Warshall default-threading normalized benchmark

Please note: markers are slightly horizontally offset for readability; the actual benchmark sizes are 16, 32, 64, 128, 256,  and 512 nodes.
 
Across both graph families, `DynamicGraph<uint32_t, double>` is consistently faster than the `uint64_t` baseline. The average speedup is roughly **12–15%**, both with `OMP_NUM_THREADS=1` and with the default threading configuration. This suggests that the observed improvement is not limited to a particular threading configuration.

The 512-node data point suggests that the benefit of `uint32_t` node IDs may increase with graph size.

Raw JSON benchmark output is available here (both files also include results for `int32_t`):

- [floyd_warshall_omp1_repeated.json](https://github.com/user-attachments/files/27756478/floyd_warshall_omp1_repeated.json)
- [floyd_warshall_all_threads_repeated.json](https://github.com/user-attachments/files/27756489/floyd_warshall_all_threads_repeated.json)

My takeaway: the repeated measurements consistently support the intuition that, within the templated mplementation, a 32-bit node-ID graph can provide a measurable runtime benefit for the migrated Floyd-Warshall implementation. Since Floyd-Warshall is limited to relatively small graphs by its `O(n^3)` complexity, the next useful step would be to validate the effect on algorithms that scale to larger node counts and are more dominated by graph traversal and adjacency storage. A separate comparison against the original non-templated implementation would also be useful to check whether the templating/refactoring itself introduces any overhead for the default `uint64_t,double` case (I will do that next).

The benchmark implementation used for these measurements is available here: https://github.com/networkit/networkit/commit/8c26782462a7309ac7b7130f301cb9cf67117917.

Schwarf:
As promised, I also compared the rebased templated implementation against the original non-templated Floyd-Warshall implementation. To keep the comparison focused, I used only a complete weighted graph with `n = 512`.

Benchmark setup:

- benchmark: `BM_FloydWarshallCompleteGraph`
- graph: complete weighted graph
- graph size: `n = 512`
- repetitions: `10`
- Google Benchmark random interleaving enabled: `--benchmark_enable_random_interleaving=true`
- metric compared: mean CPU time
- two OpenMP settings:
  - `OMP_NUM_THREADS=1`
  - `OMP_NUM_THREADS=24`

For the original `Graph` implementation, I measured:

| Implementation | `OMP_NUM_THREADS` | Mean CPU time | CV |
|---|---:|---:|---:|
| original `Graph` | 1 | ~496.6 ms | ~0.42% |
| original `Graph` | 24 | ~70.7 ms | ~1.57% |

For the migrated `DynamicGraph<uint64_t, double>` path, I measured:

| Implementation | `OMP_NUM_THREADS` | Mean CPU time | CV |
|---|---:|---:|---:|
| `DynamicGraph<uint64_t, double>` | 1 | ~496.0 ms | ~1.25% |
| `DynamicGraph<uint64_t, double>` | 24 | ~499.2 ms | ~1.05% |

The single-threaded numbers are essentially identical, which is encouraging: the default-width templated path does not appear to introduce a meaningful single-threaded regression in this benchmark.

However, the original implementation gets a clear OpenMP speedup, from about **496.6 ms** to about **70.7 ms**, roughly a **7× improvement**. The migrated `DynamicGraph<uint64_t, double>` path does not reproduce this speedup; the `OMP_NUM_THREADS=24` result remains close to the single-threaded runtime.

My interpretation is that the current migrated Floyd-Warshall path is likely not using the original parallel execution path, or that parallelization is bypassed/lost in the templated version. I cannot rule out that I made a mistake in the benchmark setup or in the migration, so this should be checked carefully.

That said, I think this issue should be resolved before proceeding much further with the `DynamicGraph` direction. If the templated graph approach cannot preserve the existing parallel performance characteristics, then in my opinion the introduction of `DynamicGraph` would not be worth it, even if smaller node IDs provide measurable single-threaded benefits.

angriman:
Thank you both for the great feedback! Special thanks to @Schwarf for taking the time to run those benchmarks. To me it looks suspicious that the template graph performs worse on 24 threads, we're running the exact same code as the original graph, parallelism should not be affected at all.

One thing to note is that the graphs used for these benchmarks are extremely small, so we're likely measuring thread synchronization overhead rather than the actual performance of the new data structure.
Further, such small graphs likely fit entirely in the CPU cache. It would be interesting to see how this performs on larger graphs that exceed the cache capacity. Since moving to 32-bit halves the memory footprint, we should see a much more significant speedup in those cases due to fewer cache misses.

I'm perfectly fine with syncing up about this topic via chat or a call so that we can define a more detailed migration plan. Let me know what works best for you all.

Schwarf:
Thanks for the feedback!

I would also be happy to sync up. Today, tomorrow, and Saturday almost any time between 09:00 and 21:00 CEST works for me. 
Until then, I will try to gather more insights regarding the benchmark results.

angriman:
I created this poll to find some suitable time slot: https://app.rallly.co/poll/2itFniBWO68y
If you have a Gmail account, could you please add it to the poll so I can send you an invite?

Schwarf:
Small update regarding the benchmark results: I think I found the reason for the initially suspicious loss of parallel speedup.

Since the refactored Floyd-Warshall implementation is header-only/template-based, the OpenMP pragmas are compiled in the translation unit where the template is instantiated. For the benchmark this means that `networkit_benchmarks` itself needs to be compiled/linked with OpenMP.

After adding `OpenMP::OpenMP_CXX` to the benchmark target I get:

| Graph size | Original `Graph` | `DynamicGraph<uint64_t, double>` | Relative difference |
|---:|---:|---:|---:|
| 512 | ~71.7 ms | ~72.5 ms | ~1.1% slower |
| 2048 | ~5.09 s | ~5.05 s | ~0.8% faster |

As a sanity check, I removed the CMake change again, deleted the build directory, rebuilt, and reran the benchmark. The slowdown reappeared: for `n = 512`, `DynamicGraph<uint64_t, double>` went from roughly 72 ms back to roughly 495 ms.

So I think the earlier benchmark result was not caused by the new graph data structure itself, but by the benchmark target not receiving OpenMP correctly after the algorithm became header-only. 😅

angriman:
> I created this poll to find some suitable time slot: https://app.rallly.co/poll/2itFniBWO68y If you have a Gmail account, could you please add it to the poll so I can send you an invite?

@fabratu do you have time today or this weekend to discuss about this?

Schwarf:
I adapted `LeftRightPlanarityCheck` to work with the new templated `DynamicGraph` types.

To exercise this on file-backed graphs, I also made a minimal/interim adjustment to `METISGraphReader`: it is now templated and no longer derives from `GraphReader`, because the base interface fixes the return type to the default `Graph`. 

I also ran benchmarks on several file-backed input graphs and graph type combinations. For these results, I focused on `OMP_NUM_THREADS=1`, because `LeftRightPlanarityCheck` does not benefit from OpenMP.

What I did for the benchmark experiment:

- tested three input graphs:
- benchmarked the following graph type combinations:
  - `uint32_t, float`
  - `uint32_t, double`
  - `uint64_t, float`
  - `uint64_t, double`
 
left_right_omp1_4elt

left_right_omp1_airfoil

left_right_omp1_hep_th



The results suggest that using `uint32_t` instead of `uint64_t` can provide a small but consistent runtime improvement for this algorithm on these graphs. In contrast, the `float` vs. `double` choice does not show a meaningful algorithmic effect, which is expected: `LeftRightPlanarityCheck` does not use edge weights, so the edge-weight type should not matter except for incidental effects such as type layout, cache behavior, or measurement noise.

While adapting `METISGraphReader` and `LeftRightPlanarityCheck`, a few minor design and migration questions came up on my side. Maybe we can discuss them during our sync.

fabratu:
I am currently travelling and will be back and available from Sunday ~6p.m. 

Of course, feel free to sync/meet. If more input is needed, I can also participate next week. 

Concerning graph sizes, you likely need to use larger instances still. For `n=2048`, if all nodes are in the cache and assuming 64bit, you have 128 KB of data. Given a moderate modern CPU, the L1/L2 caches are 384KB/3MB (example data from a Ryzen 5600H), so likely there is no cache miss.

Schwarf:
Hi @fabratu,

thank you for your feedback. I think your point is correct if we only look at storing the node IDs themselves. 
However, for the Floyd-Warshall benchmark the relevant working set is not just the node storage of the graph. The implementation allocates several `n^2` matrices. In particular, the distance matrix stores `EdgeWeightType`, and the path reconstruction matrix stores `NodeType`.

For `n = 512`, the distance matrix alone already has size:

- `512 * 512 * 8 bytes ≈ 2 MiB` for `double`
- `512 * 512 * 4 bytes ≈ 1 MiB` for `float`

In addition, the path reconstruction matrix has the same `n^2` shape and stores the node type:

- `512 * 512 * 8 bytes ≈ 2 MiB` for `uint64_t`
- `512 * 512 * 4 bytes ≈ 1 MiB` for `uint32`

So for this algorithm, even `n = 512` should already be large enough to put pressure on L1/L2 cache, at least on my machine. 

Independently of the exact cache explanation, I reran the benchmark with all four relevant type combinations:
The plot below shows the normalized mean runtime relative to `DynamicGraph<uint64_t, double>` as baseline. The lower panel shows the coefficient of variation (CV) across 10 Google Benchmark repetitions.

floyd_warshall_complete_omp1_4_combinations_normalized_with_cv

Interestingly, in this benchmark the effect of switching from `uint64_t` to `uint32_t` is larger than the effect of switching from `double` to `float`.

Schwarf:
Hi @fabratu, hi @angriman,

after our sync yesterday I got a better understanding of the benchmark goal.

I first looked into `APSP`, but on my machine this is still limited to a few ten thousand nodes because of the `O(n^2)` distance matrix. Therefore, I also tried two traversal/search-style benchmarks that scale to larger graphs: breadth-first search and Dijkstra. Both are single-threaded in this benchmark setup.

I used graphs with `2^18` and `2^19` nodes, resulting in roughly one million and two million edges, respectively. The plots compare `DynamicGraph<uint64_t, double>` against `DynamicGraph<uint32_t, float>`, normalized to the `uint64_t, double` baseline.

bfsbase_edges_1thread_separate_normalized_with_cv

dijkstrabase_edges_1thread_normalized_with_cv

For BFS, I see a large performance improvement of roughly **35–45% lower runtime** with the smaller type combination, although the measurement noise is high. For Dijkstra, however, there is essentially no performance improvement.

My current interpretation (after our discussion yesterday) is that BFS benefits more from cache locality, because it is dominated by scanning adjacency data. Dijkstra, on the other hand, has additional costs from priority-queue operations, distance updates, and more irregular memory access.

Please feel free to comment. I am off for today.

angriman:
Thank you @Schwarf for the additional benchmarks. I think that the results are strong enough to justify this migration. I started #1415 to summarize the migration plan and track progress.

hmeyerhenke:
Sounds like a big chunk. We should relieve Fabian of this burden and get it done when Mikhail is in full command.

One question: in the past, highly templated code was harder to deal with due to cryptic compiler messages. Is this better now?

fabratu:
Also, thanks from my side for the benchmarking - the results are informative.

Just for the record: My initial remark about graph sizes was based on the idea that the innermost loop (or thread) normally looks only at one line of the adjacency matrix; therefore, the complexity for memory fetching in this part is `O(n)` (which could either be IDs or weights, depending on the use case).

I would second a justification for modification. This gives both the algorithm engineer and the end user an additional tool.

Since resource planning is also part of the process: As @hmeyerhenke mentioned, likely another maintainer is joining. Implicitly mentioned, I (might) be shifting my time/focus in the upcoming months. As a result, my plan for the next time is to cover core maintenance like bugs and releases (or finish my own algorithmic contributions finally :D ) mostly. For the migration, I likely can only offer reviews.

angriman:
> One question: in the past, highly templated code was harder to deal with due to cryptic compiler messages. Is this better now?

Compiler diagnostics have improved quite a bit. While template-heavy code can still produce verbose errors, here we’re only adding two template parameters and I haven’t encountered such issues while working on this PR. This should be further mitigated if we introduce concepts as suggested by @Schwarf in [#1416](https://github.com/networkit/networkit/pull/1416#discussion_r3298032289).

angriman:
Closing this as the idea was merged in #1416

PR #1325, opened 2025-06-08
Replace const std::vector<T>& with std::span<const T> in functions input parameters.

angriman:
This PR replaces `const std::vector<T>&` with `std::span<T>` in function parameters where a non-owning view o a contiguous sequence is needed.

[`std::span`](https://www.cppreference.com/w/cpp/container/span.html) is a lightweight non-owning object representing a view on a contiguous sequence of elements. Compared to `std::vector`, it is more versatile and expresses the intent of passing a non-owning view on a contiguous sequence of elements more clearly.

The changes are backward-compatible with existing code that passes `std::vector`.


angriman:
Thank you, I've dropped the last commit.

Issue #1331, opened 2025-07-07
Optimize for Arrow in-memory format


adsharma:
As seen in #1328 there is a substantial difference in performance in using a graph stored using C++ vectors (~2 seconds) vs when benchmarking end to end from python + pandas (~9 seconds). [Benchmark script](https://github.com/adsharma/leiden-communities-openmp/blob/python2/test/benchmark_nk.py)

A lot of data scientists use python and arrow to read graphs into memory using a standardized columnar format such as arrow.

Polars (implemented in rust) uses arrow natively. Pandas optionally since 2.0:

```
df = pd.DataFrame({"a": [1, 2, 3]}, dtype_backend="pyarrow")
```

Optimizing for this use case could unlock even more performance for networkit users.

adsharma:
Just to clarify the numbers being discussed here:

* 9 secs - python wrapper around GVE - Leiden
* 9 secs - networkit ParallelLiden end-to-end benchmark

It's not surprising that the two numbers are so close. Previous discussion about these numbers: https://github.com/nodef/leiden-communities-openmp.py/pull/1

* ~2 sec number comes from running [C++ directly](https://github.com/nodef/leiden-communities-openmp.py/blob/main/main.cxx)

Breakdown of the rest of the time:

* 6 secs in `create_graph_from_arrays` [here](https://github.com/adsharma/leiden-communities-openmp/blob/python2/src/python_bindings.cpp#L30)
* Rest in [copying out the result ](https://github.com/adsharma/leiden-communities-openmp/blob/python2/src/python_bindings.cpp#L183)

The copy in/out overhead can be optimized if we use `py_array_t` -> `std::span` instead of `std::vector`. The pybind11 method I've used here and cython method used in networkit should be comparable.

adsharma:
Example code. The idea is to support read-only graphs with zero copy efficiently and read-write graphs backed by vectors in one Graph class.

[spans.cc.txt](https://github.com/user-attachments/files/21130373/spans.cc.txt)

PR #1335, opened 2025-07-17
Introduce immutable/mutable Graphs

adsharma:
`Graph`: immutable graph. Its internals can be owned by another module
`GraphW`: mutable graph. Allows you to call `addEdge()` and other mutable methods

This refactor shouldn't change any semantics in networkit. But it enables future zero-copy optimizations discussed in #1331


adsharma:
> Anyway, having a smaller user base than Python does not justify breaking user's C++ code.

Thanks for the feedback @angriman!
Agreed. We shouldn't gratuitously break people's code if it's avoidable. 

But immutability by default is a powerful principle. It feels natural to write:

```
 TEST_F(GraphToolsGTest, testIsBipartiteEmptyGraph) {
    Graph graph;
    EXPECT_TRUE(GraphTools::isBipartite(graph));
 }
```

as opposed to `GraphR graph`. This way, we make people think twice when they're mutating the graph and use the non-default `GraphW` construct.

adsharma:
I'm in the process of fixing the build failures in tests. It's onerous even with all the excellent tools available these days.

How about we take a look at the resulting commit and look at the cost of `Graph -> GraphR` rename vs `Graph -> GraphW` rename. Specifically how many renames would happen in the existing test code.

I suspect the work to separate the read-only and read-write cases in tests is worth it, regardless of how we decide to name things.

fabratu:
> > Anyway, having a smaller user base than Python does not justify breaking user's C++ code.
> 
> Thanks for the feedback @angriman! Agreed. We shouldn't gratuitously break people's code if it's avoidable.
> 
> But immutability by default is a powerful principle. It feels natural to write:
> 
> ```
>  TEST_F(GraphToolsGTest, testIsBipartiteEmptyGraph) {
>     Graph graph;
>     EXPECT_TRUE(GraphTools::isBipartite(graph));
>  }
> ```
> 
> as opposed to `GraphR graph`. This way, we make people think twice when they're mutating the graph and use the non-default `GraphW` construct.

While I see the convenience of handling `Graph` as a synonym for a immutable datastructure, such fundamental changes should be done carefully. As pointed out by angriman, there are use-cases for NetworKit as C++ library. We don't have exact numbers, however there are several projects using the library API.

Another argument is the opt-in approach of converting the code base towards this new API. In this sense we could migrate functions / classes, which only need an immutable graph to explicitly change to a new name `GraphR`. From a policy perspective, this can be done via deprecating the old API and finally removing the old API after two releases (usually done in the next major release).

I also looked more into detail of the PR changes. Separating read-only from mutable functions in graph helps in making the datastructure more readable in general (apart from the intended goal of avoiding copies). @adsharma Can you already provide an example how one would be able to use the new immutable graph order to avoid copies? I imagine for C++ -> C++ it is something similar to what the maps of [Eigen](https://libeigen.gitlab.io/docs/group__TutorialMapClass.html) provide. However, we normally use pointer/references anyway. How would this look like in Cython/Python? If I got this correctly, the major use case would be conversion / views from or to pyarrow and pandas?

In addition we should also keep in mind #1324. I think its valid approach to implement edge id/weight width via template parameters and use different classes for separating immutable/mutable graphs. One could also implement the mutability as a template parameter (for uniformity). If I am not missing something, this approach would lead to a lot of functions labeled with `enable_if` or more modern (like `requires`), increasing complexity of the graph class. Also, it is unlikely imho, that a function or class both supports mutable and immutable graphs. In any case, the user / developer has to cope with increased complexity when using the graph class. I would be interested, what you guys think. 

```
template <class NodeType, class EdgeWeightType, bool is_mutable>
class DynamicGraph final {
...
template <typename M = is_mutable, typename = std::enable_if_t<M>>
    void addEdge(/* parameters */) {
    ...
}
...
Graph = DynamicGraph<uint64_t, double, false>
GraphW = DynamicGraph<uint64_t, double, true>
```

adsharma:
> Can you already provide an example how one would be able to use the new immutable graph order to avoid copies

Basic ideas [here](https://github.com/adsharma/supercsr/blob/4d139047b7ca4d920a4f0cf010e6aafda7a30608/supercsr.h#L94)

In this example `graph` is backed by memory externally owned. Example: a db with graph + pyarrow capability such as DuckDB or KuzuDB. 

After `constructOptimized` `SuperCSR::col_idx` is pointing to that externally owned memory. There are a few `std::vector<int>` member variables which are offsets into  `SuperCSR::col_idx`. They can be 32 bit offsets for smaller graphs.

The idea is that `GraphW` will continue to use the current data structures to hold the graph to support mutability, sorting/indexing edges. And `Graph` could switch to a new `std::span` based one. It could be `SuperCSR` or something else that performs better.

`Graph` and `GraphW` remain API compatible. It should be possible to pass `GraphW` to a function that expects `Graph`. Algorithms use `Graph` when possible.

While this design works at the C++ level, it doesn't work with Cython. If we pass `GraphW` to a function expecting `Graph`, I get type errors. Looking for workarounds/suggestions here.

Naming, immediate vs gradual conversion: I've explained the rationale for the choices here, but I'll accept whatever you decide.

adsharma:
Some data here on what happens after removing gcc/clang-10 builds from CI:

https://github.com/adsharma/networkit/actions/runs/16690596913/job/47247768965?pr=3

Problems to solve:

* GeneratorsGTest.testDynamicForestFireGenerator times out
* Cython tests that can't handle Graph -> GraphW [substitution](https://en.wikipedia.org/wiki/Liskov_substitution_principle)

adsharma:
I got a chance to look into this again. Summary: 5x memory savings and 5x CPU savings possible.

Here are my experiments:

Problem: run page rank and parallel Leiden on a 0.9B node graph on Google cloud. The graph is stored in a 5GB duckdb file.

* n2-standard-16 (64GB) machine ran out of memory.
* n2-standard-32 (128GB) could run the algorithms, but took 440 secs to construct the graph from db and 40 secs to run the algo.

After implementing Graph (read-only) on top of [Arrow CSR](https://github.com/adsharma/networkit/blob/read_write_graph3/include/networkit/graph/Graph.hpp#L743-L746) arrays:

* page rank ran in 16GB
* parallel Leiden is not working yet. Need to debug memory corruption/double-free etc.

It could be further optimized by using `uint32` if the graph has less than 4B nodes and even more via `SuperCSR`.

For now, I got what I needed. I'll spend some time to get Parallel Leiden working without memory corruption. But I'm not sure if I have time to get it all cleaned up for a review/potential merge.

Some notes if someone wants to explore this in the future:

*  `ParallelLeidenView` (code in the branch above) works on the idea of `CoarseningGraphView`, where we create a view on top of a read-only graph, by maintaining a mapping instead of creating new `GraphW` objects.

* Parallel iteration interfaces need to be made virtual functions. This is why `GraphW` is broken in the linked branch. This will add some cost, but likely a lot smaller than the 5x savings above.

* Graph DBs need to support efficient creation of CSR arrow memory format. I'm not aware of any. Had to build it in python.

PR #1336, opened 2025-07-18
Add unit tests for Aux::Spinlock.

Schwarf:
## Summary

This PR adds a test suite for the internal `Aux::Spinlock` class. No production code was changed.

### DIFF
--- a/networkit/cpp/auxiliary/test/SpinLockGTest.cpp
+++ b/networkit/cpp/auxiliary/test/SpinLockGTest.cpp
@@ -1,0 +1,52 @@
+/*
+ * SpinLockGTest.cpp
+ *
+ *  Created on: 18.07.2025
+ *      Author: Andreas Scharf (andreas.b.scharf@gmail.com)
+ */
+
+#include <thread>
+#include <gtest/gtest.h>
+#include <networkit/auxiliary/SpinLock.hpp>
+
+class SpinlockGTest : public testing::Test {};
+
+namespace NetworKit {
+
+TEST_F(SpinlockGTest, testSpinLockSingleThread) {
+    Aux::Spinlock lock;
+    EXPECT_NO_THROW({
+        lock.lock();
+        lock.unlock();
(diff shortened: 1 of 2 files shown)


networkit/cpp/auxiliary/test/SpinLockGTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+            ++counter;
+            lock.unlock();
+        }
+    };
+
+    std::vector<std::thread> threads;
+    threads.reserve(numberOfThreads);
+    for (unsigned int i = 0; i < numberOfThreads; ++i) {
+        threads.emplace_back(worker);
+    }
```
angriman, line 43:
```suggestion
    std::vector<std::thread> threads(numberOfThreads);
    std::ranges::iota(threads, 0);
```

networkit/cpp/auxiliary/test/SpinLockGTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    for (auto &t : threads) {
```
angriman, line 45:
```suggestion
    for (std::thread &thread : threads) {
```

angriman (changes requested):

networkit/cpp/auxiliary/test/SpinLockGTest.cpp
angriman: If the `ranges` version is not available in C++20 you can always use `std::iota`, https://en.cppreference.com/w/cpp/algorithm/iota.html

angriman: Uh my bad you are not creating a vector of integers. Sorry my comment doesn't make sense, you can ignore it.

angriman (dismissed):

fabratu:
Merging this now, since the original approval is not recent and the rebase is only minor.

PR #1338, opened 2025-07-26
Fix crash when writing empty graph in NetworkitBinaryWriter

Schwarf:
This PR fixes a crash when writing an empty graph (numberOfNodes() == 0) using NetworkitBinaryWriter.
Closes [#909](https://github.com/networkit/networkit/issues/909).

## Root Cause
When serializing an empty graph, the internal logic attempted to compute offset values based on the number of chunks. However, if nodes == 0, then chunks == 0, and expressions like (chunks - 1) underflowed due to unsigned arithmetic. This led to corrupted headers and potential segmentation faults during read/write operations.

## Fix
This PR addresses the issue by:
- Zero-initializing all uint64_t fields of the header using brace initialization
- Explicitly handling the empty graph case at the beginning of NetworkitBinaryWriter::writeData():

### DIFF
--- a/include/networkit/io/NetworkitBinaryWriter.hpp
+++ b/include/networkit/io/NetworkitBinaryWriter.hpp
@@ -57,11 +57,11 @@ public:
     std::vector<uint8_t> writeToBuffer(const Graph &G);
 
 private:
+    static constexpr const char *FILE_FORMAT = "nkbg003";
     count chunks;
     NetworkitBinaryWeights weightsType;
     NetworkitBinaryEdgeIDs edgeIndex;
     bool preserveEdgeIndex;
-
     template <class T>
     void writeData(T &outStream, const Graph &G);
 };
--- a/networkit/cpp/io/NetworkitBinaryWriter.cpp
+++ b/networkit/cpp/io/NetworkitBinaryWriter.cpp
@@ -161,6 +161,7 @@ void NetworkitBinaryWriter::writeData(T &outStream, const Graph &G) {
         outStream.write(reinterpret_cast<char *>(&zero), sizeof(zero));
         return;
     }
+
     if (nodes < chunks) {
         chunks = nodes;
         INFO("reducing chunks to ", chunks, " chunks");
--- a/networkit/cpp/io/NetworkitBinaryWriter.cpp
+++ b/networkit/cpp/io/NetworkitBinaryWriter.cpp
@@ -154,12 +154,11 @@ void NetworkitBinaryWriter::writeData(T &outStream, const Graph &G) {
 
     count nodes = G.numberOfNodes();
     if (nodes == 0) {
-        // build header.features etc. as you do now…
         strncpy(header.magic, "nkbg003", 8);
         setFeatures();
         writeHeader();
         uint64_t zero = 0;
-        outStream.write(reinterpret_cast<char *>(&zero), sizeof(zero)); // adjListSize
+        outStream.write(reinterpret_cast<char *>(&zero), sizeof(zero));
         return;
     }
     if (nodes < chunks) {
--- a/networkit/cpp/io/NetworkitBinaryWriter.cpp
+++ b/networkit/cpp/io/NetworkitBinaryWriter.cpp
@@ -154,7 +154,7 @@ void NetworkitBinaryWriter::writeData(T &outStream, const Graph &G) {
 
     count nodes = G.numberOfNodes();
     if (nodes == 0) {
-        strncpy(header.magic, "nkbg003", 8);
+        strncpy(header.magic, FILE_FORMAT, 8);
         setFeatures();
         writeHeader();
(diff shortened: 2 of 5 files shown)


networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1422,4 +1422,34 @@ TEST_F(IOGTest, testMatrixMarketReaderIntegerWeights) {
...
+    const std::filesystem::path path = "output/empty_graph.nkb";
+    writer.write(graph, path.string());
+
+    NetworkitBinaryReader reader;
+    Graph graph_read = reader.read(path.string());
+    EXPECT_EQ(graph.numberOfNodes(), graph_read.numberOfNodes());
+    EXPECT_EQ(graph.numberOfEdges(), graph_read.numberOfEdges());
+    EXPECT_EQ(graph.isWeighted(), graph_read.isWeighted());
+    EXPECT_EQ(graph.isDirected(), graph_read.isDirected());
+    EXPECT_EQ(graph.hasEdgeIds(), graph_read.hasEdgeIds());
```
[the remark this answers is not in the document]
Schwarf: Done

networkit/cpp/io/NetworkitBinaryWriter.cpp
```diff
@@ -153,6 +153,14 @@ void NetworkitBinaryWriter::writeData(T &outStream, const Graph &G) {
     };
 
     count nodes = G.numberOfNodes();
+    if (nodes == 0) {
+        strncpy(header.magic, "nkbg003", 8);
```
angriman, line 157: Let's define this string as a static member variable.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1422,4 +1423,32 @@ TEST_F(IOGTest, testMatrixMarketReaderIntegerWeights) {
...
+           && arg.numberOfEdges() == expected.numberOfEdges()
+           && arg.isWeighted() == expected.isWeighted() && arg.isDirected() == expected.isDirected()
+           && arg.hasEdgeIds() == expected.hasEdgeIds();
+}
+
+TEST_F(IOGTest, testNetworkitBinaryWriteReadEmptyGraph) {
+    Graph graph(0, false, true);
+    NetworkitBinaryWriter writer;
+    const std::filesystem::path path = "output/empty_graph.nkb";
+    writer.write(graph, path.string());
```
angriman, line 1437:
```suggestion
    const std::filesystem::path path = "output/empty_graph.nkb";
    NetworkitBinaryWriter{}.write(graph, path.string());
```
Same below.

networkit/cpp/io/test/IOGTest.cpp
```diff
@@ -1422,4 +1423,32 @@ TEST_F(IOGTest, testMatrixMarketReaderIntegerWeights) {
...
+
+    NetworkitBinaryReader reader;
+    Graph graph_read = reader.read(path.string());
+    EXPECT_THAT(graph_read, GraphFeaturesEqual(graph));
```
angriman, line 1441:
```suggestion
    const Graph graph_read = NetworkitBinaryReader{}.read(path.string());
    EXPECT_THAT(graph_read, GraphFeaturesEqual(graph));
```
Same below.

networkit/cpp/io/NetworkitBinaryWriter.cpp
Schwarf: Thanks for the suggestion!

I'd prefer to postpone this change for now. If I start replacing the "nkbg003" constant in this PR, we'd also need to consider "nkbg002" used in NetworkBinaryReader, which opens up a broader discussion about where and how to define and group these constants.

Since this isn't directly related to the bug fix at hand, I suggest we keep the current literal here for now. If you’d like to raise a separate issue with a concrete proposal for organizing these constants, I’d be happy to take it up in a follow-up PR.

include/networkit/io/NetworkitBinaryWriter.hpp
```diff
@@ -57,11 +57,11 @@ class NetworkitBinaryWriter final : public GraphWriter {
     std::vector<uint8_t> writeToBuffer(const Graph &G);
 
 private:
+    static constexpr const char *FILE_FORMAT = "nkbg003";
```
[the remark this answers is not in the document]
Schwarf: Hmm, I don't think so.
`constexpr const char*` -->  pointer itself is constant and points to a `const char`.
`constexpr char*` → pointer is constant, but points to non-const `char`.
If I drop the explicit `const`, the gcc compilers emit a warning because, see here:
[godbolt](https://godbolt.org/z/P9hTesvWf)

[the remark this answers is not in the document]
angriman: Ah good point, then let's keep it as it is.

PR #1347, opened 2025-09-04
CI: Update several ghas, compiler version, fix errornuous runs.

fabratu:


domischweisgut (approved):
I am not the greatest expert on these CI matters but everything seems to be consistent.

PR #1360, opened 2025-10-16
Add unit tests for class EdgeScoreAsWeight.

Schwarf:
In this PR the missing unit tests for the class `EdgeScoresAsWeight` are added to improve coverage. 

I’m trying to understand the purpose of this class and other classes in the `edgescores` folder. Some expose APIs without implementations (e.g., double `EdgeScoreBlender::score(node, node)`), while others—like `EdgeScoresAsWeight`—lack API documentation about constraints; for example, `EdgeScoresAsWeight::calculate()` does not work with directed graph inputs.

I’d appreciate clarification before I proceed to write tests for these classes.

### DIFF
--- a/networkit/cpp/edgescores/EdgeScoreAsWeight.cpp
+++ b/networkit/cpp/edgescores/EdgeScoreAsWeight.cpp
@@ -15,10 +15,10 @@ EdgeScoreAsWeight::EdgeScoreAsWeight(const Graph &G, const std::vector<double> &
     : G(&G), score(&score), squared(squared), offset(offset), factor(factor) {}
 
 Graph EdgeScoreAsWeight::calculate() {
-    WARN("The class EdgeScoreAsWeight is deprecated and will be removed in future releases; "
+    WARN("The class EdgeScoreAsWeight is deprecated and will be removed in future releases."
          "Use EdgeScore<T>::calculate(...) instead.");
     if (!G->hasEdgeIds()) {
-        throw std::runtime_error("edges have not been indexed - call indexEdges first");
+        throw std::runtime_error("Edges have not been indexed - call indexEdges first");
     }
 
     Graph result(*G, true, false);


fabratu:
It's overdue to refactor the whole `edgescores` module. Both the API and the class structure are not optimal and somewhat unfinished.

The `score` function can be deprecated, since not one inherited class implements it. Also, the "implementations" for the inherited classes can be moved to the base class. The class `EdgeScoresAsWeight` just provides means to transform scores, and we could integrate it into `EdgeScores` as a helper function (+ deprecation of the complete class). Furthermore, as you mentioned, proper documentation concerning constraints is missing (and maybe on other occasions).

Concerning your contribution, I would recommend moving the `score` implementation to the base class and adding tests once for calling `score` in order to get the coverage. For the other parts, add test cases where missing. In case a particular function or code raises questions, you can mention it here. If there are simple additions, like raising errors for certain constraints, it would be good if you just added them directly. 

Everything can be dealt with in a separate PR, tentatively overhauling the module.

fabratu:
This is the first PR where Coveralls tracks the coverage directly and posts feedback. From the report (apart from artifacts of randomness in `DynTopHarmonicCloseness` tests), your test seems not to be executed. Likely, this is due to a missing definition of the `EdgeScoreAsWeight` class.

```
class EdgeScoreAsWeightGTest : public testing::Test {};

TEST(EdgeScoreAsWeightGTest, testThrowsIfEdgesNotIndexed)
...
```

Schwarf:
Hi @fabratu,
I’ve updated the PR according to your suggestions (centralized EdgeScore API, added calculate, deprecated EdgeScoreAsWeight, and added tests).
Could you please check if this goes in the intended direction?

networkit/cpp/edgescores/test/EdgeScoreAsWeightGTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+static std::vector<double> makeSequentialScores(const Graph &G) {
+    std::vector<double> s(G.numberOfEdges());
+    for (edgeid eid = 0; eid < G.numberOfEdges(); ++eid) {
+        s[eid] = static_cast<double>(eid + 1); // 1.0, 2.0, 3.0, ...
+    }
+    return s;
+}
+
+TEST(EdgeScoreAsWeight, testThrowsIfEdgesNotIndexed) {
```
fabratu: The test name should be `EdgeScoreAsWeightGTest` (as well for the other occasions).

networkit/cpp/edgescores/test/EdgeScoreGTest.cpp
```diff
@@ -0,0 +1,183 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include <networkit/edgescores/EdgeScore.hpp>
+#include <networkit/graph/Graph.hpp>
+
+namespace NetworKit {
+
+class EdgeScoreTest : public ::testing::Test {
```
fabratu: The name of the class should be `EdgeScoreGTest` as referenced in the `CMakeLists.txt` file.

networkit/cpp/edgescores/test/EdgeScoreGTest.cpp
```diff
@@ -0,0 +1,183 @@
...
+            scoreData.assign(G->upperEdgeIdBound(), 0.0);
+            G->parallelForEdges(
+                [&](node, node, edgeid eid) { scoreData[eid] = static_cast<double>(eid + 1); });
+        }
+
+        hasRun = true;
+    }
+};
+
+TEST_F(EdgeScoreTest, testScoresAndScoreRequireRun) {
```
fabratu: Also refactor the names of test cases (here and on the following occasions).

include/networkit/edgescores/EdgeScoreAsWeight.hpp
```diff
@@ -12,7 +12,9 @@
 
 namespace NetworKit {
 
-class EdgeScoreAsWeight final {
+class [[deprecated(
```
fabratu: Better use `TLX_DEPRECATED` for deprecating the class (see [here](https://github.com/networkit/networkit/blob/e326d93121f699902266c64ff2bbd1b16c6e61e6/include/networkit/graph/UnionMaximumSpanningForest.hpp#L62) for an example used for a method.)

In addition, add a DEPRECATION hint in the documentation of the constructor.

fabratu:
Thanks again. Your changes go in the intended directions. My comments are mostly concerned with minor details.

PR #1362, opened 2025-10-18
Fix clang-format version parsing.

### LINKED ISSUE #1361: check_code.sh fails to detect clang-format 17 on Ubuntu
Running `./check_code.sh -w` on Ubuntu with `clang-format-17` installed fails at the `CppClangFormat.py` step. The helper script extracts the wrong token from clang-format --version, so the major version cannot be parsed as an integer. Consequently it reports “clang-format binary not found” even though the binary is present and on PATH.

Image

Schwarf:
Fix clang-format version detection in extrafiles/tooling/CppClangFormat.py that caused ./check_code.sh -w to fail on Ubuntu.
Fixes #1361

### DIFF
--- a/extrafiles/tooling/CppClangFormat.py
+++ b/extrafiles/tooling/CppClangFormat.py
@@ -10,7 +10,7 @@ a change is necessary.
 import filecmp
 import nktooling as nkt
 import tempfile
-import sys
+import re
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


fabratu:
The new logic seems more robust; thanks for the contribution.

fabratu (approved):

PR #1376, opened 2025-11-25
CI: Automate uploading stable builds as wheels

fabratu:
This PR aims to automate the process of creating nightly wheel builds using `cibuildwheel` and test.pypi.org as the index.


domischweisgut (commented):
wrong PR :D

PR #1377, opened 2025-11-25
Speed up `LeftRightPlanarityCheck` by replacing hash maps with vector-based storage

Schwarf:
## Summary

This PR significantly improves the performance of `LeftRightPlanarityCheck`, originally introduced in [PR#1276](https://github.com/networkit/networkit/pull/1276).
The implementation now eliminates all hashmap lookups on the hot path by replacing per-edge `unordered_map` structures with `std::vector` containers indexed by `edgeid`. This removes hashing overhead and greatly improves cache locality.
Dynamic instruction count is reduced by **~12.7×**, and the planarity test-suite runtime improves by up to **~20×**.
The API remains unchanged, but the algorithm now **requires the input graph to have edge IDs**.

---

## Motivation

Profiling with `perf` showed that a large portion of runtime was spent hashing `Edge` objects. These hash-based structures stored per-edge metadata and dominated the hot path.

Since the `Graph` can provide stable `edgeid`s, all such hash maps and sets can be replaced by flat vectors indexed by `edgeid`, removing hashing and improving memory locality.

---

### 1. Replace `Edge`-keyed hash maps with `edgeid`-keyed ones  
This removed hashing of `Edge` objects and simplified lookups.

### 2. Replace all `edgeid` hash maps and sets with `std::vector`  
All per-edge metadata is now stored in contiguous arrays indexed by `edgeid`.  
This removes hashing entirely from the algorithm’s core operations.

---

### Measurement setup

Dynamic instruction counts were measured with Valgrind/Callgrind on a small standalone benchmark that links NetworKit as a static library and reads a test graph with ~45,000 edges (`4elt.graph`) from `networkit/input/`.

The benchmark:

- reads the graph via `METISGraphReader`,
- calls `graph.indexEdges()` once up-front,
- runs `LeftRightPlanarityCheck::run()` 

Callgrind was invoked as:

```bash
OMP_NUM_THREADS=1 valgrind --tool=callgrind ./leftright
```
**Dynamic instruction counts:**
| Version | Instructions | Improvement |
|--------|--------------|-------------|
| Original | 2,131,945,098 | — |
| Step 1: replace `Edge` → `edgeid` maps | 357,162,806 | **5.97×** fewer |
| Step 2: replace remaining hash maps/sets with vectors | 167,420,035 | **12.7×** fewer than original |

**Test-suite runtime** (on my machine):

- **Debug:** 4291 ms → 464 ms (**~9× faster**)  
- **Release:** 1119 ms → 54 ms (**~20× faster**)

(Approximate wall-clock times, but fully consistent with the instruction-count reductions.)

---

## API contract: graph must have edge IDs
The previous implementation did not require indexed edges. After the optimization, however, the algorithm relies on `edgeid`-indexed vectors, so edge IDs are now necessary. This requirement is realized in the constructor.

```cpp
if (!G.hasEdgeIds()) {
    throw std::runtime_error("LeftRightPlanarityCheck requires a graph with edge ids.");
}
```
This is a minor breaking change, but the fix is trivial (G.indexEdges() once up-front), and the performance improvements rely on this precondition.

## Summary
This PR
- removes hash-map bottlenecks from the planarity test,
- reduces dynamic instructions by ~92%,
- improves runtime by up to 20×
- introduces the requirement for indexed edges

### DIFF
--- a/networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
+++ b/networkit/cpp/planarity/LeftRightPlanarityCheck.cpp
@@ -213,7 +213,7 @@ void LeftRightPlanarityCheck::removeBackEdges(const edgeid edgeId, const node pa
     }
 }
 
-bool LeftRightPlanarityCheck::dfsTesting(const node startNode) {
+bool LeftRightPlanarityCheck::dfsTesting(node startNode) {
     std::stack<node> dfsStack;
     dfsStack.push(startNode);
 
@@ -290,7 +290,7 @@ bool LeftRightPlanarityCheck::dfsTesting(const node startNode) {
     return true;
 }
 
-void LeftRightPlanarityCheck::dfsOrientation(const node startNode) {
+void LeftRightPlanarityCheck::dfsOrientation(node startNode) {
     std::stack<node> dfsStack;
     dfsStack.push(startNode);
 


Schwarf:
Hi @fabratu,
the [Build wheels / macOS CPython (x86_64) (pull_request)](https://github.com/networkit/networkit/actions/runs/19682593877/job/56380067759?pr=1377) was cancelled. Not sure why. Do need to trigger CI again?

fabratu (changes requested):

Schwarf:
### 📌 Post-merge performance update (additional measurements)

After merging this PR, I re-measured the performance of `LeftRightPlanarityCheck` using a more isolated benchmark setup to quantify the *algorithm runtime* directly (single run on `4elt.graph`, Release build, `OMP_NUM_THREADS=1`).

**Additional result:**
- **Isolated algorithm runtime:** ~860 ms → ~7 ms (**~100× speedup**)  
  (measured with Google Benchmark, CPU time per iteration)

The previously reported **~12.7× reduction in dynamic instruction count** (measured with Valgrind/Callgrind) and the **~20× improvement in end-to-end planarity test-suite runtime** remain unchanged and correct.

The larger ~100× speedup becomes visible when measuring the planarity check in isolation and is consistent with the earlier instruction-count reduction.

This comment is for documentation and clarity only; it does not imply any additional code changes beyond the merged PR.

PR #1397, opened 2026-03-11
Doc: Enable dynamic doc builds (stable, nightly), triggered by CI and branch combination.

fabratu:
### Background

Currently, the development documentation (API, notebooks, ...) is built separately from the rest of the NetworKit website. It gets updated every time we push to master. Hence, the documentation might contain information or API functionality that is not yet available in a release version. While switching to RTD or similar would lead to more robust results, this would also create quite some overhead in reworking the complete website deployment/setup.

As a bridge solution, this PR proposes a switch between two builds of the dev-doc: stable (most recent release) and nightly (last push to master).

The PR is currently WIP; most of the work was done with codex.
### Summary of changes

- The deployment of the documentation is removed from `ci.yml` and instead handled in an additional CI description file `docs-pages.yml`.
- A deployment script is created, supporting the deployment to different paths and setting variables (version for stable, date for nightly)
- Frontend adjustments: add a selector and logic to maintain the state of choice.
- Adjust deployment to maintain two versions of the dev-docs.

### Result

Screenshot 2026-03-10 at 08 59 53
Screenshot 2026-03-10 at 08 59 57


angriman (approved):

fabratu:
Thanks for the review! 

I will merge this right away, since it only concerns the technical documentation and should not break anything. After the merge, I will populate the stable version (11.2.1) manually. For future releases, this should then work automatically.

PR #1401, opened 2026-04-01
Expose isBipartite to python

Schwarf:
Expose `GraphTools::isBipartite` in the Python API via Cython bindings and add
Python tests for the new functionality.

### DIFF
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <span>
+#include <queue>
 
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/Random.hpp>
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -1,6 +1,6 @@
 #include <algorithm>
-#include <span>
 #include <queue>
+#include <span>
 
 #include <networkit/auxiliary/Log.hpp>
 #include <networkit/auxiliary/Random.hpp>
--- a/networkit/cpp/graph/GraphTools.cpp
+++ b/networkit/cpp/graph/GraphTools.cpp
@@ -1,5 +1,4 @@
 #include <algorithm>
-#include <queue>
 #include <span>
 
 #include <networkit/auxiliary/Log.hpp>
--- a/networkit/test/test_graphtools.py
+++ b/networkit/test/test_graphtools.py
@@ -578,7 +578,115 @@ class TestGraphTools(unittest.TestCase):
 				self.assertLessEqual(v2, v1)
 				self.assertLessEqual(v3, v1)			
 
-	
+	def testIsBipartiteDirectedGraphThrows(self):
+		G = nk.Graph(0, False, True)
+		with self.assertRaises(RuntimeError):
+			nk.graphtools.isBipartite(G)
+
+	def testIsBipartiteEmptyGraph(self):
+		G = nk.Graph()
+		self.assertTrue(nk.graphtools.isBipartite(G))
+
+	def testIsBipartiteSingleNodeGraphs(self):
+		for n in range(1, 10):
+			G = nk.Graph(n)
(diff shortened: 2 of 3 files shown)


networkit/cpp/graph/GraphTools.cpp
```diff
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <queue>
```
angriman, line 2: This change is unrelated to the PR.

networkit/test/test_graphtools.py
```diff
@@ -578,7 +578,115 @@ def testVolume(self):
...
 
-	
+	def testIsBipartiteDirectedGraphThrows(self):
+		G = nk.Graph(0, False, True)
+		with self.assertRaises(RuntimeError):
+			nk.graphtools.isBipartite(G)
+
+	def testIsBipartiteEmptyGraph(self):
+		G = nk.Graph()
+		self.assertTrue(nk.graphtools.isBipartite(G))
```
angriman: Nit: `self.assertTrue(nk.graphtools.isBipartite(nk.Graph()))`

networkit/test/test_graphtools.py
```diff
@@ -578,7 +578,115 @@ def testVolume(self):
...
+
+	def testIsBipartiteSingleNodeGraphs(self):
```
angriman: `nk.Graph(n)` creates a graph with `n` nodes, so this test should be renamed to something like `testIsBipartiteSingletonNodesGraph`.

Also, I'd recommend to check only a few cases, for example:
```py
self.assertTrue(nk.graphtools.isBipartite(nk.Graph(1)))
self.assertTrue(nk.graphtools.isBipartite(nk.Graph(2)))
self.assertTrue(nk.graphtools.isBipartite(nk.Graph(3)))
```

networkit/test/test_graphtools.py
```diff
@@ -578,7 +578,115 @@ def testVolume(self):
...
+		for n in range(1, 10):
+			G = nk.Graph(n)
+			self.assertTrue(nk.graphtools.isBipartite(G))
+
+	def testIsBipartiteBinaryTreeGraphs(self):
+		def binaryTree(numNodes):
```
angriman: While it's a good idea to check binary graphs, this test will be hard to debug in case it starts to fail.

Instead, prefer constructing a single binary graph and document the structure in the comments, for example:

```py
	def testIsBipartiteBinaryTreeGraphs(self):
		# 7-nodes binary tree:
		#        0
		#      /   \
		#     1     2
		#    / \   / \
		#   3   4 5   6
		G = nk.Graph(7)
		
		G.addEdge(0, 1)
		G.addEdge(0, 2)
		
		G.addEdge(1, 3)
		G.addEdge(1, 4)
		
		G.addEdge(2, 5)
		G.addEdge(2, 6)

        self.assertTrue(nk.graphtools.isBipartite(G))
```

networkit/test/test_graphtools.py
```diff
@@ -578,7 +578,115 @@ def testVolume(self):
...
+				if rightChild < numNodes:
+					G.addEdge(i, rightChild, float(i))
+			return G
+
+		for n in range(1, 10):
+			G = binaryTree(n)
+			self.assertTrue(nk.graphtools.isBipartite(G))
+
+	def testIsBipartiteCompleteGraphs(self):
+		def completeGraph(numNodes):
```
angriman: As my comment above, prefer defining a simple complete graph in this test instead of generating them with a function.

networkit/cpp/graph/GraphTools.cpp
Schwarf: I agree that this change is unrelated to the PR. At the same time, I think the missing header may make the production code a bit fragile, for example by relying on transitive includes.

Would you prefer that I remove it from this PR and open a separate small PR for it?

angriman: Yes, let's move it to another PR.

Schwarf: Done here: https://github.com/networkit/networkit/pull/1405

angriman (approved):

PR #1402, opened 2026-04-01
Add keyword argument support for Closeness

### LINKED ISSUE #1154: Closeness requires exactly 3 positional arguments
Is there any reason why `Closeness` requires only positional arguments, not keyword arguments? For example, `normalized` or `checkConnectdedness` would be really nice to pass as keyword arguments.

Schwarf:
Add keyword argument support for `centrality.Closeness`

This adds support for the keyword arguments `checkConnectedness` and `variant` in the Python binding of `networkit.centrality.Closeness` (as suggested in #1154) without changing the existing positional API.

Closes #1154

### DIFF
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -772,7 +772,7 @@ cdef class Closeness(Centrality):
 		elif isinstance(arg, int):
 			self._this = new _Closeness(G._this, normalized, <_ClosenessVariant> arg)
 		else:
-			raise TypeError("Expected checkConnectedness to be bool or variant to be ClosenessVariant")
+			raise TypeError("Expected checkConnectedness to be a bool or variant to be a ClosenessVariant")
 
 cdef extern from "<networkit/centrality/ApproxCloseness.hpp>" namespace "NetworKit::ApproxCloseness":
 
--- a/networkit/centrality.pyx
+++ b/networkit/centrality.pyx
@@ -756,14 +756,23 @@ cdef class Closeness(Centrality):
 		is automatically set to False.
 	"""
 
-	def __cinit__(self, Graph G, normalized, third):
+	def __cinit__(self, Graph G, normalized, checkConnectedness=None, variant=None):
 		self._G = G
-		if isinstance(third, int):
-			self._this = new _Closeness(G._this, normalized, <_ClosenessVariant> third)
-		elif isinstance(third, bool):
-			self._this = new _Closeness(G._this, normalized, <bool_t> third)
+
+		if checkConnectedness is not None and variant is not None:
+			raise TypeError("Use either checkConnectedness or variant, not both")
+
+		arg = variant if variant is not None else checkConnectedness
+
+		if arg is None:
+			raise TypeError("Missing third argument: use either checkConnectedness or variant")
+
+		if isinstance(arg, bool):
+			self._this = new _Closeness(G._this, normalized, <bool_t> arg)
+		elif isinstance(arg, int):
+			self._this = new _Closeness(G._this, normalized, <_ClosenessVariant> arg)
 		else:
-			raise Exception("Error: the third parameter must be either a bool or a ClosenessVariant")
+			raise TypeError("Expected checkConnectedness to be bool or variant to be ClosenessVariant")
 
 cdef extern from "<networkit/centrality/ApproxCloseness.hpp>" namespace "NetworKit::ApproxCloseness":
 
--- a/networkit/test/test_centrality.py
+++ b/networkit/test/test_centrality.py
@@ -1,343 +1,332 @@
 #!/usr/bin/env python3
+import random
 import unittest
-import os
-import networkit as nk
+
 import numpy as np
 import scipy
-import random
+
+import networkit as nk
 
 
 class TestCentrality(unittest.TestCase):
 
-    def checkCovers(self, c1, c2):
-        if not c1.numberOfElements() == c2.numberOfElements():
-            return False
-        if not c1.numberOfSubsets() == c2.numberOfSubsets():
-            return False
-        for i in range(0, c1.numberOfElements()):
-            if not c1.subsetsOf(i) == c2.subsetsOf(i):
-                return False
-        return True
+	def checkCovers(self, c1, c2):
+		if not c1.numberOfElements() == c2.numberOfElements():
+			return False
+		if not c1.numberOfSubsets() == c2.numberOfSubsets():
+			return False
+		for i in range(0, c1.numberOfElements()):
+			if not c1.subsetsOf(i) == c2.subsetsOf(i):
+				return False
+		return True
 
-    def setUp(self):
-        self.L = nk.readGraph(
-            "input/looptest1.gml", nk.Format.GML
-        )  # without self-loops
-        self.LL = nk.readGraph(
-            "input/looptest2.gml", nk.Format.GML
-        )  # with self-loops sprinkled in
+	def setUp(self):
+		self.L = nk.readGraph(
+			"input/looptest1.gml", nk.Format.GML
+		)  # without self-loops
+		self.LL = nk.readGraph(
+			"input/looptest2.gml", nk.Format.GML
+		)  # with self-loops sprinkled in
 
-    def testScoreAsArray(self):
-        CL = nk.centrality.ApproxBetweenness(self.L, epsilon=0.01, delta=0.1)
-        CL.run()
-        listScores = CL.scores()
-        arrayScores = CL.scores(asarray=True)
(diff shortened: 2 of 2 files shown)


angriman (changes requested):
Since keyword arguments are a language feature, I'd recommend to not implement tests that check keyword argument correctness. Instead, we could use keyword arguments in existing tests and also ensure that we're covering all cases (e.g., tests with `checkConnectedness=True` are missing).

Schwarf:
Updated `testCloseness` to use keyword arguments in the existing test and expanded it to cover the `variant` and `checkConnectedness` cases, while keeping the original positional API calls for backward-compatibility.

networkit/test/test_centrality.py
```diff
@@ -164,6 +164,7 @@ def testBetweenness(self):
         self.assertEqual(CLL.maximum(), 28.0)
 
     def testCloseness(self):
+        # Original positional API with variant should still work
```
angriman: This comment can be dropped, the code is already clear. We could add keyword arguments to the Closeness object below.

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
         CLL.run()
         self.assertEqual(CL.ranking(), CLL.ranking())
 
+        # Keyword API for variant should behave the same
```
angriman: Let's not add keyword argument variants. What I meant in my previous review is to not add more tests just to check that keyword arguments work as intended. Instead, we could use them in all Closeness tests, and cover other uncovered cases (for example with `checkConnectedness=True`).

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+        variantKeyword = nk.centrality.Closeness(
+            self.L, True, variant=nk.centrality.ClosenessVariant.GENERALIZED
+        )
+        variantKeyword.run()
+        self.assertEqual(CL.ranking(), variantKeyword.ranking())
+
+        # Original positional API with checkConnectedness=True should still work
+        positionalCheckTrue = nk.centrality.Closeness(self.L, False, True)
+        positionalCheckTrue.run()
```
angriman: Let's move this version on a separate test and check the resulting ranking.

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
         self.assertEqual(CL.ranking(), CLL.ranking())
 
+        # Keyword API for variant should behave the same
+        variantKeyword = nk.centrality.Closeness(
+            self.L, True, variant=nk.centrality.ClosenessVariant.GENERALIZED
+        )
+        variantKeyword.run()
+        self.assertEqual(CL.ranking(), variantKeyword.ranking())
+
+        # Original positional API with checkConnectedness=True should still work
```
angriman: This comment can be dropped, the code is self-explanatory.

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+        positionalCheckTrue.run()
+
+        # Keyword API with checkConnectedness=True should behave the same
+        checkConnectednessKeywordTrue = nk.centrality.Closeness(
+            self.L, False, checkConnectedness=True
+        )
+        checkConnectednessKeywordTrue.run()
+        self.assertEqual(
+            positionalCheckTrue.ranking(), checkConnectednessKeywordTrue.ranking()
+        )
```
angriman: As mentioned previously, let's drop this test since this variant of Closeness is already covered in the test above,

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
         CLL.run()
         self.assertEqual(CL.ranking(), CLL.ranking())
 
+        # Keyword API for variant should behave the same
+        variantKeyword = nk.centrality.Closeness(
+            self.L, True, variant=nk.centrality.ClosenessVariant.GENERALIZED
+        )
+        variantKeyword.run()
+        self.assertEqual(CL.ranking(), variantKeyword.ranking())
```
angriman: Let's not add keyword argument variants. What I meant in my previous review is that we should not run the same test twice, one with and one without keyword arguments. Instead, we could use keyword arguments in all Closeness tests, and cover other uncovered cases (for example with checkConnectedness=True).

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+        # Keyword API with checkConnectedness=True should behave the same
+        checkConnectednessKeywordTrue = nk.centrality.Closeness(
+            self.L, False, checkConnectedness=True
+        )
+        checkConnectednessKeywordTrue.run()
+        self.assertEqual(
+            positionalCheckTrue.ranking(), checkConnectednessKeywordTrue.ranking()
+        )
+
+        # Original positional API with checkConnectedness=False should still work
```
angriman: This comment can be removed.

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+        positionalCheckFalse = nk.centrality.Closeness(self.L, False, False)
+        positionalCheckFalse.run()
```
angriman: Let's move this in a separate test and check the ranking.

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+
+        # Keyword API with checkConnectedness=False should behave the same
+        checkConnectednessKeywordFalse = nk.centrality.Closeness(
+            self.L, False, checkConnectedness=False
+        )
+        checkConnectednessKeywordFalse.run()
+        self.assertEqual(
+            positionalCheckFalse.ranking(), checkConnectednessKeywordFalse.ranking()
+        )
```
angriman: Same here, this test case is already covered above by `positionalCheckFalse`

networkit/test/test_centrality.py
```diff
@@ -174,6 +175,48 @@ def testCloseness(self):
...
+
+        # Passing both keyword arguments should fail
+        with self.assertRaises(TypeError):
+            nk.centrality.Closeness(
+                self.L,
+                True,
+                checkConnectedness=True,
+                variant=nk.centrality.ClosenessVariant.GENERALIZED,
+            )
```
angriman: Let's move this in a separate test.

networkit/centrality.pyx
```diff
@@ -756,14 +756,23 @@ cdef class Closeness(Centrality):
...
+		if arg is None:
+			raise TypeError("Missing third argument: use either checkConnectedness or variant")
+
+		if isinstance(arg, bool):
+			self._this = new _Closeness(G._this, normalized, <bool_t> arg)
+		elif isinstance(arg, int):
+			self._this = new _Closeness(G._this, normalized, <_ClosenessVariant> arg)
 		else:
-			raise Exception("Error: the third parameter must be either a bool or a ClosenessVariant")
+			raise TypeError("Expected checkConnectedness to be bool or variant to be ClosenessVariant")
```
angriman:
```suggestion
			raise TypeError("Expected checkConnectedness to be a bool or variant to be a ClosenessVariant")
```

Schwarf:
Hi @angriman, 
Thanks for the review — I removed the inline comments, split the `checkConnectedness=True` / `False` cases into separate tests, and added a test that verifies passing both keyword arguments fails.

I now check the resulting ranking directly instead of comparing positional and keyword-based calls. For the scores, I use `assertAlmostEqual` rather than exact float equality.

I’m not fully sure this is exactly the test structure you had in mind, so I’d appreciate a quick confirmation.

angriman:
Thank you, the structure of the centrality tests looks ok to me, but now the entire test_centrality.py file is changed (maybe you unintentionally swapped all spaces with tabs). Can you update the PR to only include the changes relative to the centrality tests?

networkit/test/test_centrality.py
```diff
@@ -174,6 +174,37 @@ def testCloseness(self):
         CLL.run()
         self.assertEqual(CL.ranking(), CLL.ranking())
 
+    def testClosenessCheckConnectednessTrue(self):
+        ranking = nk.centrality.Closeness(self.L, False, True).run().ranking()
+        expected_ranking = [(4, 0.6153846153846154), (6, 0.5714285714285714), (1, 0.5), (2, 0.47058823529411764),
```
angriman, line 179: Let's only keep a couple of decimal places, same in the tests below.

angriman (dismissed):

angriman (approved):

PR #1405, opened 2026-04-19
Add explicit <queue> include to GraphTools.cpp

Schwarf:
This PR adds the missing header as discussed here: https://github.com/networkit/networkit/pull/1401#discussion_r3030250606


networkit/cpp/graph/GraphTools.cpp
```diff
@@ -13,6 +14,7 @@ namespace GraphTools {
 
 count computeMaxDegree(const Graph &G, bool inDegree = false) {
     count result = 0;
+
```
angriman: Let's drop this change and only keep the one relative to `queue`

angriman (changes requested):

angriman (approved):

PR #1406, opened 2026-05-02
Add C++ SimRank edge scoring algorithm

### LINKED ISSUE #1381: Add support for Edge-Ranking
NetworKit currently lacks any module for **edge-ranking** or **edge-centric similarity** computations, despite offering several node-ranking and centrality tools. This is a gap because a number of link analysis, link prediction, and structural similarity methods fundamentally depend on ranking edges rather than nodes.

A natural starting point is to expose an **edge-ranking API** and implement the classical **Jeh and Widom (2002) SimRank** formulation as the first concrete algorithm under this interface. This provides a baseline method, is mathematically simple, and gives users a reference point before more advanced or faster variants are added.

### **Motivation**

1. Many modern graph-mining tasks like link prediction, structural role analysis, and semantic similarity require ordering **pairs of nodes** or **edges**, not individual vertices.
2. NetworKit has strong performance for large networks, but currently forces users to compute pairwise scores manually or rely on external Python libraries, which is inefficient and kills scalability.
3. Jeh and Widom’s original SimRank recurrence is a well-known baseline. Even if not the fastest, it is straightforward to implement and gives an initial, correct ground-truth method for edge-ranking.
4. Adding an official edge-ranking abstraction opens the path for later contributions:

   * optimized SimRank variants
   * SimRank*
   * Monte Carlo approximations
   * graph sparsification-based similarity computation
   * edge embedding-based ranking

### **Proposed API Direction**

Introduce a new module, for example:

```python
networkit.similarity
```

with a base class:

```python
class EdgeRanking:
    def run(self);
    def score(u, v);
```

Then implement:

```python
class SimRankJW(EdgeRanking):
    # classical iterative Jeh–Widom formulation
```

Expected features:

* Iterative fixed-point computation for all-pairs SimRank
* Damping factor ( C ) configurable
* Convergence tolerance and iteration cap
* Optional restriction to a candidate edge set to avoid full all-pairs computation

References:
[1] NetworkX's SimRank documentation -> https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.similarity.simrank_similarity.html#networkx.algorithms.similarity.simrank_similarity
[2] G. Jeh and J. Widom. “SimRank: a measure of structural-context similarity”, In KDD’02: Proceedings of the Eighth ACM SIGKDD International Conference on Knowledge Discovery and Data Mining, pp. 538–543. ACM Press, 2002.

Schwarf:
This PR adds the C++ implementation for the SimRank-based edge scoring part of #1381.

It introduces `SimRankScore`, an `EdgeScore<double>` implementation that assigns each indexed edge `(u, v)` the SimRank similarity between its endpoints.

The implementation follows the iterative SimRank recurrence: two nodes are considered similar if their neighbors are similar. For directed graphs, in-neighbors/predecessors are used; for undirected graphs, ordinary neighbors are used.

Let `z = G.upperNodeIdBound()` and `m` be the number of edges. The implementation uses `O(z^2)` space for the pairwise score matrices plus `O(m)` for `scoreData`. The dense worst-case runtime is `O(k * z^4)`, where `k` is the number of iterations.

The implementation is parallelized with OpenMP where feasible: diagonal initialization, pairwise score updates, and final edge-score extraction are parallelized. The convergence check uses thread-local maximum differences followed by a reduction.

In follow-up PRs, I plan to add the Python API and investigate performance improvements.
This PR does not yet close #1381 because the Python API is still missing.

### DIFF
--- a/include/networkit/edgescores/SimRankScore.hpp
+++ b/include/networkit/edgescores/SimRankScore.hpp
@@ -15,13 +15,13 @@ namespace NetworKit {
 
 class SimRankScore final : public EdgeScore<double> {
 public:
-    SimRankScore(const Graph &G, double damping = 0.9, count maxIterations = 100,
+    SimRankScore(const Graph &G, double similarityPropagationFactor = 0.9, count maxIterations = 100,
                  double tolerance = 1e-4);
 
     void run() override;
 
 private:
-    double damping;
+    double similarityPropagationFactor;
     count maxIterations;
     double tolerance;
 
--- a/include/networkit/edgescores/SimRankScore.hpp
+++ b/include/networkit/edgescores/SimRankScore.hpp
@@ -1,13 +1,18 @@
-//
-// Created by andreas on 01.05.26.
-//
+/*  SimRankScore.hpp
+ *
+ *  Created on: 01.05.2026
+ *  Authors: Andreas Scharf (andreas.b.scharf@gmail.com)
+ *
+ */
 
-#ifndef NETWORKIT_SIMRANKSCORE_H
-#define NETWORKIT_SIMRANKSCORE_H
+#ifndef NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
+#define NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
 
 #include <networkit/edgescores/EdgeScore.hpp>
 #include <networkit/graph/Graph.hpp>
+
 namespace NetworKit {
+
 class SimRankScore final : public EdgeScore<double> {
 public:
     SimRankScore(const Graph &G, double damping = 0.9, count maxIterations = 100,
@@ -22,5 +27,6 @@ private:
 
     count iterations;
 };
+
 } // namespace NetworKit
-#endif // NETWORKIT_SIMRANKSCORE_H
+#endif // NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
--- a/include/networkit/edgescores/SimRankScore.hpp
+++ b/include/networkit/edgescores/SimRankScore.hpp
@@ -5,28 +5,67 @@
  *
  */
 
-#ifndef NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
-#define NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
+#ifndef NETWORKIT_EDGESCORES_SIM_RANK_SCORE_HPP_
+#define NETWORKIT_EDGESCORES_SIM_RANK_SCORE_HPP_
 
 #include <networkit/edgescores/EdgeScore.hpp>
 #include <networkit/graph/Graph.hpp>
 
 namespace NetworKit {
 
+/**
+ * Computes a SimRank similarity score for each edge of the input graph.
+ *
+ * The score assigned to an edge (u, v) is the SimRank similarity between
+ * its endpoints u and v.
+ *
+ * This implementation follows the classical iterative SimRank recurrence:
+ * two nodes are similar if their neighbors/predecessors are similar.
+ *
+ * The result is stored as one score per edge id in scoreData.
+ *
+ * Requires indexed edges, i.e. G.indexEdges() must have been called before run().
+ */
 class SimRankScore final : public EdgeScore<double> {
 public:
-    SimRankScore(const Graph &G, double similarityPropagationFactor = 0.9, count maxIterations = 100,
-                 double tolerance = 1e-4);
-
+    /**
+     * Constructs a SimRank edge-score algorithm.
+     *
+     * @param G The input graph.
+     * @param similarityPropagationFactor Factor in [0, 1] that scales the contribution
+     *        of neighboring node-pair similarities in each SimRank iteration.
+     *        Larger values preserve more propagated similarity.
+     * @param maxIterations Maximum number of iterations.
+     * @param tolerance Convergence tolerance for the maximum absolute score change
+     *        between two consecutive iterations.
+     *
+     * @throws std::invalid_argument if similarityPropagationFactor is not in [0, 1].
+     * @throws std::invalid_argument if maxIterations is 0.
+     * @throws std::invalid_argument if tolerance is negative.
+     */
+    SimRankScore(const Graph &G, double similarityPropagationFactor = 0.9,
+                 count maxIterations = 100, double tolerance = 1e-4);
+    /**
+     * Computes one SimRank score per indexed edge.
+     *
+     * After run(), scores can be accessed through:
+     * - scores()
+     * - score(edgeid)
+     * - score(u, v)
+     *
+     * @throws std::runtime_error if the graph has no indexed edges.
+     *         Call G.indexEdges() before running the algorithm.
+     * @throws std::overflow_error if the internal pairwise score matrix would
+     *         exceed the supported index range.
+     */
     void run() override;
 
 private:
     double similarityPropagationFactor;
     count maxIterations;
     double tolerance;
-
     count iterations;
 };
 
 } // namespace NetworKit
-#endif // NETWORKIT_EDGESCORES_SIMRANK_SCORE_HPP
+#endif // NETWORKIT_EDGESCORES_SIM_RANK_SCORE_HPP_
--- a/include/networkit/edgescores/SimRankScore.hpp
+++ b/include/networkit/edgescores/SimRankScore.hpp
@@ -1,0 +1,26 @@
+//
+// Created by andreas on 01.05.26.
+//
+
+#ifndef NETWORKIT_SIMRANKSCORE_H
+#define NETWORKIT_SIMRANKSCORE_H
+
+#include <networkit/edgescores/EdgeScore.hpp>
+#include <networkit/graph/Graph.hpp>
+namespace NetworKit {
+class SimRankScore final : public EdgeScore<double> {
+public:
+    SimRankScore(const Graph &G, double damping = 0.9, count maxIterations = 100,
+                 double tolerance = 1e-4);
+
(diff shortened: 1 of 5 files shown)


networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+void SimRankScore::run() {
+    if (!G->hasEdgeIds()) {
+        throw std::runtime_error("Graph must have edge IDs for SimRank calculation");
+    }
+
+    const index n = G->upperNodeIdBound();
+
+    // This is practically unreachable in ordinary tests because constructing a graph
+    // with n >= 2^32 nodes usually fails first, but the check prevents n * n from
+    // wrapping before allocating the dense SimRank matrix.
```
angriman, line 39: What happens in case the number of nodes is less than 2? Should we return an error? Let's also add tests to cover this case.

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    const index matrixSize = n * n;
+
+    auto matrixIndex = [n](node u, node v) -> index { return u * n + v; };
+
+    std::vector<double> oldScore(matrixSize, 0.0);
+    std::vector<double> newScore(matrixSize, 0.0);
+
+    G->parallelForNodes([&](node u) { oldScore[matrixIndex(u, u)] = 1.0; });
```
angriman: Since this is also used below, prefer wrapping it in a lambda (e.g., `initDiagonal`).

include/networkit/edgescores/SimRankScore.hpp
```diff
@@ -0,0 +1,71 @@
...
+     * @throws std::overflow_error if the internal pairwise score matrix would
+     *         exceed the supported index range.
+     */
+    void run() override;
+
+private:
+    double similarityPropagationFactor;
+    count maxIterations;
+    double tolerance;
+    count iterations;
```
angriman: This does not seem to be used outside `run()`. In that case, prefer moving it as a local variable in `run()`.

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    iterations = 0;
+    double maxDifference = std::numeric_limits<double>::max();
+
+    for (count iter = 0; iter < maxIterations; ++iter) {
+        std::fill(newScore.begin(), newScore.end(), 0.0);
```
angriman: Nit: Prefer `std::ranges::fill`

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        G->parallelForNodes([&](node u) { newScore[matrixIndex(u, u)] = 1.0; });
+
+        std::vector<double> threadMaxDifferences(omp_get_max_threads(), 0.0);
```
angriman: Nit: Prefer declaring this vector outside the for-loop to avoid dynamic allocations. Here you can reset it with `fill`.

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        G->parallelForNodePairs([&](node u, node v) {
```
angriman: This loop is likely the hottest part of the algorithm, as it always performs N^2 operations with a lot of branching. You can reduce branching by splitting it in two loops and removing the first `if` (the sum is zero in case any of the two nodes is isolated), here is an idea:
```cpp
const auto computeSum = [&](node u, node v, auto getRange) -> double {
  double sum = 0.0;
  for (node a : getRange(u)) {
    for (node b : getRange(v)) {
      sum += oldScore[matrixIndex(a, b)];
    }
  }
  return sum;
};

const auto updateMaxDifferences = [&](node u, node v, double sum) -> void {
  const double value = ...
  // Update threadMaxDifference...
};

if (G->isDirected()) {
  G->parallelForNodePairs([&](node u, node v) {
    const double sum = computeSum(u, v, [&G](node n) { return G->inNeighborRange(n); });
    updateMaxDifference(u, v, sum);
  };
} else {
  // Same, but use G->neighborRange(n) in the lambda.
}
```

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
...
+                std::max(std::abs(value - oldScore[uv]), std::abs(value - oldScore[vu]));
+
+            const int tid = omp_get_thread_num();
+            threadMaxDifferences[tid] = std::max(threadMaxDifferences[tid], difference);
+        });
+
+        double iterationMaxDifference = 0.0;
+        for (const double localDifference : threadMaxDifferences) {
+            iterationMaxDifference = std::max(iterationMaxDifference, localDifference);
+        }
```
angriman: You can simplify this with std::ranges::max
```suggestion
        const double iterationMaxDifference = std::ranges::max(threadMaxDifferences);
```

networkit/cpp/edgescores/test/SimRankScoreGTest.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+TEST_F(SimRankScoreGTest, testRunProducesOneScorePerIndexedEdge) {
+    Graph G(3, false, false, true);
+    G.addEdge(0, 1);
+    G.addEdge(1, 2);
+
+    SimRankScore simrank(G);
+    simrank.run();
+
+    EXPECT_EQ(simrank.scores().size(), G.upperEdgeIdBound());
```
angriman: Prefer using SizeIs:
```suggestion
    EXPECT_THAT(simrank.scores(), testing::SizeIs(G.upperEdgeIdBound()));
```

networkit/cpp/edgescores/test/SimRankScoreGTest.cpp
```diff
@@ -0,0 +1,262 @@
...
+    const auto firstScores = simrank.scores();
+
+    simrank.run();
+    const auto secondScores = simrank.scores();
+
+    EXPECT_EQ(firstScores.size(), secondScores.size());
+
+    for (index i = 0; i < firstScores.size(); ++i) {
+        EXPECT_NEAR(firstScores[i], secondScores[i], 1e-12);
+    }
```
angriman: This can be simplified with
```suggestion
  EXPECT_THAT(firstScores, Pointwise(DoubleNear(1e-12), secondScores));
```

networkit/cpp/edgescores/test/SimRankScoreGTest.cpp
```diff
@@ -0,0 +1,262 @@
...
+    multiThreaded.run();
+    const auto multiThreadedScores = multiThreaded.scores();
+
+    omp_set_num_threads(previousThreads);
+
+    EXPECT_EQ(singleThreadedScores.size(), multiThreadedScores.size());
+
+    for (index i = 0; i < singleThreadedScores.size(); ++i) {
+        EXPECT_NEAR(singleThreadedScores[i], multiThreadedScores[i], 1e-12);
+    }
```
angriman: This can be simplified as mentioned in the previous test.

networkit/cpp/edgescores/test/SimRankScoreGTest.cpp
```diff
@@ -0,0 +1,262 @@
...
+#endif
+}
+
+TEST_F(SimRankScoreGTest, testRepeatedParallelRunsProduceSameScores) {
```
angriman: I think that this test can be dropped, it covers the same case as the previous one.

networkit/cpp/edgescores/SimRankScore.cpp
Schwarf: I added tests for graphs with zero and one node. I think these are valid inputs: the algorithm should run without throwing, and the resulting score vector should be empty because there are no edges to score.

networkit/cpp/edgescores/SimRankScore.cpp
Schwarf: Done

networkit/cpp/edgescores/SimRankScore.cpp
Schwarf: Done

networkit/cpp/edgescores/SimRankScore.cpp
Schwarf: Done

networkit/cpp/edgescores/test/SimRankScoreGTest.cpp
Schwarf: Done

networkit/cpp/edgescores/SimRankScore.cpp
Schwarf: Done

angriman (approved):

networkit/cpp/edgescores/SimRankScore.cpp
```diff
@@ -0,0 +1,124 @@
+/*  SimRankScore.cpp
+ *
+ *  Created on: 01.05.2026
+ *  Authors: Andreas Scharf (andreas.b.scharf@gmail.com)
+ *
+ */
+
+#include <networkit/edgescores/SimRankScore.hpp>
+
+#include "networkit/auxiliary/NumericTools.hpp"
```
angriman:
```suggestion
#include <networkit/auxiliary/NumericTools.hpp>
#include <networkit/edgescores/SimRankScore.hpp>
```

angriman (dismissed):

angriman (approved):

Issue #1172, opened 2024-01-30
Some Python tests fail for higher number of threads


fabratu:
When executed with 48 threads, several unittests for the Python interface fail:

```
======================================================================
FAIL: testApproxElectricalCloseness (test_centrality.TestCentrality.testApproxElectricalCloseness)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/brandtfa/scripts/test/test_centrality.py", line 55, in testApproxElectricalCloseness
    self.assertLessEqual(abs(apx[u] - pinv[u]), eps)
AssertionError: 0.11044452053538456 not less than or equal to 0.1

======================================================================
FAIL: testApproxSpanningEdge (test_centrality.TestCentrality.testApproxSpanningEdge)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/brandtfa/scripts/test/test_centrality.py", line 68, in testApproxSpanningEdge
    self.assertLessEqual(abs(apxScore - exactScore), 2*eps)
AssertionError: 0.24007940334904146 not less than or equal to 0.2

======================================================================
FAIL: testForest (test_centrality.TestCentrality.testForest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/brandtfa/scripts/test/test_centrality.py", line 292, in testForest
    self.assertLessEqual(abs(apx - exact), eps)
AssertionError: 0.06750794287494666 not less than or equal to 0.05

----------------------------------------------------------------------
```

Reducing the number of threads mitigates this problem.

domischweisgut:
Fixed via PR #1411

PR #1411, opened 2026-05-11
Feature/fix 1172 thread sensitive tests

domischweisgut:
This PR addresses Issue #1172 

testApproxElectricalCloseness, testApproxSpanningEdge, and testForest exercise randomized approximations whose error bound is only probabilistic. With many threads the per-thread sample splits and thread-id-derived RNG seeds shift, occasionally pushing the observed error past eps even with a fixed seed.

Wrap each test in a save -> setNumberOfThreads(1) -> try/finally -> restore block. The single-threaded path is deterministic for a given seed, so the strict assertion holds; the surrounding thread count is restored exactly.

Co-developed with Claude (Claude Code).

### DIFF
--- a/networkit/test/test_centrality.py
+++ b/networkit/test/test_centrality.py
@@ -119,9 +119,7 @@ class TestCentrality(unittest.TestCase):
         self.assertEqual(CL.numberOfSamples(), 63026)
 
     def testApproxElectricalCloseness(self):
-        # Pin to a single thread: the algorithm is a randomized Monte-Carlo
-        # approximation whose probabilistic error bound is only tight in
-        # expectation, so high thread counts can occasionally exceed eps (#1172).
+        # Pin to a single thread
         threadsAvailable = nk.getMaxNumberOfThreads()
         nk.setNumberOfThreads(1)
         try:
@@ -149,7 +147,6 @@ class TestCentrality(unittest.TestCase):
             nk.setNumberOfThreads(threadsAvailable)
 
     def testApproxSpanningEdge(self):
-        # See note in testApproxElectricalCloseness re: thread count and #1172.
         threadsAvailable = nk.getMaxNumberOfThreads()
         nk.setNumberOfThreads(1)
         try:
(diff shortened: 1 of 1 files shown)


fabratu (dismissed):

networkit/test/test_centrality.py
```diff
@@ -119,39 +119,50 @@ def testApproxBetweennessNumberOfSamples(self):
...
-                    L[u, v] = -1
-                    L[v, u] = -1
-
-            pinv = np.linalg.pinv(L).diagonal()
-            for u in g.iterNodes():
-                self.assertLessEqual(abs(apx[u] - pinv[u]), eps)
+        # Pin to a single thread
+        threadsAvailable = nk.getMaxNumberOfThreads()
+        nk.setNumberOfThreads(1)
+        try:
```
angriman: Instead of repeating this for each test, you can use `contextlib` to run single-threaded tests within a `with` statement:
```py
from contextlib import contextmanager

@contextmanager
def run_single_thread():
    nk.setNumberOfThreads(1)
    try:
        yield
    finally:
        nk.setNumberOfThreads(nk.getMaxNumberOfThreads())

class TestCentrality(unittest.TestCase):
    def testApproxElectricalCloseness(self):
        with run_single_thread():
            for seed in [1, 2, 3]: ...
```

Also, note that we don't need the `threadsAvailable` variable since `getMaxNumberOfThreads` returns `omp_get_max_threads()`: https://github.com/networkit/networkit/blob/d11dfd228edfa0f91847be04e0753b0ca50a2059/networkit/cpp/auxiliary/Parallelism.cpp#L23-L30

angriman (changes requested):

networkit/test/test_centrality.py
domischweisgut: Thank you for the hint! The fix looks much cleaner now. Extracted `run_single_thread()` as a `contextmanager` and dropped the unnecessary variable: 
```py
from contextlib import contextmanager

@contextmanager
def run_single_thread():
    nk.setNumberOfThreads(1)
    try:
        yield
    finally:
        nk.setNumberOfThreads(nk.getMaxNumberOfThreads())


class TestCentrality(unittest.TestCase):
```

angriman (approved):

Issue #949, opened 2022-05-27
community detection: PLM.get_partition behaves inconsistently 


clstaudt:
Got a chance to use networkit community detection in a real-world project and noticed that `PLM.getPartition()`  gives two different results when called twice in a row. The first call returns the actual clustering. The second call (on the same, fitted instance of PLM) returns apparently an empty partition.

![Screenshot 2022-05-27 at 13 24 28](https://user-images.githubusercontent.com/875194/170690894-d99ad5c9-4eda-42f4-8841-bf53b1dbcac9.png)
![Screenshot 2022-05-27 at 13 24 49](https://user-images.githubusercontent.com/875194/170690902-76848656-c590-46dd-a684-84bddb384629.png)

clstaudt:
@fabratu  Here's a stale issue for example... what do you think?

clstaudt:
@trIAge-bot, can you generate some test cases that would catch the error? NetworKit uses GTest.

PR #1412, opened 2026-05-12
return by value + test to confirm the fix 

domischweisgut:
This PR addresses Issue #949 

In Issue #949, the following scenario is reported: If we run an instance of PLM and use PLM.getPartition() twice in a row, the second call yields different results. In particular, the second call returns an empty partition. 

Here's my opinion on where the problem is: 
In the C++ declaration of `getPartition` we return a const reference to a partition object (to be precise: result): 
`virtual const Partition &getPartition() const;`
However, in the Cython binding we find the following: 
The Cython binding declared the return value also as a reference. The problem with this is the following: 
When we call getPartition() via Python, in community.pyx, we call 
`return Partition().setThis((<_CommunityDetectionAlgorithm*>(self._this)).getPartition())`
 `Partition()` calls `__cinit__`, which creates an empty partition. 
In `setThis` we then use std::swap, which means that `result` now holds the empty partition and the Python wrapper owns the real partition. Every subsequent call then swaps empty partitions with empty partitions.

The (quick) fix is then to declare the Cython return by value. With this, `other` is then a local copy instead of a reference to result. If we use `swap` then, it doesn't influence `result`. 

I know this might not be the performance optimal fix, but it is already used in similar scenarios such as in community.pyx: `getCover()`, so I thought I stick with this pattern. Let me know if another fix is preferred.  

There is also a new test to make sure that `getPartition()` is idempotent now. 

Co-developed with Claude (Claude Code).

### DIFF
--- a/networkit/community.pyx
+++ b/networkit/community.pyx
@@ -50,7 +50,7 @@ cdef extern from "<networkit/community/CommunityDetectionAlgorithm.hpp>":
 
 	cdef cppclass _CommunityDetectionAlgorithm "NetworKit::CommunityDetectionAlgorithm"(_Algorithm):
 		_CommunityDetectionAlgorithm(const _Graph &_G)
-		_Partition &getPartition() except +
+		_Partition getPartition() except +
 
 cdef class CommunityDetector(Algorithm):
 	""" 
--- a/networkit/test/test_community.py
+++ b/networkit/test/test_community.py
@@ -248,11 +248,22 @@ class TestCommunity(unittest.TestCase):
 
 	def testPLMProlong(self):
 		PLM = nk.community.PLM(self.L)
-		PLM.run()	
-		PLMP = PLM.getPartition()		
+		PLM.run()
+		PLMP = PLM.getPartition()
 		coarse = PLM.coarsen(self.L, PLMP)
 		part = PLM.prolong(coarse[0], PLMP, self.L, coarse[1])
 		self.assertDictEqual(part.subsetSizeMap(), {0:9})
+
+	def testPLMGetPartitionIdempotent(self):
+		plm = nk.community.PLM(self.L)
+		plm.run()
+		p1 = plm.getPartition()
+		p2 = plm.getPartition()
+		self.assertGreater(p1.numberOfElements(), 0)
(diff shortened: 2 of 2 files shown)


fabratu:
Your explanation makes sense, and the most forward fix would be to replicate the copy of memory from `Cover` (and other examples in NetworKit). Given the time, I would prefer a view approach if possible (haven't yet practically tried a full integration).

Here is a small POC:

```
cdef class PartitionView:
	cdef const _Partition* p
	cdef object owner   # keeps original object alive

	@staticmethod
	cdef PartitionView make(const _Partition* p, object owner):
		cdef PartitionView v = PartitionView()
		v.p = p
		v.owner = owner
		return v

	def __len__(self):
		return self.p.numberOfElements()

	def numberOfSubsets(self):
		return self.p.numberOfSubsets()

cdef class CommunityDetector(Algorithm):

	cdef Graph _G
	def __init__(self, *args, **namedargs):
		if type(self) == CommunityDetector:
			raise RuntimeError("Error, you may not use CommunityDetector directly, use a sub-class instead")

	def getPartition(self):
		"""  
		getPartition()
		
		Returns a partition of the clustering.

		Returns
		-------
		networkit.Partition
			A Partition of the clustering.
		"""
		if self._this == NULL:
			raise RuntimeError("Error, object not properly initialized")
		cdef _Partition* p = &(<_CommunityDetectionAlgorithm*>(self._this)).getPartition()
		return PartitionView.make(p, self)
```

```
>>> import networkit as nk
>>> G = nk.generators.BarabasiAlbertGenerator(5,100).generate()
>>> plm = nk.community.PLM(G)
>>> plm.run()
<networkit.community.PLM object at 0x124f90670>
>>> part = plm.getPartition()
>>> part
<networkit.community.PartitionView object at 0x125560240>
>>> part.numberOfSubsets()
6
>>> len(part)
100
>>> part = plm.getPartition()
>>> part.numberOfSubsets()
6
>>> len(part)
100
```

With this approach, we could also exchange other referenced large return types whenever these are assumed as read-only.

fabratu:
Gentle ping ... do you still intend to work on this feature? :)

domischweisgut:
Yes, I will try to close it asap, sorry for the delay :)

Issue #1415, opened 2026-05-17
Templated graph library proposal


angriman:
## Motivation

NetworKit’s current `Graph` uses fixed 64-bit types for nodes and edge weights, limiting flexibility and scalability. #1324 showed that templating node and edge weight types can yield consistent performance improvements. Further, by templating the graph type in algorithms, it will possible to leverage other more performant graph data structures like the one proposed in #1246. Here we propose an incremental plan for migrating NetworKit to a template library.

## Goals

- Add new template parameters for node and edge weight types to `Graph`.
- Migrate core data structures and algorithms to use template node and edge weight types.
- Expose new "lightweight" graphs to Python (e.g., 32-bit nodes). In the future, we might want to migrate to 32-bit graphs by default in Python.

### Templated graph

Introduce `AdjListGraph<NodeT, EdgeWeightT>` and define `using Graph = AdjListGraph<node, edgeweight>;` to preserve the existing Graph name and behavior.

Further, we propose to use the template implementation patter to avoid maintaining too large header files.

### Data structures

Data structures used in algorithms will have to be templated following the same pattern as in `Graph`, i.e., introduce a new name for the template data structure and add an alias to maintain the behavior of the non-template data structure unchanged. For example, BucketPQ:

```cpp
// BucketPQ.hpp

#include <networkit/Globals.hpp>

namespace Aux {

// New template bucket priority queue.
template <class KeyType, class ValueType>
class BucketPriorityQueue : public PrioQueue<KeyType, ValueType> {
    static_assert(std::is_signed_v<KeyType> && std::is_integral_v<KeyType>,
              "KeyType must be a signed integer");
   // ...
};

// Alias for the original BucketPQ. This is necessary to maintain compatibility with non-template algorithms.
using BucketPQ = BucketPriorityQueue<int64_t, NetworKit::index>;

} // namespace Aux
```

### Algorithms

Algorithms will have a new `GraphT` template parameter and use aliases for node and edge types:
  ```cpp
  template <class GraphT>
  class MyAlgo {
      using NodeT = typename GraphT::NodeT;
      using EdgeWeightT = typename GraphT::EdgeWeightT;
  };
  ```

### Tests

After templating a data structure / algorithm, tests should be parametrized on the new template parameters. I'd avoid any combination of all possible template parameters, that would make tests too expensive. I'd suggest covering common use cases, for example in #1324 we used:

```cpp
[...]

PR #1416, opened 2026-05-17
Add template node and edge types to Graph

angriman:
This PR adds `NodeT` and `EdgeWeightT` template parameters to `Graph`, iterator classes, and to utility classes such as `Edge`, `WeightedEdge`, etc. Further details are available at #1324 and  #1415.


include/networkit/graph/Graph.hpp
```diff
@@ -1494,7 +1472,7 @@ class Graph final {
      * @return @a i-th (outgoing) neighbor of @a u, or @c none if no such
      * neighbor exists.
      */
-    node getIthNeighbor(Unsafe, node u, index i) const { return outEdges[u][i]; }
+    node getIthNeighbor(Unsafe, NodeT u, index i) const { return outEdges[u][i]; }
```
Schwarf: This returns `node`. Is this correct or should it be `NodeT`?

include/networkit/graph/EdgeUtils.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+template <class NodeT, class EdgeWeightT>
+inline bool operator<(const WeightedEdge<NodeT, EdgeWeightT> &e1,
+                      const WeightedEdge<NodeT, EdgeWeightT> &e2) {
+    return e1.weight < e2.weight;
+}
+
+// We expose to Cython the default 64-bit Edge and WeightedEdge.
+using _CythonEdge = Edge<node>;
+using _CythonWeightedEdge = WeightedEdge<node, edgeweight>;
```
Schwarf: I think the `Globals.hpp` include is missing here.
Neither `node`, nor `edgeweight` are defined here.

include/networkit/graph/Graph.hpp
```diff
@@ -447,11 +425,11 @@ class Graph final {
         typename std::enable_if<
             (Aux::FunctionTraits<F>::arity >= 2)
             && std::is_same<edgeid, typename Aux::FunctionTraits<F>::template arg<2>::type>::value
-            && std::is_same<node, typename Aux::FunctionTraits<F>::template arg<1>::type>::
+            && std::is_same<NodeT, typename Aux::FunctionTraits<F>::template arg<1>::type>::
                 value /* prevent f(v, weight, eid)
                        */
             >::type * = (void *)0>
-    auto edgeLambda(F &f, node u, node v, edgeweight, edgeid id) const -> decltype(f(u, v, id)) {
+    auto edgeLambda(F &f, NodeT u, NodeT v, EdgeWeightT, edgeid id) const -> decltype(f(u, v, id)) {
```
Schwarf: We discussed before that we do not want to rule out integer types as weights. That means `NodeT` and `EdgeWeightT` can be the same type e.g. `int32_t`. I am not sure if the `edgeLambda` mechanism relies on distinguishable types for `NodeT` and `EdgeWeightT`. Therefore, I would add unit tests that cover these cases.

include/networkit/graph/Graph.hpp
```diff
@@ -763,11 +745,11 @@ class Graph final {
                     // initialize both inEdgeWeights and outEdgeWeights with the
                     // defaultEdgeWeight
                     inEdgeWeights.resize(z);
-                    for (node u = 0; u < z; ++u) {
+                    for (NodeT u = 0; u < z; ++u) {
```
Schwarf, line 690: In the line below, `defaultEdgeweight` is defined as `double` in `Globals.hpp`. That would break for graphs using `float` as `EdgeWeightT`

include/networkit/graph/NodeIterators.hpp
```diff
@@ -17,16 +17,16 @@ namespace NetworKit {
 /**
  * Class to iterate over the nodes of a graph.
  */
-template <typename GraphType>
+template <template <class, class> class GraphType, class NodeT, class EdgeWeightT>
 class NodeIteratorBase {
 
-    const GraphType *G;
+    const GraphType<NodeT, EdgeWeightT> *G;
     node u{none};
```
Schwarf: `NodeT` instead of `node`?

include/networkit/graph/NodeIterators.hpp
```diff
@@ -44,7 +44,7 @@ class NodeIteratorBase {
     // Own type.
     using self = NodeIteratorBase;
 
-    NodeIteratorBase(const GraphType *G, node u) : G(G), u(u) {
+    NodeIteratorBase(const GraphType<NodeT, EdgeWeightT> *G, node u) : G(G), u(u) {
```
Schwarf: `NodeT u ` instead of `node u`?

include/networkit/graph/Graph.hpp
```diff
@@ -306,7 +269,12 @@ class Graph final {
      * @return The edge id
      */
     template <bool graphHasEdgeIds>
-    inline edgeid getOutEdgeId(node u, index i) const;
+    inline edgeid getOutEdgeId(NodeT u, index i) const {
+        if constexpr (graphHasEdgeIds)
+            return outEdgeIds[u][i];
+        else
+            return none;
```
Schwarf: In the `else`-branch the function returns `none` of type `NodeT` but the return-type of the function is `edgeid`

include/networkit/graph/Graph.hpp
```diff
@@ -317,7 +285,12 @@ class Graph final {
      * @return The edge id
      */
     template <bool graphHasEdgeIds>
-    inline edgeid getInEdgeId(node u, index i) const;
+    inline edgeid getInEdgeId(NodeT u, index i) const {
+        if constexpr (graphHasEdgeIds)
+            return inEdgeIds[u][i];
+        else
+            return none;
```
Schwarf: In the else-branch the function returns none of type `NodeT` but the return-type of the function is `edgeid`

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+    if (!directed) {
+        return indexInOutEdgeArray(v, u);
+    }
+    for (index i = 0; i < inEdges[v].size(); i++) {
+        NodeT x = inEdges[v][i];
+        if (x == u) {
+            return i;
+        }
+    }
+    return none;
```
Schwarf, line 2381: The function return-type is `index` but `none` is of type `NodeT`. I think we need a `none-index` or something similar.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+
+template <class NodeT, class EdgeWeightT>
+index AdjListGraph<NodeT, EdgeWeightT>::indexInOutEdgeArray(NodeT u, NodeT v) const {
+    for (index i = 0; i < outEdges[u].size(); i++) {
+        NodeT x = outEdges[u][i];
+        if (x == v) {
+            return i;
+        }
+    }
+    return none;
```
Schwarf, line 2392: The function return-type is `index` but `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+
+template <class NodeT, class EdgeWeightT>
+edgeid AdjListGraph<NodeT, EdgeWeightT>::edgeId(NodeT u, NodeT v) const {
+    if (!edgesIndexed) {
+        throw std::runtime_error("edges have not been indexed - call indexEdges first");
+    }
+
+    index i = indexInOutEdgeArray(u, v);
+
+    if (i == none) {
```
Schwarf, line 2462: This comparison would not work as expected if `none` is not `std::numeric_limits<uint64_t>::max()` which is not guaranteed since `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+}
+
+template <class NodeT, class EdgeWeightT>
+EdgeWeightT
+AdjListGraph<NodeT, EdgeWeightT>::computeWeightedDegree(NodeT u, bool inDegree,
+                                                        bool countSelfLoopsTwice) const {
+    if (weighted) {
+        EdgeWeightT sum = 0.0;
+        auto sumWeights = [&](NodeT v, EdgeWeightT w) {
+            sum += (countSelfLoopsTwice && u == v) ? 2. * w : w;
```
Schwarf: `2. * w` promotes to double, then narrows back to `EdgeWeightT`.Maybe use `EdgeWeightT{2}` .

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+
+    const auto isLoop = (u == v);
+    m--; // decrease number of edges
+    if (isLoop)
+        storedNumberOfSelfLoops--;
+
+    // remove edge for source node
+    erase<NodeT>(u, vi, outEdges);
+    if (weighted) {
+        erase<EdgeWeightT>(u, vi, outEdgeWeights);
```
Schwarf: The function template `erase` has as template arguments `class NodeT, typename T`. Maybe call `erase<NodeT, EdgeWeightT>(...)`

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+    }
+
+    index vi = indexInOutEdgeArray(u, v); // index in outEdges array
+    index ui = indexInInEdgeArray(v, u);  // index in inEdges array
+
+    if (edgesIndexed) {
+        deletedID = edgeId(u, v);
+    }
+
+    if (vi == none) {
```
Schwarf, line 2909: As above `vi` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+        });
+        // use erase to remove data entry at index `deletedID` and compact the data vector again
+        auto &theMap = edgeAttributeMap.attrMap;
+        for (auto it = theMap.begin(); it != theMap.end(); ++it) {
+            auto attributeStorageBase = it->second.get();
+            attributeStorageBase->erase(deletedID);
+        }
+    }
+    if (directed) {
+        assert(ui != none);
```
Schwarf, line 2994: As above `ui` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+    }
+
+    m -= removedEdges + removedSelfLoops;
+    storedNumberOfSelfLoops -= removedSelfLoops;
+}
+
+template <class NodeT, class EdgeWeightT>
+void AdjListGraph<NodeT, EdgeWeightT>::swapEdge(NodeT s1, NodeT t1, NodeT s2, NodeT t2) {
+    index s1t1 = indexInOutEdgeArray(s1, t1);
+    if (s1t1 == none)
```
Schwarf, line 3092: As above `s1t1` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+        throw std::runtime_error("The first edge does not exist");
+    index t1s1 = indexInInEdgeArray(t1, s1);
+
+    index s2t2 = indexInOutEdgeArray(s2, t2);
+    if (s2t2 == none)
```
Schwarf, line 3097: As above `s2t2` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+        return indexInOutEdgeArray(u, v) != none;
+    }
+}
+
+/** EDGE ATTRIBUTES **/
+
+template <class NodeT, class EdgeWeightT>
+EdgeWeightT AdjListGraph<NodeT, EdgeWeightT>::weight(NodeT u, NodeT v) const {
+    index vi = indexInOutEdgeArray(u, v);
+    if (vi == none) {
```
Schwarf, line 3145: As above `vi` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+}
+
+template <class NodeT, class EdgeWeightT>
+void AdjListGraph<NodeT, EdgeWeightT>::setWeight(NodeT u, NodeT v, EdgeWeightT ew) {
+    if (!weighted) {
+        throw std::runtime_error("Cannot set edge weight in unweighted graph.");
+    }
+
+    index vi = indexInOutEdgeArray(u, v);
+    if (vi == none) {
```
Schwarf, line 3159: As above `v1` is of type `index` and  `none` is of type `NodeT`.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2332,1026 @@ void Graph::sortEdges(Lambda lambda) {
...
+}
+
+template <class NodeT, class EdgeWeightT>
+void AdjListGraph<NodeT, EdgeWeightT>::increaseWeight(NodeT u, NodeT v, EdgeWeightT ew) {
+    if (!weighted) {
+        throw std::runtime_error("Cannot increase edge weight in unweighted graph.");
+    }
+
+    index vi = indexInOutEdgeArray(u, v);
+    if (vi == none) {
```
Schwarf, line 3183: As above `vi` is of type `index` and  `none` is of type `NodeT`.

Schwarf (changes requested):

include/networkit/graph/Graph.hpp
angriman: Right, good catch, done.

include/networkit/graph/Graph.hpp
angriman: I've added some minimal tests for `AdjListGraph` where we use the same type for `NodeT` and `EdgeWeightT`.

angriman:
Rebased on top of #1420.

include/networkit/graph/EdgeUtils.hpp
angriman: Correct, added.

include/networkit/graph/NodeIterators.hpp
angriman: Done.

include/networkit/graph/NodeIterators.hpp
angriman: Right, done.

include/networkit/graph/Graph.hpp
angriman: Good point, I defined a new `nullEdgeId` constant for an undefined edge id. This is the same as the original `none` constant (infinity unit64_t).

include/networkit/graph/Graph.hpp
angriman: Right, I think a safer approach here is to replace `none` with `nullNodeId` where used with `NodeT` and keep `none` when used for indices.

include/networkit/graph/Graph.hpp
angriman: resolved by avoiding too override `none` as `NodeT`

include/networkit/graph/Graph.hpp
angriman: Right, done.

include/networkit/graph/Graph.hpp
angriman: Actually, I think that we can remove explicit template parameters here, they are automatically deduced.

include/networkit/graph/Graph.hpp
angriman: This was indeed causing problems due to name clashes when the two types are the same, I've restructured `edgeLambda` to use a compile-time dispatcher logic to avoid clashes.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2276,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+/** CONSTRUCTORS **/
+
+template <class NodeT, class EdgeWeightT>
+AdjListGraph<NodeT, EdgeWeightT>::AdjListGraph(count n, bool weighted, bool directed,
+                                               bool edgesIndexed)
+    : n(n), m(0), storedNumberOfSelfLoops(0), z(n), omega(0), t(0),
+
+      weighted(weighted), // indicates whether the graph is weighted or not
+      directed(directed), // indicates whether the graph is directed or not
+      edgesIndexed(edgesIndexed), deletedID(nullNodeId),
```
Schwarf: `deletedID` is of type `edgeId` so I think `nullEdgeId` should be used.

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2276,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+template <class NodeT, class EdgeWeightT>
+void AdjListGraph<NodeT, EdgeWeightT>::indexEdges(bool force) {
+    if (edgesIndexed && !force)
+        return;
+
+    omega = 0; // reset edge ids (for re-indexing)
+
+    outEdgeIds.clear(); // reset ids vector (for re-indexing)
+    outEdgeIds.resize(outEdges.size());
+    forNodes([&](NodeT u) { outEdgeIds[u].resize(outEdges[u].size(), nullNodeId); });
```
Schwarf: Should this be initialized with `nullEdgeId` instead of `nullNodeId`?

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2276,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+
+    if (directed) {
+        inEdgeIds.resize(inEdges.size());
+        forNodes([&](NodeT u) { inEdgeIds[u].resize(inEdges[u].size(), nullNodeId); });
```
Schwarf: See above: `nullNodeId` --> `nullEdgeId`?

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2276,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+    });
+
+    // copy edge ids for the edges in the other direction. Note that
+    // "indexInOutEdgeArray" is slow which is why this second loop in parallel
+    // makes sense.
+    if (!directed) {
+        balancedParallelForNodes([&](NodeT u) {
+            for (index i = 0; i < outEdges[u].size(); ++i) {
+                NodeT v = outEdges[u][i];
+                if (v != nullNodeId && outEdgeIds[u][i] == nullNodeId) {
```
Schwarf: Should the second comparison be against `nullEdgeId` instead?
`outEdgeIds[u][i] == nullEdgeId`

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2276,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+    }
+}
+
+template <class NodeT, class EdgeWeightT>
+bool AdjListGraph<NodeT, EdgeWeightT>::hasEdge(NodeT u, NodeT v) const noexcept {
+    if (u >= z || v >= z) {
+        return false;
+    }
+    if (!directed && outEdges[u].size() > outEdges[v].size()) {
+        return indexInOutEdgeArray(v, u) != nullNodeId;
```
Schwarf: Should this comparison (and the two following) be made with `none`

include/networkit/graph/Graph.hpp
```diff
@@ -1637,11 +1559,11 @@ class Graph final {
...
      * @param i index; should be in [0, degreeOut(u))
      * @return pair: i-th (outgoing) neighbor of @a u and the corresponding
-     * edge weight, or @c defaultEdgeWeight if unweighted.
+     * edge weight, or @c defaultEdgeWeightT if unweighted.
      */
-    std::pair<node, edgeweight> getIthNeighborWithWeight(node u, index i) const {
+    std::pair<NodeT, EdgeWeightT> getIthNeighborWithWeight(NodeT u, index i) const {
         if (!hasNode(u) || i >= outEdges[u].size())
-            return {none, none};
+            return {nullNodeId, nullNodeId};
```
Schwarf: I think the second return value should be `nullEdgeId`

include/networkit/graph/Graph.hpp
```diff
@@ -104,7 +52,10 @@ class CurveballMaterialization;
  * @ingroup graph
  * A graph (with optional weights) and parallel iterator methods.
  */
-class Graph final {
+template <class NodeT, class EdgeWeightT>
```
Schwarf, line 55: Minor: Using C++20 concepts this could be written as:
```
template <class NodeT, class EdgeWeightT>
    requires std::integral<NodeT> && std::arithmetic<EdgeWeightT>
class AdjListGraph final {
```

or even more concise:
```
template <std::integral NodeT, std::arithmetic EdgeWeightT>
class AdjListGraph final {
```
Then the `static_asserts` below can be dropped.

include/networkit/Globals.hpp
```diff
@@ -30,11 +30,19 @@ using node = index;        ///< node indices are 0-based
...
+inline constexpr edgeweight defaultEdgeWeight = 1.0;
+inline constexpr edgeweight nullWeight = 0.0;
+inline constexpr index none =
+    std::numeric_limits<index>::max(); ///< value for not existing nodes/edges
+
+inline constexpr double PI =
+    3.141592653589793238462643383279502884197169399375105820974944592307816406286;
+
+template <class NodeT>
+struct NodeTraits {
```
Schwarf: Minor (simplification): I think we can omit the struct and define `nullNodeId` directly as:
```
static constexpr NodeT nullNodeId = std::numeric_limits<NodeT>::max();
```
in `Graph.hpp`.

include/networkit/graph/Graph.hpp
```diff
@@ -139,29 +90,31 @@ class Graph final {
...
     //!< same schema (and same order!) as outEdges
-    std::vector<std::vector<edgeweight>> outEdgeWeights;
+    std::vector<std::vector<EdgeWeightT>> outEdgeWeights;
 
     //!< only used for directed graphs, same schema as inEdges
     std::vector<std::vector<edgeid>> inEdgeIds;
     //!< same schema (and same order!) as outEdges
     std::vector<std::vector<edgeid>> outEdgeIds;
 
+    static constexpr NodeT nullNodeId = NodeTraits<NodeT>::null;
```
Schwarf: Minor simplification:
```
static constexpr NodeT nullNodeId = std::numeric_limits<NodeT>::max();
```

Schwarf (changes requested):

include/networkit/Globals.hpp
angriman: I prefer keeping it in `Globals.hpp` because `nullNodeId` is not something specific to `AdjListGraph` but should be used by other graph implementations and algorithms, WDYT?

Schwarf: Oh, yes you are right. Then maybe we can use a variable template (with concept requirements):
```
template <std::integral NodeT>
inline constexpr NodeT nullNodeId = std::numeric_limits<NodeT>::max();
```
or without:
```
template <class NodeT>
inline constexpr NodeT nullNodeId = std::numeric_limits<NodeT>::max();
```

include/networkit/graph/Graph.hpp
Schwarf: As discussed above, not an option.

include/networkit/graph/Graph.hpp
angriman: True, fixed this and also similar comments below.

include/networkit/graph/Graph.hpp
angriman: This is a potentially very interesting idea. I'm ok in using it, but I noticed that:
- `std::arithmetic` does not exist, we have to define a custom concept for that, e.g.,:
```cpp
template <typename T>
concept EdgeWeight = std::integral<T> || std::floating_point<T>;
```
- It requires using those concepts not only in `AdjListGraph`, but in any place we use `NodeT` and `EdgeWeightT`.
- There are issues with GCC, I haven't figure it out yet how to fix them: Screenshot 2026-05-25 at 19 25 53

The gcc issue is a blocker for me, unless we manage to fix it, I'd suggest keeping the `static_assert`s for now.

Schwarf: Hmm, GCC 10 seems to be the first GCC version with C++20 concepts support, so this might be related to early concepts support.

I agree that this should not block the first refactoring. We can discuss the options and next steps in our meeting.

include/networkit/graph/Graph.hpp
Schwarf: I think the second comparison in the `if`-condition is not correct since `NodeT != edgeid`

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2275,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+        if (directed) {
+            if (degreeIn(u) == 0) {
+                inEdges[u].clear();
+                if (weighted)
+                    inEdgeWeights[u].clear();
+                if (edgesIndexed)
+                    inEdgeIds[u].clear();
+            } else {
+                for (index i = 0; i < inEdges[u].size(); ++i) {
+                    while (i < inEdges[u].size() && inEdges[u][i] == nullEdgeId) {
```
Schwarf: `inEdges[u][i]` returns a `NodeT`, so the second comparison should use `nullNodeId` instead of an edge ID: `inEdges[u][i] == nullNodeId

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2275,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+    this->parallelForNodes([&](NodeT u) {
+        if (degreeOut(u) == 0) {
+            outEdges[u].clear();
+            if (weighted)
+                outEdgeWeights[u].clear();
+            if (edgesIndexed)
+                outEdgeIds[u].clear();
+        } else {
+            for (index i = 0; i < outEdges[u].size(); ++i) {
+                while (i < outEdges[u].size() && outEdges[u][i] == nullEdgeId) {
```
Schwarf: `outEdges[u][i]` returns a `NodeT`, so the second comparison should use `nullNodeId` instead of an edge ID: `outEdges[u][i] == nullNodeId`

include/networkit/graph/Graph.hpp
```diff
@@ -2419,6 +2275,1025 @@ void Graph::sortEdges(Lambda lambda) {
...
+    }
+}
+
+template <class NodeT, class EdgeWeightT>
+bool AdjListGraph<NodeT, EdgeWeightT>::hasEdge(NodeT u, NodeT v) const noexcept {
+    if (u >= z || v >= z) {
+        return false;
+    }
+    if (!directed && outEdges[u].size() > outEdges[v].size()) {
+        return indexInOutEdgeArray(v, u) != nullEdgeId;
```
Schwarf: Since `indexInOutEdgeArray` returns an `index` comparing with `none` would be preferable to `nullEdgeId`. `nullEdgeId` works for now only because `edgeid` is currently an alias for `index`.

Same in lines 3134 and 3136.

include/networkit/graph/GraphUtils.hpp
```diff
@@ -0,0 +1,81 @@
...
+        if (sorted) {
+            std::tie(u, v) = std::minmax(_u, _v);
+        } else {
+            u = _u;
+            v = _v;
+        }
+    }
+
+private:
+    static constexpr NodeT none = std::numeric_limits<NodeT>::max();
```
Schwarf: This might cause confusion. In `Globals.hpp` we defined 
```
template <class NodeT>
inline constexpr NodeT NullNodeId = std::numeric_limits<NodeT>::max();
```
and
```
inline constexpr index none =  std::numeric_limits<index>::max();
```
Here `none` is redefined 

I suggest to remove this line. The default constructor above could use `NullNodeId<NodeT>` directly

Schwarf (changes requested):

include/networkit/Globals.hpp
angriman: Used the version without concepts for now.

include/networkit/graph/Graph.hpp
angriman: right, done.

include/networkit/graph/Graph.hpp
angriman: Thanks, done.

include/networkit/graph/GraphUtils.hpp
angriman: I defined a private `nullNodeId` as done in `Graph.hpp`.

Schwarf:
Hi @angriman,
I am almost ready to approve. There is just one remaining finding that seems to have been marked as resolved, but I think it has not been addressed yet.

In `Graph.hpp`, around line 2432, the second comparison looks incorrect to me:

```cpp
if (v != nullNodeId && outEdgeIds[u][i] == nullNodeId)
```
Since `outEdgeIds[u][i]` refers to an edge id, I think this should compare against `nullEdgeId` instead:
```cpp
if (v != nullNodeId && outEdgeIds[u][i] == nullEdgeId)
```

By the way, for some reason I do not seem to be able to resolve review findings myself.

include/networkit/graph/Graph.hpp
fabratu: If that helps: Since GCC 11 came out April 27th 2021, we can switch compiler versions.

angriman:
> Hi @angriman, I am almost ready to approve. There is just one remaining finding that seems to have been marked as resolved, but I think it has not been addressed yet.
> 
> In `Graph.hpp`, around line 2432, the second comparison looks incorrect to me:
> 
> ```c++
> if (v != nullNodeId && outEdgeIds[u][i] == nullNodeId)
> ```
> 
> Since `outEdgeIds[u][i]` refers to an edge id, I think this should compare against `nullEdgeId` instead:
> 
> ```c++
> if (v != nullNodeId && outEdgeIds[u][i] == nullEdgeId)
> ```
> 
> By the way, for some reason I do not seem to be able to resolve review findings myself.

Right, I missed that. It should now be solved.

include/networkit/graph/Graph.hpp
angriman: Thanks, let's do that in another PR, for now we can keep static asserts.

Schwarf (approved):

include/networkit/graph/Graph.hpp
```diff
@@ -139,29 +89,31 @@ class Graph final {
     edgeid deletedID;
 
     // per node data
-    //!< exists[v] is true if node v has not been removed from the graph
+    //!< exists[v] is true if NodeT v has not been removed from the graph
```
fabratu: I think "node" is meant as a structural element of a graph here. Using "NoteT" might be confusing.

include/networkit/graph/Graph.hpp
```diff
@@ -242,26 +194,27 @@ class Graph final {
...
-    using EdgeStringAttribute = Attribute<PerEdge, Graph, std::string, false>;
+    using EdgeIntAttribute = Attribute<PerEdge, AdjListGraph, int, false>;
+    using EdgeDoubleAttribute = Attribute<PerEdge, AdjListGraph, double, false>;
+    using EdgeStringAttribute = Attribute<PerEdge, AdjListGraph, std::string, false>;
 
 private:
+    static constexpr EdgeWeightT defaultEdgeWeightT = EdgeWeightT{1};
     /**
-     * Returns the index of node u in the array of incoming edges of node v.
+     * Returns the index of NodeT u in the array of incoming edges of NodeT v.
```
fabratu: Same as above. There are also more occasions below (not marking everything). I think in the doc we should separate the description of graph entities (nodes, edges) and their respective data types. As a rule of thumb, I suggest we use graph entities in doc descriptions and data types in the params section.

include/networkit/graph/GraphUtils.hpp
```diff
@@ -0,0 +1,81 @@
+#ifndef NETWORKIT_GRAPH_GRAPH_UTILS_HPP_
```
fabratu: I suggest not naming this file `GraphUtils` in order not to confuse it with `GraphTools`. Maybe something like `EdgeUtils.hpp`?

include/networkit/graph/EdgeUtils.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <networkit/Globals.hpp>
+
+namespace NetworKit {
+
+struct Unsafe {};
+static constexpr Unsafe unsafe{};
+
+template <class NodeT>
+struct Edge {
```
fabratu: This basically breaks public code using Edge/WeightedEdge (and so on). Maybe better use the same patterns as with `using Graph = AdjListGraph<node, edgeweight>`?

fabratu (changes requested):
I could not spot any more bugs/problematic decisions except for a few comments, which we can deal with.

Apart from that, we also discussed splitting implementation and header code. Especially for `Graph.hpp` I would suggest doing so. How about having `Graph.hpp` and `Graph_impl.hpp`? Maybe even better, use `AdjListGraph.hpp` and `AdjListGraph_impl.hpp` and have `Graph.hpp` just define the template specialization?

include/networkit/graph/Graph.hpp
angriman: Thanks for spotting it, fixed it also in other places.

include/networkit/graph/Graph.hpp
angriman: Sure, changed in all other places.

include/networkit/graph/GraphUtils.hpp
angriman: Changed to `EdgeUtils`.

include/networkit/graph/EdgeUtils.hpp
angriman: Good point, I've renamed those to `EdgeT` and `WeightedEdgeT` and left `Edge` and `WeightedEdge` untouched.

fabratu:
Imho, the PR is in good shape now and a valid starting point. 

The only thing left is the splitting of `Graph.hpp` into two files. Are we all on the same page concerning this? I am also fine with postponing this design question and focusing on finalizing this PR.

Schwarf:
I agree that the `Graph.hpp` split is not required to finalize this PR. I would therefore prefer to keep the current PR focused and handle the split separately. I suggest creating a follow-up issue so that the discussion and the remaining work are not lost.

angriman:
@fabratu sorry I missed the comment about splitting `Graph`. 

I created #1415 to describe the plan and keep track of progress. It does not yet include details about how to split `Graph`, I'll add them shortly. To me having `AdjListGraph.hpp` and `AdjListGraph_impl.hpp` looks like a good solution. @Schwarf do you agree?

Schwarf:
Yes, I agree. `AdjListGraph.hpp` and `AdjListGraph_impl.hpp` sound like a good solution.

angriman:
@fabratu Should we merge this?

fabratu (approved):

PR #1422, opened 2026-06-05
feat: `GraphFromCsr` method

ilan-gold:
An optimized version of this means that graph construction is very fast and doesn't involve an intermediate copy at the expense of a slightly messy C++ API.

Open to suggestions here, but this is very fast for big graphs and `GraphFromCoo`'s iteration is much slower. This seems to be related to general wishes for faster construction times: #1335 and #1331 without introducing any new dependencies

~I suppose this construction procedure could also be applied to COO matrices.  I'd imagine you only want to maintain one of these fast paths since going from COO->CSR or CSR->COO is fast/cheap (aside from the memory)~ No, this sort of procedure probably can't be applied in general to COO matrices because they are not guaranteed to be in any sort of sensible order (where nodes' edges to are grouped together).  You could in theory do a first pass to get the number of in/out edges per node and allocate followed by parallel filling of the buffers, but it's quite different and probably doesn't parallelize amazingly since the memory access wouldn't be guaranteed to be contiguous.  But maybe I'm wrong here.


fabratu:
Hi, thanks for your contribution! 

This would be a nice addition.

Instead of using raw C-style pointers for `indptr` etc., use `const &vector[index]`. This avoids pitfalls from unsafe pointers and mimics the API from `CSRGeneralMatrix`.

It should be of no additional cost compared to your proposed solution.

```
...
indptr_arr = np.ascontiguousarray(indptr_arr, dtype=np.int64)
rowIdx = vector[int64_t](&indptr_arr [0], &indptr_arr[0] + indptr_arr.shape[0])
...
```

ilan-gold:
> It should be of no additional cost compared to your proposed solution.

Just to make sure I understand, the current implementation (as far as I could try) uses pointers to the underlying buffers of the user-provided CSR matrix to construct the graph (i.e., not intermediate buffers generated).  But would

```
rowIdx = vector[int64_t](&indptr_arr [0], &indptr_arr[0] + indptr_arr.shape[0])
```

generate a copy? I'm happy to do this, but just want to make sure we're on the same page (and also so I can learn something - first c++ PR!).

Thanks for the reply! We're really looking forward to using networkit!

fabratu:
You are right concerning the copy for vectors, since they become owners of the data. I was not thinking to the end here. After doing so (the thinking part), I don't see how we can avoid the pointer-based API for Cython -> C++.

A cleaner approach from the C++ perspective would be, though, to use `span` (essentially a view of the data) for the public API:

In C++:

```
Graph fromCSR(
    std::span<const index> indptr,
    std::span<const index> indices,
    std::span<const double> data,
    count nRows,
    count nCols,
    bool directed,
    bool weighted
) { ...}

Graph _fromCSRRaw(
    const index* indptr,
    std::size_t indptrSize,
    const index* indices,
    std::size_t indicesSize,
    const double* data,
    std::size_t dataSize,
    count nRows,
    count nCols,
    bool directed,
    bool weighted
) {
    return fromCSR(
        std::span<const index>(indptr, indptrSize),
        std::span<const index>(indices, indicesSize),
        std::span<const double>(data, dataSize),
        nRows,
        nCols,
        directed,
        weighted
    );
}
```

And in Cython:

```
cdef const index* indptr_ptr = <const index*> indptr_arr.data
cdef const index* indices_ptr = <const index*> indices_arr.data
cdef const double* data_ptr = <const double*> data_arr.data

cdef size_t indptr_size = indptr_arr.shape[0]
cdef size_t indices_size = indices_arr.shape[0]
cdef size_t data_size = data_arr.shape[0]

cdef _Graph tmp = _Graph._fromCSRRaw(
    indptr_ptr,
    indptr_size,
    indices_ptr,
    indices_size,
    data_ptr,
    data_size,
    nRows,
    nCols,
    directed,
    weighted
)
...
```

In `_fromCSRRaw` in C++ the documentation should include a hint, that the function is only added due to Cython compatibility and is not intended to use for on its own.

fabratu:
The upcoming Cython 3.3.0 will include `std::span` in libcpp. This could then also avoid the helper function in a future PR.

ilan-gold:
Hi @fabratu I used `span` - I also removed the `nRows` argument from the C++ because I think it can be resolved from the `indptr` directly but maybe I have this wrong? Or you'd still like it?

include/networkit/graph/Graph.hpp
```diff
@@ -880,6 +881,29 @@ class Graph final {
         return *this;
     };
 
+    /**
+     * Create a Graph from CSR (compressed sparse row) arrays. The spans are
+     * views over the compressed row-pointer array (@a indptr), the column
+     * indices array (@a indices) and the non-zero values (@a data). The number
+     * of rows is derived from @a indptr, whose length is the row count plus
```
fabratu: `The number of nodes ...` (since we construct a graph from CSR)

include/networkit/graph/Graph.hpp
```diff
@@ -880,6 +881,29 @@ class Graph final {
...
+     * one. The referenced data is copied into internal storage, so the spans
+     * need not outlive the call.
+     */
+    static Graph fromCSR(std::span<const index> indptr, std::span<const index> indices,
```
fabratu: Not a strong one, but I suggest we mimic the naming scheme of `CSRGeneralMatrix` and add a hint about the nature of the parameter:

indptr -> rowIdxView
indices -> columnIdxView
data -> nonZerosView

include/networkit/graph/Graph.hpp
```diff
@@ -880,6 +881,29 @@ class Graph final {
...
+                         bool isWeighted = false);
+
+    /**
+     * Raw-pointer overload of fromCSR(), forwarding to it via std::span.
+     *
+     * This exists ONLY for Cython interoperability: Cython cannot yet
+     * construct a std::span directly (support is planned for Cython 3.3.0).
+     * Prefer fromCSR() in C++ code; do not use this function on its own.
+     */
+    static Graph _fromCSRRaw(const index *indptr, std::size_t indptrSize, const index *indices,
```
fabratu: Same as above (indptr -> rowIdxPtr ...)

networkit/cpp/graph/Graph.cpp
```diff
@@ -990,4 +991,120 @@ bool Graph::checkConsistency() const {
...
+        if (rowEnd < rowStart) {
+            throw std::invalid_argument("indptr must be non-decreasing");
+        }
+        rowSizes[i] = static_cast<size_t>(rowEnd - rowStart);
+        graph.outEdges[i].resize(rowSizes[i]);
+        if (isWeighted)
+            graph.outEdgeWeights[i].resize(rowSizes[i]);
+    }
+
+#pragma omp parallel for schedule(static)
```
fabratu: Since workload for threads is usually imbalanced here, better use `schedule(guided)` instead of `static`.

networkit/cpp/graph/Graph.cpp
```diff
@@ -990,4 +991,120 @@ bool Graph::checkConsistency() const {
...
+    return graph;
+}
+
+Graph Graph::_fromCSRRaw(const index *indptr, std::size_t indptrSize, const index *indices,
+                         std::size_t indicesSize, const double *data, std::size_t dataSize,
+                         bool directed, bool isWeighted) {
+    // This helper only adapts raw pointers to std::span for callers (e.g. Cython)
+    // that cannot construct a span directly; the real work happens in fromCSR().
+    if (indptr == nullptr || indices == nullptr) {
+        throw std::invalid_argument("null CSR pointer passed to _fromCSRRaw");
```
fabratu: If `isWeighted == true`, we should also check for `nullptr` in `data`.

networkit/graph.pxd
```diff
@@ -23,6 +24,14 @@ cdef extern from "<networkit/Globals.hpp>" namespace "NetworKit":
 
 	index _none "NetworKit::none"
 
+cdef extern from "<networkit/algebraic/CSRGeneralMatrix.hpp>" namespace "NetworKit":
```
fabratu: Imho, you can drop this addition. If this is needed, it should be included in `algebraic.pyx/pxd` (not yet present). Also, the declaration misses its implementation part.

fabratu (changes requested):

fabratu:
> Hi @fabratu I used `span` - I also removed the `nRows` argument from the C++ because I think it can be resolved from the `indptr` directly but maybe I have this wrong? Or you'd still like it?

It's fine to drop it. Like you said, it can be inferred.

networkit/graph.pxd
ilan-gold: Sorry about this, meant to remove it.  Was from a previous attempt to use this directly

networkit/cpp/graph/Graph.cpp
```diff
@@ -990,4 +991,124 @@ bool Graph::checkConsistency() const {
...
+    std::vector<size_t> rowSizes(nRows);
+    graph.outEdges.resize(nRows);
+    if (isWeighted) {
+        graph.outEdgeWeights.resize(nRows);
+    }
+
+    for (index i = 0; i < nRows; ++i) {
+        index rowStart = rowIdxView[i];
+        index rowEnd = rowIdxView[i + 1];
+        if (rowEnd < rowStart) {
```
fabratu: Such an `if` is costly and we don't expect to consume custom-made CSR-structures here. Better use `assert(...)`, such that this is tested in debug builds.

fabratu (changes requested):
Imho, the PR is in a good shape.

Before we can merge it, please add tests as well for C++. For this add tests in `GraphGTest.cpp`, checking for both succesfull and unsuccesfull cases in the implementation. You don't need to check `_graphFromCsr`, since it is only used for Cython.

For "successfull" add a small hand-crafted example graph, for "unsuccessful" try to trigger throws/corner cases.

fabratu (dismissed):

fabratu:
@ilan-gold We recently merged a larger feature (see PR #1416), changing Graph to be a templated class. For merging your contribution, we will need to rebase. Sry, for the additional work :(

ilan-gold:
So it turns out I got confused about some inputs from scanpy and hardcoded symmetry into the CSR constructor, which threw off calculations like `totalEdgeWeight` - `Graph` adds both in and out edge weights for undirected and weighted graphs on `addEdge`.

So I've updated this PR to mimic that for CSR.

There was also a bug in the `GraphFromCoo` implementation that ignored edge weights from the data? I don't think that was intentional. I have a clean commit at the end here for that: https://github.com/networkit/networkit/pull/1422/changes/5daf1809d2f5dd0062e198cfbb7831f77dbd55c4

fabratu:
Good catch concerning `graphFromCoo`. So far, the function has assumed that edges are given in COO-format without any weights. Looking in the specification, this assumption seems incorrect in the current `scipy` version. Can you also adjust the documentation of the function?

fabratu (approved):

fabratu:
The last substantial feature code change was pushed more than 7 days ago and current changes from other PRs add the recurring need for rebasing this PR. Hence will merge right away.

PR #1424, opened 2026-06-10
graph: prevent edge callback crash for invalid nodes.

### LINKED ISSUE #1419: G.forEdgesOf segfaults for unknown node
Testcase:
```py
import networkit as nk

G = nk.graphio.EdgeListReader("\t", 0, "#", continuous=False).read("out.edgelist")

def edgeFunc(u, v, weight, edgeId):
    print("Edge from {} to {} has weight {} and id {}".format(u, v, weight, edgeId))

G.forEdgesOf(912, edgeFunc)
```

`out.edgelist` contains only

```
1008	1056
```

Node 912 doesn't exist, so `G.forEdgesOf` segfaults. It would be preferrable to throw an exception or otherwise gracefully handle the error.

I encountered this bug on networkit  `v11.2.1`

Schwarf:
Fixes a crash in the Python API when `Graph.forEdgesOf` or `Graph.forInEdgesOf` is called with a non-existing node.

Invalid or removed nodes now raise a RuntimeError instead of reaching unchecked C++ adjacency access.

Fixes #1419

### DIFF
--- a/networkit/graph.pyx
+++ b/networkit/graph.pyx
@@ -766,6 +766,9 @@ cdef class Graph:
 			Any callable object that takes the parameter tuple(int, int, float, int).
 			Parameter list refering to (node id, node id, edge weight, edge id).
 		"""
+		if not self._this.hasNode(u):
+			raise RuntimeError("Cannot iterate over edges of non-existing node {}".format(u))
+
 		cdef EdgeCallBackWrapper* wrapper
 		try:
 			wrapper = new EdgeCallBackWrapper(callback)
@@ -790,6 +793,9 @@ cdef class Graph:
 			Any callable object that takes the parameter tuple(int, int, float, int).
 			Parameter list refering to (node id, node id, edge weight, edge id).
 		"""
+		if not self._this.hasNode(u):
+			raise RuntimeError("Cannot iterate over in-edges of non-existing node {}".format(u))
+
 		cdef EdgeCallBackWrapper* wrapper
 		try:
 			wrapper = new EdgeCallBackWrapper(callback)
--- a/networkit/test/test_graph.py
+++ b/networkit/test/test_graph.py
@@ -672,6 +672,65 @@ class TestGraph(unittest.TestCase):
         self.assertEqual(un.getUMSF().numberOfEdges(), 5)
         self.assertTrue(un.inUMST(0, 1))
 
+    def testForInEdgesOfInvalidNodeRaises(self):
+        G = nk.Graph(2, directed=True)
+        G.addEdge(0, 1)
+
+        def edgeFunc(*args):
+            self.fail("Callback should not be called for a non-existing node")
+
+        with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot iterate over in-edges of non-existing node 912",
+        ):
+            G.forInEdgesOf(912, edgeFunc)
+
+
+    def testForInEdgesOfRemovedNodeRaises(self):
+        G = nk.Graph(3, directed=True)
+        G.addEdge(0, 1)
+        G.addEdge(2, 1)
+
+        G.removeNode(1)
+
+        def edgeFunc(*args):
+            self.fail("Callback should not be called for a removed node")
+
+        with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot iterate over in-edges of non-existing node 1",
+        ):
+            G.forInEdgesOf(1, edgeFunc)
+
+    def testForEdgesOfInvalidNodeRaises(self):
+        G = nk.Graph(2)
+        G.addEdge(0, 1)
+
+        def edgeFunc(*args):
+            self.fail("Callback should not be called for a non-existing node")
+
+        with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot iterate over edges of non-existing node 912",
+        ):
+            G.forEdgesOf(912, edgeFunc)
+
+
+    def testForEdgesOfRemovedNodeRaises(self):
+        G = nk.Graph(3)
+        G.addEdge(0, 2)
+
+        G.removeNode(1)
+
+        def edgeFunc(*args):
+            self.fail("Callback should not be called for a removed node")
+
+        with self.assertRaisesRegex(
+                RuntimeError,
+                "Cannot iterate over edges of non-existing node 1",
+        ):
+            G.forEdgesOf(1, edgeFunc)
+
 
 if __name__ == "__main__":
     unittest.main()


fabratu:
When `forEdgesOf` was implemented, testing whether a node exists was left out intentionally. The rationale/design behind it is to avoid safety checks for iterators in the graph datastructures for performance reasons.

However, there is inconsistency, since some other iterators like `forNodes` check whether the node exists, while most of the other iterators don't. Is it possible that you measure how much the additional check costs in terms of running time (especially if the degree of a node is high)? Also, what does it look like if we use try/catch blocks? I would be interested in how much this differs for modern systems.

Furthermore, if we employ such a check, it should be done in `forEdgesOf` in C++.

Schwarf:
Thanks, agreed. I will move the check into the public C++ `forEdgesOf` / `forInEdgesOf` APIs so C++ callers are protected too, and remove the duplicate Python-side checks.

I will also add a small benchmark.

angriman (dismissed):

fabratu:
@angriman Are you suggesting to accept the PR as is? I would at least move the checks to the C++ iterators.

angriman:
@fabratu sorry I did not read thoroughly the comments in this PR. I would not introduce more safety checks in C++ as iterators are heavily used in algorithms. We could use assertions instead and add the checks in Python, WDYT?

hmeyerhenke:
Let's be very careful with performance-related changes such as safety checks in iterators. I prefer @angriman's suggestion of using assertions on the C++ side. If that's too much of a burden for users, we should think about different settings. Losing performance can hurt us at other important places, however.

fabratu:
Yes, using `assert` was also my initial comment. However, I would be curious what the performance implications are. Especially since we have this kind of check in other iterators (the aforementioned `forNodes` as one example). So I also see consistency in code as one argument. Of course, not losing performance is more important.

```
template <typename L>
void Graph::forNodes(L handle) const {
    for (node v = 0; v < z; ++v) {
        if (exists[v]) {
            handle(v);
        }
    }
}
```

@angriman Hm, using the safety checks only in Python also sounds interesting. In this way the core is unaffected. However, I would also try to avoid a major regression of performance in Python. We had quite a few issues and code examples from users with time-intensive functions/algorithms written in Python.

@Schwarf Given you still have the time, I think a small benchmark might be viable. Taking the other comments into consideration, it might be interesting to measure the difference for: (i) no safety checks, (ii) safety checks in C++ and (iii) safety checks in Python. Thx for your contribution!

angriman:
> However, I would also try to avoid a major regression of performance in Python. We had quite a few issues and code examples from users with time-intensive functions/algorithms written in Python.

These checks are lightweight; unless someone is implementing algorithms directly in Python, there shouldn't be any noticeable performance degradation. IMO, the NetworKit Python API is meant to be a safe, user-friendly interface for our C++ algorithms, so I believe the safety improvements these checks provide are well worth the minimal overhead.

Anyway, I'm in favor of doing some benchmarking before merging this.

Schwarf:
I added a follow-up commit, `7d14689a`, which moves the runtime checks into the public C++ iterator APIs:
- `Graph::forEdgesOf`
- `Graph::forInEdgesOf`

The checks are now performed before dispatching to the internal iterator implementations:
```cpp
if (!hasNode(u)) {
    throw std::runtime_error("Node does not exist.");
}
```
So this benchmark compares the current baseline (`607880e`) against the actual C++ runtime-check implementation (`7d14689a`), not the earlier wrapper simulation.

### Benchmark setup

All benchmarks use unweighted graphs without edge IDs. Results are from Release builds (`-O3 -DNDEBUG`) compiled with GCC/G++ 11.4.0. CPU scaling was disabled and runs were pinned with `taskset -c 4`, using 20 repetitions and `--benchmark_min_time=1s`.

### `forEdgesOf` on high-degree node

| Graph | Nodes | Edges | Max Degree | Statistic | Baseline `607880e` | C++ check `7d14689a` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| undirected star | 100,000 | 99,999 | 99,999 | median | 12,171 ns | 12,261 ns | +0.74% |
| undirected star | 100,000 | 99,999 | 99,999 | mean | 12,193 ns | 12,244 ns | +0.42% |
| undirected star | 100,000 | 99,999 | 99,999 | stddev | 204 ns | 145 ns | |

### `forEdgesOf` on all nodes

| Graph | Nodes | Edges | Max Degree | Statistic | Baseline `607880e` | C++ check `7d14689a` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | median | 9,551,170 ns | 9,462,171 ns | -0.93% |
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | mean | 9,687,231 ns | 9,465,944 ns | -2.28% |
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | stddev | 362,131 ns | 45,123 ns | |

### `forInEdgesOf` on all nodes

| Graph | Nodes | Edges | Max In-Degree | Statistic | Baseline `607880e` | C++ check `7d14689a` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | median | 14,355,671 ns | 14,415,845 ns | +0.42% |
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | mean | 14,387,216 ns | 14,415,081 ns | +0.19% |
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | stddev | 105,216 ns | 23,420 ns | |

Overall, I do not see a meaningful slowdown from adding the C++ runtime checks in these microbenchmarks. These results were obtained with GCC/G++ 11.4.0 on my machine, so code generation and performance may differ with other compilers, compiler versions, optimization settings, or hardware. Since `hasNode` is a small member function defined in the class body, the compiler can likely inline it and eliminate the function-call overhead. The validity check itself must generally still be evaluated for a runtime node ID, but its cost appears negligible compared with the cost of traversing the adjacency lists in these workloads.

Schwarf:
Here are the python results:
The Python-side check is commit `607880e`; the baseline is PR base `9659a950`.

### `forEdgesOf` on high-degree node

| Graph | Nodes | Edges | Max Degree | Statistic | Baseline `9659a950` | Python check `607880e` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| undirected star | 100,000 | 99,999 | 99,999 | median | 6,781,711 ns | 6,734,281 ns | -0.70% |
| undirected star | 100,000 | 99,999 | 99,999 | mean | 6,794,436 ns | 6,741,671 ns | -0.78% |
| undirected star | 100,000 | 99,999 | 99,999 | stddev | 84,744 ns | 117,287 ns | |


### `forEdgesOf` on all nodes

| Graph | Nodes | Edges | Max Degree | Statistic | Baseline `9659a950` | Python check `607880e` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | median | 1,892,095,729 ns | 1,944,286,128 ns | +2.76% |
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | mean | 1,894,533,635 ns | 1,951,505,007 ns | +3.01% |
| undirected sparse band graph | 100,000 | 12,791,744 | 256 | stddev | 7,580,948 ns | 28,418,136 ns | |

### `forInEdgesOf` on all nodes

| Graph | Nodes | Edges | Max In-Degree | Statistic | Baseline `9659a950` | Python check `607880e` | Delta |
|---|---:|---:|---:|---|---:|---:|---:|
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | median | 984,219,537 ns | 985,021,362 ns | +0.08% |
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | mean | 981,174,885 ns | 988,091,343 ns | +0.70% |
| directed cyclic band graph | 100,000 | 12,800,000 | 128 | stddev | 5,662,778 ns | 29,039,391 ns | |

The Python-side benchmark also does not indicate a substantial overhead from the additional `hasNode` checks.

fabratu:
@Schwarf thanks for doing the benchmarks! There could be some still hidden observations to be found; however, the general trend becomes clear. 

Even though the C++ code does not create any overhead, I am in favor of following the comment/suggestion from @angriman -> Python = user-friendly environment, C++ = minimalistic algorithm environment.

fabratu:
After rebasing and removing the check in C++, I think we can merge the PR.

fabratu (approved):

angriman (approved):

fabratu:
Merging this PR since the last code change is older than 7 days.

PR #1430, opened 2026-06-23
Update clang-tidy to clang-tidy-22 

mkirilin:
Upgrade of the Clang/LLVM version used in CI and clang-tidy checks from 17 to 22 with corresponding code tweaks and minor refactorings for clarity and consistency in accordance with the updated clang-tidy policies.

**CI and clang-tidy configuration updates:**
* Made the LLVM version configurable via the `LLVM_VERSION` environment variable in both `.github/workflows/ci.yml` and `.github/workflows/scripts/clang_tidy.sh`. 
* Added the `NETWORKIT_CLANG_TIDY_EXECUTABLE` CMake option to specify which clang-tidy executable to use. Documentation in `DevGuide.rst` was updated to explain this option.

**Minor modernization and consistency:**
* Replaced `.size() == 0` checks with `.empty()` in quadtree-related classes for improved readability and consistency.
* Removed a couple of unused variables.
* Take strings by const references in parsers and a reader. Also take a predecessor object in 'ApproxGroupBetweenness.cpp` by const reference.
* Pre-allocate a vector in a couple of places.
* Pass a shared pointer by const reference (no much sense but satisfies clang).
* In the `Attribute` class, initialize `bool valid` by value instead of moving, since it's a simple bool and may even degrade performance due to the lack of compiler optimization.

**More peculiar changes:**
Found with indirect help of new clang.
* Test isolation for betweenness GTests.
* Fixed BarabasiAlbert graph generator edge cases, when a graph is generated by an empty of edgeless graph. Before these cases led to undefined behavior and infinite loop. Added tests `testBarabasiAlbertEmptyInit` and `testBarabasiAlbertParallelKOne` that catch this behavior.

# **For debates:**
Clang-22 complains about `Graph`'s methods' argument passed by value, e.g. 
```
template <typename L>
void forNodes(L handle) const;
```
It comes from `performance-unnecessary-value-param` flag and implies to pass it by (universal) reference. The latter might affect user's codes, so I decided to suppress the clang in the affected lines so far. Would be interesting to get a second opinion on that.

### DIFF
--- a/include/networkit/graph/Attributes.hpp
+++ b/include/networkit/graph/Attributes.hpp
@@ -274,7 +274,7 @@ private:
         index idx;
     }; // class IndexProxy
 public:
-    explicit Attribute(std::shared_ptr<AttributeStorage_type> ownedStorage = nullptr,
+    explicit Attribute(const std::shared_ptr<AttributeStorage_type> &ownedStorage = nullptr,
                        const GraphType *graph = nullptr)
         : ownedStorage{ownedStorage}, theGraph{graph},
           valid{ownedStorage != nullptr && graph != nullptr} {}
@@ -299,14 +299,14 @@ public:
 
     Attribute(Attribute &&other) noexcept
         : ownedStorage{std::move(other.ownedStorage)}, theGraph{std::move(other.theGraph)},
-          valid{std::move(other.valid)} {
+          valid{other.valid} {
         other.valid = false;
     }
 
     template <bool ic = isConst, std::enable_if_t<ic, int> = 0>
     Attribute(Attribute<NodeOrEdge, GraphType, T, false> &&other) noexcept
         : ownedStorage{std::move(other.ownedStorage)}, theGraph{std::move(other.theGraph)},
-          valid{std::move(other.valid)} {
+          valid{other.valid} {
         other.valid = false;
     }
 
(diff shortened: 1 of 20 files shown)


include/networkit/graph/Attributes.hpp
```diff
@@ -298,14 +298,14 @@ class Attribute {
 
     Attribute(Attribute &&other) noexcept
         : ownedStorage{std::move(other.ownedStorage)}, theGraph{std::move(other.theGraph)},
-          valid{std::move(other.valid)} {
```
fabratu, line 302: What is rationale of this change? What was clang-tidy complaining about?

fabratu (commented):
Your changes involve quite some changes from `string` towards `const string &`. Better, use `string_view` where possible.

include/networkit/graph/Attributes.hpp
mkirilin: clang-tidy was triggered by `performance-move-const-arg` due to `std::move` of copyable-type. For fundamental types move semantics is not faster than copying.

mkirilin:
Changed affected `const string &` to `string_view` where possible. `const string &` is left only for those, that are being passed as a key to a map that expects `const string &`.

One CI fails due to code style, #1433 is supposed to fix it.

fabratu (approved):

Issue #1431, opened 2026-06-25
Temporarily pin GCC 15 macOS CI to macos-15


mkirilin:
## Problem

The `macOS latest (gcc-15): core` CI job fails when landing at a runner with  `macOS 26.4` after GitHub Actions updated `macos-latest` to `macos-26`.

Specifically, it's Google Test's own source that fails to build and it seems to be a GCC + macOS 26 SDK compatibility issue:

  ```text
2026-06-24T13:29:20.8193990Z In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include/mach/message.h:78,
2026-06-24T13:29:20.8296370Z                  from /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include/mach/host_info.h:66,
2026-06-24T13:29:20.8397790Z                  from /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include/mach/mach_types.h:80,
2026-06-24T13:29:20.8500020Z                  from /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include/mach/mach_init.h:61,
2026-06-24T13:29:20.8624230Z                  from /Users/runner/work/networkit/networkit/extlibs/googletest/googletest/src/gtest-port.cc:59,
2026-06-24T13:29:20.8772320Z                  from /Users/runner/work/networkit/networkit/extlibs/googletest/googletest/src/gtest-all.cc:45:
  /Library/Developer/CommandLineTools/SDKs/MacOSX26.sdk/usr/include/mach/
  message.h:299:1:
  error: expected constructor, destructor, or type conversion before '(' token
    299 | xnu_static_assert_struct_size(mach_msg_type_descriptor_t, 12);
        | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Proposed solution

Pin the GCC 15 macOS job to the  the newest successful `macos-15` runner.

Keep the LLVM job on macos-latest so we still test the newest macOS image where the compiler supports the SDK.

## Follow-up

Remove this pin once the compatibility issue is resolved. Until then keep the issue open.

PR #1432, opened 2026-06-25
Pin MacOs to version 15

mkirilin:
Addresses the #1431 issue.


fabratu (approved):
Only nitpicky: the commit message should say "15" instead of "16". Approving + merging anyway :)

PR #1438, opened 2026-07-04
Using C++20 concepts to constrain NodeT and EdgeWeightT.

Schwarf:
## Proposal: constrain `AdjListGraph` template parameters with C++20 concepts

This PR is part of this [proposal](https://github.com/networkit/networkit/issues/1415) and replaces the existing `static_assert` checks in `AdjListGraph` with C++20 template constraints.

Current constraints:
- `NodeT` must satisfy `std::integral`
- `EdgeWeightT` must satisfy the new `GraphEdgeWeight` concept, currently defined as integral or floating point

This makes the type contract visible directly at the template declaration instead of inside the class body.

One consequence is that every out-of-class member definition now also carries the constrained template head:
```cpp
template <std::integral NodeT, GraphEdgeWeight EdgeWeightT>
```
This is more explicit, but also adds some repetition throughout `Graph.hpp`.

A further advantage of named concepts is that graph-related type requirements can be defined once and reused consistently by graph implementations and generic algorithms. This also makes them easier to refine centrally in the future. For example, if `bool` should not be accepted as a node or edge-weight type, dedicated GraphNode and GraphEdgeWeight concepts could express that rule, and algorithms using those concepts would automatically follow it.

@angriman @fabratu What do you think about this direction? Is the clearer template contract worth the additional verbosity in the out-of-class definitions?

### DIFF
--- a/include/networkit/graph/GraphConcepts.hpp
+++ b/include/networkit/graph/GraphConcepts.hpp
@@ -5,8 +5,8 @@
  *
  */
 
-#ifndef NETWORKIT_GRAPH_CONCEPTS_HPP_
-#define NETWORKIT_GRAPH_CONCEPTS_HPP_
+#ifndef NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
+#define NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
 
 #include <concepts>
 #include <type_traits>
@@ -20,4 +20,4 @@ concept GraphEdgeWeight = std::integral<T> || std::floating_point<T>;
 
 } // namespace NetworKit
 
-#endif // NETWORKIT_GRAPH_CONCEPTS_HPP_
+#endif // NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
--- a/include/networkit/graph/GraphConcepts.hpp
+++ b/include/networkit/graph/GraphConcepts.hpp
@@ -10,6 +10,7 @@
 
 #include <concepts>
 #include <type_traits>
+
 namespace NetworKit {
 
 template <typename T>
--- a/include/networkit/graph/GraphConcepts.hpp
+++ b/include/networkit/graph/GraphConcepts.hpp
@@ -9,7 +9,6 @@
 #define NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
 
 #include <concepts>
-#include <type_traits>
 
 namespace NetworKit {
 
--- a/include/networkit/graph/GraphConcepts.hpp
+++ b/include/networkit/graph/GraphConcepts.hpp
@@ -1,0 +1,23 @@
+/*  GraphConcepts.hpp
+ *
+ *  Created on: 05.07.2026
+ *  Authors: Andreas Scharf (andreas.b.scharf@gmail.com)
+ *
+ */
+
+#ifndef NETWORKIT_GRAPH_CONCEPTS_HPP_
+#define NETWORKIT_GRAPH_CONCEPTS_HPP_
+
+#include <concepts>
+#include <type_traits>
+namespace NetworKit {
+
+template <typename T>
+concept GraphNode = std::integral<T>;
+
+template <typename T>
+concept GraphEdgeWeight = std::integral<T> || std::floating_point<T>;
+
+} // namespace NetworKit
+
+#endif // NETWORKIT_GRAPH_CONCEPTS_HPP_
--- a/networkit/cpp/graph/test/AdjListGraphGTest.cpp
+++ b/networkit/cpp/graph/test/AdjListGraphGTest.cpp
@@ -35,7 +35,36 @@ TYPED_TEST_P(AdjListGraphGTest, testDefaultConstructor) {
     EXPECT_EQ(G.isDirected(), TypeParam::directed);
 }
 
-REGISTER_TYPED_TEST_SUITE_P(AdjListGraphGTest, testDefaultConstructor);
+TYPED_TEST_P(AdjListGraphGTest, testNodeAndEdgeIterators) {
+    using Graph = AdjListGraph<typename TestFixture::NodeT, typename TestFixture::EdgeWeightT>;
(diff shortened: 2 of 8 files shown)


Schwarf:
Hi @mkirilin, hi @fabratu,

I am currently investigating the `clang-format` failures in this PR. After switching to `clang-format-22`, the formatting of my changes is correct, but the checker still reports a few unrelated files that were not touched by this PR as non-compliant.

Would you prefer me to address those files in a separate formatting PR, or should they be handled differently?

angriman:
@Schwarf I like this solution more than the static_asserts. A couple of comments:
- Since we use `NodeT` and `EdgeWeightT` in node iterators and edge iterators too, I'd recommend migrating those classes to use concepts as well.
- If we also migrate the iterators to use concepts, should we define these concepts in a separate file?

Schwarf:
Hi @angriman,
I will migrate node iterators and egde iterators to use the same concepts and move these concepts to a new file "GraphConcepts.hpp". What do you think?

angriman:
Sounds good!

angriman:
> Hi @mkirilin, hi @fabratu,
> 
> I am currently investigating the `clang-format` failures in this PR. After switching to `clang-format-22`, the formatting of my changes is correct, but the checker still reports a few unrelated files that were not touched by this PR as non-compliant.
> 
> Would you prefer me to address those files in a separate formatting PR, or should they be handled differently?

Done in #1441

Schwarf:
Hi @angriman,

I applied the changes as discussed. I added iterator coverage to `AdjListGraphGTest.cpp`, which exposed an inconsistency in `upperNodeIdBound()`.

I changed the return type from
```cpp
index upperNodeIdBound() const noexcept { return z; }
```
to
```cpp
NodeT upperNodeIdBound() const noexcept { return z; }
```
since `z` is also stored as `NodeT`.

networkit/cpp/graph/test/AdjListGraphGTest.cpp
```diff
@@ -35,7 +35,36 @@ TYPED_TEST_P(AdjListGraphGTest, testDefaultConstructor) {
...
+    using EdgeWeightT = typename TestFixture::EdgeWeightT;
+
+    Graph G(4, TypeParam::weighted, TypeParam::directed);
+    G.addEdge(NodeT{0}, NodeT{1}, EdgeWeightT{2});
+    G.addEdge(NodeT{1}, NodeT{2}, EdgeWeightT{3});
+
+    count nodes = 0;
+    for ([[maybe_unused]] const auto u : G.nodeRange()) {
+        ++nodes;
+    }
```
angriman: Instead of just checking the number of nodes and edges, consider checking the actual elements, for example:
```cpp

using ::testing::ElementsAre;

const std::vector<NodeT> nodes(G.nodeRange().begin(), G.nodeRange().end());
EXPECT_THAT(nodes, ElementsAre(0, 1, 2, 3));

```

networkit/cpp/graph/test/AdjListGraphGTest.cpp
```diff
@@ -35,7 +35,36 @@ TYPED_TEST_P(AdjListGraphGTest, testDefaultConstructor) {
...
+
+    count edges = 0;
+    for ([[maybe_unused]] const auto edge : G.edgeRange()) {
+        ++edges;
+    }
```
angriman: As above, but here you might prefer defining a custom matcher:

```cpp
MATCHER_P2(EdgeEq, expected_u, expected_v, "") {
    return arg.u == expected_u && arg.v == expected_v;
}

const std::vector<Edge<NodeT>> edges(G.edgeRange().begin(), G.edgeRange().end());
EXPECT_THAT(edges, ElementsAre(Edge<NodeT>{0, 1}, Edge<NodeT>{1, 2}));
```

You can do something similar below for weighted edges.

Schwarf: Done

networkit/cpp/graph/test/AdjListGraphGTest.cpp
Schwarf: Done

angriman (dismissed):

Schwarf:
Hi @angriman,

I rebased this PR after #1440 was merged and adapted the concept changes to the new `AdjListGraph.hpp` / `AdjListGraphImpl.hpp` split.

Could you review/approve again when you have time? It would be nice to get this merged soon too.

include/networkit/graph/GraphConcepts.hpp
```diff
@@ -0,0 +1,23 @@
...
+ *  Authors: Andreas Scharf (andreas.b.scharf@gmail.com)
+ *
+ */
+
+#ifndef NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
+#define NETWORKIT_GRAPH_GRAPH_CONCEPTS_HPP_
+
+#include <concepts>
+#include <type_traits>
+namespace NetworKit {
```
angriman, line 13: Nit: add an empty line between the includes and the namespace.

angriman (dismissed):

include/networkit/graph/GraphConcepts.hpp
Schwarf: Can we do a clean-up in a small follow-up PR? I also recognized some unused header includes. 
I could raise the follow-up PR tomorrow.

angriman: Sure, but since this is a new file I prefer that we address this now.

Schwarf: Done

angriman (dismissed):

Schwarf:
Hi @fabratu, 
the remaining CI failure appears to occur while installing prerequisites from the Microsoft Ubuntu package repository, before the build starts:
`Clearsigned file isn't valid, got 'NOSPLIT'`

angriman (dismissed):

fabratu:
@Schwarf PR needs a rebase before merging

angriman (approved):

PR #1439, opened 2026-07-05
Template `BucketPQ` key and value types

Schwarf:
This PR is part of this [proposal](https://github.com/networkit/networkit/issues/1415) and generalizes `Aux::BucketPQ` from the fixed `int64_t`/`index` pairing to a templated implementation:

```cpp
template <SignedIntegral KeyType = int64_t, IntegralValue ValueType = index>
class BucketPQ
```

The default specialization preserves the previous API behavior, while making the type requirements explicit:
- keys must be signed integral types
- values must be non-bool integral types

The implementation now separates:
- `KeyType`: priority/key type
- `ValueType`: stored value type
- `BucketIndex`: internal bucket-array index type

Additionally, `PrioQueue` was made sign-compare clean for signed value types. This is needed because `BucketPQ<KeyType, ValueType>` derives from `PrioQueue<KeyType, ValueType>`, so typed tests such as `BucketPQ<int64_t, int>` also instantiate inherited `PrioQueue` methods. The indexing logic now converts values through a single helper before comparing against or indexing into the backing vector.

This clarifies that `myBucket` stores bucket indices, not keys or values.

I also added `BucketPQGTest.cpp`, which provides a typed test suite similar to `AdjListGraphGTest.cpp` and verifies the implementation across several key/value type combinations.

I measured the default specialization with GCC 11.4 and did not observe a visible performance regression.

CI note: the clang-format failure appears to be caused by the same pre-existing, unrelated files also mentioned in #1438.


include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -19,8 +25,12 @@ namespace Aux {
...
 using index = NetworKit::index;
 using count = NetworKit::count;
-using Bucket = std::list<index>;
-constexpr int64_t none = std::numeric_limits<int64_t>::max();
+
+template <typename T>
+concept SignedIntegral = std::integral<T> && std::is_signed_v<T>;
+
+template <typename T>
+concept IntegralValue = std::integral<T> && !std::same_as<std::remove_cvref_t<T>, bool>;
```
fabratu: @angriman @Schwarf Is it reasonable to generalize `GraphConcepts` from #1438 even more and reuse concepts throughout NetworKit? One canonical file for storing concepts would be `Globals.hpp`.

include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -29,37 +39,45 @@ constexpr int64_t none = std::numeric_limits<int64_t>::max();
  * the obvious constraint minPrio <= maxPrio.
  * Amortized constant running time for each operation.
  */
-class BucketPQ : public PrioQueue<int64_t, index> {
+template <SignedIntegral KeyType = int64_t, IntegralValue ValueType = index>
+class BucketPQ : public PrioQueue<KeyType, ValueType> {
```
fabratu: Rename the class to `BucketPriorityQueue` + add an alias for support of older code.

include/networkit/auxiliary/BucketPQ.hpp
```diff
@@ -69,7 +87,22 @@ class BucketPQ : public PrioQueue<int64_t, index> {
     /**
      * Called from various constructors for initializing members.
      */
-    void construct(uint64_t capacity);
+    void construct(uint64_t capacity) {
```
fabratu: @Schwarf @angriman IMO it would be more consistent to split all template classes (independent of LOCs) into a declaration + implementation file. What do you think?

fabratu (changes requested):

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: Hi @fabratu, 
I am generally in favor of reusing concepts where they express the same contract.

For now, I would keep `BucketPQ` concepts separate from `GraphConcepts`: these are auxiliary data-structure constraints, while `GraphConcepts` are graph-domain constraints.

For `Globals.hpp`, I do not remember a strong technical reason against it. In #1438 we chose `GraphConcepts.hpp` because the concepts were graph-specific. I would move concepts to `Globals.hpp` once we have a clearer convention for NetworKit-wide concepts.

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: I have no strong opinion here. Splitting can make larger class templates easier to scan, but for smaller ones I prefer a single file to avoid extra file switching.

Since template implementations still need to be included anyway, I would keep `BucketPQ` in one file unless we want this split consistently for all class templates.

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: I can do this, but I want to clarify first:

Renaming the implementation to `BucketPriorityQueue` and adding
```cpp
template <SignedIntegral KeyType = int64_t, IntegralValue ValueType = index>
using BucketPQ = BucketPriorityQueue<KeyType, ValueType>;
```
keeps old code working as `BucketPQ<>`.

If we also want default uses to become bare `BucketPQ q(...)`, we need extra deduction guides, because `ValueType` is not deducible from the constructor arguments and some `BucketPQ` constructor calls mix key argument types like `int` and `int64_t`.

I tried this locally, and supporting all existing constructor calls without `<>` required several deduction guides. My preference would be to keep the alias and use `BucketPQ<>` for the default specialization, unless you specifically prefer bare `BucketPQ`.

What do you think?

include/networkit/auxiliary/BucketPQ.hpp
angriman: I think that we can be flexible and split only where it helps improving readability. In this case the implementation is complex so I believe that we should split.

include/networkit/auxiliary/BucketPQ.hpp
angriman: To me it makes sense to keep concepts temporarily separated and consolidate them later once they are used more widely across the platform.

networkit/cpp/auxiliary/test/BucketPQGTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+TYPED_TEST_P(BucketPQGTest, testConstructFromKeysSkipsNone) {
+    using KeyType = typename TestFixture::KeyType;
+    using ValueType = typename TestFixture::ValueType;
+
+    std::vector<KeyType> keys = {KeyType{3}, std::numeric_limits<KeyType>::max(), KeyType{-2},
+                                 KeyType{1}};
+    Aux::BucketPQ<KeyType, ValueType> prioQ(keys, KeyType{-5}, KeyType{5});
+
+    EXPECT_EQ(prioQ.size(), 3u);
```
angriman: Prefer:
```cpp
using ::testing::SizeIs;

EXPECT_THAT(prioQ, SizeIs(3));
```

networkit/cpp/auxiliary/test/BucketPQGTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    EXPECT_TRUE(prioQ.contains(ValueType{3}));
+
+    EXPECT_EQ(prioQ.getKey(ValueType{0}), KeyType{3});
+    EXPECT_EQ(prioQ.getKey(ValueType{2}), KeyType{-2});
+    EXPECT_EQ(prioQ.getKey(ValueType{3}), KeyType{1});
+
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{-2}, ValueType{2}}));
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{1}, ValueType{3}}));
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{3}, ValueType{0}}));
+    EXPECT_TRUE(prioQ.empty());
```
angriman: Prefer:
```cpp
using ::testing::IsEmpty;

EXPECT_THAT(prioQ, IsEmpty());
```

Same below.

Schwarf: Done

networkit/cpp/auxiliary/test/BucketPQGTest.cpp
Schwarf: Done

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: Done

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: Renaming is done. The compatibility alias is available as `BucketPQ<>`; I did not add deduction guides for bare `BucketPQ q(...)`.

include/networkit/auxiliary/BucketPQ.hpp
Schwarf: For now I left the `BucketPriorityQueue` concepts in the header. We can consolidate them later once it is clearer which concepts are needed more broadly.

include/networkit/auxiliary/BucketPQ.hpp
fabratu: Hm, I cannot find callers using the mixed key argument types for `BucketPQ`. Can you point to them? Overall I favor `BucketPQ` providing the same API promise (key = `int64_t`, value = `index`), hence dropping the template arguments for the alias completely. With the general restructuring, the name `BucketPQ` will be deprecated anyway.

Schwarf: Of course, you are right. I missed the obvious solution here.

For backwards compatibility, a simple non-template alias is the right approach:
```cpp
using BucketPQ = BucketPriorityQueue<int64_t, index>;
```
Users who want custom key/value types can use `BucketPriorityQueue<KeyType, ValueType>` directly.

fabratu (dismissed):

networkit/cpp/auxiliary/test/BucketPriorityQueueGTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+    EXPECT_EQ(prioQ.getKey(ValueType{3}), KeyType{1});
+
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{-2}, ValueType{2}}));
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{1}, ValueType{3}}));
+    EXPECT_EQ(prioQ.extractMin(), (std::pair<KeyType, ValueType>{KeyType{3}, ValueType{0}}));
+    EXPECT_THAT(prioQ, IsEmpty());
+}
+
+TYPED_TEST_P(BucketPriorityQueueGTest, testChangeKeyAndRemove) {
+    using ::testing::IsEmpty;
```
angriman: Prefer declaring these aliases on top of the file (e.g., right after the opening of the anonymous namespace)

angriman (dismissed):

angriman (approved):

networkit/cpp/auxiliary/test/BucketPriorityQueueGTest.cpp
Schwarf: Done

Schwarf:
Hi @angriman,
could we merge this once the CI builds pass? I can raise the remaining follow-up PRs tomorrow.

Issue #1434, opened 2026-06-30
CI: Template node and edge types lead to degredation in documentation build


fabratu:
With the merge of PR #1416, the documentation build fails due to errors in `exhale`:

```
[9](https://github.com/networkit/networkit/actions/runs/28432921579/job/84251742448#step:6:7590)
[~] Exhale: changing directories back to [/home/runner/work/networkit/networkit/core_build] after Doxygen.
[+] Exhale: doxygen ran successfully in 2.56 seconds.
[~] Exhale: adding tree view css / javascript.
[+] Exhale: added tree view css / javascript.
(!) Invalid C++ template: NetworKit::AdjListGraph::SafeArg< TraitsT, N, typename std::enable_if<(N< TraitsT::arity)>::type > has 3 '<' and 2 '>', exhale does not know what to do with your code.Traceback (most recent call last):
  File "/home/runner/work/networkit/networkit/pyenv/lib/python3.13/site-packages/exhale/graph.py", line 2460, in initializeNodeFilenameAndLink
    raise Exception(str(node))
Exception: Struct(NetworKit::AdjListGraph::SafeArg< TraitsT, N, typename std::enable_if<(N< TraitsT::arity)>::type >, template=<typename TraitsT, size_t N>, n_kids=0)
make[3]: *** [CMakeFiles/general_docs.dir/build.make:71: CMakeFiles/general_docs] Error 1
make[2]: *** [CMakeFiles/Makefile2:1604: CMakeFiles/general_docs.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:1643: CMakeFiles/docs.dir/rule] Error 2
make: *** [Makefile:208: docs] Error 2
Error: Process completed with exit code 2.
```
(see run: https://github.com/networkit/networkit/actions/runs/28432921579/job/84251742448)

PR #1443, opened 2026-07-09
Fix error when generating C++ doc strings (exhale)

fabratu:
CI on master currently shows failures for the documentation build (see issue #1434).

Reason: Doxygen emits that comparison as part of the template name, and Exhale's (plugin for C++ doc generation) template parser counts the '<' in `N < TraitsT::arity` as a template opener. That gives Exhale an unbalanced template string, causing the 3 '<' and 2 '>' error.

The current fix leaves the logic of the code as is; however, it changes the detection for Exhale.


angriman (approved):
Thank you!

PR #1447, opened 2026-07-13
Fix BucketPriorityQueue clear state reset

### LINKED ISSUE #1442: `Aux::BucketPQ::clear()` does not clear bucket queue state
I found this while refactoring `BucketPQ` in #1439 and during subsequent local performance tests.

`Aux::BucketPQ` inherits `clear()` from `Aux::PrioQueue`, but `BucketPQ` stores its actual queue state in its own members.

As a result, calling `clear()` on a concrete `BucketPQ` does not clear the queue.

Minimal reproducer:
```cpp
TEST_F(AuxGTest, testBucketPQClear) {
    Aux::BucketPQ prioQ(3, 0, 2);

    prioQ.insert(0, 0);
    prioQ.insert(1, 1);
    prioQ.clear();

    EXPECT_TRUE(prioQ.empty());
    EXPECT_EQ(prioQ.size(), 0);
    EXPECT_FALSE(prioQ.contains(0));
    EXPECT_FALSE(prioQ.contains(1));
}
```
A fix would be to override `clear()` in `BucketPQ` and reset its bucket storage, value-position state, and min/max bookkeeping.

I will fix this issue after #1439 is merged.

Schwarf:
This PR adds a `clear()` override to `BucketPriorityQueue`.

Previously, `BucketPriorityQueue` inherited `clear()` from `PrioQueue`, which only reset the base-class state. The new override resets the bucket queue state and adds a regression test.

Fixes #1442

### DIFF
--- a/include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
+++ b/include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
@@ -184,13 +184,10 @@ void BucketPriorityQueue<KeyType, ValueType>::remove(const ValueType &value) {
 
 template <SignedIntegral KeyType, IntegralValue ValueType>
 void BucketPriorityQueue<KeyType, ValueType>::clear() {
-    for (std::list<ValueType> &bucket : buckets) {
-        bucket.clear();
-    }
-    for (OptionalIterator &ptr : nodePtr) {
-        ptr.reset();
-    }
-    std::ranges::fill(myBucket.begin(), myBucket.end(), noneBucket);
+    std::ranges::fill(bucketHead, noneValue);
+    std::ranges::fill(previous, noneValue);
+    std::ranges::fill(next, noneValue);
+    std::ranges::fill(myBucket, noneBucket);
     currentMinKey = noneKey;
     currentMaxKey = std::numeric_limits<KeyType>::min();
     numElems = 0;
--- a/include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
+++ b/include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
@@ -183,6 +183,20 @@ void BucketPriorityQueue<KeyType, ValueType>::remove(const ValueType &value) {
 }
 
 template <SignedIntegral KeyType, IntegralValue ValueType>
+void BucketPriorityQueue<KeyType, ValueType>::clear() {
+    for (auto &bucket : buckets) {
+        bucket.clear();
+    }
+    for (auto &ptr : nodePtr) {
+        ptr.reset();
+    }
+    std::fill(myBucket.begin(), myBucket.end(), noneBucket);
+    currentMinKey = noneKey;
+    currentMaxKey = std::numeric_limits<KeyType>::min();
+    numElems = 0;
+}
+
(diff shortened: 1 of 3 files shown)


include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
```diff
@@ -158,6 +158,20 @@ void BucketPriorityQueue<KeyType, ValueType>::remove(const ValueType &value) {
     }
 }
 
+template <SignedIntegral KeyType, IntegralValue ValueType>
+void BucketPriorityQueue<KeyType, ValueType>::clear() {
+    for (auto &bucket : buckets) {
```
angriman: Avoid `auto` here and below, the type is not obvious.

include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
```diff
@@ -158,6 +158,20 @@ void BucketPriorityQueue<KeyType, ValueType>::remove(const ValueType &value) {
...
+        bucket.clear();
+    }
+    for (auto &ptr : nodePtr) {
+        ptr.reset();
+    }
+    std::fill(myBucket.begin(), myBucket.end(), noneBucket);
```
angriman: Here you can use std::ranges::fill

networkit/cpp/auxiliary/test/BucketPriorityQueueGTest.cpp
```diff
@@ -91,8 +91,32 @@ TYPED_TEST_P(BucketPriorityQueueGTest, testEmptySentinelsUseTemplateTypes) {
...
+    prioQ.insert(KeyType{2}, ValueType{1});
+    prioQ.insert(KeyType{-1}, ValueType{2});
+
+    prioQ.clear();
+
+    EXPECT_THAT(prioQ, IsEmpty());
+    EXPECT_FALSE(prioQ.contains(ValueType{0}));
+    EXPECT_EQ(prioQ.getMin(),
+              (std::pair<KeyType, ValueType>{std::numeric_limits<KeyType>::max(),
+                                             std::numeric_limits<ValueType>::max()}));
```
angriman: Here you can use the `Pair` matcher (https://google.github.io/googletest/reference/matchers.html#member-matchers):

```cpp
using ::testing::Pair;

EXPECT_THAT(
    prioQ.getMin(),
    Pair(std::numeric_limits<KeyType>::max(),
         std::numeric_limits<ValueType>::max()));
```

Schwarf: Done

include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
Schwarf: Done

include/networkit/auxiliary/BucketPriorityQueueImpl.hpp
Schwarf: Done

Schwarf:
Hi @angriman,
I would prefer to merge #1446 before this one, since there will be merge conflicts and it would be easier for me to adapt this PR afterwards.

angriman (approved):

Issue #1428, opened 2026-06-22
visualisation with iplotx


iosonofabio:
Dear networkit developers,

Hope you are well. I recently created [iplotx](https://iplotx.readthedocs.io/), a Python library that focuses on visualising graphs/networks independent of where they come from. It uses a sort of adaptor class to ingest network data structuers from all kinds of libraries (e.g. igraph, networkx, graph-tool, etc.). Additional libraries can signal how to connect to iplotx via custom [data providers](https://iplotx.readthedocs.io/en/latest/providers.html#creating-a-custom-data-provider).

I wonder whether you would consider adding a data provider to networkit. I would be happy to help, the code would preferrably reside within networkit and it would enable your users to visualise your networks with iplotx's extensive [style specification](https://iplotx.readthedocs.io/en/latest/api/complete_style_specification.html). We have about 100 examples on our [gallery](https://iplotx.readthedocs.io/en/latest/gallery/index.html).

Would you be interested?

hmeyerhenke:
Sounds interesting. @mkirilin, could you please take a look and discuss further steps with @fabratu and me?

hmeyerhenke:
And thank you for bringing this to our attention, @iosonofabio!

iosonofabio:
No problem. I created this after refactoring the whole viz stack for igraph and learning a few lessons from networkx and matplotlib folks too. It isn't perfect but I find it better than anything else. Also, contributions and requests welcome of course.

mkirilin:
@iosonofabio, thanks for suggesting this and for offering to help.

Before we decide on next steps, we would like to better understand the value for users from the NetworKit perspective. NetworKit already has a few visualization options, so it would be helpful to see where `iplotx` could complement or improve the workflow for our users.

Could you share a couple of concrete examples where using `iplotx` with NetworKit would provide a clear advantage? Relevant NetworKit capabilities to compare against might be:

  - `networkit.vizbridges.widgetFromGraph`: provides notebook-based graph visualization.
  - `networkit.plot`: provides plots for common results analysis.
  - `networkit.viz`: contains layout algorithms.
  - `networkit.graphio` and `networkit.gephi`: support export workflows for external visualization tools.

This would help us understand the right scope and ownership model for a possible integration. Our current preference would be to keep the integration small and optional, with clear value for users.

iosonofabio:
[...]

PR #1448, opened 2026-07-14
Adapter for iplotx visualisation (issue #1428)

iosonofabio:
Hi @mkirilin, hi @hmeyerhenke, hi devs,

This is an initial PR addressing issue #1428, i.e. an initial implementation of the adapter for network visualisation via iplotx.

I tried to follow your code conventions, apologies if some of it is not conventional.

This works on my machine with both of the following snippets in a fresh environment:

```python
import numpy as np
import networkit as nk
import iplotx as ipx
from networkit import vizbridges
import matplotlib

matplotlib.use("Qt5Agg")
import matplotlib.pyplot as plt

G = nk.graph.Graph(10)
G.addEdges((np.arange(3), np.arange(3) + 2))
layout = np.arange(20).reshape((10, 2)).astype(np.float64)
fig, ax = plt.subplots()
vizbridges.iplotxFromGraph(G, layout, ax=ax)
fig.tight_layout()
plt.ion(); plt.show()
```

and

```python
import numpy as np
import networkit as nk
import iplotx as ipx
import matplotlib

matplotlib.use("Qt5Agg")
import matplotlib.pyplot as plt

G = nk.graph.Graph(10)
G.addEdges((np.arange(3), np.arange(3) + 2))
layout = np.arange(20).reshape((10, 2)).astype(np.float64)
fig, ax = plt.subplots()
ipx.network(G, layout, ax=ax)
fig.tight_layout()
plt.ion(); plt.show()
```

Outcome (after saving to file):

iplotx_networkit_example

The second option is *iplotx*'s preferred way to interact with external libraries because it involves less pollution on your end. In other words, using the second option makes the function `vizbridges.iplotxFromGraph` fully redundant, which is generally nice because it's somewhat of an empty shell anyway.

Personally, I would only implement the adapter subclass and entry point in `pyproject.toml` and not add  `vizbridges.iplotxFromGraph` because if we add it then we have to document it, but its keyword arguments are very rich, already documented in iplotx, and their documentation and/or implementation might occasionally change on iplotx's side which would require manual resyncing here.

In addition to this code, as mentioned in the issue, I would like to suggest some kind change to your docs to let users know this is an option for them.

Happy to answer any questions and thanks for your time.

### DIFF
--- a/networkit/vizbridges.py
+++ b/networkit/vizbridges.py
@@ -33,6 +33,20 @@ except ImportError:
 else:
     hasSeaborn = True
 
+try:
+    from iplotx.typing import LayoutType
+    from iplotx.ingest.typing import (
+        NetworkDataProvider as IplotxNetworkDataProvider,
+        NetworkData as IplotxNetworkData,
+    )
+except (ImportError, ModuleNotFoundError):  # pragma: no cover
+    LayoutType = object  # type: ignore[assignment,misc]
+    IplotxNetworkDataProvider = object  # type: ignore[assignment,misc]
+    IplotxNetworkData = object  # type: ignore[assignment,misc]
+    hasIplotx = False
+else:
+    hasIplotx = True
+
 
 class Dimension(Enum):
     """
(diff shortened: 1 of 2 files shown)


iosonofabio:
Would it be OK to please take a look now?

Happy to take a look at the CI things but just to know if you like the design

Thanks

hmeyerhenke:
Hi @iosonofabio,

@mkirilin is on vacation leave until tomorrow and I am busy with other tasks. I _think_ @mkirilin can take a look on Wednesday.

Thanks for your patience,
Henning

iosonofabio:
Thanks and no rush, just wanted to make sure I'm not doing the wrong thing but there's no urgency

PR #1445, opened 2026-07-12
Migrate SSSP to templates

angriman:
SSSP is used as base class for algorithms such as BFS and Dijkstra, which will be migrated in future PRs. As mentioned in #1415, for backward compatibility we rename SSSP to SingleSourceShortestPaths and introduce an SSSP alias based on `AdjListGraph<node, edgeweight>`.

Further, add aliases for `NodeT` and `EdgeWeightT` in `AdjListGraph`, which are necessary for accessing them in `SingleSourceShortestPaths`.

PR #1449, opened 2026-07-15
Migrate BFS to templates

angriman:
As discussed in #1415, this PR introduces a template version of BFS. It depends on #1445 to be merged first.


Schwarf:
Hi @angriman, 
This PR looks like it is based on an outdated version of #1445. Could you please rebase/update it onto the latest #1445 first? That should make the BFS-specific diff easier to review and avoid reintroducing older SSSP changes.

For BFS specifically, I would prefer keeping such a compact and fundamental algorithm in a single header, e.g. `BFS.hpp`, instead of splitting the implementation into `BFSImpl.hpp`.

include/networkit/distance/BFSImpl.hpp
```diff
@@ -0,0 +1,87 @@
...
+ *
+ *  Created on: Jul 23, 2013
+ *      Author: Henning
+ */
+
+#ifndef NETWORKIT_DISTANCE_BFS_IMPL_HPP_
+#define NETWORKIT_DISTANCE_BFS_IMPL_HPP_
+
+#include <queue>
+
```
Schwarf: If you still prefer the split (see [here](https://github.com/networkit/networkit/pull/1449#issuecomment-5035990372)), then `BFSImpl.hpp` should include the header:
```cpp
#include <networkit/distance/BFS.hpp>

Schwarf (changes requested):

include/networkit/distance/BFSImpl.hpp
angriman: I think it's ok to have the implementation in the same `BFS.hpp` file, moved it there.

fabratu (approved):

PR #1456, opened 2026-07-31
Generalize Cover to GenericCover

Schwarf:
This PR is part of the refactoring proposed in https://github.com/networkit/networkit/issues/1415 and introduces `GenericCover<IndexType>`, a generalization of the `Cover` class.

The existing `Cover` APIs are preserved as an alias for `GenericCover<index>`.


include/networkit/structures/GenericCover.hpp
angriman, line 1: I think that we can implement this in `Cover.hpp`, no need to add another file.

include/networkit/structures/GenericCoverImpl.hpp
```diff
@@ -0,0 +1,215 @@
...
+    std::map<IndexType, count> mapping;
+    std::vector<count> sizes;
+    count newIndex = 0;
+    for (IndexType e = 0; e <= this->z; ++e) { // stores sizes in a vector
+        for (IndexType t : data[e]) {
+            if (mapping.find(t) == mapping.end()) {
+                mapping[t] = newIndex++;
+                sizes.push_back(1);
+            } else {
+                sizes[mapping[t]]++;
```
angriman: Nit: Prefer pre-increments, same below.

networkit/cpp/structures/test/CoverGTest.cpp
angriman, line 1: As above, I think that we can implement these tests in `CoverGTest.cpp`.

include/networkit/structures/GenericCover.hpp
Schwarf: Done

include/networkit/structures/GenericCoverImpl.hpp
Schwarf: Done

networkit/cpp/structures/test/CoverGTest.cpp
Schwarf: Done

angriman (approved):

PR #1459, opened 2026-08-21
Fix issue #1458

### LINKED ISSUE #1458: LAMG segfaults for a disconnected component with one node
Reproduce:

```cpp
#include <cmath>
#include <iostream>

#include <networkit/algebraic/CSRMatrix.hpp>
#include <networkit/algebraic/Vector.hpp>
#include <networkit/numerics/LAMG/Lamg.hpp>

int main() {
    using namespace NetworKit;

    constexpr count starSize = 201;
    Graph graph(starSize + 1); // nodes 0..200 form a star; node 201 is isolated
    for (node leaf = 1; leaf < starSize; ++leaf) {
        graph.addEdge(0, leaf);
    }

    const auto laplacian = CSRMatrix::laplacianMatrix(graph);
    Vector rhs(graph.numberOfNodes(), -1.0);
    rhs[0] = starSize - 1;
    rhs[starSize] = 0.0;
    Vector result(graph.numberOfNodes(), 0.0);

    Lamg<CSRMatrix> lamg;
    lamg.setup(laplacian);
    const auto status = lamg.solve(rhs, result);
    const double residual = (laplacian * result - rhs).length();

    std::cout << "dimension=" << result.getDimension() << '\n'
              << "residual=" << residual << '\n'
              << "reported_residual=" << status.residual << '\n'
              << "converged=" << std::boolalpha << status.converged << '\n';

    const bool valid = result.getDimension() == graph.numberOfNodes()
                       && std::isfinite(residual) && std::isfinite(status.residual)
                       && residual <= 1e-6 && status.converged;
    return valid ? 0 : 1;
}
```
Observed behavior is a SIGSEGV (exit status 139). A condensed backtrace is:

```text
Program received signal SIGSEGV, Segmentation fault.
#0  NetworKit::Vector::operator[](unsigned long)
#1  NetworKit::Lamg<NetworKit::CSRMatrix>::solveThread(...)
    at include/networkit/numerics/LAMG/Lamg.hpp  // component writeback
#2  NetworKit::Lamg<NetworKit::CSRMatrix>::solve(...)
#3  main
```

Schwarf:
This PR fixes the bug by:
- Replacing an unintended `Vector` assignment from `0.0` with `fill(0.0)`.
- The old assignment was only accepted because `Vector` had a non-explicit size constructor, so that constructor is now explicit. 

A regression test was added for this case.

Fixes:  #1458

### DIFF
--- a/networkit/cpp/numerics/test/SolverLamgGTest.cpp
+++ b/networkit/cpp/numerics/test/SolverLamgGTest.cpp
@@ -102,7 +102,9 @@ TEST_F(SolverLamgGTest, testSolveWithSingletonEliminationLevelSetsZeroResult) {
 
     solver.solve(x, b, status);
 
-    EXPECT_THAT(my_vector, AllOf(SizeIs(2), Each(DoubleEq(0.0))));
+    EXPECT_EQ(x.getDimension(), 2);
+    EXPECT_DOUBLE_EQ(x[0], 0.0);
+    EXPECT_DOUBLE_EQ(x[1], 0.0);
 }
 
 Vector SolverLamgGTest::randVector(count dimension) const {
--- a/networkit/cpp/numerics/test/SolverLamgGTest.cpp
+++ b/networkit/cpp/numerics/test/SolverLamgGTest.cpp
@@ -102,9 +102,7 @@ TEST_F(SolverLamgGTest, testSolveWithSingletonEliminationLevelSetsZeroResult) {
 
     solver.solve(x, b, status);
 
-    EXPECT_EQ(x.getDimension(), 2);
-    EXPECT_DOUBLE_EQ(x[0], 0.0);
-    EXPECT_DOUBLE_EQ(x[1], 0.0);
+    EXPECT_THAT(my_vector, AllOf(SizeIs(2), Each(DoubleEq(0.0))));
 }
 
 Vector SolverLamgGTest::randVector(count dimension) const {
--- a/networkit/cpp/numerics/test/SolverLamgGTest.cpp
+++ b/networkit/cpp/numerics/test/SolverLamgGTest.cpp
@@ -17,6 +17,7 @@
 #include <networkit/io/METISGraphReader.hpp>
 #include <networkit/io/METISGraphWriter.hpp>
 #include <networkit/numerics/GaussSeidelRelaxation.hpp>
+#include <networkit/numerics/LAMG/Level/EliminationStage.hpp>
 #include <networkit/numerics/LAMG/Lamg.hpp>
 #include <networkit/numerics/LAMG/MultiLevelSetup.hpp>
 #include <networkit/numerics/LAMG/SolverLamg.hpp>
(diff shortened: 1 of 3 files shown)


networkit/cpp/numerics/test/SolverLamgGTest.cpp
```diff
@@ -80,6 +81,32 @@ TEST_F(SolverLamgGTest, testSmallGraphs) {
...
+    SolverLamg<CSRMatrix> solver(hierarchy, smoother);
+    Vector x(2, 1.0);
+    const Vector b(2, 0.0);
+    LAMGSolverStatus status;
+
+    solver.solve(x, b, status);
+
+    EXPECT_EQ(x.getDimension(), 2);
+    EXPECT_DOUBLE_EQ(x[0], 0.0);
+    EXPECT_DOUBLE_EQ(x[1], 0.0);
```
angriman:
```suggestion
    EXPECT_THAT(my_vector, AllOf(SizeIs(2), Each(DoubleEq(0.0))));
```

angriman (dismissed):

networkit/cpp/numerics/test/SolverLamgGTest.cpp
Schwarf: Hi @angriman, 
This does not work because `Vector` does not provide the APIs needed by that matcher. I reverted the suggestion.

Schwarf:
Hi @fabratu,
the CI failures seem unrelated to my change.

This seems also the case for PR#1462 and PR#1463.

AI suggests: 
 CI fails while compiling the Cython-generated `networkit/centrality.cpp`, before running tests. The build dependency uses unpinned `cython`, so CI may have picked up a Cython version that generates duplicate vector conversion helpers.

networkit/cpp/numerics/test/SolverLamgGTest.cpp
angriman: Would `ElementsAre` work instead? If not, we can keep your version.

Schwarf: EXPECT_THAT is fine, but the suggested matcher does not fit `Vector` directly.
• `SizeIs` expects `.size()`, while `Vector` has `getDimension()`.
• `Each/ElementsAre` need STL-style iteration via `begin()`/`end()`, which `Vector` does not expose.

So it won't work with `ElementsAre` either

angriman: I see, if no `begin()`/`end()` is exposed then those matchers won't work. It would be a nice feature to have in `Vector`.

angriman (approved):

fabratu:
> Hi @fabratu, the CI failures seem unrelated to my change.
> 
> This seems also the case for PR#1462 and PR#1463.
> 
> AI suggests: CI fails while compiling the Cython-generated `networkit/centrality.cpp`, before running tests. The build dependency uses unpinned `cython`, so CI may have picked up a Cython version that generates duplicate vector conversion helpers.

Yep, just came back from a business trip and saw it also on the master build. I will look into it and merge this PR.

PR #1462, opened 2026-08-22
Template `LeftRightPlanarityCheck`

Schwarf:
This PR generalizes `LeftRightPlanarityCheck` to work with `AdjListGraph<NodeT, EdgeWeightT>` as part of the templated graph work discussed in #1415.

The implementation was moved from the `.cpp` file to a new `LeftRightPlanarityCheckImpl.hpp`. The existing `LeftRightPlanarityCheck` name is kept as an alias for the default `Graph` type.

The tests for generated graphs are now typed tests covering different node and edge-weight types.

### DIFF
--- a/networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
+++ b/networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
@@ -4,7 +4,6 @@
  *  Created on: 05.01.2025
  *      Author: Andreas Scharf (andreas.b.scharf@gmail.com)
  */
-#include <cmath>
 #include <stdexcept>
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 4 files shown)


networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
```diff
@@ -4,370 +4,400 @@
...
+    GraphType graphWithNodes(count numNodes) {
+        return GraphType(numNodes, TestT::weighted, false, true);
+    }
+
+    GraphType pathGraph(count numNodes) {
+        GraphType graph = graphWithNodes(numNodes);
         for (count i = 0; i < numNodes - 1; ++i) {
-            graph.addEdge(i, i + 1);
+            graph.addEdge(static_cast<NodeT>(i), static_cast<NodeT>(i + 1));
         }
```
angriman: Consider changing numNodes from `count` to `NodeT`, same in other functions below.

angriman (dismissed):

networkit/cpp/planarity/test/LeftRightPlanarityCheckGTest.cpp
Schwarf: Done

angriman (approved):

PR #1466, opened 2026-09-01
Template `NetworKitBinaryGraph` (reader / writer)

fabratu:
The PR should template `NetworKitBinaryGraph` and its reader/writer. I opened it as a draft since I am not sure about some design decisions.

- Introduced v5 for supporting variable bit widths, adjusted documentation
- v5 adds four new header bits: `NODE_TYPE_WIDTH_MASK`, `NODE_TYPE_WIDTH_SHIFT`, `TABLE_WIDTH_MASK` and `TABLE_WIDTH_SHIFT`. All four are handling lower bit widths for node IDs (two for node IDs directly, two for chunking)
- Lower bit widths for edge weights are (to some extent) already present in nkbg via `VARINT`. Instead of introducing new header bits for fixed lower bit widths, `VARINT` is reused. Cons: For small types like `uint8_t` the space consumption is doubled (2 bytes instead of 1 for numbers in the range `[128, 255]`). The effect diminishes, of course, for larger int types. Another option would be to also introduce header bits for edge weights (similar to node IDs). If we introduce new header bits, the auto detection would neglect `VARINT` completely. But maybe that is what we want in the end.

Current weight detection in `NetworKitBinaryWriterImpl.hpp`:

```
    auto detectWeightsType = [&]() -> nkbg::WEIGHT_FORMAT {
        if (!G.isWeighted())
            return nkbg::WEIGHT_FORMAT::NONE;

        if constexpr (std::is_integral_v<EdgeWeightT>) {
            return std::is_signed_v<EdgeWeightT> ? nkbg::WEIGHT_FORMAT::SIGNED_VARINT
                                                 : nkbg::WEIGHT_FORMAT::VARINT;
        }
        if constexpr (std::is_floating_point_v<EdgeWeightT>) {
            return std::is_same_v<EdgeWeightT, float> ? nkbg::WEIGHT_FORMAT::FLOAT
                                                      : nkbg::WEIGHT_FORMAT::DOUBLE;
        }
        throw std::runtime_error("Unsupported NetworkitBinaryGraph weight type");
    };
```

- Instead of changing `GraphReader` / `GraphWriter` to support templates and hence removing the virtual inheritance model, `GraphT readGraph` is adopted directly in `NetworKitBinaryReader` / `NetworKitBinaryWriter`. I am not sure what other design option there might be.
- `GraphT readGraph` needs to manually set the node and edge weight types without knowing the required bitwidth per se. I added a function `AnyBinaryGraph readCompact`, whereas `AnyBinaryGraph` represents `std::variant` of all common bitwidth combinations of node IDs and edge weights. Another option would be to add a function that returns the best fit tuple `(node IDs, edge weights)` based on the network file. This could be incorporated via additional header fields with the required information. With this, a linear scan over the file can be avoided.


include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
 static constexpr uint64_t WGHT_SHIFT = 0x1;
 static constexpr uint64_t INDEX_MASK = 0x10; // bit 4
 static constexpr uint64_t INDEX_SHIFT = 0x4;
+static constexpr uint64_t NODE_TYPE_WIDTH_MASK = 0x60; // bit 5-6
+static constexpr uint64_t NODE_TYPE_WIDTH_SHIFT = 0x5;
```
Schwarf, line 48: `NODE_TYPE_WIDTH_SHIFT` and `NODE_TYPE_WIDTH_MASK` ... They are only used in the NetworkitBinaryWriter ... is that on purpose?

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
...
+static constexpr uint64_t TABLE_WIDTH_MASK = 0x600; // bit 9-10
+static constexpr uint64_t TABLE_WIDTH_SHIFT = 0x9;
+
+inline uint8_t widthCode(uint8_t bytes) noexcept {
```
Schwarf, line 52: IMHO its hard to grasp what this function does. Even a descriptive name would need an explaining comment. 
My proposal:
```
// v5 stores selected byte widths as 2-bit codes in the feature word.
// The code is log2(byteWidth):
//   0 -> 1 byte
//   1 -> 2 bytes
//   2 -> 4 bytes
//   3 -> 8 bytes
inline uint8_t byteWidthToWidthCode(uint8_t byteWidth) noexcept {
    return byteWidth <= 1 ? 0 : byteWidth <= 2 ? 1 : byteWidth <= 4 ? 2 : 3;
}

```

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
...
+    return bytes <= 1 ? 0 : bytes <= 2 ? 1 : bytes <= 4 ? 2 : 3;
+}
+
+inline uint8_t widthBytes(uint8_t code) noexcept {
+    return static_cast<uint8_t>(1u << code);
```
Schwarf, line 57: My proposal:
```
inline uint8_t widthCodeToByteWidth(uint8_t widthCode) noexcept {
    return static_cast<uint8_t>(1u << widthCode);
}
```

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
...
+}
+
+inline uint8_t getFitWidthBytes(uint64_t value) noexcept {
```
Schwarf, line 60: Name proposal:
`requiredUintByteWidth`

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
...
+    if (value <= std::numeric_limits<uint8_t>::max())
+        return 1;
+    if (value <= std::numeric_limits<uint16_t>::max())
+        return 2;
+    if (value <= std::numeric_limits<uint32_t>::max())
+        return 4;
+    return 8;
+}
+
+inline void writeUint(std::ostream &out, uint64_t value, uint8_t bytes) {
```
Schwarf, line 70: Name proposal:
`writeFixedWidthUintLE`

include/networkit/io/NetworkitBinaryGraph.hpp
```diff
@@ -42,6 +44,42 @@ static constexpr uint64_t WGHT_MASK = 0xE;  // bit 1-3
...
+    for (uint8_t i = 0; i < bytes; ++i) {
+        const auto byte = static_cast<uint8_t>((value >> (i * 8)) & 0xFF);
+        out.write(reinterpret_cast<const char *>(&byte), sizeof(byte));
+    }
+}
+
+inline uint64_t readUint(const char *it, uint8_t bytes) noexcept {
```
Schwarf, line 77: Name proposal:
`readFixedWidthUintLE`

include/networkit/io/NetworkitBinaryReader.hpp
```diff
@@ -9,28 +9,46 @@
...
 #include <string>
+#include <variant>
 #include <networkit/graph/Graph.hpp>
 #include <networkit/io/GraphReader.hpp>
 #include <networkit/io/MemoryMappedFile.hpp>
+#include <networkit/io/NetworkitBinaryGraph.hpp>
 
 namespace NetworKit {
 
+using AnyBinaryGraph = std::variant<
```
Schwarf, line 20: I think there are cleaner ways to write this, although they may also add unnecessary complexity.

IMHO we should consider dropping the 1-byte types altogether, i.e. `uint8_t` for node IDs and `uint8_t`/`int8_t` for edge weights. That would reduce this list from 40 to 24 alternatives (40%), and I would not expect the additional memory overhead to be relevant in practice.

If we decide that 8-bit node/weight types are generally not worth supporting, we could enforce this directly through the `AdjListGraph` constraints and only allow 16-, 32-, and 64-bit integral types.
What do you think @fabratu & @angriman ?

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+        version = 3;
+    } else if (!memcmp("nkbg004", header.magic, 8)) {
+        version = 4;
+    } else if (!memcmp("nkbg005", header.magic, 8)) {
+        version = 5;
+    } else {
+        throw std::runtime_error("Reader expected another magic value");
+    }
+
+    weightFormat =
```
Schwarf, line 71: Could we split this into two statements for readability, e.g.:
```
const auto weightFormatCode =
    (header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT;
weightFormat = static_cast<nkbg::WeightFormat>(weightFormatCode);
```

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+        static_cast<nkbg::WeightFormat>((header.features & nkbg::WGHT_MASK) >> nkbg::WGHT_SHIFT);
+    tableWidth = 8;
+    if (version >= 5)
+        tableWidth = nkbg::widthBytes(static_cast<uint8_t>(
```
Schwarf, line 75: As above, could we split this in two statements.

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+
+namespace NetworKit {
+
+namespace nkbg {
+inline const char *accessData(const MemoryMappedFile &source) {
+    return source.cbegin();
+}
+
+inline const char *accessData(const std::vector<uint8_t> &source) {
+    return reinterpret_cast<const char *>(source.data());
```
Schwarf, line 26: `accessData` is useful to handle both memory-mapped files and in-memory buffers in the same way. This is probably out of scope for this PR, but it only returns a raw pointer and loses the size information.
The reader then uses offsets from the file header without checking whether they are still inside the mapped file / buffer. For truncated or corrupted files, this could lead to reads past the end.
Maybe this is something we should clean up in a follow-up, for example by keeping both the data pointer and the size and validating offsets before reading.

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+    if (version > 2) {
+        memcpy(&header.offsetAdjIdLists, it, sizeof(uint64_t));
+        it += sizeof(uint64_t);
+        memcpy(&header.offsetAdjIdTranspose, it, sizeof(uint64_t));
+    }
+
+    uint64_t maxUnsignedWeight = 1;
+    int64_t minSignedWeight = 1;
+    int64_t maxSignedWeight = 1;
+    if (header.nodes && weightFormat == nkbg::WeightFormat::VARINT) {
```
Schwarf, line 101: There is quite a bit of duplicated code between the unsigned and signed `VARINT` cases. Maybe this is an alternative

For example:

```cpp
const bool signedWeights =
    weightFormat == nkbg::WeightFormat::SIGNED_VARINT;

if (header.nodes
    && (weightFormat == nkbg::WeightFormat::VARINT || signedWeights)) {

    const char *adjIt = startIt + header.offsetAdjLists;
    const uint64_t weights =
        nkbg::readUint(adjIt + (header.chunks - 1) * tableWidth, tableWidth);

    uint64_t weightOff = (header.chunks - 1) * tableWidth;
    const char *weightIt = startIt + header.offsetWeightLists;

    for (uint64_t i = 0; i < weights; ++i) {
        uint64_t encodedWeight;
        weightOff += nkbg::varIntDecode(
            reinterpret_cast<const uint8_t *>(weightIt + weightOff),
            encodedWeight);

        if (signedWeights) {
            const int64_t weight = nkbg::zigzagDecode(encodedWeight);
            minSignedWeight = std::min(minSignedWeight, weight);
            maxSignedWeight = std::max(maxSignedWeight, weight);
        } else {
            maxUnsignedWeight =
                std::max(maxUnsignedWeight, encodedWeight);
        }
    }
}
```

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+        };
+
+        auto readAdjWeight = [&]() -> EdgeWeightT {
+            EdgeWeightT weight{1};
+            switch (weightFormat) {
+            case nkbg::WeightFormat::VARINT: {
+                uint64_t unsignedWeight;
+                wghtOff += nkbg::varIntDecode(
+                    reinterpret_cast<const uint8_t *>(adjWghtIt + wghtOff), unsignedWeight);
+                weight = static_cast<EdgeWeightT>(unsignedWeight);
```
Schwarf, line 373: Since `EdgeWeightT` is now selected be the caller of this function, should we verify that the stored weights are representable by the requested type before casting? Currently values can silently narrow or change signedness, e.g. reading 1000 into `uint8_t` or `-1` into an unsigned weight type.

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+                double doubleWeight;
+                memcpy(&doubleWeight, adjWghtIt + wghtOff, sizeof(double));
+                wghtOff += sizeof(double);
+                weight = static_cast<EdgeWeightT>(doubleWeight);
+            } break;
+            case nkbg::WeightFormat::SIGNED_VARINT: {
+                uint64_t unsignedWeight;
+                wghtOff += nkbg::varIntDecode(
+                    reinterpret_cast<const uint8_t *>(adjWghtIt + wghtOff), unsignedWeight);
+                weight = static_cast<EdgeWeightT>(nkbg::zigzagDecode(unsignedWeight));
```
Schwarf, line 385: Same as above, we cast but its unclear if we cast e.g. 1000 to 'uint8_t'?

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+            it += sizeof(uint64_t);
+            memcpy(&header.offsetAdjIdTranspose, it, sizeof(uint64_t));
+            it += sizeof(uint64_t);
+        }
+    };
+
+    readHeader();
+    nodes = header.nodes;
+    chunks = header.chunks;
+    if (nodes > static_cast<uint64_t>(std::numeric_limits<NodeT>::max()))
```
Schwarf: I think this check is off by one. `header.nodes` stores the upper node ID bound, not the maximum node ID itself.  For example, 256 node IDs (`0..255`) still fit into `uint8_t`, but this check would reject them because `256 > 255`.

Should this instead compare `nodes - 1` against `std::numeric_limits<NodeT>::max()` for non-empty graphs?

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+            uint64_t encodedWeight;
+            weightOff += nkbg::varIntDecode(reinterpret_cast<const uint8_t *>(weightIt + weightOff),
+                                            encodedWeight);
+            const int64_t weight = nkbg::zigzagDecode(encodedWeight);
+            minSignedWeight = std::min(minSignedWeight, weight);
+            maxSignedWeight = std::max(maxSignedWeight, weight);
+        }
+    }
+
+    if (header.nodes <= static_cast<uint64_t>(std::numeric_limits<uint8_t>::max()))
```
Schwarf: I think all of these checks are off by one. `header.nodes` stores the upper node ID bound, not the maximum node ID itself.

For example, 256 node IDs (`0..255`) still fit into `uint8_t`, 65536 node IDs (`0..65535`) still fit into `uint16_t`, and the same applies to `uint32_t`.

So the type selection should probably be based on the maximum node ID, i.e. `header.nodes - 1` for non-empty graphs, rather than comparing `header.nodes` directly against `std::numeric_limits<NodeT>::max()`.

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+GraphT NetworkitBinaryReader::readData(const T &source) {
+    using NodeT = typename GraphT::NodeT;
+    using EdgeWeightT = typename GraphT::EdgeWeightT;
+
+    nkbg::Header header;
+    nkbg::WeightFormat weightFormat;
+
+    const char *startIt = nkbg::accessData(source);
+    const char *it = startIt;
+    auto readHeader = [&]() {
```
Schwarf, line 186: I am a bit confused here. The same header data was already parsed in `readCompactData()`. Is there a reason why this parsing logic exists twice? Could we introduce a common helper for it?

include/networkit/io/NetworkitBinaryReaderImpl.hpp
```diff
@@ -0,0 +1,527 @@
...
+            for (uint64_t j = 0; j < outNbrs; j++) {
+                edgeid id = 0;
+                uint64_t add;
+                off += nkbg::varIntDecode(reinterpret_cast<const uint8_t *>(adjIt + off), add);
+                const EdgeWeightT weight = readAdjWeight();
+                if (indexed) {
+                    indexOff += nkbg::varIntDecode(
+                        reinterpret_cast<const uint8_t *>(adjIdIt + indexOff), id);
+                    if (id > omega)
+                        omega = id;
```
Schwarf, line 471: It seems to me there could be a serious problem here, although it is probably out of scope for this PR since it already existed before.
`omega` is captured by reference in `constructGraph()` and written from each iteration of the OpenMP parallel loop:
Since multiple threads can read and write `omega` concurrently without synchronization, this can cause a data race and therefore undefined behavior.

include/networkit/io/NetworkitBinaryWriterImpl.hpp
```diff
@@ -0,0 +1,444 @@
...
+    auto detectWeightsType = [&]() -> nkbg::WEIGHT_FORMAT {
+        if (!G.isWeighted())
+            return nkbg::WEIGHT_FORMAT::NONE;
+
+        if constexpr (std::is_integral_v<EdgeWeightT>) {
+            return std::is_signed_v<EdgeWeightT> ? nkbg::WEIGHT_FORMAT::SIGNED_VARINT
+                                                 : nkbg::WEIGHT_FORMAT::VARINT;
+        }
+        if constexpr (std::is_floating_point_v<EdgeWeightT>) {
+            return std::is_same_v<EdgeWeightT, float> ? nkbg::WEIGHT_FORMAT::FLOAT
```
Schwarf, line 53: I just realize that we also have no constrained in `AdjListGraph` for `long double` that would be narrowed here to `double`.

networkit/cpp/io/test/NetworkitBinaryGTest.cpp
```diff
@@ -0,0 +1,745 @@
+/*
+ * NetworkitBinaryGTest.cpp
```
Schwarf, line 2: I think the tests should cover all allowed combinations of `NodeT` and `EdgeWeightT` listed in `AnyBinaryGraph`, ideally with non-default weights and for both directed and undirected graphs.

Schwarf (changes requested):

Schwarf:
Thanks a lot for this PR, @fabratu. It is quite a large change, and I worked through the reader/writer code in some detail.

While reviewing it, I also found a few issues in the existing binary I/O code that are not really in scope of this PR. I think it would make sense to track a follow-up issue where we clean up the reader and writer implementation separately.

If you would like support with this PR, for example with adding templated/parameterized tests for the different `AdjListGraph<NodeT, EdgeWeightT>` combinations, let me know. I would be happy to contribute.

PR #1495, opened 2026-09-11
Template PrimMSF with Agent Skill for networkit template migration 

Schwarf:
This PR migrates `PrimMSF` as part of #1415 and adds an initial agent skill for future template migrations under `docs/AGENT_SKILLS`.

I intentionally chose a small algorithm, but `PrimMSF` also depends on the spanning-tree/forest infrastructure. Applying the skill with Codex correctly identified this dependency and adapted it as well.

The generated migration includes:

* templating the affected code while preserving the existing API;
* typed tests for representative graph types;
* the established test conventions
* removal of obsolete `.cpp` files and corresponding CMake entries where required.

The skill is based on patterns and review findings from previous migration PRs and is meant as a starting point that can be refined with future migrations.

I placed it in `docs/AGENT_SKILLS` for now because I am not aware of a single agent-agnostic location automatically recognized by Codex, Copilot, Claude, etc.

Please have a look — I think this could significantly speed up the remaining template migration work.

FYI: @angriman , @fabratu, @clstaudt


fabratu:
Unless I missed it, it would be good to add to the agent skill that the coverage of the involved files (changed / created) should not decrease.

If there is so far not a standardized way to include agent skills, I would favor having `.agents-skills` as hidden folder under root instead of docs. I guess in any case it will not be hard for the toolkit to locate the folder / files.

include/networkit/graph/SpanningForestImpl.hpp
```diff
@@ -0,0 +1,83 @@
...
+    for (NodeT u = 0; u < G.upperNodeIdBound(); ++u) {
+        if (!G.hasNode(u)) {
+            copy.removeNode(u);
+        }
+    }
+    return copy;
+}
+
+template <typename GraphT>
+void GenericSpanningForest<GraphT>::run() {
```
fabratu, line 32: The former implementation used `Traversal::BFSEdgesFrom`, which is not yet templated. Might need to tweak the skill so that AI also templates all includes exhaustively.

networkit/cpp/graph/test/SpanningGTest.cpp
```diff
@@ -21,10 +28,92 @@ namespace NetworKit {
...
+
+    std::vector<NodeT> graphNodes;
+    graphNodes.reserve(g.numberOfNodes());
+    g.forNodes([&](NodeT u) { graphNodes.push_back(u); });
+
+    std::vector<NodeT> forestNodes;
+    forestNodes.reserve(t.numberOfNodes());
+    t.forNodes([&](NodeT u) { forestNodes.push_back(u); });
+
+    EXPECT_THAT(forestNodes, testing::UnorderedElementsAreArray(graphNodes));
```
fabratu, line 43: I think it is not necessary to do this additional check since the forest is based on the graph.

fabratu (changes requested):

include/networkit/graph/SpanningForestImpl.hpp
Schwarf: That’s a valid point. I can include those changes here, but I think migrating BFS.hpp may be beyond the original scope of this PR.
Alternatively, I could migrate BFS.hpp in a separate PR first, and then we can continue with this one on top of that.
What do you think, @fabratu?

networkit/cpp/graph/test/SpanningGTest.cpp
Schwarf: Removed

include/networkit/graph/SpanningForestImpl.hpp
fabratu: A separate PR sounds more sane.

Issue #1457, opened 2026-08-05
Feature request: Subgraph isomorphism (VF2) for pattern matching


chiangchenghsin-hash:
## Motivation

We're building a software architecture governance system on top of a graph database (LadybugDB, a Kùzu fork). A core operation is architecture conformance checking: given an "as-is" call graph extracted from source code and a "should-be" design model defined in C4/DSL, detect whether the design model's sub-patterns exist (or don't exist) in the actual implementation.

Concretely: "Does the codebase contain a dependency chain matching Controller → Service → Repository, without going through the authorized gateway?" This is subgraph isomorphism — exactly what VF2 solves.

## Current state

NetworKit has an excellent algorithm catalog (centrality, community detection, graph difference, etc.), but subgraph isomorphism / pattern matching is not currently covered. The VF2 algorithm (Cordella et al. 2001, IEEE TPAMI) is the standard reference for this and already exists in sibling projects (igraph's `igraph_subisomorphic_vf2`, NetworkX's `DiGraphMatcher.subgraph_isomorphisms_iter`).

## Proposed addition

`SubgraphIsomorphism` (or `VF2`) as a new algorithm class, accepting a pattern graph and a target graph, returning all isomorphic mappings. Directed graph support is essential for the architecture use case.

## Who benefits

- Software architecture governance tools (the motivating use case)
- Pattern-based code search / clone detection
- Graph database ecosystem (LadybugDB plans to bridge NetworKit algorithms via the icebug project)

## Willing to contribute

Happy to prototype an implementation based on the paper and the igraph reference, if there's openness to the feature.

## References

- Cordella, L. P., Foggia, P., Sansone, C., & Vento, M. (2001). An Improved Algorithm for Matching Large Graphs. IEEE TPAMI.
- igraph: https://igraph.org/c/doc/igraph-Isomorphism.html
- NetworkX VF2: https://networkx.org/documentation/stable/reference/algorithms/isomorphism.vf2.html

mkirilin:
Sounds interesting, I will take a look at it.

PR #1521, opened 2026-09-17
Add a new isomorphism module

mkirilin:
# Add subgraph isomorphism module

This PR adds `networkit.isomorphism` module that finds every occurrence of a small pattern graph inside a large target graph. It ships three algorithms in C++ and Python, with tests and docs for both: VF2, RI, ParallelRI).

## Algorithms

- **`VF2`** follows Cordella et al. (2004).
- **`RI`** follows Bonnici et al. (2013). On sparse targets it is usually faster than VF2.
  - `RI.Variant.RI_DS` is the RI-DS-SI-FC variant from Kimmig, Meyerhenke and Strash (2017). Before the search, it pre-computes a set of candidate target nodes for every pattern node and prunes those sets as nodes get mapped.
- **`ParallelRI`** is the shared-memory parallel RI from Kimmig et al. (2017).

## Shared API

All three algorithms derive from the abstract `SubgraphIsomorphism`, which derives from `Algorithm`. They offer the same options:

- **Semantics.** `INDUCED` (the default) also requires pattern non-edges to map to target non-edges. `MONOMORPHISM` allows extra target edges.
- **Match limit.** `maxMatches` in the constructor stops the search early, and `hasMatch()` together with `maxMatches = 1` answers a yes/no question.
- **Labels.** `setNodeLabels()` and `setEdgeLabels()` restrict matches to equal labels. The label `none` matches any label. Edge labels need edge ids (`Graph::indexEdges()`).
- **Results.** By default, `getMatches()` returns the stored matches. `setStoreMatches(false)` only counts them, and `numberOfMatches()` works in both cases.
- **Callbacks.** In C++, `setCallback()` takes one of two forms. A `MatchCallback` is never called concurrently. A `ParallelMatchCallback` also receives the worker id and may run concurrently. Python exposes them as `setCallback(match)` and `setParallelCallback(workerId, match)`.

A match is a list indexed by pattern node, and `match[u]` is the target node that pattern node `u` maps to. The search reports each occurrence once per automorphism of the pattern, so a triangle occurs six times. The pattern must not contain self-loops. The search ignores edge weights and target self-loops, and it merges parallel edges. It throws an exception if two parallel edges carry different labels, because merging them would lose a label.

```python
import networkit as nk
from networkit.isomorphism import RI, ParallelRI, Semantics

target = nk.readGraph("input/karate.graph", nk.Format.METIS)
triangle = nk.Graph(3)
triangle.addEdge(0, 1); triangle.addEdge(1, 2); triangle.addEdge(2, 0)

ri = RI(triangle, target, semantics=Semantics.MONOMORPHISM)
ri.run()
print(ri.numberOfMatches())          # 270 = 45 triangles x 6 automorphisms

nk.setNumberOfThreads(4)
perWorker = [0] * 4
pri = ParallelRI(triangle, target)
pri.setParallelCallback(lambda worker, match: perWorker.__setitem__(worker, perWorker[worker] + 1))
pri.run()
print(sum(perWorker))                # 270
```


mkirilin:
Inspired by #1457.

PR #1522, opened 2026-09-17
Fix: DHBGraph edge and self-loop counters

mkirilin:
`DHBGraph::addEdges` and `DHBGraph::removeSelfLoops` updated edge and self-loop counts threads without synchronization. I fixed it by letting each thread incrementing in its own slot and then adding them up. 

For directed graphs, `DHBGraph::removeSelfLoops` also searched the incoming edges of every node, scanning every block in the graph. However, it could never find anything because, the entry for outgoing edge (and therefore incoming) has already been removed.

The tests hid the races because setup `Aux::setNumberOfThreads(1)`, and that setting leaked into later test suites. Now it restores the original number of threads at the end. I also added two more tests, that test correctness of the counters.


mkirilin:
Waiting for https://github.com/hu-macsy/dhb/pull/15.

fabratu:
> Waiting for [hu-macsy/dhb#15](https://github.com/hu-macsy/dhb/pull/15).

Done

fabratu (approved):

fabratu:
Good catch.

PR #1523, opened 2026-09-17
Fix narrowing of the weighted degree in LaplacianCentrality

UpstreamMath:
## Problem

`LaplacianCentrality::run()` assigns the node's own weighted degree to `count`:

```cpp
count degreeU = G.weightedDegree(u);
```

`Globals.hpp` defines `count` as `uint64_t` and `edgeweight` as `double`, so this narrows the value
to an integer, while the neighbour term two lines below uses `G.weightedDegree(v)` as a full
`double`. The `d_u^2` term is therefore computed from the truncated value, in both
`energyLossOnNodeDrop` and `totalLaplacianEnergy`. The conversion is also undefined for a negative,
NaN or infinite weighted degree, and `Graph` does not forbid negative edge weights.

## Example

A 3-node weighted path with `w(0, 1) = 2.5` and `w(1, 2) = 1.25`, so the weighted degrees are
2.5, 3.75 and 1.25:

| | node 0 | node 1 | node 2 |
|---|---|---|---|
| current | 29.0 | 32.4375 | 11.9375 |
| expected | 31.25 | 37.5 | 12.5 |

With `normalized = true` the same graph returns `0.9789, 1.0949, 0.4030` where the correct values
are `0.8333, 1.0, 0.3333`. `Centrality` documents normalized scores as "normalized in the interval
[0,1]", so a score above 1 contradicts the base class.

## Fix

Keep the weighted degree in `edgeweight`, as the neighbour term already does, and drop the two
`static_cast<double>` calls that are then redundant.

## Scope

Unweighted and integer-weighted graphs are unaffected: the three existing tests use integer weights
and are value-identical before and after. Values do change for graphs with fractional weighted
degrees, so `ranking()` order can change there. No API or signature change.

## Tests

Two cases added to `CentralityGTest.cpp` covering fractional weights, unnormalized and normalized.
Both fail before this change and pass after. Built locally against `master` with GCC 13 and ran
`CentralityGTest`; the only failure is `benchCoreDecompositionLocal`, which fails identically on an
unmodified checkout.


UpstreamMath:
The macOS 14 arm64 failure is a pypi timeout, not a code issue — pip gave up fetching ipycytoscape after five ReadTimeoutErrors. ctest had already finished with 100% tests passed out of 2 beforehand. Could someone re-run the job when convenient?

fabratu (approved):

mkirilin (approved):

